PR #104, opened 2018-12-28
fix(manifest.go): fix Name property, add Description, add test

vdice:
The main change here (mapping `Manifest.Name` to `name` in yaml, as opposed to `image`), fixes the issue wherein the `name` field in the `bundle.json` would be rendered empty after a `porter build`.  This empty field prevents successful installation of the bundle (it cannot be found.)


carolynvs:
Let's hold off on this for a sec until I can clear my head from vacation. I think there's a bit of a mixup between how porter resolves bundle dependencies, and how people _expect_ it to that prompted this fix. 

I want to make sure that we have that clear in our heads before we merge any fixes related to that.

pkg/porter/build_test.go
```diff
@@ -80,3 +82,35 @@ func TestPorter_prepareDockerFilesystem(t *testing.T) {
 	require.NoError(t, err)
 	assert.True(t, execMixinExists, "The exec-runtime mixin wasn't copied into %s", wantExecMixin)
 }
+
+func TestPorter_buildBundle(t *testing.T) {
```
carolynvs: Whatever else happens, let's make sure to keep a new test. 👍

pkg/config/manifest.go
```diff
@@ -20,7 +20,8 @@ type Manifest struct {
 	path    string
 	outputs map[string]string
 
-	Name         string                 `yaml:"image,omitempty"`
```
carolynvs: Looks like image is no longer being populated anymore?

carolynvs:
@vdice Just in case this was in response #108, just wanted you to see that I created #109 to give us a space to figure out how to resolve bundle dependencies.

I think this was to fix `duffle list`? If you could clarify, that would be helpful! 😀

vdice:
@carolynvs this PR originated with my using `porter` to build a bundle (`bundle.json`) from a `porter.yaml` file.  Without this change, I noticed that the `build.json` had an empty string for the `name` property -- due to, as seen in the changeset, `Manifest.Name` mapping to `image` in the yaml (`porter.yaml`), when it should (or I assumed it should) map to `name` in the yaml.  (I was referencing example `porter.yaml` [here](https://github.com/deislabs/porter/blob/master/examples/azure-mysql-wordpress/porter.yaml)).

Later, upon reading the issue in https://github.com/deislabs/porter/issues/108, I theorized this user perhaps was experiencing the effects of the same issue described above.  So, not so much around the larger issue(s) of how Porter resolves dependencies, but more just a simple bug fix to ensure the go struct (`Manifest`) gets the data it needs (from `porter.yaml`) to output a correctly-formed bundle (`bundle.json`).

pkg/config/manifest.go
vdice: Indeed, if there was an `image` field in the example [porter.yaml](https://github.com/deislabs/porter/blob/master/examples/azure-mysql-wordpress/porter.yaml) at one point, it is no longer there.  I've assumed this should point to the `name` field, which is populated.

carolynvs:
Thanks for clarifying! I'm going to let @jeremyrickard take this from here, since it's not dependency related then.

PR #107, opened 2019-01-02
feat(examples): add azure-ark example

vdice:
Adds Porter example utilizing Azure Storage and a Helm-based Ark installation

Uses porter-azure mixin functionality PR'ed in https://github.com/deislabs/porter-azure/pull/6

For development/testing, I built the porter-azure mixin binaries (from branch above) and copied them into the default porter location before a `porter build`, `duffle install`, etc.:

```
cd path/to/porter-azure && make build && cp bin/mixins/azure/azure* path/to/porter/bin/mixins/azure/
cd path/to/porter/examples/azure-ark && porter build
duffle bundle sign -f bundle.json
duffle creds generate ark-creds porter-azure-ark
duffle install ark porter-azure-ark -c ark-creds
```

TODOs (assuming PR a candidate for merge):
 - [x] update image repo to `deislabs`


examples/azure-ark/porter.yaml
```diff
@@ -0,0 +1,82 @@
+mixins:
+  - azure
+  - helm
+ 
+name: porter-azure-ark
+description: "A Porter example using Azure Storage and Ark"
+version: 0.1.0
+# TODO: change to deislabs
```
carolynvs: You have push rights on deislabs, let me know if you run into trouble

examples/azure-ark/porter.yaml
```diff
@@ -0,0 +1,82 @@
...
+
+install:
+  - description: "Create Azure Storage Account and Container"
+    azure:
+      type: storage
+      name: porter-azure-ark
+      resourceGroup:
+        source: bundle.credentials.RESOURCE_GROUP
+      parameters:
+        location: "eastus"
```
carolynvs: Will this run into trouble if the resource group isn't in eastus?

examples/azure-mysql-wordpress/porter.yaml
```diff
@@ -56,7 +56,7 @@ install:
     helm:
       name: porter-ci-wordpress
       chart: stable/wordpress
-      repalce: true
+      replace: true
```
carolynvs: 🙇‍♀️

carolynvs:
OK, step away from the keyboard, E_TOO_MUCH_COUGH_SYRUP

examples/azure-ark/porter.yaml
vdice: Ah, good point; I'd imagine it would.  Perhaps we parameterize this value?  If so, should this parameter be non-required, with a default of `eastus` (but changeable) or required with no default?

examples/azure-mysql-wordpress/porter.yaml
vdice: Just a quick search-and-repalce :wink:

examples/azure-ark/porter.yaml
jeremyrickard: +1 default to eastus or soemthing.  Out of curiosity, why the credentials instead of a parameter to the bundle? they seem like conceptually different things.

vdice: Mmm, good call -- indeed, resource group _should_ be a parameter.  Will fixup as well.

carolynvs: 👍 Seems like a good way to do it, a parameter with a default.

examples/azure-ark/porter.yaml
vdice: 👍 Yup, no problems; thank you!

examples/azure-ark/porter.yaml
vdice: Now updated; PTAL

carolynvs:
> It seems the porter-azure mixin from the most recent master commit may not have been published or was not published to the latest URL 

The way the builds are setup here's what get published when:

* canary: tip of master
* vX.Y.Z: tagged commits
* latest: most recent tagged commit pushed

So we can either tag a commit on porter-azure, then porter will pick it up. Or we can say that really porter should be relying on canary (tip of master) from porter-azure in its makefile, not latest (last tag).

What do you think?

carolynvs:
> Anyways, for my purposes, no matter: workaround is to rm these files before a subsequent make build

I didn't want to re-download the mixins on every build. Suggestions welcome on how to manage that better. I do have plans to make it easier to install plugins for users, but didn't have much on the backlog for managing this for ourselves (beyond what you just did, rm and redo the build).

How about clean should remove the cached mixins?

This was all put together 6 hours before KubeCon in the hotel lounge when we found out that we were going to show porter to people, so don't assume anything was by design. 😉

carolynvs:
I've turned my comment explaining the release tags into documentation on the readme. https://github.com/deislabs/porter/pull/110

vdice:
Ah, thank you @carolynvs for the clarification; indeed, I'm not surprised to hear both cases had been considered *and* that the defaults make sense!

> So we can either tag a commit on porter-azure, then porter will pick it up. Or we can say that really porter should be relying on canary (tip of master) from porter-azure in its makefile, not latest (last tag).

This totally makes sense -- we might not want porter to grab the very latest (canary here).  I agree, I think porter should grab the latest release of the azure mixin that we're confident in -- so, the first route of tagging when ready, to trigger the proper publishing.  Good call.

> How about clean should remove the cached mixins?

I agree here, too -- when playing around with PHONY-izing, indeed, it didn't feel right to re-fetch external dependencies on every build.  I think just adding a few lines to `make clean` sounds perfect!  I'll PR.

carolynvs:
MAGIC! 🎩 ✨ 

Let me know when you feel good about a tag on porter-azure. In the meantime you can manually grab the canary binaries and use them locally to test. Maybe we need a new makefile target to make that easier to do?

vdice:
> Let me know when you feel good about a tag on porter-azure

In my testing of this example, master branch of porter-azure looks good (newly-added storage template/functionality working swimmingly).  If test automation (`test-cli` in this repo -- other?) agrees, I'd say let's tag it!

PR #112, opened 2019-01-08
fix(Makefile): add REGISTRY default

vdice:
This variable is used in the `test-cli` target but didn't have a default set...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 SHELL = bash
 
+REGISTRY ?= $(USER)
 VERSION ?= $(shell git describe --tags 2> /dev/null || echo v0)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,5 +1,6 @@
 SHELL = bash
 
+REGISTRY ?= deislabs
```
carolynvs: Oops! Thanks for fixing. Would you mind changing the default to `$(USER)` instead? That way when I run `make test-cli` I don't accidentally push to deislabs. 😊 

The CI build already sets REGISTRY to deislabs explicitly so it shouldn't hurt it.

vdice: Definitely!  Pushed up amended commit.

PR #117, opened 2019-01-16
docs: Initial invocation image build document

### LINKED ISSUE #1: Document how porter builds the invocation image
* copying in stuff for helm, arm
* generating dockerfile

jeremyrickard:
This document includes details on how the invocation image is constructed.
It includes details on how the porter.yaml is used to determine which dependnecies and
mixins to include in the bundle, how the mixins are used to contribute to the resulting
Dockerfile and how dependencies are included in a bundle. Throughout the document, it
uses several example porter.yaml files and the resulting `porter build` command output
to explain each of these scenarios.

Closes #1


jeremyrickard:
I added a couple of #tbd links to add placeholders. As we build out documentation, i think it would be good to link to Mixin and dependency documentation from this page.

docs/content/build-image.md
```diff
@@ -0,0 +1,358 @@
...
+descriptions: How does Porter build an Invocation Image?
+---
+
+When you build Cloud Native Application Bundle (CNAB) with Porter, a bundle.json and an invocation image are created for you. How does Porter turn your _porter.yaml_ into an invocation image? This walkthrough will explain how Porter constructs the invocation image, including how mixins and other bundles allow you to compose functionality.
+
+## Starting From Scratch
+
+When you create a new bundle with Porter, your project is bootstrapped with a sample _porter.yaml_ and a new _cnab_ directory. This scaffolding provides almost everything you need to generate your CNAB, including the invocation image. Let's use this to explain how the invocation image is built. 
+
+To create a new CNAB with Porter, you first run `porter create`. The generated `porter.yaml` will look like this:
```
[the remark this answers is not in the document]
jeremyrickard: I'll open a follow up issue to address putting these into the bin maybe with packr.

docs/content/build-image.md
```diff
@@ -0,0 +1,358 @@
...
+uninstall:
+- description: "Say Goodbye"
+  exec:
+    command: bash
+    arguments:
+      - -c
+      - echo Goodbye World
+```
+
+This bundle, for example, declares a dependency on a bundle named `mysql`. The CNAB specification doesn't provide a mechanism for handling dependency resolution. Porter addresses this capability by resolving any dependencies at build time, including the contents of each dependency in the invocation image. At runtime the contents of that bundle will therefore be in the bundle and the Porter runtime component can execute them successfully.
```
carolynvs: Change `Port addresses this capability` to `gap`

carolynvs: Or to be more positive, `Porter supplements the CNAB spec to support dependencies by resolving ....`

docs/content/build-image.md
```diff
@@ -0,0 +1,358 @@
...
+    │   │   ├── helm
+    │   │   └── helm-runtime
+    │   └── porter
+    │       ├── porter
+    │       └── porter-runtime
+    ├── porter-runtime
+    └── run
+```
+
+Porter found the `mysql` bundle by first looking in the `./bundles` directory. If nothing was found, it then checked `$PORTER_HOME\bundles` for the `mysql` bundle. In this case, it found the bundle and was able to copy it into the directory. Porter also copied the `helm` mixin into our bundle, despite the `porter.yaml` declaring only the `exec` mixin. It did this because the `mysql` dependency requires the `helm` mixin. As we can also see from the Dockerfile, it included the mixin for the `mysql` before the `exec` mixin.
```
carolynvs: I think we need a final paragraph summarizing the state of the invocation image and wrapping things up? Maybe something like 

`In the end, regardless of whether there are dependencies, the result is a single invocation image with all of the necessary pieces (config files, binaries, charts, etc included in the image) which can be executed by any tool that supports the CNAB spec... blah blah`

PR #130, opened 2019-01-24
List installed mixins

### LINKED ISSUE #116: Provide a way to list available mixins
It is often useful for tooling to know which mixins are present, for example to validate a `mixins` section in YAML or to query each mixin for its schema.  At the moment this requires tooling to locate the Porter binary and traverse subdirectories from there.  It would be useful to have a command such as `porter mixin list` to get the info from Porter itself.

carolynvs:
- [x] Make it work
- [x] Ok, now really make it work for everybody
- [x] Document it (preview @ https://deploy-preview-130--porter.netlify.com/cli/#list-mixins)
- [x] Test it
- [x] Go back and fix all the bugs

Closes #116

### DIFF
--- a/cmd/porter/mixins_test.go
+++ b/cmd/porter/mixins_test.go
@@ -1,0 +1,38 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/deislabs/porter/pkg/porter"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBuildListMixinsCommand_DefaultFormat(t *testing.T) {
+	p := porter.NewTestPorter(t)
+	cmd := buildMixinListCommand(p.Porter)
+	cmd.ParseFlags([]string{"-o", "json"})
+
+	err := cmd.PreRunE(cmd, []string{})
+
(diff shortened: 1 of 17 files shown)


cmd/porter/mixins_test.go
```diff
@@ -0,0 +1,38 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/deislabs/porter/pkg/porter"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBuildListMixinsCommand_DefaultFormat(t *testing.T) {
```
vdice: nit: should this actually be named the `*_AlternateFormat` test?  (non-default `-o json` passed)

carolynvs: D'oh! I think I got interrrupted mid-test and these 2 aren't named right and one isn't finished! 😊 Lemme fix.

carolynvs: OK, all fixed in the latest commit. Thanks for noticing!

PR #133, opened 2019-02-01
feat(pkg/printer): add yaml and raw printers

vdice:
Ref: https://github.com/deislabs/porter-helm/pull/12

Adds `yaml` and `raw` printers to the printer pkg.


pkg/printer/raw.go
```diff
@@ -0,0 +1,12 @@
+package printer
+
+import (
+	"fmt"
+	"io"
+)
+
+// PrintRaw is a printer that prints the provided value as is
```
carolynvs: So raw is really a dump of the go structure? Let me know if I don't have that right.

What's the use case for this format? json and yaml I see people using that in scripts. If they are chaining the output of this new raw format into a Go program I want to make sure that maybe there isn't an easier way (like making it so that they can use our libraries and functions directly from their code). Or at least making sure that it's a consumable format (e.g. maybe we should be using `%#v` instead).

Also once we have hammered out the use case, let's pick a name that describes the format more clearly, e.g. "go" or something along those lines.

vdice: Updated to change to `FormatPlaintext` per https://github.com/deislabs/porter-helm/pull/12#discussion_r254059588

pkg/printer/plaintext.go
```diff
@@ -0,0 +1,12 @@
+package printer
+
+import (
+	"fmt"
+	"io"
+)
+
+// PrintPlaintext is a printer that prints the provided value as is
+func PrintPlaintext(out io.Writer, v interface{}) error {
+	fmt.Fprintf(out, `%#v`, v)
```
carolynvs: I just tried it out on the [go playground](https://play.golang.org/p/zoN94WdrPS2) and realized that my suggestion of `%#v` won't work well with what we are trying to do with helm status. It's going to put quotes around the passed value when the goal is the pass through the helm output as-is.

Sorry about that, let's go back to what you had before. I don't recall if it was `fmt.Fprintln(out, v)` which adds a trailing newline or `fmt.Fprintf(out, "%v", v)` which relies on the passed input having a newline at the end.

vdice: Ah good point; switched back to `fmt.Fprintf(out, "%v", v)`

PR #134, opened 2019-02-01
feat(Makefile/mixin.mk): add install target

vdice:


carolynvs:
Oh shoot! I didn't see this when I submitted #140... 🤦‍♀️ 

I think that there are some key things from both that we may want to consider smooshing together, but I'm not 100% sure. The most important thing is that porter can figure out on its own where its home directory is, without PORTER_HOME being explicitly set.

The way I did that with my 2 PRs is to symlink, and then make sure porter knew how to follow the symlink back. 

How does yours work?

vdice:
Oooh, I do like the symlink approach; we should go that route.  My approach was not as nice (simply copied `porter` cli bin to `/usr/local/bin`, only moved mixin/runtime bins to `~/.porter`/`$PORTER_HOME`.

Plus, you've updated documentation! 😅 

I'd say, if we add mixin/runtime installation (to `~/.porter`/`$PORTER_HOME`) in https://github.com/deislabs/porter/pull/140, we'll be golden.  I'll close this one in favor of that route.

PR #135, opened 2019-02-01
feat(*): add upgrade action support

vdice:
Adds `upgrade` action support to porter (and the `exec` mixin).

Closes https://github.com/deislabs/porter/issues/123

### DIFF
--- a/pkg/config/manifest.go
+++ b/pkg/config/manifest.go
@@ -27,6 +27,7 @@ type Manifest struct {
 	Mixins       []string               `yaml:"mixins,omitempty"`
 	Install      Steps                  `yaml:"install"`
 	Uninstall    Steps                  `yaml:"uninstall"`
+	Upgrade      Steps                  `yaml:"upgrade"`
 	Parameters   []ParameterDefinition  `yaml:"parameters,omitempty"`
 	Credentials  []CredentialDefinition `yaml:"credentials,omitempty"`
 	Dependencies []*Dependency          `yaml:"dependencies,omitempty"`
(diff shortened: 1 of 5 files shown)


pkg/config/manifest.go
```diff
@@ -315,16 +318,22 @@ func (m *Manifest) MergeDependency(dep *Dependency) error {
...
 }
 
 func (m *Manifest) MergeInstall(dep *Dependency) {
 	dep.m.Install.setDependency(dep)
 
-	result := make(Steps, len(m.Install)+len(dep.m.Install))
-	copy(result[:len(m.Install)], dep.m.Install)
-	copy(result[len(m.Install):], m.Install)
-	m.Install = result
+	m.Install = prependSteps(dep.m.Install, m.Install)
```
carolynvs: 👍 Yay! Thank you for tidying this up. It sparks much joy

pkg/mixin/exec/uninstall.go
```diff
@@ -1,9 +1,6 @@
 package exec
 
 func (m *Mixin) Uninstall(commandFile string) error {
-	err := m.LoadInstruction(commandFile)
-	if err != nil {
-		return err
-	}
-	return m.Execute()
+	// re-use Install's logic
+	return m.Install(commandFile)
```
carolynvs: In a follow-on PR, it's up for grabs to anyone (doesn't have to be you!) who would like to extract the logic from `m.Install` into a function that is named a bit more generically so that all of these functions can call into it, maybe `m.ExecuteScript` so that we don't need this comment.

It's a bit confusing to see uninstall calling to install here. I know it's because exec is an odd-ball mixin. 😀

carolynvs:
FYI, it's up to you if you want to prefix your PRs with feat/chore/fix. I don't mind, but I don't get too much out of it either. Just throwing that out there in case you weren't sure if it's required in these porter repos.

PR #139, opened 2019-02-05
Resolve PORTER_HOME when it's symlinked

carolynvs:
When someone symlinks /usr/local/bin/porter from ~/.porter/porter, we need to handle that properly
so that we can still find the mixins and other assets in the original directory.

Also cache the resolved PORTER_HOME so we don't have to do that more than once.

PR #140, opened 2019-02-05
Add install target

carolynvs:
The install target dumps the bin into ~/.porter and symlinks it to /usr/local/bin.

Note that this doesn't work with porter until #139 is merged.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,6 +13,9 @@ on top of your work.
 1. `make build`, and the resulting binaries are located in `./bin`.
 1. `./bin/porter COMMAND`, such as `./bin/porter build`.
 
+If you would like to install a developer build, run `make install`.
+This copies a dev build to `~/.porter` and symlinks it to `/usr/local/bin`.
+
 # Documentation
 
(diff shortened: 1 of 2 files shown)


carolynvs:
The `install` target runs `build` first, which is handling getting all of the mixins into bin before we dump the entire thing over to `~/.porter`.

Here's what's in `~/.porter` after I deleted my bin, and `~/.porter` and then ran `make install`:

```
/Users/carolynvs/.porter
├── mixins
│   ├── azure
│   │   ├── azure
│   │   └── azure-runtime
│   ├── exec
│   │   ├── exec
│   │   └── exec-runtime
│   ├── helm
│   │   ├── helm
│   │   └── helm-runtime
│   └── porter
│       ├── porter
│       └── porter-runtime
├── porter
└── templates
    ├── porter.yaml
    └── run
```

Do you see anything missing?

Yeah, I originally expected that `PORTER_HOME` at the top of the Makefile would point to `~/.porter`, then realized that it doesn't due to the CI tests. I will follow-up in another PR to change that so that it works like the other Makefiles (and still runs the CI tests from inside bin instead of against the system).

vdice:
Ah, now I gotcha 👍.  Thank you for the detailed explanation.

PR #146, opened 2019-02-07
Add porter schema root command

### LINKED ISSUE #76: Porter should have a way to report schema both for itself and its mixins
The parameters and output format for the Mixins can vary so we should provide a way for users to interrogate the mixins to get the schema.

carolynvs:
```console
$ ./bin/porter schema
{
    "type": "object",
    "properties": {
      "name": {
        "type": "string"
      },
      "description": {
        "type": "string"
      },
      "version": {
        "type": "string"
      },
      "image": {
        "type": "string"
      },
      "mixins": {
        "type": "array",
        "items": {
          "type": "string"
        }
      }
    },
    "additionalProperties": true
}
```

Note: This isn't the full json schema. We will fill that in once we validate this these commands are plumb together (between an IDE, porter and the mixins) to help provide autocomplete/intellisense.

~I'm following up on this PR with another that adds `porter schema --mixin NAME` but figured it would be easier to review incrementally.~ EDIT: After talking with @itowlson this is on hold, may not need a subcommand.

Closes #76 

Here's all the things I found when shaving this yak that need to be split out into separate PRs most likely:

- [x] Porter not mixin anymore
- [ ] Go generate or packr build or packr clean. We aren't using packr right in our builds. (#193)
- [ ] Remaining mixin functions to interface as issue (#192)
- [x] Helm and azure schema sync up (this didn't fix the schema to report all the fields, just to have them report the _correct shape_ as per #182)


pkg/porter/schema/manifest.json
```diff
@@ -0,0 +1,24 @@
...
+      "name": {
+        "type": "string"
+      },
+      "description": {
+        "type": "string"
+      },
+      "version": {
+        "type": "string"
+      },
+      "image": {
```
vdice: Should this be `invocationImage`?  (Referencing the two `porter.yaml` examples under `examples`) . Or would that  fall under `additionalProperties`?

Speaking of which, what is the delineation of properties that are 'top level' and that are grouped under `additionalProperties`?

itowlson: @vdice `additionalProperties: false/true` simply means "do/don't error if you see properties other than those listed".  It's not parking a second set of properties away somewhere; it's saying "it's okay if the document contains other properties too".

carolynvs: oops! Thanks I accidentally looked at the struct field instead of the yaml. 🤦‍♀️ 

There shouldn't be anything under additional properties. I just have it turned on for now because it's not the full schema while we test this out with Ivan and VS Code. Once we have the full schema in place, we'll turn it to false.

pkg/porter/schema/manifest.json
```diff
@@ -0,0 +1,24 @@
+{
```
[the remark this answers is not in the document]
carolynvs: I hand wrote it for now. Once you bless what we are doing, I'm going to generate it programmatically to make it look like whatever you tell me to. 😇

pkg/porter/schema/manifest.json
```diff
@@ -0,0 +1,24 @@
...
+        "type": "string"
+      },
+      "mixins": {
+        "type": "array",
+        "items": {
+          "type": "string"
+        }
+      }
+    },
+    "additionalProperties": true
```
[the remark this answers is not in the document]
carolynvs: I did that for now until it actually represents the real schema. Honestly I don't know json schema much at all right now, so I'm just focusing on the wiring up of the commands, and plan on circling back on the contents of the schema in a second step. If that seems wildly unhelpful, I can do it differently though.

pkg/porter/schema/manifest.json
```diff
@@ -0,0 +1,24 @@
...
+      "version": {
+        "type": "string"
+      },
+      "image": {
+        "type": "string"
+      },
+      "mixins": {
+        "type": "array",
+        "items": {
+          "type": "string"
```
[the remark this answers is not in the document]
carolynvs: Are you cool with it being the list of installed mixins? So that the results of the call to `porter schema` change based on what is installed on the system? If that's okay I can change to an enum.

There isn't a centralized list of all mixins, anyone can write one.

cmd/porter/schema.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"github.com/deislabs/porter/pkg/porter"
+	"github.com/spf13/cobra"
+)
+
+func buildSchemaCommand(p *porter.Porter) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "schema",
+		Short: "Print the json schema for the Porter manifest",
```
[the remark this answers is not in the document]
carolynvs: roger! 👍

pkg/porter/schema/manifest.json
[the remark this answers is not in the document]
itowlson: For the VS Code POC I was thinking I would have an enum of installed mixins.  That means that Intellisense can suggest installed mixins and flag if a mixin name is unrecognised.  Hopefully I can still have a code fix for "find and install this mixin" that can coexist with the more 'raw' JSON Schema validation error.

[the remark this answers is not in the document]
carolynvs: ooh "find and install this mixin". I'm going to split that off into another issue. It's a good scenario but since currently mixins aren't distributed via a centralized repository, we need to plan on how that would be achievable.

carolynvs:
Switching back to WIP. After talking with @itowlson I'm going to have it work this way instead:

```console
$ porter schema
< dump a consolidate / merged JSON schema for the entire porter.yaml taking into consideration the mixins installed on the system>
```

The other command `porter schema --mixin FOO` is on hold, may not ever be needed.

carolynvs:
@itowlson Here is what the command looks like now. Let's ignore the helm mixin definition for now, I'm going to need to go back and fix it, since it's in another repository. I did do the exec mixin since it's in this repo. How well does this match what you expected and what VS Code requires?

The list of definitions is dynamic based on what mixins are installed, and so is the mixins enum, and the refs under install. Once we have install working, I'll do upgrade and uninstall as well.

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "additionalProperties": false,
  "definitions": {
    "exec.install": {
      "properties": {
        "description": {
          "type": "string"
        },
        "exec": {
          "properties": {
            "arguments": {
              "items": {
                "type": "string"
              },
              "type": "array"
            },
            "command": {
              "type": "string"
            }
          },
          "type": "object"
        }
      },
      "type": "object"
    },
    "helm.install": {
      "additionalProperties": true,
      "properties": {
        "chart": {
          "type": "string"
        },
        "name": {
          "type": "string"
        }
      },
      "type": "object"
    }
  },
  "properties": {
    "description": {
      "type": "string"
    },
    "install": {
      "anyOf": [
        {
          "$ref": "#/definitions/exec.install"
        },
        {
          "$ref": "#/definitions/helm.install"
        }
      ],
      "type": "array"
    },
    "invocationImage": {
      "type": "string"
    },
    "mixins": {
      "items": {
        "enum": [
          "exec",
          "helm"
        ],
        "type": "string"
      },
      "type": "array"
    },
    "name": {
      "type": "string"
    },
    "version": {
      "type": "string"
    }
  },
  "type": "object"
}
```

If it's easier than reading the schema, I can give you a binary to try out locally. Just let me know.

itowlson:
That looks like it's going the right way, but I'd consider refactoring the `install` definitions so that `description` is part of `install` rather than each mixin having to report it.  (Unless you want mixins to be able to define properties outside the scope of the `<mixin_name>:` element of course.)

I think the way I had this in my POC evaluated to something like:

```json
// in the definitions section (the actual schema
// for install would be an array of this)
"install": {
  "type": "object",
  "properties": {
    "description": { ... },
    "outputs": { ... },
    "exec": { "type": "object", "$ref": "#/definitions/exec-install" },
    "helm": { "type": "object", "$ref": "#/definitions/helm-install" },
    "oneOf": [
      { "required": "exec" },
      { "required": "helm" }
    ]
  }
}
```

This meant that (what I took to be) common stuff like `description` and `outputs` would be declared at the step level and it was only the `exec: ...` or `helm: ...` bodies that would have to be declared by the mixin.  I now realise I did not check my assumptions with you though!

Hope this makes sense - happy to explore in more detail if you want.  I'm not trying to mandate what I came up with - in fact I prefer `anyOf` to the `oneOf: [ { required... } ]` trick - just trying to keep the mixin implementation side simple if I have understood it right.

carolynvs:
I just double checked and moving outputs under the mixin would not have any negative effect on porter. Description was being used by porter for logging progress such as "now executing the DESCRIPTION step". But that's ok to move too.

It is tidier to move it all under the mixin. I'll try that, and keep the "anyOf" since you like that better. 

Hopefully this is getting closer to an end. Sorry it's taking so long!

itowlson:
Sounds good - I'll load it up into VS Code and see how it plays out in terms of proferred popups and validation error messages.

itowlson:
VS Code plays nicely with the proposed output - or at least no worse than it did with any of the other ways I've tried expressing the schema.  The main difference I noticed was if I had both a `helm` and `exec` in a step: my old schema reported `Matches multiple schemas when only one must validate` and the new one reports `Unexpected property helm`.  I think the new one is probably more accurate and informative.

I did need to make a few tweaks, but those were mostly relating to the Helm stuff which you hadn't done yet.  We also need some `required` elements to get Code to whinge if e.g. you supply a step description with no actual body!  Here's where I ended up in my testing:

```json
  {
    "$schema": "http://json-schema.org/draft-07/schema#",
    "additionalProperties": false,
    "definitions": {
      "exec.install": {
        "properties": {
          "description": {
            "type": "string"
          },
          "exec": {
            "properties": {
              "arguments": {
                "items": {
                  "type": "string"
                },
                "type": "array"
              },
              "command": {
                "type": "string"
              }
            },
            "type": "object",
            "additionalProperties": false
          }
        },
        "required": ["exec"],
        "type": "object",
        "additionalProperties": false
      },
      "helm.install": {
        "properties": {
          "description": {
            "type": "string"
          },
          "helm": {
            "properties": {
              "chart": {
                  "type": "string"
              },
              "name": {
                  "type": "string"
              }
            },
            "type": "object",
            "additionalProperties": false
          }
        },
        "required": ["helm"],
        "type": "object",
        "additionalProperties": false
      }
    },
    "properties": {
      "description": {
        "type": "string"
      },
      "install": {
        "type": "array",
        "items": {
          "anyOf": [
            {
              "$ref": "#/definitions/exec.install"
            },
            {
              "$ref": "#/definitions/helm.install"
            }
          ]
        }
      },
      "invocationImage": {
        "type": "string"
      },
      "mixins": {
        "items": {
          "enum": [
            "exec",
            "helm"
          ],
          "type": "string"
        },
        "type": "array"
      },
      "name": {
        "type": "string"
      },
      "version": {
        "type": "string"
      }
    },
    "type": "object"
  }
```

carolynvs:
@jeremyrickard Rebased for great glory

carolynvs:
Okee, one more rebase to pick up the go generate PR and resolve some conflicts.

cmd/porter/run.go
```diff
@@ -23,7 +23,7 @@ func buildRunCommand(p *porter.Porter) *cobra.Command {
 			if opts.rawAction == "" {
 				opts.rawAction = os.Getenv(config.EnvACTION)
 				if p.Debug {
-					fmt.Fprintf(p.Out, "DEBUG: defaulting action to %s (%s)\n", config.EnvACTION, opts.rawAction)
+					fmt.Fprintf(p.Err, "DEBUG: defaulting action to %s (%s)\n", config.EnvACTION, opts.rawAction)
```
vdice: Curious as to the decision to print to stderr (if that is what `p.Err` equates to) for this informational/non-error debug message?

pkg/mixin/runner.go
```diff
@@ -50,10 +50,10 @@ func (r *Runner) Validate() error {
 
 func (r *Runner) Run() error {
 	if r.Debug {
-		fmt.Fprintf(r.Out, "DEBUG mixin:    %s\n", r.Mixin)
-		fmt.Fprintf(r.Out, "DEBUG mixinDir: %s\n", r.mixinDir)
-		fmt.Fprintf(r.Out, "DEBUG file:     %s\n", r.File)
-		fmt.Fprintf(r.Out, "DEBUG stdin:\n%s\n", r.Step)
+		fmt.Fprintf(r.Err, "DEBUG mixin:    %s\n", r.Mixin)
```
vdice: Similar question as posed above; wondering if informational (non-error) debug messages should in fact be sent to stderr?

pkg/porter/schema.go
```diff
@@ -0,0 +1,111 @@
...
+	if !ok {
+		return nil, errors.Errorf("root porter manifest schema has invalid properties.mixins.items type, expected map[string]interface{} but got %T", mixinSchema["items"])
+	}
+
+	enumSchema, ok := itemsSchema["enum"].([]interface{})
+	if !ok {
+		return nil, errors.Errorf("root porter manifest schema has invalid properties.mixins.items.enum type, expected []interface{} but got %T", itemsSchema["enum"])
+	}
+
+	installSchema, ok := propertiesSchema["install"].(map[string]interface{})
```
vdice: Do we also need `upgradeSchema` and `uninstallSchema`?

I ran `porter schema` off of this branch and see: https://gist.github.com/vdice/114bec04b6bfbf7dabf14027372e591a

Or, to put it another way, I currently don't see the other actions of `upgrade` and `uninstall` represented in the generated schema for mixins that currently implement both (`exec` and `helm`).

pkg/config/actions.go
```diff
@@ -1,21 +1,38 @@
...
 	switch action {
 	case ActionInstall, ActionUpgrade, ActionUninstall:
 		return action, nil
 	default:
-		return "", fmt.Errorf("invalid action %q", value)
+		return "", fmt.Errorf("%s %q", ErrInvalidAction, value)
 	}
 }
+
+// IsInvalidActionError determines if an error is the error returned by ParseAction when
```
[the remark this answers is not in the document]
carolynvs: I added this in case it would be useful later. do you think I should take it out since it's not used now?

cmd/porter/run.go
jeremyrickard: Agree, but this can get fixed in a followup AFAIAMC

carolynvs: Yeah, this wasn't something I originally intended to do but then the schema command kind of forced me to go down this route. Let me share how it came about and maybe you have ideas on if it was a good idea or if there is another way to deal with it?

When you run `porter schema --debug` because a mixin is acting up or something isn't right, if everything goes to stdout, the debug output is intermixed with the manifest json schema. Then it's impossible to get a clean schema out. By putting the debug statements into stderr, you can do this

`porter schema --debug 2> err.log`

so that you can see the manifest and the debug lines and manage them separately

Thoughts? OK to keep it like this for now? Better way to deal?

pkg/porter/schema.go
carolynvs: The plan was to let Ivan get his hands on this, try it out for install, and if it works, then finish the other two in a follow-up.

carolynvs:
Merge like this PR is the albatross that ruined the month of February. 🐦

carolynvs:
I've created #199 and #200 to track the follow-up work from this PR.

cmd/porter/run.go
vdice: Ah, I see.  This reasoning makes sense to me!  Maybe we can add that example to docs around schema/debugging.

carolynvs: How about in the contributing guide? I can add a note about how to do logging there.

vdice: Good idea 👍

PR #154, opened 2019-02-08
Initial refactor to support packr to bundle templates

### LINKED ISSUE #118: Use Packr to put the templates into the porter binary
Currently, Porter expects the templates/ directory to exist at a path relative to the binary or in PORTER_HOME. It would simplify things, including as new versions of Porter are released, if we used something like [packr v2](https://github.com/gobuffalo/packr/tree/master/v2) to bundle the templates into the binary. The ARM templates for the porter-azure mixin are handled this way for reference.

jeremyrickard:
This PR removes the templates directory and introduces a dependency on Packr2 in order to bundle those things into the Porter binary directly. 

Previously, `porter create` would read those files from the PORTER_HOME and put them into the bundle. Now they are easy bake oven'ed into pkg/config and are fetched that way, rather than reading the file system.

The PR updates the existing tests that read those and write into FS for things like build tests. Now they are fetched from the config and written to the FS via a new test helper method.

A Test as also added to config to validate that what we get from the generated box matches what's in pkg/config/templates directory. 

Fixes #118


pkg/config/config.go
```diff
@@ -52,36 +53,16 @@ func (c *Config) GetHomeDir() (string, error) {
...
-// GetTemplatesDir determines the path to the templates directory.
-func (c *Config) GetTemplatesDir() (string, error) {
-	home, err := c.GetHomeDir()
-	if err != nil {
-		return "", err
-	}
-	return filepath.Join(home, "templates"), nil
-}
-
 // GetPorterConfigTemplate reads templates/porter.yaml from the porter home directory.
```
vdice: Godoc/function comments here and elsewhere could use updates (no longer fetching from the porter home directory)

carolynvs:
If you incorporate the changes from my PR, the build will pass (or at least I think that's what is missing)

https://github.com/deislabs/porter/pull/146/files#diff-836546cc53507f6b2d581088903b1785

https://github.com/deislabs/porter/pull/146/files#diff-b67911656ef5d18c4ae36cb6741b7965

Makefile
```diff
@@ -35,12 +35,10 @@ build-client:
 	$(MAKE) build-client MIXIN=porter -f mixin.mk
 	$(MAKE) build-client MIXIN=exec -f mixin.mk
 	cp bin/mixins/porter/porter$(FILE_EXT) bin/
-	cp -R templates bin/
```
carolynvs: I am so happy to see these go! 🎉

cmd/porter/run_test.go
```diff
@@ -15,12 +15,14 @@ func TestRun_Validate(t *testing.T) {
 
 	p := porter.NewTestPorter(t)
 	p.TestConfig.SetupPorterHome()
-	p.TestConfig.TestContext.AddTestFile("../../templates/porter.yaml", config.Name)
+	configTpl, err := p.TestConfig.GetPorterConfigTemplate()
+	require.Nil(t, err)
```
carolynvs: just a FYI, @vdice taught me about require.NoError and it's nice. When there is an error, it will print out the error contents. No need to change what you have here, just sharing what I learned. 😀

pkg/config/config.go
```diff
@@ -52,36 +53,16 @@ func (c *Config) GetHomeDir() (string, error) {
...
-		return nil, err
-	}
-
-	tmplPath := filepath.Join(tmplDir, Name)
-	return c.FileSystem.ReadFile(tmplPath)
+	t := packr.New("templates", "./templates")
+	return t.Find(Name)
 }
 
 // GetRunScriptTemplate reads templates/run from the porter home directory.
```
carolynvs: Same as Vaughn's comment above, the godoc string should be updated too.

jeremyrickard:
Hey @carolynvs and @vdice take a look. I think I got all the doc changes.

PR #156, opened 2019-02-11
Add doc page on dependencies

### LINKED ISSUE #148: Document how porter merges dependency manifests into the root manifest
Here are some notes on how the merge works, but more is needed. Here's the code in question that all the doc is explaining

https://github.com/deislabs/porter/blob/05c04ad319a8595d2464b007dc49d1ee74cdfc7f/pkg/config/manifest.go#L284

* When specifying a dependency upon another bundle, a bundle can override the default value of a parameter defined by the dependency.
* Porter merges parameters and credentials from dependencies into the parameters and credentials of the "root" bundle. When those parameters are required, then duffle (or whatever tool is handling the runtime) will enforce that it is supplied.
* When multiple bundles define a parameter inconsistently, `porter build` reports the inconsistency and fails. We don't currently have defined how to resolve these situations.
* Porter handles merging dependencies and their steps in the proper order based on the action so that during install, dependencies are executed first. For example with uninstall, the steps are executed in reverse order (so that the app is uninstalled before the database that it depends upon).

carolynvs:
Closes #148 

This will evolve as we work on it further but it documents how it works today.

Preview available at: https://deploy-preview-156--porter.netlify.com/dependencies/


docs/content/dependencies.md
```diff
@@ -0,0 +1,75 @@
...
+
+* Install - The dependency's steps are executed _before_ the steps in the root bundle.
+* Upgrade - The dependency's steps are executed _before_ the steps in the root bundle.
+* Uninstall - The dependency's steps are executed _after_ the steps in the root bundle.
+
+## Dependencies and the CNAB Spec
+
+The CNAB Spec that Porter implements does not define how dependencies between
+bundles works (or that they can even exist). Porter has a trick for getting
+around that: Porter identifies dependencies between bundles and then builds a
```
glyn: The fact that porter takes Porter-authored bundles (with a Porter manifest) and combines them to produce a single CNAB bundle (presumably also with a Porter manifest) is crucial to understanding Porter. I think it deserves spelling out clearly somewhere near the start of the documentation.

docs/content/dependencies.md
```diff
@@ -0,0 +1,75 @@
...
+single composite bundle during the `porter build` command. At runtime there is a
+single invocation image, so that any tool can work with bundles created by
```
glyn: Although the composite bundle has a single invocation image, I got the impression that this delegates actions to the dependent bundles' invocation images. So it's a little confusing to say that there is a single invocation image at runtime because that gives the impression that the invocation images may have been somehow smashed together, which I doubt is particularly feasible.

docs/content/dependencies.md
```diff
@@ -0,0 +1,75 @@
...
+will be revisited.
+
+* bundles (in the same directory as the Porter manifest)
+* PORTER_HOME/bundles
+
+Parameters defined in a dependent bundle can be defaulted from the root bundle.
+In the example below, the mysql bundle defined the `database_name` and
+`mysql_user` parameters, and the root bundle (Wordpress) decided to default them
+to a specific value, so that the user wouldn't be required to select a value for
+those parameters when installing Wordpress.
```
glyn: Is it true to say that parameters defined in dependent bundles which are not defaulted from the root bundle end up being parameters of the resultant composite bundle?

What about parameters defined in dependent bundles which are defaulted from the root bundle? Do these end up being parameters of the resultant composite bundle too, but with default values?

glyn: It may be worth moving this paragraph into the "Consolidating Parameters and Credentials" section since that section already answers my questions.

docs/content/dependencies.md
```diff
@@ -0,0 +1,75 @@
...
+```
+
+## Consolidating Parameters and Credentials
+
+Porter consolidates the parameters and credentials from the dependencies and
+promotes them to the root bundle. This allows the user executing the bundle to
+interact with the parameter or credential from a dependency just as they would
+from one defined on the root bundle.
+
+When a parameter or credential has the same type and name, then it is
```
glyn: This paragraph should explain which parameter or credential is being talked about. Presumably it is one which is defined in at least two bundles out of the root bundle and its dependent bundles. With this stated, the paragraph could then be clarified to talk about a parameter having the same type and name (or not) in all the places where it is defined.

docs/content/dependencies.md
```diff
@@ -0,0 +1,75 @@
...
+consolidated. When the same name is used, but not the same type then the bundle
+build fails.
+
+_Note: Porter doesn't yet have a conflict resolution mechanism in place to
+allow an author to force consolidating parameters/credentials, or resolve a
+conflict._
+
+## Dependency Graph
+
+Currently Porter does not build a dependency graph and only supports direct
```
glyn: It does build a dependency graph, but the graph just consists of a root node and its immediate dependents. :-)

docs/content/dependencies.md
```diff
@@ -0,0 +1,75 @@
...
+dependencies. Dependencies of dependencies, transitive dependencies, are
+ignored. See [Design: Dependency Graph Resolution](https://github.com/deislabs/porter/issues/69) for more information.
+
+## Dependency Ordering
+
+Depending on the action being performed, Porter handles executing the dependencies
```
glyn: "dependencies'"?

docs/content/dependencies.md
```diff
@@ -0,0 +1,75 @@
...
+steps either before or after the steps in the root bundle.
```
glyn: What's the difference between a step and an action? If they are the same thing, it would be better to use the same term throughout.

docs/content/dependencies.md
```diff
@@ -0,0 +1,75 @@
...
+
+* Install - The dependency's steps are executed _before_ the steps in the root bundle.
+* Upgrade - The dependency's steps are executed _before_ the steps in the root bundle.
+* Uninstall - The dependency's steps are executed _after_ the steps in the root bundle.
```
glyn: What's the ordering for custom actions?

docs/content/dependencies.md
carolynvs: This is one reason why the dependent bundles must be authored by porter. Porter does not execute invocation images from those bundles, because there is no way (currently) to extract information from them after they are run.

Porter reads in the porter manifest from those dependencies, merges the manifests, and creates a single composite bundle. So there really is just a single invocation image.

docs/content/dependencies.md
carolynvs: All parameters defined in the dependencies are merged into the final bundle. I'll reword and maybe move around a bit so that it's introduced in the proper order.

docs/content/dependencies.md
carolynvs: They are different.

* Action = CNAB action: install, upgrade and uninstall
* Step = One of the entries listed under the action in the yaml in the porter manifest

```yaml
install:
- description: "Say Hello"
  exec:
    command: bash
    arguments:
      - -c
      - echo Hello World
```

In the example above the install action has a single step.

```

carolynvs: We need to document the manifest, and when we do, that will be explained... :-( Sorry there are so many doc TODOs!

docs/content/dependencies.md
carolynvs: Porter doesn't support undefined actions. So the actions you see listed are the ones supported right now.

Issue #158, opened 2019-02-12
Can't pass source variables to the exec arguments list


carolynvs:
This was first implemented in https://github.com/deislabs/porter/issues/31 but doesn't seem to work anymore.

When you run this, replace TODO with your username.

```yaml
mixins:
  - exec

name: HELLO
version: 0.1.0
description: "An example Porter configuration"
invocationImage: TODO/porter-hello:latest

parameters:
  - name: cmd
    type: string
    default: "echo hello world"

install:
  - description: "Install Hello World"
    exec:
      command: bash
      arguments:
        - -c
        - source: bundle.parameters.cmd

uninstall:
  - description: "Uninstall Hello World"
    exec:
      command: bash
      arguments:
        - -c
        - echo Goodbye World
```

```console
$ porter build && duffle install FOO -f bundle.json --insecure
Executing install action...
executing porter install configuration from /cnab/app/porter.yaml
Install Hello World
Error: yaml: unmarshal errors:
  line 6: cannot unmarshal !!map into string
err: yaml: unmarshal errors:
  line 6: cannot unmarshal !!map into string
```

nunix:
just for information, I got the exact same issue by running with porter only:
```bash
porter run --action install --file ./porter.yaml --debug
```

really looking forward to this fix :)

jeremyrickard:
This turns out not to be a problem with the way the parameter sourcing is working, it's a "how go is parsing YAML" problem. You _can_ pass the params, with slightly different YAML right now. 

The yaml above fails (here it is again):

```
install:
  - description: "Install Hello World"
    exec:
      command: bash
      arguments:
        - -c
        - source: bundle.parameters.cmd
```

However.....this does work

```
install:
  - description: "Install Hello World"
    exec:
      command: bash
      arguments:
        - -c
        - "source: bundle.parameters.cmd"
```

When the YAML blob gets loaded to process, it's failing to handle that correctly. Working on a fix now.

nunix:
I tried the workaround and the values are empty (still based on my yaml file). Which I think it's because I do directly the command `porter run --action install --file ./porter.yaml --debug` instead of piping it like Carolyn did in her example.

My porter version is:
![image](https://user-images.githubusercontent.com/905874/52700487-48794980-2f78-11e9-8407-5b5ebbcc06ae.png)

carolynvs:
I think I know what you are running into!

Short answer: 

[...]

PR #162, opened 2019-02-13
Enhances how we're handling sourcing

jeremyrickard:
This PR updates how we handle slice elements for doing parameter/credential/ouput sourcing.
We probably should think about revisiting the entire process for this in the future.

In summary, we are using reflectwalk to handle walking the manifest. This involves handling a couple
call backs, notably "SliceElement" and "MapElement". The Slice Element assumed that it would see something
of the form source: <x.y.z> as a string. In reality, if that's not wrapped in quote, Go and YAML see that
as a Map, but the map callback doesn't get called to handle that. This was resulting in the block not getting
rewritten and the exec mixin was exploding (probably any would have). This PR updates the logic of the slice elem processing to check and see if it's a map and handle it in a similar manner to a top level map entry in the yaml.

Closes: #158

### DIFF
--- a/pkg/config/manifest.go
+++ b/pkg/config/manifest.go
@@ -597,16 +597,13 @@ func (m *Manifest) SliceElem(index int, val reflect.Value) error {
 						value := kv.String()
 						replacement, err := m.resolveValue(value)
 						if err != nil {
-							errors.Wrap(err, "unable to resolve value for map")
+							errors.Wrap(err, fmt.Sprintf("unable to resolve value for key: %s", value))
 						}
 						v.Set(reflect.ValueOf(replacement))
 					}
 				}
 			}
-		} else {
-			fmt.Println("not sure what this is...")
-			fmt.Printf("eh? %s", reflect.TypeOf(val).String())
-		}
+		} // it's not a map, we shouldn't try and process it.
 
 	}
 	return nil
--- a/pkg/config/manifest.go
+++ b/pkg/config/manifest.go
@@ -546,6 +546,24 @@ func (m *Manifest) SliceElem(index int, val reflect.Value) error {
 		return nil
 	}
 
+	// There are two cases possible in the YAML. The first is when the element in the slice is a string. That might look like:
+	// install:
+	//   - description: "Install Hello World"
+	//     exec:
+	// 	     command: bash
+	// 	     arguments:
+	// 	       - -c
+	// 	       - "source:  bundle.parameters.command"
+	// The second case is when the declaration interpreted to represent a Map. This is the case when there are no quotes:
+	// install:
+	//   - description: "Install Hello World"
+	//     exec:
+	// 	     command: bash
+	// 	     arguments:
+	// 	       - -c
+	// 	       - source:  bundle.parameters.command
+	// Branching logic below first checks the easy case (a string) then falls through into the case where we have a map with a single key.
+	// This is fairly similar to how we process map elements, but the replacement is done differently.
 	v, ok := val.Interface().(string)
 	if ok {
 		//if the array entry is a string that matches source:...., we should replace it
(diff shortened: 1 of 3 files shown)


pkg/config/manifest.go
```diff
@@ -546,6 +546,24 @@ func (m *Manifest) SliceElem(index int, val reflect.Value) error {
...
 	}
 
+	// There are two cases possible in the YAML. The first is when the element in the slice is a string. That might look like:
+	// install:
+	//   - description: "Install Hello World"
+	//     exec:
+	// 	     command: bash
+	// 	     arguments:
+	// 	       - -c
+	// 	       - "source:  bundle.parameters.command"
```
carolynvs: I don't want to make this harder, because I agree that we need to rethink how we do sourcing anyway. But when I look at the two here,

- "source: bundle.parameters.command"
- source: bundle.parameters.command

My reaction is that the first one shouldn't even be parsed by porter and should be ignored really. It's only the second that matches how source works everywhere else.

Does that simplify this PR? If so, can we nix the first case "source as a string"? If it makes it harder, then leave it as it. 😀

pkg/config/manifest.go
```diff
@@ -559,6 +577,34 @@ func (m *Manifest) SliceElem(index int, val reflect.Value) error {
...
+		if val.Kind() == reflect.Interface {
+			val = val.Elem()
+		}
+		if kind := val.Kind(); kind == reflect.Map {
+			if len(val.MapKeys()) == 1 {
+				sk := val.MapKeys()[0]
+				if sk.Kind() == reflect.Interface {
+					sk = sk.Elem()
+				}
+				//if the key is a string, and the string is source, then we should try
```
carolynvs: Like if we stop supporting `- "source: bundle...."` does this entire block go away?

pkg/config/testdata/slice-test.yaml
```diff
@@ -0,0 +1,29 @@
...
+    type: string
+    default: "echo Hello World"
+
+install:
+  - description: "Install Hello World"
+    exec:
+      command: bash
+      arguments:
+        - -c
+        - source:  bundle.parameters.command
```
carolynvs: This looks natural to me, based on how we do source elsewhere.

Ok I've said it 3 ways. I'll stop now. 😇

pkg/config/manifest.go
jeremyrickard: No, the opposite. This got added to support the other case where there are no quotes. 

This is the part that handles the quotes:

```
v, ok := val.Interface().(string)
	if ok {
		//if the array entry is a string that matches source:...., we should replace it
		re := regexp.MustCompile("source:\\s?(.*)")
		matches := re.FindStringSubmatch(v)
		if len(matches) > 0 {
			source := matches[1]
			r, err := m.resolveValue(source)
			if err != nil {
				return errors.Wrap(err, "unable to source value")
			}
			val.Set(reflect.ValueOf(r))
		}
	}
```

carolynvs: Oh noes, I've been found out. I have a really hard time reading this reflection code.

I'm ok merging this as-is. Up to you if you think we can just remove the "source: bundle...." case and make things simpler or not.

pkg/config/testdata/slice-test.yaml
jeremyrickard: Yeah, to support this form, the change i added is needed. Both this and "source:...." are valid YAML, so I think we should support both for right now.

We should probably fix this in general though..but also:

```yaml
arguments: 
   - asdadadad
   - {bundle.params.blah}
```

will probably have similar stupid YAML processing outcomes for example, that block above gets turned into:

```yaml 
args: 
  - blah
  - 
    bundle.parameters.blah: ~
```

by a validator, which is even worse. I think in that case it might require quotes to make things simpler.

carolynvs: rawr 🦁 Ok I see why this is hard.

I say ship it and then reasses later.

PR #164, opened 2019-02-14
Document porter manifest

carolynvs:
Huh, I guess we didn't have an issue for documenting the manifest? 🤷‍♀️ 

Preview at https://deploy-preview-164--porter.netlify.com/authoring-bundles/

Closes #34

PR #168, opened 2019-02-14
Introduce a document describing Param and credential wiring

jeremyrickard:
WIP document showing how to use parameter, credentials and outputs in bundles. Also covers how to wire things up.


jeremyrickard:
The preview is available here: https://deploy-preview-168--porter.netlify.com/wiring/

docs/content/wiring.md
```diff
@@ -0,0 +1,321 @@
...
+
+You can also provide any other attributes, as specified by the CNAB [parameters](https://github.com/deislabs/cnab-spec/blob/master/101-bundle-json.md#parameters) specification. To specify a default value, for example, you could provide the following parameter definition:
+
+```yaml
+- name: database_name
+  type: string
+  default: "wordpress"
+```
+
+This would result in the following `bundle.json` parameter definition:
```
carolynvs: Is it useful to spell out in the doc here how it gets turned into parameter in bundle.json? We don't do that for anything else.

docs/content/wiring.md
```diff
@@ -0,0 +1,321 @@
...
+credentials:
+- name: SUBSCRIPTION_ID
+  env: AZURE_SUBSCRIPTION_ID
+```
+
+## Wiring Credentials
+
+The same mechanism for declaring how to use a parameter can be used for credentials. To declare a credential usage, references are defined with the following syntax: `source: bundle.credentials.CREDENTIAL_NAME`.
+
+When the bundle is executed, the Porter runtime will locate the parameter definition in the `porter.yaml` to determine where the parameter value has been stored. The Porter runtime will then rewrite the YAML block before it is passed to the mixin.
```
carolynvs: It feels like this is more information that is relevant to who is reading it (people who want to know how to wrire up a bundle). Maybe it belongs more in the runtime architecture?

Just worried that it's a bit overwhelming in addition to everything else. But it's okay to override me if you have strong feels.

docs/content/wiring.md
```diff
@@ -0,0 +1,321 @@
...
+
+## Outputs
+
+In addition to parameters and credentials, Porter allows you to declare how to store mixin outputs. The form used to declare these varies by mixin, but might look something like:
```
carolynvs: We should explain what an output is.

docs/content/wiring.md
```diff
@@ -0,0 +1,321 @@
...
+        databaseName:
+          source: bundle.parameters.database_name
+    outputs:
+      - name: "MYSQL_URL"
+        key: "MYSQL_HOST"
+```
+
+In this example, a new output will be created named `MYSQL_URL`. The Azure mixin allows you to specity the key to fetch the output from, in this case it is `MYSQL_HOST`. Each mixin can provide different ways of addressing outputs, so refer to the schema for each mixin. The Porter runtime will keep a map in memory with each of the outputs declared.
+
+TODO: What happens if someone overwrites one? Should we fail the `porter build`?
```
carolynvs: Let's change this TODO into a design issue.

docs/content/wiring.md
```diff
@@ -0,0 +1,321 @@
+---
+title: Using Parameters, Credentials and Outputs
+description: How to wire parameters, credentials and outputs into steps
+---
+
+# Parameters, Credentials and Outputs in Porter
+
+In the Porter manifest, you can declare both [parameters](https://github.com/deislabs/cnab-spec/blob/master/101-bundle-json.md#parameters) and [credentials](https://github.com/deislabs/cnab-spec/blob/master/101-bundle-json.md#credentials), which are defined in the CNAB spec. The CNAB specification specifies how both credentials and parameters can be provided to an [invocation image](https://github.com/deislabs/cnab-spec/blob/master/102-invocation-image.md) and how they should be declared in a bundle manifest.
```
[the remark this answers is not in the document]
jeremyrickard: #164 got merged so I'll update. 

In general, are you in favor of not surfacing CNAB? I have mixed feelings on that.

docs/content/wiring.md
jeremyrickard: 🌮 

☝️  that's my new thumbs up.

docs/content/wiring.md
jeremyrickard: Removed, as we fill out that other doc we can incorporate how things are handled at runtime.

docs/content/wiring.md
jeremyrickard: Removed.

docs/content/wiring.md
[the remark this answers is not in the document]
jeremyrickard: I removed most of this. I don't think it's overly confusing as I would guess that most (but obviously not all) people looking at Porter are doing it in the context of building CNAB bundles. But we aren't really consistent so I've removed most of this (i left a later reference to the parameter definition in the spec)

PR #172, opened 2019-02-18
Use Hugo Asset Pipeline

carolynvs:
I originally copied the site over from helm.sh, and didn't figure out how to copy the gulp asset pipeline (which handled converting scss to css). This finally uses the Hugo pipeline so that we don't need a separate process or set of tools to deal with the conversion.

Preview at https://deploy-preview-172--porter.netlify.com/

Basically the images and css on the homepage work so you know I didn't mess it up. 😀 

A few things to keep in mind:

* After modifying the scss files, do a `make docs` and check in the resources directory. Netlify relies on that for reasons I'm not 100% sure about.
* Images in the assets directory (symlinked to the theme's src directory) are not automatically available from a static URL (e.g. /img). So I updated the scss to reference absolute paths. Seems to work, there is probably a better way, so suggestions and improvements welcome.
* A huge chunk of this PR, is just adding in the missing foundation scss files, so ignore that commit when reviewing.


flynnduism:
Checked this out locally and all shiny and working.

I tested editing the `scss` and rebuilding the site, but can't get them to compile (even after `make docs` and re-running things). Not sure why!

Not a blocker for this, merge ahoy!

carolynvs:
Which version of hugo are you using and how did you install it? I'm using 

```
$ hugo version
Hugo Static Site Generator v0.53/extended darwin/amd64 BuildDate: unknown
```

Maybe I'm cheating?

flynnduism:
@carolynvs using Hugo v0.54.0 (I upgraded to test this out)

Hugo build stuff all worked fine, no errors. I just couldn't figure out how to get edits in the SCSS to apply to the site. Gonna dig into the Pipelines stuff more today, I'm probably just doing something dumb

carolynvs:
I think the extended version of hugo may be important? I got that by installing from source vs homebrew. That let's me check in the resources directory.

The other part is the pipelines part

```
 <!-- the sass directory is relative to the assets directory under the theme, I cheated here and did a symlink to the existing src directory -->
  {{ $style := resources.Get "sass/styles.scss" | resources.ToCSS | resources.Minify }}
  <link rel="stylesheet" href="{{ $style.Permalink }}">
```

PR #178, opened 2019-02-20
Prints generated Dockerfile lines nicely

youreddy:
Resolves https://github.com/deislabs/porter/issues/111
/cc @ryanmoran

### DIFF
--- a/pkg/porter/build_test.go
+++ b/pkg/porter/build_test.go
@@ -1,7 +1,6 @@
 package porter
 
 import (
-	"bytes"
 	"testing"
 
 	"encoding/json"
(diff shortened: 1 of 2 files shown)


pkg/porter/build_test.go
```diff
@@ -39,6 +40,36 @@ func TestPorter_buildDockerfile(t *testing.T) {
 	assert.Equal(t, wantlines, gotlines)
 }
 
+func TestPorter_buildDockerfile_output(t *testing.T) {
+	buffer := bytes.NewBuffer([]byte{})
```
carolynvs: I would like to show you a special function that I built into TestPorter that assists with collecting output during a test.

You can remove lines 44 and 46. Then on line 70, change `buffer.String` to `p.TestConfig.TestContext.GetOutput()`. 

The TestPorter struct does the same thing that you did in the test (make a byte buffer) and then exposes it as a string with `GetOputput()`.

carolynvs:
I'm excited to have you contributing to porter! Please let me know if you are interested in more issues, and I can help you find another. 😀

pkg/porter/build_test.go
youreddy: That makes sense to me! Just updated the branch to reflect those changes.

youreddy:
> I'm excited to have you contributing to porter! Please let me know if you are interested in more issues, and I can help you find another. 😀

@carolynvs thank you for being so welcoming. We're excited to be contributing in this space.

PR #187, opened 2019-02-26
Accept urls and file paths for porter run

youreddy:
Resolves https://github.com/deislabs/porter/issues/174


pkg/config/manifest.go
```diff
@@ -151,6 +157,38 @@ func (c *Config) ReadManifest(path string) (*Manifest, error) {
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "could not read from url %s", path)
+	}
+
+	m := &Manifest{}
+	err = yaml.Unmarshal(body, m)
+	if err != nil {
+		return nil, errors.Wrapf(err, "could not parse manifest yaml in %q", path)
+	}
+	m.path = path
```
carolynvs: Hmm, we will need to think a bit more about how to handle this variable right here.

The `path` variable holds the location of the manifest on the local filesystem. Porter uses it to resolve references to dependencies. For example, if the bundle depended upon another bundle named "mysql", one place Porter would look for the mysql bundle would be in the same directory as the bundle on disk, in a directory named "bundles"

```
- wordpress/
--- bundles/
---- porter.yaml
-- porter.yaml
```

Once we have registries, and porter uses the local store, this won't be a problem. But until then, we should at least set this to a value that doesn't cause an error. 

Perhaps we can set it to "", and then add a check in config.go (GetBundleDir) to skip looking for the local bundle dir when it's empty?

What do you think?

youreddy: @carolynvs just revisited the code and saw what you're describing. We could continue setting`path` to the url and have `GetBundleDir` skip the lookup if it sees that `path` resembles a url. The benefit of doing that might be a little bit more clarity on why the skip is happening. The snippet of code could look something like:
`if !strings.HasPrefix(localDir, "http"){ ... }`

What do you think?

carolynvs: Good idea, I'm fine with that solution. 👍

youreddy: Awesome! Just updated the PR branch.

carolynvs:
Looks good, thank you for this improvement! It will make it much easier for people to try out examples.

Issue #193, opened 2019-02-27
Fix how we use packr in our builds


carolynvs:
Options:

* Use the go generate pattern https://github.com/gobuffalo/packr/issues/48#issuecomment-439528246 and then use go generate + go build in our makefile instead of calling packr directly
* Replace go build with `packr2 build`. It's not clear if that results in requiring a `packr2 clean` afterwards.
* Run `packr2` + `packr2 clean`. It's not clear if this automatically traverses the entire directory structure or if we need to call it repeatedly for subpackages.

Goals
* Do not check-in `*-packr.go` files as they recommend.
* Avoid lots people needing to know how to use packr even more.
* We are getting lots of merge conflicts and they suck. Not checking in those files should fix that, but also doing a clean will fix tools complaining about them being overwritten and whatnot.
* World peace

PR #194, opened 2019-02-27
Do not check in packr files

carolynvs:
This is part of #193 (split off from #146). Packr recommends to not check in any `*-packr.go` files because they are generated and needed just during the build. Either way they are causing me lots of merge conflicts to boot in my mega PR of doom. So I'm hoping we can merge this soon.


carolynvs:
Merge like it's a small change, no one is around and multiple PRs build on top of this. 😇

PR #196, opened 2019-02-27
Help people find good first issues

carolynvs:
To make life easier for people, good first issue and help wanted no longer overlap. I've explained how they work in the contributing guide and link to a search against our repo for them. They are filtered by backlog so that they don't pick up issues that people are working on already.

I was going to link to a waffle query, but there's a bug (or misunderstanding) that I'm working with them on. Waffle would be better because it can do a cross repo query more easily.


carolynvs:
If you join the Good First Issue Brigade... 👀

PR #201, opened 2019-03-01
Verify vendor directory during CI

### LINKED ISSUE #155: CI: Fail build if `dep ensure` finds changes
Proposing we add a check in our CI pipeline to run `dep ensure` and verify no new changes come in.  If they do, the build will be failed and it should be apparent to the dev that they need to update vendor accordingly.

carolynvs:
This works but Azure Pipelines does odd things with the git directory. I'll wait until after we move to brigade to merge it.

---

This checks that all dependencies that should be in the vendor/ directory are present.

When you import something new in your code, and don't check it into vendor, or change something in Gopkg.toml/lock and don't update the vendored code, you will see something like this:

```console
$ make verify
dep check
# Gopkg.lock is out of sync:
github.com/deislabs/duffle/pkg/credentials: imported or required, but missing from Gopkg.lock's input-imports
github.com/deislabs/duffle/pkg/duffle/home: imported or required, but missing from Gopkg.lock's input-imports
github.com/deislabs/duffle/pkg/utils/crud: imported or required, but missing from Gopkg.lock's input-imports
github.com/gobuffalo/packr/v2/file/resolver: in Gopkg.lock's input-imports, but neither imported nor required

Makefile:96: recipe for target 'verify-vendor' failed
```

Closes #155


vdice:
@carolynvs we should be ready to add this to our Brigade pipeline (`brigade.js`) when convenient.  Perhaps in the [build](https://github.com/deislabs/porter/blob/master/brigade.js#L47-L49) or [unit test](https://github.com/deislabs/porter/blob/master/brigade.js#L67-L69) job?

carolynvs:
Oh yeah! I even found a better way to do it. I'll update this today.

PR #210, opened 2019-03-06
ref(ci): add brigade

vdice:
Transitions CI to use a [Brigade](https://github.com/Azure/brigade)-based pipeline.

Implements https://github.com/deislabs/porter/issues/176


vdice:
Ok, this is in a state for review.  Tested via a PR to my fork for GitHub interactions/setup of corresponding Brigade project.  See https://github.com/vdice/porter/pull/1

Some `TODO`/etc. comments remain; hoping to spur conversation from reviewers.

Afterwards, and assuming we're good at a code level, plan on working with @carolynvs to hook our main deislabs GH App up to this repo and configure the corresponding Brigade project.

Makefile
```diff
@@ -1,10 +1,13 @@
 SHELL = bash
 
+# --no-print-directory avoids verbose logging when invoking targets that utilize sub-makes
+MAKE_OPTS ?= --no-print-directory
```
carolynvs: This doesn't seem to work on my machine. I don't get different output when running make build. Is this for a particular make command?

Makefile
```diff
@@ -48,17 +51,25 @@ ifndef HAS_PACKR2
 	go get -u github.com/gobuffalo/packr/v2/packr2
 endif
 
+HAS_DEP := $(shell command -v dep)
+dep:
+ifndef HAS_DEP
+	go get -u github.com/golang/dep/cmd/dep
```
carolynvs: We should use a different command to install dep

```
curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
```

This gets the latest tagged release so that dep version works, and not the tip of master.

vdice: For now, we've decided to remove fetching `dep` (This may change and/or be revisited in https://github.com/deislabs/porter/pull/201)

Makefile
vdice: This was added to address verbose output from CI agents.  Indeed, my local version doesn't see any change with this added, either.

brigade.js
```diff
@@ -0,0 +1,237 @@
...
+  constructor (name) {
+    super(name);
+
+    const gopath = "/go";
+    const localPath = gopath + `/src/github.com/${projectOrg}/${projectName}`;
+
+    // Here using the large-but-useful deis/go-dev image as we have a need for deps
+    // already pre-installed in this image, e.g. helm, az, docker, etc.
+    // TODO: replace with lighter-weight image (Carolyn)
+    this.image = "deis/go-dev";
```
vdice: @carolynvs did you find a better fit for a base image for our job pods here?

carolynvs: Is it okay if we merge as-is and I'll make a slimmer image this week? I'd like to see this shiny new brigade job 🚢 without waiting for me to do hard work first.

vdice: Definitely!  Let's do it.  We can improve iteratively.

PR #212, opened 2019-03-07
Include all supported actions in the schema

### LINKED ISSUE #199: Include remaining actions in porter schema command
After we've tried out #146 with VS Code, and have worked out any kinks, lets finish it with upgrade and uninstall.

carolynvs:
Adds in support for the other actions: upgrade and uninstall.

I am validating the schema in the unit test. Next step after this is to do this validation in porter when we read in the porter.yaml file.

TODO:
- [x] Pass the wrapping action to the mixin
- [x] Change schema merge to look for actions in "properties.ACTION"
- [x] Merge mixin schema and fixup relative refs (since vs code doesn't support ids properly) 

Closes #199

### DIFF
--- a/pkg/exec/schema/exec.json
+++ b/pkg/exec/schema/exec.json
@@ -1,7 +1,73 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
   "definitions": {
-    "execStep": {
+    "installStep": {
+      "type": "object",
+      "properties": {
+        "exec": {
+          "type": "object",
+          "properties": {
+            "description": {
+              "type": "string",
+              "minLength": 1
+            },
+            "command": {
+              "type": "string"
+            },
+            "arguments": {
+              "type": "array",
+              "items": {
+                "type": "string",
+                "minItems": 1
+              }
+            }
+          },
+          "additionalProperties": false,
+          "required": [
+            "description",
+            "command"
+          ]
+        }
+      },
+      "additionalProperties": false,
+      "required": [
+        "exec"
+      ]
+    },
+    "upgradeStep": {
+      "type": "object",
+      "properties": {
+        "exec": {
+          "type": "object",
+          "properties": {
+            "description": {
+              "type": "string",
+              "minLength": 1
+            },
+            "command": {
+              "type": "string"
+            },
+            "arguments": {
+              "type": "array",
+              "items": {
+                "type": "string",
+                "minItems": 1
+              }
+            }
+          },
+          "additionalProperties": false,
+          "required": [
+            "description",
+            "command"
+          ]
+        }
+      },
+      "additionalProperties": false,
+      "required": [
+        "exec"
+      ]
+    },
+    "uninstallStep": {
       "type": "object",
       "properties": {
         "exec": {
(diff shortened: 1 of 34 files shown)


pkg/porter/schema.go
```diff
@@ -5,10 +5,12 @@ import (
 	"fmt"
 
 	"github.com/deislabs/porter/pkg/config"
-
 	"github.com/pkg/errors"
 )
 
+type jsonSchema = map[string]interface{}
```
carolynvs: I am abusing type aliases just to make my tests easier to write/read. Am I a bad person? I can remove if this seems like a bad idea. I just got tired of writing out map[string]interface a hundred times. 😀

carolynvs:
OK, I've worked with Ivan and come up with something that works with VS Code. The plan is to not use ids. The mixins use relative references, and then porter rewrites them when it merges in their schema.

pkg/porter/schema.go
vdice: I'm down w/ this; to me it definitely makes sense for comprehension.

pkg/porter/schema.go
```diff
@@ -86,24 +84,39 @@ func (p *Porter) GetManifestSchema() (map[string]interface{}, error) {
 			continue
 		}
 
-		mixinSchemaMap := make(map[string]interface{})
+		// Update relative refs with the new location and reload
```
vdice: Worth mentioning that this is done to play nicely with older versions of jsonschema?  Perhaps if we have intentions to remove this logic once we're able to update to a more recent jsonschema version.

pkg/exec/schema/exec.json
```diff
@@ -1,84 +1,60 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
```
vdice: Is `draft-07` still the version that these schemas represent?

pkg/porter/schema.go
carolynvs: I can't even say why at this point. It's to work around bugs found in VS Code's support for json schema validation which comes from the RH Yaml extension which sounds like it's supported by a college student?

https://github.com/redhat-developer/vscode-yaml/issues/129

They don't know what version of json schema it is, so it may be a bug?

I'm tracking a list of bugs and workarounds [here](https://gist.github.com/carolynvs/c57c3517bd228d4c5b4d0dc24334fe7d). Since it's a lot of conjecture I didn't want to bake accusations into the code though, you know?

pkg/exec/schema/exec.json
carolynvs: I was working off of the current json schema draft (07) which has been out for years when I wrote these schemas. So I could try to diff the drafts and figure out if they are actually compatible with older drafts but that's a ton of work. 

I'm inclined to chalk up the problems so far to bugs, not older draft versions until I get clear answers on what version of the draft the libs used by VS Code is actually using.

pkg/porter/schema.go
carolynvs: Sorry not trying to be defensive and I realize that I said it's due to older versions of json schema previously. I've been digging further with Ivan (and just thinking about this way too much), and I change my mind about 20 times a day about what is wrong! 😝

PR #219, opened 2019-03-12
Fix canary install

carolynvs:
* Fix install script for non-latest version 
* Fix where porter is built and published


carolynvs:
Merge like releases are borked and I wanna release tonight. 😇

Issue #224, opened 2019-03-14
porter build should always create secure bundles


carolynvs:

carolynvs:
Hold off on this for a bit while CNAB/duffle redoes how it handles signing.

jeremyrickard:
This seems to still be unsettled in CNAB/duffle..maybe we flip install to assume unsigned for now? Experience will be a little easier until it's all settled.

carolynvs:
Yeah I hate always having to use `--insecure` when it's the only thing that works right now. EXECUTE! 👍

carolynvs:
Closing as the security spec is not complete.

PR #229, opened 2019-03-18
Add porter install command

### LINKED ISSUE #106: Add porter install command
This should link to functionality behind `duffle install` so that a user doesn't have to switch between porter and duffle. No code copy/pasta!

UPDATE: 

I've built out more of the commands that need to be implemented for this epic. So for this issue in particular I'm going to start with just 

```
porter install --insecure [-f path/to/bundle.json] [--name CLAIM] [-c CREDS]`
```

* `--insecure` has to be specified for now since #224 isn't implemented yet.
* `--name` will default to the bundle name, and is an optional flag.
* `-f` will default to the bundle file in the current direction, just like it does for `porter run`.
* the creds will still be handled by duffle until porter has that implemented.

The goal is that a developer could have a feedback cycle of `porter build` -> `porter install` -> edit ...

Tasks:
- [x] scaffold the command (#229)
- [x] call out to duffle for the install
- [ ] read creds from our own store in PORTER_HOME
- [x] --name defaults to the name of the bundle in the manifest
- [x] -f defaults to the bundle in the current directory

carolynvs:
This is the skeleton for the install command.

```
$ porter bundle install --help
Install a bundle

Usage:
  porter install [flags]

Examples:
  porter install
  porter install --insecure
  porter install --file myapp/bundle.json
  porter install --name MyAppInDev
  porter install --param-file base-values.txt --param-file dev-values.txt --param test-mode=true --param header-color=blue
  porter install --cred azure --cred kubernetes


Flags:
  -c, --cred strings         Credential to use when installing the bundle. May be either a named set of credentials or a filepath.
  -f, --file string          Path to the CNAB definition to install (default "bundle.json")
  -h, --help                 help for install
      --insecure             Allow installing untrusted bundles
      --name string          Name of the claim, defaults to the name of the bundle
      --param strings        Define an individual parameter in the form NAME=VALUE. Overrides parameters set with the same name using --param-file. May be specified multiple times.
      --param-file strings   Path to a parameters definition file for the bundle, each line in the form of NAME=VALUE. May be specified multiple times.

```

* alias: `porter install`
* `--insecure` is required until #224 is implemented

Part of #106.

### DIFF
--- a/cmd/porter/bundle.go
+++ b/cmd/porter/bundle.go
@@ -1,0 +1,57 @@
+package main
+
+import (
+	"github.com/deislabs/porter/pkg/porter"
+	"github.com/spf13/cobra"
+)
+
+func buildBundleCommands(p *porter.Porter) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "bundle",
+		Short: "bundle commands",
+	}
+
+	cmd.AddCommand(buildBundleInstallCommand(p))
+
+	return cmd
+}
+
+func buildBundleInstallCommand(p *porter.Porter) *cobra.Command {
+	opts := porter.InstallOptions{}
+	cmd := &cobra.Command{
+		Use:   "install",
+		Short: "Install a bundle",
+		Example: `  porter install
+  porter install --insecure
+  porter install --file myapp/bundle.json
+  porter install --name MyAppInDev
+  porter install --param-file base-values.txt --param-file dev-values.txt --param test-mode=true --param header-color=blue
+  porter install --cred azure --cred kubernetes
+`,
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			return opts.Prepare()
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return p.InstallBundle(opts)
+		},
+	}
+
+	f := cmd.Flags()
+	f.BoolVar(&opts.Insecure, "insecure", false,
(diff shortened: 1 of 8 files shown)


cmd/porter/bundle.go
```diff
@@ -0,0 +1,57 @@
...
+		"Allow installing untrusted bundles")
+	f.StringVarP(&opts.File, "file", "f", "bundle.json",
+		"Path to the CNAB definition to install")
+	f.StringVar(&opts.Name, "name", "",
+		"Name of the claim, defaults to the name of the bundle")
+	f.StringSliceVar(&opts.ParamFiles, "param-file", nil,
+		"Path to a parameters definition file for the bundle, each line in the form of NAME=VALUE. May be specified multiple times.")
+	f.StringSliceVar(&opts.RawParams, "param", nil,
+		"Define an individual parameter in the form NAME=VALUE. Overrides parameters set with the same name using --param-file. May be specified multiple times.")
+	f.StringSliceVarP(&opts.CredentialSets, "cred", "c", nil, "Credential to use when installing the bundle. May be either a named set of credentials or a filepath.")
```
vdice: Worth mentioning that this can be specified multiple times, as the two above?

carolynvs: Oh yes! I want to call that out for anything that repeatable since most people don't know how to figure that out from "stringArray" or whatever cobra prints out in the help text.

Thanks, I'll fix that right now.

PR #239, opened 2019-03-26
Default --debug to the porter-debug parameter at runtime

carolynvs:
Look for the new porter-debug parameter that are included in porter bundles now and use it to default the porter run --debug flag.

### DIFF
--- a/cmd/porter/run.go
+++ b/cmd/porter/run.go
@@ -1,47 +1,26 @@
 package main
 
 import (
-	"fmt"
-	"os"
-
-	"github.com/deislabs/porter/pkg/config"
 	"github.com/deislabs/porter/pkg/porter"
 	"github.com/spf13/cobra"
 )
 
 func buildRunCommand(p *porter.Porter) *cobra.Command {
-	var opts struct {
-		file      string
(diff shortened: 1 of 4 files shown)


cmd/porter/run.go
```diff
@@ -1,47 +1,26 @@
...
-			}
-
-			var err error
-			opts.action, err = config.ParseAction(opts.rawAction)
-			if err != nil {
-				return err
-			}
-
-			return nil
+			return opts.Validate()
```
carolynvs: I moved logic out of the cobra command while I was at it, didn't belong there.

pkg/config/config.go
```diff
@@ -22,6 +22,9 @@ const (
 
 	// EnvACTION is the request
 	EnvACTION = "CNAB_ACTION"
+
+	// EnvDEBUG is a custom porter parameter that signals that --debug flag has been pass through from the client to the runtime.
```
vdice: nit: typo in comment.  `s/pass/passed`

carolynvs:
Fixed the typo! 💪

PR #240, opened 2019-03-26
Implement porter uninstall

### LINKED ISSUE #243: porter uninstall
`porter uninstall [CLAIM] [--insecure] [-f bundle.json] [--cred azure] [--param header=blue] [--param-file dev-values]`

All the arguments and flags work the same as `porter install` where everything can be defaulted when in the directory of the bundle.

carolynvs:
Closed #243

### DIFF
--- a/build/testdata/bundles/wordpress/porter.yaml
+++ b/build/testdata/bundles/wordpress/porter.yaml
@@ -42,5 +42,6 @@ install:
 uninstall:
 - helm:
     description: "Uninstall Wordpress"
-    name:
-      source: bundle.parameters.wordpress-name
+    purge: true
+    releases:
+      - "source: bundle.parameters.wordpress-name"
--- a/pkg/porter/cnab.go
+++ b/pkg/porter/cnab.go
@@ -1,10 +1,262 @@
 package porter
 
-import cnabprovider "github.com/deislabs/porter/pkg/cnab/provider"
+import (
+	"bufio"
+	"github.com/deislabs/porter/pkg/config"
+	"github.com/deislabs/porter/pkg/parameters"
+	"github.com/pkg/errors"
+	"io/ioutil"
+	"os"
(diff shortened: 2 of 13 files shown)


build/testdata/bundles/wordpress/porter.yaml
```diff
@@ -42,5 +42,6 @@ install:
 uninstall:
 - helm:
     description: "Uninstall Wordpress"
-    name:
-      source: bundle.parameters.wordpress-name
+    purge: true
+    releases:
+      - "source: bundle.parameters.wordpress-name"
```
vdice: Ooh, this works?  Handy to know...

pkg/porter/cnab.go
```diff
@@ -1,10 +1,262 @@
...
 type CNABProvider interface {
 	Install(arguments cnabprovider.InstallArguments) error
-	//Upgrade() error
-	//Uninstall() error
+	Uninstall(arguments cnabprovider.UninstallArguments) error
+}
+
+
+// sharedOptions are common options that apply to multiple CNAB actions.
+type sharedOptions struct {
```
vdice: sharing is caring!

build/testdata/bundles/wordpress/porter.yaml
carolynvs: When you are in the yaml list, you need to put quotes around `"source: bundle...."` to make it work. But yeah it's neat!

pkg/porter/cnab.go
carolynvs: Yeah I think we'll want to do this for the duffle Install/Uninstall args too but I wasn't confident so I held off on that for now. It seems that pretty much all the cnab actions take the same options and just treat things slightly differently. I didn't want to consolidate too quickly so I wanted to wait until we had all the actions implemented fully, then see if they are all the same.

PR #251, opened 2019-04-05
feat(*): mask sensitive values in console output

vdice:
Mask values in console output if value is:

- a credential
- an output
- a parameter if designated as sensitive

Implements https://github.com/deislabs/porter/issues/132

### DIFF
--- a/pkg/config/manifest.go
+++ b/pkg/config/manifest.go
@@ -19,8 +19,9 @@ import (
 
 type Manifest struct {
 	// path where the manifest was loaded, used to resolve local bundle references
-	path    string
-	outputs map[string]string
+	path            string
+	outputs         map[string]string
+	sensitiveValues []string
 
 	Name         string                 `yaml:"name,omitempty"`
 	Description  string                 `yaml:"description,omitempty"`
@@ -48,6 +49,7 @@ type ParameterDefinition struct {
 	MaxLength     *int              `yaml:"maxLength,omitempty"`
 	Metadata      ParameterMetadata `yaml:"metadata,omitempty"`
 	Destination   *Location         `yaml:"destination,omitempty"`
+	Sensitive     bool              `yaml:"sensitive"`
 }
 
 type CredentialDefinition struct {
@@ -97,6 +99,10 @@ func (d *Dependency) resolveValue(key string) (interface{}, error) {
 	switch sourceType {
 	case "outputs":
 		replacement = d.m.outputs[sourceName]
+		// Porter considers all outputs as sensitive
+		if replacement != nil {
+			d.m.sensitiveValues = append(d.m.sensitiveValues, reflect.ValueOf(replacement).String())
+		}
 	case "parameters":
 		for _, param := range d.m.Parameters {
 			if param.Name == sourceName {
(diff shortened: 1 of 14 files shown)


pkg/config/manifest.go
```diff
@@ -723,16 +744,30 @@ func (m *Manifest) resolveValue(key string) (interface{}, error) {
...
+			if replacement != nil {
+				m.sensitiveValues = append(m.sensitiveValues, reflect.ValueOf(replacement).String())
+			}
 		}
 	case "dependencies":
 		for _, dep := range m.Dependencies {
 			if dep.Name == sourceName {
-				return dep.resolveValue(key)
+				replacement, err := dep.resolveValue(key)
+				// Retrieve updated list of sensitive values from dependency and add to our list
```
carolynvs: I think I see why you put this here, but it really will need to be refactored or improved later. It's problematic that we need to remember to do this after calling `dep.resolveValue` and points to a leaky abstraction with how we are either managing our dependencies or how we are collecting and managing these sensitive values I think.

pkg/context/context.go
```diff
@@ -24,15 +25,42 @@ type Context struct {
 	NewCommand CommandBuilder
 }
 
+type AuditableWriter struct {
```
carolynvs: I hate to nit on the name but I think this name is misleading. Audit implies that it will record who performed actions so that there is accountability. This is more of a "censored" writer, who will bleep stuff out. 😀

pkg/mixin/runner_test.go
```diff
@@ -62,3 +65,33 @@ func TestRunner_Run(t *testing.T) {
...
+	// Capture the output
+	r.Out = auditableWriter
+	r.Err = auditableWriter
+
+	err := r.Validate()
+	require.NoError(t, err)
+
+	err = r.Run()
+	assert.NoError(t, err)
+	assert.Contains(t, string(output.Bytes()), "Hello *******")
```
carolynvs: 🎉

pkg/porter/run.go
```diff
@@ -113,6 +113,9 @@ func (p *Porter) Run(opts RunOptions) error {
 		if err != nil {
 			return errors.Wrap(err, "unable to resolve sourced values")
 		}
+		// Hand over values needing masking in context output streams
+		p.Context.SetSensitiveValues(p.Manifest.GetSensitiveValues())
```
carolynvs: Ok, so this is in reference to the comment above about refactoring. What do you think about having `p.Context.AppendSensitiveValue("poop")`? Then as the manifest finds sensitive words, instead of the manifest tracking them in a slice, and having to coordinate passing it back to the context here (and then also after resolving the dependency's values), it could append to them as it finds them?

You are closer to this, so let me know if you think that would help reduce complexity or not.

pkg/context/context.go
vdice: You're right, this needed renaming.  To that end, I've updated to `CensoredWriter`.

pkg/porter/run.go
vdice: In talking about this offline, I believe we came to the consensus that we're good with revisiting this area of the code after some near-term refactoring.

PR #264, opened 2019-04-12
use mustache to wire up params, creds, outputs

### LINKED ISSUE #161: Iterate on variable references
As mentioned in #136, I feel like there is some interest in enabling composability in variable references. Instead of a strictly 1:1 mapping, I was suggesting using a format taking to how string interpolation is usually done:

```yaml
parameter.Name: |-
  jdbc:mysql://${outputs.hostname}:${outputs.port}/${outputs.dbName}
```

jeremyrickard:
This PR changes how we wire params, credentials and outputs between steps. Instead of the previous "source:" bundle.x.y.z, now we use Mustache templates. 

- [x] Introduce 
- [x] Update docs
- [x] Fix e2e tests

Closes #161


jeremyrickard:
@carolynvs @vdice take a look. 

The only thing I'm kind of bummed about in this is that the mustache templates need to be quoted, otherwise the go marshalling isn't happy. Maybe we need to change how the marshalling works and only marshall the steps after they've been fully resolved? This seems like a big enough change to review on it's own though, tackle that in another refactor.

carolynvs:
I'm not sure I see how to resolve before we've parsed the file, but you are closer than I am to it for sure. Yeah let's go with this for now and iterate.

jeremyrickard:
> I'm not sure I see how to resolve before we've parsed the file, but you are closer than I am to it for sure. Yeah let's go with this for now and iterate.

I was thinking more that we wouldn't fully try and rehydrate a struct with the yaml marsheller, but I don't know if that's possible. just thinking out loud

build/testdata/bundles/mysql/porter.yaml
```diff
@@ -28,10 +28,8 @@ install:
     version: 0.10.2
     replace: true
     set:
-      mysqlDatabase:
-        source: bundle.parameters.database-name
-      mysqlUser:
-        source: bundle.parameters.mysql-user
+      mysqlDatabase: "{{ bundle.parameters.database-name}}"
```
carolynvs: muy bien!

docs/content/authoring-bundles.md
```diff
@@ -122,10 +122,8 @@ install:
     chart: stable/mysql
     version: 0.10.2
     set:
-      mysqlDatabase:
-        source: bundle.parameters.database-name
-      mysqlUser:
-        source: bundle.parameters.mysql-user
+      mysqlDatabase: {{ bundle.parameters.database-name }}
```
carolynvs: should these have quotes?

docs/content/mixin-architecture.md
```diff
@@ -92,10 +92,8 @@ helm:
   version: 0.10.2
   replace: true
   set:
-    mysqlDatabase:
-      source: bundle.parameters.database-name
-    mysqlUser:
-      source: bundle.parameters.mysql-user
+    mysqlDatabase: {{ bundle.parameters.database-name }}
```
carolynvs: should these have quotes?

docs/content/porter-or-duffle.md
```diff
@@ -95,23 +95,18 @@ install:
           key: mysql-password
   - helm:
       description: "Install Wordpress"
-      name:
-        source: bundle.parameters.wordpress-name
+      name: {{ bundle.parameters.wordpress-name }}
```
carolynvs: Seems like all the markdown files are missing quotes around the moustash template

docs/content/wiring.md
```diff
@@ -29,7 +29,7 @@ You can also provide any other attributes, as specified by the CNAB [parameters]
 
 ## Wiring Parameters
 
-Once a parameter has been declared in the `porter.yaml`, Porter provides a simple mechanism for specifying how the parameter value should be wired into the mixin. To reference a parameter in the bundle, you use a notation of the form `source: bundle.parameters.PARAM_NAME`. This can be used anywhere within step definition. For example, to use the `database_name` parameter defined above in the `set` block of the helm mixin:
+Once a parameter has been declared in the `porter.yaml`, Porter provides a simple mechanism for specifying how the parameter value should be wired into the mixin. To reference a parameter in the bundle, you use a notation of the form `{{ bundle.parameters.PARAM_NAME }}`. This can be used anywhere within step definition. For example, to use the `database_name` parameter defined above in the `set` block of the helm mixin:
```
carolynvs: we should mention the surrounding quotes

docs/content/wiring.md
```diff
@@ -259,19 +244,31 @@ The wordpress bundle declares a dependency on the `mysql` bundle, which we saw a
...
+When the install is executed for this bundle, the steps defined in the `mysql` bundle are completed first. Once those steps have run, any outputs defined are available. In this case, we want to use the `mysql-password` output from the `mysql` dependency. As the example YAML indicates, we do so with the declaration `{{ bundle.dependencies.mysql.outputs.mysql-password }}`. The Porter runtime uses the `mysql` manifest to determine how to obtain the values from the dependency output and parameters.
 
-For more information on how dependencies are handled, refer to the [dependencies](/dependencies) documentation.
\ No newline at end of file
+For more information on how dependencies are handled, refer to the [dependencies](/dependencies) documentation.
+
+## Combining References
+
+It is possible to reference multiple parameters, credentials and/or outputs in a single place. You simply combine the expressions as follows:
```
carolynvs: nit: Let's not use terms like "simply" in the docs. Bridget just [tweeted](https://twitter.com/bridgetkromhout/status/1116730265647370241) about why.

 As a replacement you can say "You can combine the expressions as follows"

pkg/config/manifest.go
```diff
@@ -84,57 +82,61 @@ func (d *Dependency) Validate() error {
...
-		return nil, errors.New(fmt.Sprintf("no value found for source specification: %s", key))
+	creds := make(map[string]interface{})
+	depVals["credentials"] = creds
+	for _, cred := range d.m.Credentials {
+		val, err := resolveCredential(cred)
+		if err != nil {
+			return nil, nil, errors.Wrap(err, fmt.Sprintf("could not handle bundle dependency %s", d.Name))
+		}
+		sensitiveStuff = append(sensitiveStuff, val)
+		params[cred.Name] = val
```
carolynvs: shouldn't this be `creds[cred.Name] = val`?

pkg/config/manifest.go
```diff
@@ -566,220 +568,73 @@ func (s *Step) GetMixinName() string {
...
-					if err != nil {
-						return errors.Wrap(err, fmt.Sprintf("unable to set value for %s", k.String()))
-					}
-					mp.SetMapIndex(k, reflect.ValueOf(replacement))
-				}
-			}
+	params := make(map[string]interface{})
+	bundle["parameters"] = params
+	for _, param := range m.Parameters {
+		//pe := strings.ToUpper(param.Name)
```
carolynvs: If this comment isn't needed, let's remove this line so I don't get confused. 😀

pkg/config/manifest.go
```diff
@@ -566,220 +568,73 @@ func (s *Step) GetMixinName() string {
...
-// ResolveStep will walk through the Step's data and resolve any placeholder
-// data using the definitions in the manifest, like parameters or credentials.
-func (m *Manifest) ResolveStep(step *Step) error {
-	return reflectwalk.Walk(step, m)
-}
+func (m *Manifest) buildSourceData() (map[string]interface{}, error) {
+	data := make(map[string]interface{})
+	m.sensitiveValues = []string{}
+	bundle := make(map[string]interface{})
+	data["bundle"] = bundle
```
carolynvs: This is a question about mustache. Looks like we are building up a map that has keys for parameters, outputs, etc inside of it, then that's what the template is operating upon.

If we had a struct called "TemplateMagicPants" (or whatever) and it had fields for parameters, outputs, credentials, and everything else we needed, would that work too as input data to mustache? Or would that result in different syntax in the template?

This is just a question for me to understand mustache, not asking you to change anything. 😀

pkg/config/manifest.go
```diff
@@ -566,220 +568,73 @@ func (s *Step) GetMixinName() string {
...
+	for _, param := range m.Parameters {
+		//pe := strings.ToUpper(param.Name)
+		pe := param.Name
+		var val string
+		val, err := resolveParameter(param)
+		if err != nil {
+			return nil, err
+		}
+		if param.Sensitive {
+			m.sensitiveValues = append(m.sensitiveValues, val)
```
carolynvs: This isn't on you, but the sensitive values stuff is in desperate need of a refactor so that it's not tied to the manifest like this. I can't wait for your work to split out the runtime from the manifest as I believe that's something we identified as being helpful in order to do the masking in a less obtrusive way going forward. 🤞

pkg/config/manifest.go
```diff
@@ -566,220 +568,73 @@ func (s *Step) GetMixinName() string {
...
-			val.Set(reflect.ValueOf(r))
-		}
-	} else {
-		v := val
-		if val.Kind() == reflect.Interface {
-			val = val.Elem()
+	creds := make(map[string]interface{})
+	bundle["credentials"] = creds
+	for _, cred := range m.Credentials {
+		//pe := strings.ToUpper(cred.Name)
```
carolynvs: can we nuke this comment?

pkg/config/manifest.go
```diff
@@ -566,220 +568,73 @@ func (s *Step) GetMixinName() string {
...
+	sourceData, err := m.buildSourceData()
+	if err != nil {
+		return errors.Wrap(err, "unable to resolve step")
 	}
-
-	if replacement == nil {
-		return nil, errors.New(fmt.Sprintf("no value found for source specification: %s", key))
+	payload, err := yaml.Marshal(step)
+	if err != nil {
+		return errors.Wrap(err, "unable to resolve step")
```
carolynvs: I'm not sure wrapping all these various errors with the same error message is going to be helpful? Having a bit more context about what was going on may be useful when dealing with errors like "unable to resolve step: unexpected EOF", you know?

pkg/config/manifest.go
```diff
@@ -566,220 +568,73 @@ func (s *Step) GetMixinName() string {
...
+	}
+	rendered, err := mustache.Render(string(payload), sourceData)
+	if err != nil {
+		return errors.Wrap(err, "unable to resolve step")
 	}
-	return replacement, nil
+	return yaml.Unmarshal([]byte(rendered), step)
```
carolynvs: this last error isn't being wrapped

pkg/config/manifest.go
jeremyrickard: We'd need to have a struct called TemplateMagicPants that had those, that lived in a map with the key bundle. Or another struct. It's walking what ever the template syntax is. I was trying to keep it _mostly_ the same, and I think the comment from our Amazon friends used similar syntax.

I did find myself wondering what the point of the "bundle" is. everything is rooted off that....soooooo maybe just {{ parameters.x }} or {{ outputs.y }} or {{ credentials.z }} or {{ dependencies.x.y.z }} 

I was thinking along these lines for the general refactor of "the manifest" from the running state of things, but here I didn't want to change the syntax of the template since we hadn't chatted about that. Happy to go either way, but I think my opinion is the "bundle" key is not really needed?

docs/content/porter-or-duffle.md
jeremyrickard: Yeah, I forgot to update them! Thanks for pointing that out.

pkg/config/manifest.go
carolynvs: Let's not remove the `bundle.` prefix for now. I believe that it is important for both giving people a starting point in their heads for what to type, and also for the autocomplete to have something to go off of.

carolynvs:
Merge at will

PR #265, opened 2019-04-12
install mixin from url

carolynvs:
This implements the first part of mixin installation

```
$ porter mixin install NAME --version VERSION --url URL
```

for now --url is required, but once we introduce subscriptions (i.e. being able to follow all the mixins from deislabs) it won't be anymore.

Docs available at https://deploy-preview-265--porter.netlify.com/mixin-distribution/


cmd/porter/mixins.go
```diff
@@ -39,6 +41,29 @@ func buildMixinsListCommand(p *porter.Porter) *cobra.Command {
...
+			return opts.Validate(args)
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return p.Mixins.Install(opts)
+		},
+	}
+
+	cmd.Flags().StringVarP(&opts.Version, "version", "v", "latest",
+		"The mixin version. This can either be a version number, or a tagged release like 'latest' or 'canary'")
+	cmd.Flags().StringVar(&opts.URL, "url", "",
```
vdice: This may have been considered/mentioned already, but do we want to supply a default url here?

I'd think it would point to whatever our current mixin store is (as of writing, `https://deislabs.blob.core.windows.net/porter/mixins`).  Then, `porter mixin install helm` would just work.

But perhaps this isn't necessary as these same mixins that we currently track/host (`helm`, `azure`, etc.) are/will be automatically fetched for a given bundle (`porter.yaml`) that requires them...

pkg/mixin/provider/install.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "could not create the mixin at %s", destPath)
+	}
+	defer destFile.Close()
+	err = os.Chmod(destPath, 0755)
+	if err != nil {
+		return errors.Wrapf(err, "could not set the mixin as executable at %s", destPath)
+	}
+
+	if p.Debug {
```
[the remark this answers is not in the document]
carolynvs: oops that got lost in a shuffle, yes thanks!

docs/content/mixin-distribution.md
```diff
@@ -0,0 +1,72 @@
+---
```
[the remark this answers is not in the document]
carolynvs: Yeah I wasn't sure about that file or not? I realized that it was a placeholder that never got filled out and wasn't sure if I should start now or not. 😝 

I was thinking of making that page generated from the cobra help text because otherwise it's doomed to always be out of date. Let me know what you think!

cmd/porter/mixins.go
```diff
@@ -39,6 +41,29 @@ func buildMixinsListCommand(p *porter.Porter) *cobra.Command {
...
+	cmd.Flags().StringVarP(&opts.rawFormat, "output", "o", "table",
+		"Output format, allowed values are: table, json")
+
+	return cmd
+}
+
+func BuildMixinInstallCommand(p *porter.Porter) *cobra.Command {
+	opts := mixin.InstallOptions{}
+	cmd := &cobra.Command{
+		Use:   "install",
```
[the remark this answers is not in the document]
carolynvs: oops yes that was supposed to get filled out and I forgot. Thanks for the reminder!

pkg/mixin/provider/install.go
```diff
@@ -0,0 +1,94 @@
...
+		fmt.Fprintf(p.Out, "installed %s mixin\n%s", m.Name, confirmedVersion)
+	}
+
+	return nil
+}
+
+func (p *FileSystem) downloadFile(url url.URL, destPath string) error {
+	// Ensure the parent directories exist
+	parentDir := filepath.Dir(destPath)
+	err := os.MkdirAll(parentDir, 0755)
```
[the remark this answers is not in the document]
carolynvs: I've shuffled it around and added a defer so that if we can't install it properly, we remove the entire mixin directory.

cmd/porter/mixins.go
carolynvs: That's my plan for a follow-up. We can't do that immediately because so far we have been publishing our mixins into a separate directory per mixin e.g. .../helm, .../exec .../azure, so there isn't a single default URL that works.

My plan is to get to a point ASAP where the default install gives people deislabs as a default "mixin subscription" so that they get all the mixins we make. So they can type `porter mixin install helm` or `porter mixin upgrade --all` and it will be able to figure it out based on what we publish for all of our mixins.  But this PR is just the first step for installing just a single mixin from a URL.

carolynvs: I'm also going to add a vanity URL so that we aren't exposing the full ugly url for our blob storage container. 😀 So very quickly it will be a default URL of porter.sh/mixins/atom.xml

carolynvs:
Thanks for the solid review. I think all your comments have been addressed but it's now Friday at 5 sooooooooooooo 💨

docs/content/mixin-distribution.md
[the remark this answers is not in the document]
vdice: I think a generated page in its stead would be great, for the reasons you've mentioned!  👍

pkg/mixin/provider/install.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+	defer resp.Body.Close()
+
+	// Ensure the parent directories exist
+	parentDir := filepath.Dir(destPath)
+	err = p.FileSystem.MkdirAll(parentDir, 0755)
+	if err != nil {
+		errors.Wrapf(err, "unable to create parent directory %s", parentDir)
+	}
+	defer p.FileSystem.RemoveAll(parentDir) // If we can't install the mixin, don't leave traces of it
```
vdice: This area may still need attention... currently, I see:
```
 $ porter mixin install helm --url https://deislabs.blob.core.windows.net/porter/mixins/helm
Error: could not run mixin command /Users/vdice/.porter/mixins/helm/helm /Users/vdice/.porter/mixins/helm/helm version: fork/exec /Users/vdice/.porter/mixins/helm/helm: no such file or directory
```
Is it the `defer` statement?  Is there a way to only have `defer` run if err != nil?

carolynvs: 🤦‍♀️ zomg what was I thinking? That made perfect sense for some reason in the moment. 😂

carolynvs: I miss mega `try/catch/finally` blocks...

pkg/mixin/provider/install_test.go
```diff
@@ -0,0 +1,38 @@
...
+	"net/http/httptest"
+	"testing"
+
+	"github.com/deislabs/porter/pkg/config"
+	"github.com/deislabs/porter/pkg/mixin"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestFileSystem_Install(t *testing.T) {
```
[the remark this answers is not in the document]
carolynvs: I added tests for both cases

pkg/mixin/provider/install.go
carolynvs: Fixed

carolynvs:
I'm going to leave out updating the cli.md file for now. It's so momumentally out of date it's not even funny and later figure out how to code-gen it from cobra.

carolynvs:
The test is failing on brigade because it runs as root, so my little trick of making a file that the current user can't read doesn't work. We can either keep the test, and not run as root, or fix/remove the test. I'm not sure of how else to sabotage that test though. Thoughts?

vdice:
Ah.  Hmmm, I'm assuming running as a non-root user would require a non-trivial amount of work and affect all tests?  Otherwise, my only idea would be to mock out and enable use of a test filesystem such that we can make sure `Create` returns an error...

Either way, it sounds a bit tricky and may not be worth the cost.  I'll defer to your judgement here.

carolynvs:
I don't have an off the shelf fake tool for making those methods fail on demand. So for every variation of the test we'd have a hand rolled fake which is way too much effort for this test IMO. For now I vote to leave it off, unless someone else wants to try to tackle that later? We'd end up making a general purpose new type of AferoFS that can be told to sabotage any of its operations. I don't want to try to do that in this PR however, it could come in handy later if we wan to try to do fakes at that level in all of our tests.

PR #267, opened 2019-04-15
Fix CI test runs so they don't step on each other

### LINKED ISSUE #215: Support concurrent CI runs of integration tests
Currently, multiple runs of the integration tests (see the `test-cli` make target) may conflict, causing failures.  For instance, the `porter-wordpress` example bundle installs a helm chart onto a shared k8s cluster and the helm release name is currently not unique.

The work for this ticket therefore is to ensure multiple/concurrent CI runs are supported.

### LINKED ISSUE #268: Use porter uninstall to cleanup CI test runs
Right now we run `make cleanup` at the beginning of each test run to cleanup from the previous test run. Now that we have a `porter uninstall` command, and we are making new namespaces per test run, we should start cleaning up after the each test run immediately. So each of the test-X.sh scripts should run through the lifecycle of the bundle: install, upgrade (when we have that) and then uninstall.

carolynvs:
This should let the CI tests run concurrently without rando failing. The other bundles don't have any global state so they don't need to be modified.

* The k8s related bundles install into a new namespace per test run
* The test now run uninstall at the end of the test run to cleanup after themselves

Closes #215
Closes #268

### DIFF
--- a/build/testdata/bundles/mysql/porter.yaml
+++ b/build/testdata/bundles/mysql/porter.yaml
@@ -19,6 +19,9 @@ parameters:
   type: string
   destination:
     env: MYSQL_USER
+- name: namespace
+  type: string
+  default: ''
 
 install:
 - helm:
@@ -26,6 +29,8 @@ install:
     name: porter-ci-mysql
     chart: stable/mysql
     version: 0.10.2
+    namespace:
+      source: bundle.parameters.namespace
     replace: true
     set:
       mysqlDatabase:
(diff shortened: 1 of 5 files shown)


carolynvs:
Yeah I realized when I did this that we aren't using porter uninstall yet in our cleanup. Let me add another issue to tackle that.

carolynvs:
Oops, I see what you mean. Let me tackle that extra cleanup changes in this.

PR #272, opened 2019-04-15
Do not encode -runtime into mixin download filename

carolynvs:
Instead of requiring the runtime binary for the mixin to be uploaded as foo-runtime-linux-amd64, allow the client mixin for linux, which is foo-linux-amd64, to be reused as there is no difference.

This sets us up anyway for when porter needs to support a customizable runtime image that could be a different os/arch.

Note: This doesn't change the name of the file on the filesystem, just what we require the mixin author to upload it as.

- [x] Update da code
- [x] Update all da makefiles
    * generate filenames that match the new format in the porter repo
    * install using porter mixin install
    * consolidate targets for the internal / external mixins

After this PR, I'll update the azure, helm, and terraform repos, and then all of the mixins will have done a canary publish with the new naming conventiong

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,22 +24,23 @@ else
 FILE_EXT=
 endif
 
-MIXIN_TAG          ?= canary
-HELM_MIXIN_URL      = https://deislabs.blob.core.windows.net/porter/mixins/helm/$(MIXIN_TAG)/helm
-AZURE_MIXIN_URL     = https://deislabs.blob.core.windows.net/porter/mixins/azure/$(MIXIN_TAG)/azure
-TERRAFORM_MIXIN_URL = https://deislabs.blob.core.windows.net/porter/mixins/terraform/$(MIXIN_TAG)/terraform
+INT_MIXINS = exec kubernetes
+EXT_MIXINS = helm azure terraform
+MIXIN_TAG ?= canary
+MIXINS_URL = https://deislabs.blob.core.windows.net/porter/mixins
 
-build: build-client build-runtime azure helm terraform
+.PHONY: build
+build: build-porter build-mixins get-mixins
 
-build-runtime: generate
-	$(MAKE) $(MAKE_OPTS) build-runtime MIXIN=porter -f mixin.mk BINDIR=bin
-	$(MAKE) $(MAKE_OPTS) build-runtime MIXIN=exec -f mixin.mk
-	$(MAKE) $(MAKE_OPTS) build-runtime MIXIN=kubernetes -f mixin.mk
+build-porter: generate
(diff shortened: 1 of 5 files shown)


carolynvs:
Ok, boy that was harder than I expected. Silly make. This is ready for review.

Makefile
```diff
@@ -156,9 +126,11 @@ install: build
...
+clean-mixins:
 	-rm -fr bin/
-	-rm -fr cnab/
-	-rm porter.yaml Dockerfile bundle.json
-	-helm delete --purge porter-ci-mysql
-	-helm delete --purge porter-ci-wordpress
+
+clean-last-testrun:
+	-rm -fr cnab/ porter.yaml Dockerfile bundle.json
+	-helm delete --purge porter-ci-mysql porter-ci-wordpress
```
vdice: This line can be removed now, right?  (No longer applicable now that helm releases have unique names and the corresponding test script(s) invoke a `porter uninstall`, etc.)

Makefile
```diff
@@ -50,59 +51,28 @@ ifndef HAS_PACKR2
...
-	$(MAKE) $(MAKE_OPTS) xbuild-runtime MIXIN=kubernetes -f mixin.mk
+xbuild-mixins: $(addprefix xbuild-mixin-,$(INT_MIXINS))
+xbuild-mixin-%: generate
+	$(MAKE) $(MAKE_OPTS) xbuild-all MIXIN=$* -f mixin.mk
 
-xbuild-client:
-	$(MAKE) $(MAKE_OPTS) xbuild-client MIXIN=porter -f mixin.mk BINDIR=bin
-	$(MAKE) $(MAKE_OPTS) xbuild-client MIXIN=exec -f mixin.mk
-	$(MAKE) $(MAKE_OPTS) xbuild-client MIXIN=kubernetes -f mixin.mk
+get-mixins: $(addprefix get-mixin-,$(EXT_MIXINS))
```
vdice: You might've already played around with this option, but we could reduce this and the `get-mixin-*` targets below with:
```make
get-mixins:
	$(foreach MIXIN, $(EXT_MIXINS), \
		bin/porter mixin install $(MIXIN) --version $(MIXIN_TAG) --url $(MIXINS_URL)/$(MIXIN); \
	)
```
Which would be handy as the external mixin list grows (just need to update `EXT_MIXINS`, etc.)

Though, if there is a common use case for needing to fetch individual external mixins, this might not be the best approach.

carolynvs: I did it this way originally because I thought it would help for getting an individual mixin. But really the command being run `porter mixin install` isn't hard. So if someone needed to be doing one off work on a single mixin, then I don't see why they couldn't use that directly instead of relying on the makefile.

So yeah let me make that change.

PR #279, opened 2019-04-18
Generate a mixin feed based on a directory

carolynvs:
This adds two commands to porter as a build up to supporting mixin feeds. All alone they aren't really enough to add new docs or anything yet though.

Note: This doesn't update an existing atom.xml file yet. I plan on doing that in a follow-up PR.

```console
$ porter mixins feed template --help
Create an atom feed template in the current directory

Usage:
  porter mixins feed template
```

```console
$ porter mixins feed generate --help
Generate an atom feed from the mixins in a directory.

A template is required, providing values for text properties such as the author name, 
base URLs and other values that cannot be inferred from the mixin file names. You 
can make a default template by running 'porter mixins feed template'.

The file names of the mixins must follow the naming conventions required of published 
mixins:

VERSION/MIXIN-GOOS-GOARCH[FILE_EXT]

More than one mixin may be present in the directory, and the directories may be nested 
a few levels deep, as long as the file path ends with the above naming convention, porter 
will find and match it. Below is an example directory structure that porter can list to 
generate a feed:

bin/
└── v1.2.3/
    ├── mymixin-darwin-amd64
    ├── mymixin-linux-amd64
    └── mymixin-windows-amd64.exe

See https://porter.sh/mixin-distribution more details.

Usage:
  porter mixins feed generate [flags]

Examples:
porter mixin feed generate
porter mixin feed generate --dir bin --file bin/atom.xml --template porter-atom-template.xml

Flags:
  -d, --dir string        The directory to search for mixin versions to publish in the feed. Defaults to 
the current directory.
  -f, --file string       The path of the atom feed output by this command. (default "atom.xml")
  -h, --help              help for generate
  -t, --template string   The template atom file used to populate the text fields in the generated 
feed. (default "atom-template.xml")
```

### DIFF
--- a/pkg/mixin/feed/generate.go
+++ b/pkg/mixin/feed/generate.go
@@ -1,0 +1,168 @@
+package feed
+
+import (
+	"fmt"
+	"os"
+	"regexp"
+	"sort"
+	"time"
+
+	"github.com/cbroglie/mustache"
+	"github.com/deislabs/porter/pkg/context"
+	"github.com/pkg/errors"
+)
+
+type GenerateOptions struct {
+	SearchDirectory string
+	AtomFile        string
+	TemplateFile    string
+}
+
+func (o *GenerateOptions) Validate(c *context.Context) error {
+	err := o.ValidateSearchDirectory(c)
+	if err != nil {
+		return err
+	}
+
+	return o.ValidateTemplateFile(c)
+}
+
+func (o *GenerateOptions) ValidateSearchDirectory(cxt *context.Context) error {
+	if o.SearchDirectory == "" {
+		wd, err := os.Getwd()
+		if err != nil {
+			return errors.Wrap(err, "could not get current working directory")
+		}
+
+		o.SearchDirectory = wd
+	}
+
+	if _, err := cxt.FileSystem.Stat(o.SearchDirectory); err != nil {
+		return errors.Wrapf(err, "invalid --directory %s", o.SearchDirectory)
+	}
+
+	return nil
+}
+
+func (o *GenerateOptions) ValidateTemplateFile(cxt *context.Context) error {
+	if _, err := cxt.FileSystem.Stat(o.TemplateFile); err != nil {
+		return errors.Wrapf(err, "invalid --template %s", o.TemplateFile)
+	}
+
+	return nil
+}
+
+type mixinFileset struct {
+	Mixin   string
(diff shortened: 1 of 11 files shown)


pkg/mixin/feed/generate.go
```diff
@@ -0,0 +1,168 @@
...
+		wd, err := os.Getwd()
+		if err != nil {
+			return errors.Wrap(err, "could not get current working directory")
+		}
+
+		o.SearchDirectory = wd
+	}
+
+	if _, err := cxt.FileSystem.Stat(o.SearchDirectory); err != nil {
+		return errors.Wrapf(err, "invalid --directory %s", o.SearchDirectory)
```
vdice: Q: should we print `invalid --dir` or `invalid -d` to represent the flag strings supported?  Or perhaps just `invalid directory` for a more general error msg?

pkg/mixin/feed/generate.go
```diff
@@ -0,0 +1,168 @@
...
+
+type mixinFeed map[string]map[string]*mixinFileset
+
+func Generate(opts GenerateOptions, cxt *context.Context) error {
+	feedTmpl, err := cxt.FileSystem.ReadFile(opts.TemplateFile)
+	if err != nil {
+		return errors.Wrapf(err, "error reading template file at %s", opts.TemplateFile)
+	}
+
+	mixinRegex := regexp.MustCompile(`(.*/)?(.+)/([a-z]+)-([a-z0-9]+)-([a-z0-9]+)(\.exe)?`)
```
vdice: Do we want to restrict some of these regex fields to only the currently supported values?  (For instance, OS can be only one of `{linux, windows, darwin}`, arch only `amd64`, etc.)

I can also understand wanting to keep it unrestricted with a vision towards supporting other variants...

pkg/mixin/feed/generate.go
carolynvs: oops, yes thanks it should be `--dir`

pkg/mixin/feed/generate.go
carolynvs: I originally didn't want to be in the business of keeping up to date with our build matrix but honestly, it is not going to change often. So we just need to be aware that this is here. I'll update it and leave a note in the makefile's build matrix that this should be updated too when it's changed.

PR #280, opened 2019-04-18
feat(manifest): enable bundle name interpolation

vdice:
This enhancement came about as I wrote a bundle that had a use for passing the bundle name to a mixin action (via templating in the porter manifest).

If this feature gets the go ahead, do we also want to add similar for a dependency bundle name, such that one could also reference/interpolate `bundle.dependency.name` at some point in a porter manifest?


carolynvs:
Also yes, I think we should do a follow up for dependent bundle names.

PR #281, opened 2019-04-19
scripts/install: make exporting easier for users

anubhavmishra:
This PR shows users the explicit command to add `$PORTER_HOME` to their path.


jeremyrickard:
Thanks @anubhavmishra!!

PR #287, opened 2019-04-23
feat(*): add --driver opt to action commands

vdice:
Closes https://github.com/deislabs/porter/issues/252

### DIFF
--- a/cmd/porter/bundle.go
+++ b/cmd/porter/bundle.go
@@ -82,7 +82,10 @@ func buildBundleInstallCommand(p *porter.Porter) *cobra.Command {
 		Short: "Install a bundle",
 		Long: `Install a bundle.
 
-The first argument is the name of the claim to create for the installation. The claim name defaults to the name of the bundle.`,
+The first argument is the name of the claim to create for the installation. The claim name defaults to the name of the bundle.
+
+Porter uses the Docker driver as the default runtime for executing a bundle's invocation image, but an alternate driver may be supplied via '--driver/-d'.
+For instance, the 'debug' driver may be specified, which simply logs the info given to it and then exits.`,
 		Example: `  porter bundle install
   porter bundle install --insecure
   porter bundle install MyAppInDev --file myapp/bundle.json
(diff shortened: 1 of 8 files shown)


pkg/porter/cnab.go
```diff
@@ -260,3 +268,14 @@ func (o *sharedOptions) combineParameters() map[string]string {
 
 	return final
 }
+
+// Validate that the provided driver is supported
+func (o *sharedOptions) validateDriver() error {
+	// Not using duffle's driver.Lookup() as it currently does not return an error on invalid drivers
```
carolynvs: Is there an issue in duffle that this function needs to be changed? If not, can you make one, even just a placeholder, so that we don't forget?

cmd/porter/bundle.go
```diff
@@ -154,6 +158,8 @@ The first argument is the name of the claim to uninstall. The claim name default
 		"Define an individual parameter in the form NAME=VALUE. Overrides parameters set with the same name using --param-file. May be specified multiple times.")
 	f.StringSliceVarP(&opts.CredentialIdentifiers, "cred", "c", nil,
 		"Credential to use when uninstalling the bundle. May be either a named set of credentials or a filepath, and specified multiple times.")
+	f.StringVarP(&opts.Driver, "driver", "d", "docker",
+		"Specify a driver to use. Defaults to 'docker'.")
```
carolynvs: This should list the allowed drivers, e.g. "Allowed values: docker, debug". 

Also, it's not explained what the "debug" driver is, or what a driver is. Maybe not here, feel free to use your judgement, but somewhere in this helptext. Right now we have a new example that says "--driver debug" and it's really not clear what that does. We could either add text to the example to explain, we could put a paragraph into the `Long` to explain drivers, or explain in the flag text.

pkg/porter/cnab.go
vdice: Good call; filed https://github.com/deislabs/duffle/pull/723.  We'll see if this gains approval...

cmd/porter/bundle.go
vdice: Updated; let me know how it reads/looks to you.  Thanks!

PR #306, opened 2019-04-29
Reference types in deislabs/cnab-go 

carolynvs:
Previously we had been using CNAB types (such as Bundle) copied from duffle. Now that they are available in github.com/deislabs/cnab-go, we can reference them from there instead.


carolynvs:
RADU IS EVERYWHERE! 👀 

Thanks for the review!

Issue #320, opened 2019-05-03
Improve abstraction over CNAB


carolynvs:
The goal is to improve our abstraction so that as people learn
porter by just using it and seeing what happens, they are more likely to
fall into using it in a way that we recommend.

Some of the day 0 questions/comments that I have already seen and want to address
before they even pop into their heads:

* _Should I first do `porter build` and then edit bundle.json going forward
and run the commands passing in the -f bundle.json arg?_
* _Where should I copy my files into the container?_
* _What is the /cnab/app directory in the container and do I have to use it?_
* _I feel like I need to know the spec first before I can build a bundle_
* _Does porter work with just porter bundles or any CNAB bundle?_

## Bundle Directory Argument

Porter will give them the bundle directory (/cnab/app) as a build arg.

```
FROM ubuntu:latest

# Put your bundle's files here
ARG BUNDLE_DIR

# Your porter.yaml is in BUNDLE_DIR, so you can reference these with a relative path
COPY hi.txt $BUNDLE_DIR/

# You are encouraged you to copy files from their canonical location
ADD https://github.com/ashleymcnamara/gophers/raw/master/WonderWomanGopher.png $BUNDLE_DIR/
```

The `porter create` command will output a template Dockerfile and a README.

## Autocopy Files in Bundle Directory

When the user chooses to not use a template Dockerfile, all of the files in the bundle directory (the directory that contains the porter.yaml file), are automatically copied into the invocation image under /cnab/app.

## Consolidate CNAB files in cnab/

Currently we output the bundle.json to the root of the bundle directory. This
makes it easy for users to think that they should/could be doing something with
it. Since we already have a directory named cnab, we can move it there, and
explain that everything in that directory is part of our compatibility with the
CNAB spec and are generated files.

As part of that change, the cnab directory should move from being created during
`porter create` to `porter build` to reinforce that it is a generated artifact
from the build. This also helps with the question of if people need to check in
the cnab directory because Porter will automatically generate it before each
build.

## Example: Working with bundle files

Currently we hint to users that they can copy files into cnab/ and that's how
they can get files into the invocation image. Now with the Dockerfile template
generated during `porter create`, and cnab/ becoming more ephemeral, here's
what the recommendation looks like:

1. `porter create` to get a Dockerfile and porter.yaml.
1. Files can be put in the root of the bundle directory, or in subdirectories.
1. They are copied into the bundle via COPY commands in the Dockerfile.

    ```
[...]

PR #328, opened 2019-05-09
Add Credentials Skeleton and Generate Flow

### LINKED ISSUE #225: porter credentials generate

jeremyrickard:
This PR adds skeleton stuff to the porter cmd/ stuff to add credentials commands.
This PR only wires up generate, however, since it's already pretty large. Will handle
the others in follow-ons. The generate command is similar to Duffle in that it uses
survey, but the text and prompts are different.

Closes #225

### DIFF
--- a/cmd/porter/credentials.go
+++ b/cmd/porter/credentials.go
@@ -8,9 +8,10 @@ import (
 
 func buildCredentialsCommand(p *porter.Porter) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:     "credentials",
-		Aliases: []string{"cred"},
-		Short:   "Credentials commands",
+		Use:         "credentials",
+		Aliases:     []string{"cred"},
+		Annotations: map[string]string{"group": "resource"},
+		Short:       "Credentials commands",
 	}
 
 	cmd.AddCommand(buildCredentialsAddCommand(p))
@@ -25,8 +26,9 @@ func buildCredentialsCommand(p *porter.Porter) *cobra.Command {
 
 func buildCredentialsAddCommand(p *porter.Porter) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "add",
-		Short: "Add Credential",
+		Use:    "add",
+		Short:  "Add Credential",
+		Hidden: true,
 		PreRunE: func(cmd *cobra.Command, args []string) error {
 			return nil
 		},
@@ -40,8 +42,9 @@ func buildCredentialsAddCommand(p *porter.Porter) *cobra.Command {
 
 func buildCredentialsEditCommand(p *porter.Porter) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "edit",
-		Short: "Edit Credential",
+		Use:    "edit",
+		Short:  "Edit Credential",
+		Hidden: true,
 		PreRunE: func(cmd *cobra.Command, args []string) error {
 			return nil
 		},
@@ -56,9 +59,9 @@ func buildCredentialsEditCommand(p *porter.Porter) *cobra.Command {
 func buildCredentialsGenerateCommand(p *porter.Porter) *cobra.Command {
 	opts := porter.CredentialOptions{}
 	cmd := &cobra.Command{
-		Use:   "generate [name]",
+		Use:   "generate [NAME]",
 		Short: "Generate Credential Set",
-		Long: `Generate a CNAB Credential Set.
+		Long: `Generate a named set of credentials.
 
 The first argument is the name of credential set you wish to generate. If not
 provided, this will default to the bundle name. By default, Porter will 
@@ -104,8 +107,9 @@ func buildCredentialsListCommand(p *porter.Porter) *cobra.Command {
 		format    printer.Format
 	}{}
 	cmd := &cobra.Command{
-		Use:   "list",
-		Short: "List credentials",
+		Use:    "list",
+		Short:  "List credentials",
+		Hidden: true,
 		PreRunE: func(cmd *cobra.Command, args []string) error {
 			var err error
 			opts.format, err = printer.ParseFormat(opts.rawFormat)
@@ -122,8 +126,9 @@ func buildCredentialsListCommand(p *porter.Porter) *cobra.Command {
 
 func buildCredentialsRemoveCommand(p *porter.Porter) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "remove",
-		Short: "Remove a Credential",
+		Use:    "remove",
+		Short:  "Remove a Credential",
+		Hidden: true,
 		PreRunE: func(cmd *cobra.Command, args []string) error {
 			return nil
 		},
@@ -137,8 +142,9 @@ func buildCredentialsRemoveCommand(p *porter.Porter) *cobra.Command {
 
 func buildCredentialsShowCommand(p *porter.Porter) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "show",
-		Short: "Show a Credential",
+		Use:    "show",
+		Short:  "Show a Credential",
+		Hidden: true,
 		PreRunE: func(cmd *cobra.Command, args []string) error {
 			return nil
 		},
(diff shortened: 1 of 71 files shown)


pkg/porter/credentials_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+import (
+	"testing"
+
+	"github.com/deislabs/duffle/pkg/bundle"
+	cnabprovider "github.com/deislabs/porter/pkg/cnab/provider"
+	"github.com/stretchr/testify/require"
+)
+
+type TestCNABProvider struct {
```
vdice: ooh these will be helpful 👍

pkg/porter/credentials.go
```diff
@@ -0,0 +1,90 @@
...
+		g.Name = args[0]
+	} else if len(args) > 1 {
+		return errors.Errorf("only one positional argument may be specified, the credential name, but multiple were received: %s", args)
+	}
+	return nil
+}
+
+func (p *Porter) GenerateCredentials(opts CredentialOptions) error {
+
+	//TODO make this work for either porter.yaml OR a bundle
```
vdice: +1

cmd/porter/credentials.go
```diff
@@ -0,0 +1,151 @@
+package main
+
+import (
+	"github.com/deislabs/porter/pkg/porter"
+	"github.com/deislabs/porter/pkg/printer"
+	"github.com/spf13/cobra"
+)
+
```
vdice: Do we want a `creds` alias?

carolynvs: Yeah seems like it would be nice to have both aliased because we all suck at typing.

cmd/porter/credentials.go
```diff
@@ -0,0 +1,151 @@
...
+func buildCredentialsAddCommand(p *porter.Porter) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "add",
+		Short: "Add Credential",
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			return nil
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			p.PrintVersion()
```
carolynvs: I suggest setting `Hidden: true` on anything that is wired but not implemented, just so that it doesn't show up in the helptext. You can still call it, but it won't be here confusing anyone using canary.

cmd/porter/credentials.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/deislabs/porter/pkg/porter"
+	"github.com/deislabs/porter/pkg/printer"
+	"github.com/spf13/cobra"
+)
+
+func buildCredentialsCommand(p *porter.Porter) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "credentials",
+		Aliases: []string{"cred"},
+		Short:   "Credentials commands",
```
carolynvs: This needs `Annotations: map[string]string{"group": "resource"}` so that it shows up nice and purdy in the help text when you are ready for that.

cmd/porter/credentials.go
```diff
@@ -0,0 +1,151 @@
...
+			return nil
+		},
+	}
+	return cmd
+}
+
+func buildCredentialsGenerateCommand(p *porter.Porter) *cobra.Command {
+	opts := porter.CredentialOptions{}
+	cmd := &cobra.Command{
+		Use:   "generate [name]",
```
carolynvs: nit: it should be `[NAME]` to be consistent with the rest of the helptext.

cmd/porter/credentials.go
```diff
@@ -0,0 +1,151 @@
...
+		Short: "Generate Credential Set",
+		Long: `Generate a CNAB Credential Set.
```
carolynvs: No need to call it a CNAB credential set here. Elsewhere in the helptext I've been calling it a named set of credentials if that helps give you a better long set of text.

cmd/porter/credentials.go
```diff
@@ -0,0 +1,151 @@
...
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			return opts.Validate(args)
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return p.GenerateCredentials(opts)
+		},
+	}
+
+	f := cmd.Flags()
+	f.BoolVar(&opts.Insecure, "insecure", false,
```
carolynvs: This should default to true now that I just went rogue and burned --insecure to the ground in that other PR. 😀

cmd/porter/credentials.go
```diff
@@ -0,0 +1,151 @@
...
+		"Allow working with untrusted bundles.")
+	f.StringVarP(&opts.File, "file", "f", "",
+		"Path to the CNAB definition to install. Defaults to the bundle in the current directory.")
+	f.BoolVar(&opts.DryRun, "dry-run", false,
```
carolynvs: Do we need a dry-run? I'm not sure why other than because duffle has it?

pkg/credentials/generator.go
```diff
@@ -0,0 +1,141 @@
...
+		cs.Credentials = append(cs.Credentials, c)
+	}
+
+	return cs, nil
+}
+
+func genEmptyCredentials(name string) (credentials.CredentialStrategy, error) {
+	return credentials.CredentialStrategy{
+		Name:   name,
+		Source: credentials.Source{Value: "EMPTY"},
```
carolynvs: Can we use TODO instead of EMPTY? Just to drive home that it's not filled in?

pkg/porter/credentials.go
carolynvs: Yup, per our goals from #320, all of our commands should stop taking bundle.json files as arguments when a porter bundle is present and work off of the manifest first. But for now this is a great first step! 💯

pkg/porter/credentials.go
```diff
@@ -0,0 +1,90 @@
...
+		Name:        name,
+		Credentials: bundle.Credentials,
+		Silent:      opts.Silent,
+	}
+	fmt.Fprintf(p.Out, "Generating new credential %s from bundle %s\n", genOpts.Name, bundle.Name)
+	fmt.Fprintf(p.Out, "==> %d credentials required for bundle %s\n", len(genOpts.Credentials), bundle.Name)
+
+	cs, err := credentials.GenerateCredentials(genOpts)
+	if err != nil {
+		//do something with the error
```
carolynvs: I vote to take it out for a steak dinner

cmd/porter/credentials.go
jeremyrickard: The dry run is neat if you care what's going to be created and want to see it. I found it useful when testing this out at least, so I could see what was going to end up in the file.

pkg/credentials/generator.go
jeremyrickard: sure!

pkg/porter/credentials.go
jeremyrickard: I think we should come up with a good way to resolve that stuff and then wire it all up. Part of me feels like we should still work with a "Bundle" object just so everything is consistent and we can easily use duffle, have the translation from porter.yaml -> Bundle (the struct) and bundle.json ->Bundle (the struct) happen in a single place and these can all jsut say context or whatever give me my bundle please and thank you. I figured we could get all this in, then go through and make the :magic: happen

jeremyrickard: lol i wanted this to be a reply to the comment re #320

carolynvs: I think we are saying the same thing? I was just referring the to the flag names and what the user sees when executing commands. 

That `--file` would be for porter.yaml and be defaulted automatically, and --cnab-file would be for bundle.json, etc ... there's more in the issue about how it should work. Then the code would go like you said from a porter.yaml to making sure the bundle.json is up to date, rebuild if necessary, then deal with the bundle from there on a single code path.

If we aren't saying the same thing, let's chat in person since this is going back and forth a bit. 😀

jeremyrickard: > I think we are saying the same thing? I was just referring the to the flag names and what the user sees when executing commands.
 
i think we are

jeremyrickard:
@carolynvs any outstanding issues? or can I merge this with @vdice's approval?

carolynvs:
Merge at will, pew pew.

PR #329, opened 2019-05-09
Add porter bundle list command

vdice:
Adds `porter bundle list` to list installed bundles, optionally with json/yaml outputs.

In doing so, I made some updates to `table.go` from the `printer` pkg, as it appeared previously we weren't actually utilizing the power of the `tabWriter` library for table output spacing.

```
  $ porter bundle list
NAME              CREATED               MODIFIED              LAST ACTION   LAST STATUS
porter-keyvault   Fri Apr 26 11:38:27   Fri Apr 26 11:38:41   install       success

 $ porter bundle list -o json
[
  {
    "Name": "porter-keyvault",
    "Created": "Fri Apr 26 11:38:27",
    "Modified": "Fri Apr 26 11:38:41",
    "Action": "install",
    "Status": "success"
  }
]

 $ porter bundle list -o yaml
- name: porter-keyvault
  created: Fri Apr 26 11:38:27
  modified: Fri Apr 26 11:38:41
  action: install
  status: success
```

Closes https://github.com/deislabs/porter/issues/284


cmd/porter/bundle.go
```diff
@@ -77,6 +79,35 @@ func buildBuildCommand(p *porter.Porter) *cobra.Command {
 	return cmd
 }
 
+func buildBundleListCommand(p *porter.Porter) *cobra.Command {
+	opts := struct {
```
carolynvs: Heads up that I made a mistake when I first did some of the original commands, defining the structs anonymously here, instead of as a type in pkg/porter like we do for other commands like Install.

It would be really nice at some point to fix these older commands to use the new and improved pattern, instead of adding more commands that use my old mistake. I'm fine with merging this, but if you feel motivated, as a follow up maybe we could fix this one and the other one you used as a pattern for this?

cmd/porter/bundle.go
```diff
@@ -77,6 +79,35 @@ func buildBuildCommand(p *porter.Porter) *cobra.Command {
...
+		rawFormat string
+		format    printer.Format
+	}{}
+
+	cmd := &cobra.Command{
+		Use:   "list",
+		Short: "list bundles",
```
carolynvs: Let's change this to be a bit more clear what we are listing. How about "List installed bundles"?

I see that used managed bundles down below which I really like. So maybe you have better wording to suggest? "List managed bundles"? Up to you but just "List bundles" I think is too vague, especially since duffle lists bundles in a different way than porter.

cmd/porter/bundle.go
```diff
@@ -77,6 +79,35 @@ func buildBuildCommand(p *porter.Porter) *cobra.Command {
...
+		Long:  `List all bundles managed by Porter`,
```
carolynvs: We can get a bit more verbose here. Maybe let them know that they can see bundles that have been installed? They may not immediately connect the word "managed" with something that they did with "porter install", you know?

cmd/porter/bundle_test.go
```diff
@@ -66,3 +66,34 @@ func TestValidateUninstallCommand(t *testing.T) {
...
+
+func TestValidateBundleListCommand(t *testing.T) {
+	testcases := []struct {
+		name      string
+		args      string
+		wantError string
+	}{
+		{"no args", "bundle list", ""},
+		{"output json", "bundle list -o json", ""},
+		{"invalid format", "bundle list -o wingdings", "invalid format: wingdings"},
```
carolynvs: haha wingdings

pkg/cnab/provider/list.go
```diff
@@ -0,0 +1,60 @@
...
+	case printer.FormatYaml:
+		return printer.PrintYaml(d.Out, condensedClaims)
+	case printer.FormatTable:
+		printClaimRow :=
+			func(v interface{}) []interface{} {
+				cl, ok := v.(CondensedClaim)
+				if !ok {
+					return nil
+				}
+				return []interface{}{cl.Name, cl.Created, cl.Modified, cl.Action, cl.Status}
```
carolynvs: It would be great to either in this PR or in a follow-up format the dates in a human friendly format. I'm cool with either finding a go format that looks less verbose or using a package like go humanize.

pkg/porter/cnab.go
```diff
@@ -18,6 +19,7 @@ type CNABProvider interface {
 	Install(arguments cnabprovider.InstallArguments) error
 	Upgrade(arguments cnabprovider.UpgradeArguments) error
 	Uninstall(arguments cnabprovider.UninstallArguments) error
+	List(opts printer.PrintOptions) error
```
carolynvs: I'm not sure if this needs to go on the CNAB provider interface. 🤔  The others were there because they were mapping to code that was all duffle implemented stuff, that later we may want to swap out with test stubs.

This is all porter's own code that doesn't need to be put into this interface. I'd be happy moving all the code out of provider/list.go into pkg/porter/list.go and removing this interface declaration.

pkg/printer/table.go
```diff
@@ -21,10 +21,24 @@ func PrintTable(out io.Writer, v interface{}, getRow func(row interface{}) []int
...
 	for i := 0; i < rows.Len(); i++ {
-		fmt.Fprintln(table, getRow(rows.Index(i).Interface())...)
+		fmt.Fprintln(table, tabify(getRow(rows.Index(i).Interface()))...)
 	}
 	return table.Flush()
 }
+
+// tabify is a helper function which takes a slice and injects tab characters
+// between each element such that tabwriter can work its magic
+func tabify(untabified []interface{}) []interface{} {
```
carolynvs: D'oh! I always use a better package than the stdlib `text/tabwriter` and completely forgot to do this.

I think I'll do a follow-up later to switch to a table writer package that I like better but thank you, this fixes the stdlib for now. 👍

cmd/porter/bundle.go
vdice: Ah, ok.  Definitely, I'll start a follow-up.

pkg/porter/cnab.go
vdice: Ah good call; moving out 👍

cmd/porter/bundle.go
vdice: Updated; let me know what you think... meanwhile, new ticket to address dissonance between bundles and clams in: https://github.com/deislabs/porter/issues/330

pkg/cnab/provider/list.go
vdice: Updated with a friendly format; let me know what you think.

cmd/porter/bundle.go
vdice: Updated....

cmd/porter/bundle.go
```diff
@@ -80,29 +80,30 @@ func buildBuildCommand(p *porter.Porter) *cobra.Command {
 }
 
 func buildBundleListCommand(p *porter.Porter) *cobra.Command {
-	opts := struct {
-		rawFormat string
-		format    printer.Format
-	}{}
+	opts := porter.ListOptions{}
```
carolynvs: Yay! Thanks for fixing my past mistakes for me. 🙇‍♀

cmd/porter/bundle.go
```diff
@@ -80,29 +80,30 @@ func buildBuildCommand(p *porter.Porter) *cobra.Command {
...
 
 	cmd := &cobra.Command{
 		Use:   "list",
-		Short: "list bundles",
-		Long:  `List all bundles managed by Porter`,
+		Short: "list installed bundles",
+		Long: `List all bundles installed by Porter.
```
carolynvs: 💯

pkg/porter/list.go
```diff
@@ -1,10 +1,72 @@
...
-// ListBundles lists bundles using the provided printer.PrintOptions
+// ListOptions represent options for a bundle list command
+type ListOptions struct {
+	RawFormat string
+	Format    printer.Format
+}
+
+const (
+	// TimeFormat is used to generate a human-readable representation of a raw time.Time value
+	TimeFormat = "Mon Jan _2 15:04:05"
```
carolynvs: 👍 

Seriously, I'm so going to make a prettytime pkg now.

PR #331, opened 2019-05-10
ref(*): remove use of anonymous options where applicable

vdice:
A quick follow-up refactor spurred by https://github.com/deislabs/porter/pull/329

### DIFF
--- a/cmd/porter/mixins.go
+++ b/cmd/porter/mixins.go
@@ -26,25 +26,23 @@ func buildMixinsCommand(p *porter.Porter) *cobra.Command {
 }
 
 func buildMixinsListCommand(p *porter.Porter) *cobra.Command {
-	opts := struct {
-		rawFormat string
-		format    printer.Format
-	}{}
+	opts := porter.PrintMixinsOptions{}
+
 	cmd := &cobra.Command{
 		Use:   "list",
 		Short: "List installed mixins",
 		PreRunE: func(cmd *cobra.Command, args []string) error {
 			var err error
-			opts.format, err = printer.ParseFormat(opts.rawFormat)
+			opts.Format, err = printer.ParseFormat(opts.RawFormat)
 			return err
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return p.PrintMixins(printer.PrintOptions{Format: opts.format})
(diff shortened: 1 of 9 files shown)


cmd/porter/mixins.go
```diff
@@ -26,25 +26,23 @@ func buildMixinsCommand(p *porter.Porter) *cobra.Command {
...
 		Short: "List installed mixins",
 		PreRunE: func(cmd *cobra.Command, args []string) error {
 			var err error
-			opts.format, err = printer.ParseFormat(opts.rawFormat)
+			opts.Format, err = printer.ParseFormat(opts.RawFormat)
 			return err
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return p.PrintMixins(printer.PrintOptions{Format: opts.format})
+			return p.PrintMixins(printer.PrintOptions{Format: opts.Format})
```
carolynvs: The intent was that instead of matching the opts to the cli command (MixinList), was to match the opts to the porter command being executed (PrintMixins). So this would be `p.PrintMixins(opts)`, where opts would be of type `PrintMixinsOptions`.

cmd/porter/mixins.go
```diff
@@ -26,25 +26,23 @@ func buildMixinsCommand(p *porter.Porter) *cobra.Command {
...
 		},
 	}
 
-	cmd.Flags().StringVarP(&opts.rawFormat, "output", "o", "table",
-		"Output format, allowed values are: table, json")
+	cmd.Flags().StringVarP(&opts.RawFormat, "output", "o", "table",
+		"Output format, allowed values are: table, json, yaml")
```
carolynvs: yaml isn't a supported format for pkg/porter/mixins.go PrintMixins

vdice: Oops; you're right.  I added yaml support to `PrintMixins` -- though I can remove support and option if we don't wish to provide this format for this command.

cmd/porter/mixins.go
vdice: Ah, thank you.  Updated.

cmd/porter/mixins.go
carolynvs: No that's cool, thanks!

PR #332, opened 2019-05-10
ref(Makefile): mandate PORTER_HOME value for make; no build prereq for install

vdice:
* Ensure `PORTER_HOME` is set to `bin` for all applicable make targets (not yet setup to run with other, user-defined values)

* Remove `build` target prerequisite from `install` (shortens dev loop for, say, only building the porter binary and installing locally, as in `make build-porter install`)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ VERSION ?= $(shell git describe --tags 2> /dev/null || echo v0)
 PERMALINK ?= $(shell git name-rev --name-only --tags --no-undefined HEAD &> /dev/null && echo latest || echo canary)
 
 KUBECONFIG  ?= $(HOME)/.kube/config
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -115,7 +115,7 @@ publish: prep-install-scripts
 	fi
 	az storage blob upload-batch -d porter/$(PERMALINK) -s bin/$(VERSION)
 
-install: build
+install:
```
carolynvs: Should this be build-porter then instead of dropping the requirement completely? Either way please update the CONTRIBUTING.md file to let people know how to run `make install` properly as that target is documented in there.

vdice: I've updated the CONTRIBUTING doc.  I'd still vote for decoupling `install` with any of the `build*` targets -- what do you think?

PR #334, opened 2019-05-11
Use relative times for claims modified recently 

carolynvs:
Use relative times for claims modified recently

When the claim has been touched in the last day, instead of showing the
datetime stamp, just show "1 min ago" or "5 hours ago". Otherwise if
it's been more than a day, just show the date, 2019-10-03.

Also sort the list with the most recently modified claims at the top.

Here is the sample output:

```console
$ porter bundle list
NAME    CREATED          MODIFIED         LAST ACTION   LAST STATUS
farts   29 minutes ago   29 minutes ago   install       success
HELLO   2019-05-10       2019-05-10       install       success
```

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -96,6 +96,14 @@
   revision = "3a771d992973f24aa725d07868b467d1ddfceafb"
 
 [[projects]]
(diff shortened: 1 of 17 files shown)


carolynvs:
@vdice , you totally tricked me into doing this. 😀

PR #336, opened 2019-05-13
feat(brigade.js): run specific check run on check_run:rerequested event

vdice:
Refactor to support running a specific check run when handling a `check_run:rerequested` event.

### DIFF
--- a/brigade.js
+++ b/brigade.js
@@ -13,8 +13,7 @@ const projectOrg = "deislabs";
 
 events.on("check_suite:requested", runSuite);
 events.on("check_suite:rerequested", runSuite);
-// TODO: we should determine which check run is being requested and *only* run this
-events.on("check_run:rerequested", runSuite);
+events.on("check_run:rerequested", runCheck);
 
 events.on("exec", (e, p) => {
   Group.runAll([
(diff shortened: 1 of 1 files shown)


brigade.js
```diff
@@ -134,20 +133,50 @@ function publish(e, p) {
...
+
+  name = payload.body.check_run.name;
+  check = checks[name];
+
+  if (typeof check !== 'undefined') {
+    checkRun(e, p, check.runFunc, check.description)
+      .catch(e => {console.error(e.toString())});
+  } else {
+    err = new Error(`No check found with name: ${name}`);
+    // TODO: remove this console.error statement once Brigade logs err.message when thrown
```
carolynvs: TODO's are like hydra, for everyone one you slay, another shall take its place. 😂 

jk, though, this is great 👍

brigade.js
```diff
@@ -134,20 +133,50 @@ function publish(e, p) {
...
+    console.error(err.message);
+    throw err;
+  }
+}
+
 // Here we add GitHub Check Runs, which will run in parallel and report their results independently to GitHub
 function runSuite(e, p) {
-  if (e.revision.ref.includes("refs/heads/master")) {
+  if (e.revision.ref == "master" ) {
```
carolynvs: How come you can replace that check for a fully qualified ref with an equality check for just the branch name? What changed?

vdice: @carolynvs thanks for asking this -- I needed the push to get to the bottom of it.

It appears to be due to GH API differences.  Classically, indeed, the `e.revision.ref` takes the form of `refs/heads/master` for the majority of GH events, such as a [push event](https://developer.github.com/v3/activity/events/types/#pushevent).  The API call that the Brigade GH App uses is [GetRef()](https://github.com/brigadecore/brigade-github-app/blob/master/pkg/webhook/github.go#L239).  (We use the Brigade GH App as our main Brigade gateway to process incoming GH events/webhooks and produce corresponding Brigade events.)

However, for check suite and check run events, we only have access to a `head_branch` value, as seen in the [check suite event](https://developer.github.com/v3/activity/events/types/#checksuiteevent) doc.  The API call that the Brigade GH App uses is thus [GetHeadBranch()](https://github.com/brigadecore/brigade-github-app/blob/master/pkg/webhook/github.go#L129) . This `head_branch` value is `master` for said check suite/run events triggered from a merge to master.

carolynvs: Silly GitHub. 😀  Makes perfect sense, though one more question.

`runSuite` was working before, right? So since we didn't change how we called that function (No? Or did I miss that part...) I don't see where we switched from using one API to another, weren't we always using the check suite API here and would have always then been doing a bad comparison then? 🤔

vdice: Ah, good question.  In fact, unfortunately, our `if master branch` section of `runSuite` wasn't/isn't quite working.  Note the check statuses for master by navigating to https://github.com/deislabs/porter/branches (and then hover over the check status icon for `master` -- I can't seem to locate an actual separate check statuses page.) 

We see the following checks:

Screen Shot 2019-05-14 at 12 45 16 PM

Which just represent our default set as written [here](https://github.com/deislabs/porter/blob/master/brigade.js#L143-L147) -- meaning, it said, oh, `master` doesn't include `refs/heads/master`, so I'll just run the regular set, etc.

carolynvs: D'oh! Yes, I see it all now. 👍 Thank you for explaining it.

vdice:
(amended commit just to add link to Brigade issue per TODO added in this PR)

PR #342, opened 2019-05-14
Stop on error during install check

carolynvs:
The inline script wasn't stopping on errors, so I'm using
a real bash script. Here's hoping it works better.


carolynvs:
Looks weird but the failure is good. Means that it's working, we need to fix the publish for helm and azure too. Onwards!

PR #347, opened 2019-05-15
Added some words about spec compatibility

jeremyrickard:
Added some words about spec compat and linked to CNAB 1.0 milestone

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,6 @@ Porter go in lots of directions! Here are our top 4 goals at the moment:
 
     Make it easy for anyone to create and distribute mixins that porter can discover and install.
 
-4. CNAB Specification Compliance - Milestone TBC
+4. CNAB Specification Compliance - Milestone [CNAB 1.0](https://github.com/deislabs/porter/milestone/12)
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -36,3 +36,7 @@ Porter go in lots of directions! Here are our top 3 goals at the moment:
 3. Mixin Distribution - Milestone TBC
 
     Make it easy for anyone to create and distribute mixins that porter can discover and install.
+
+4. CNAB Specification Compliance - Milestone TBC
```
carolynvs: Can we replace Milestone TBC with a link to the  CNAB 1.0 milestone?

README.md
```diff
@@ -36,3 +36,7 @@ Porter go in lots of directions! Here are our top 3 goals at the moment:
...
+
+    As the [CNAB specification](https://github.com/deislabs/cnab-spec) moves toward 1.0, update Porter to be compliant with the spec. Currently, if you build a bundle with Porter, you'll be able to install it with Porter. There are some gaps with the spec that limit compatibility with other CNAB tooling. See the [CNAB 1.0 Milestone](https://github.com/deislabs/porter/milestone/12) for more information on these gaps.
```
carolynvs: This is more of a question for my own curiosity than anything else. Have we seen any porter bundles that can't be installed with duffle? Just wondering if we have a concrete scenario that we know won't work, or just a "hey that's not implemented in both" but it's not clear how that would impact things.

jeremyrickard: Probably not since duffle basically has the same limitations wrt to parameter definitions. Anyone that has updated and validated schema will fail. Not sure how far along the Pivotal and Docker tools are, sonwe need to be cognizant of those, since porter and duffle aren’t the only tools.

PR #350, opened 2019-05-15
docs(cli.md): add bundle list command

vdice:
* Adds Bundle List command
* Adjusts formatting for Create and Build commands

Closes https://github.com/deislabs/porter/issues/348

### DIFF
--- a/docs/content/cli.md
+++ b/docs/content/cli.md
@@ -14,8 +14,9 @@ description: Porter CLI Commands Explained
 This command is aliased and is available both as `porter create`
 and `porter bundle create`.
 
-```console $ porter create --help Create a bundle. This generates a porter
-manifest, porter.yaml, and the CNAB run script in the current directory.
+```console
+ $ porter create --help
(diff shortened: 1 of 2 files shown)


docs/content/cli.md
```diff
@@ -196,6 +197,33 @@ Global Flags:
...
 
+### Bundles List
+
+This command is available both as `porter bundle list` and `porter bundles list`.
+
+```console
+ $ porter bundle list --help
+List all bundles installed by Porter.
+
+A listing of bundles currently installed by Porter will be provided, along with metadata such as creation time, last action, last status, etc.
```
carolynvs: Can you hard wrap the help text line here so that it fits into the text box on the doc page? You can see about where I was wrapping it for the other commands.

Otherwise people have to sidescroll to read the text which is awkward.

Here's the preview of your helptext on the website so you can see what I'm talking about: https://deploy-preview-350--porter.netlify.com/cli/#bundles-list, if you run `make docs-preview` you can try it out locally first.

vdice: Updated w/ a hard wrap and also added an entry to the command in the `config.toml` file.

PR #359, opened 2019-05-28
feat(brigade.js): handle issue comment events

vdice:
Adding ability to handle issue/PR comment events, which utilizes a recent [Brigade-Github-App feature](https://github.com/brigadecore/brigade-github-app/pull/56).  This is still a pre-release feature, so hoping to help test/dogfood in this Porter project.

This feature may perhaps be most useful as a way to trigger CI for PRs that aren't automatically approved to run.  (Note that the comment author must be in the approved list for an action to actually take place.)

This can serve as an exercise on how we may do the same in other deislabs projects.  See also the [issue_comment_checks.js](https://github.com/brigadecore/brigade-github-app/blob/master/docs/examples/issue_comment_checks.js) example.

### DIFF
--- a/brigade.js
+++ b/brigade.js
@@ -14,6 +14,8 @@ const projectOrg = "deislabs";
 events.on("check_suite:requested", runSuite);
 events.on("check_suite:rerequested", runSuite);
 events.on("check_run:rerequested", runCheck);
+events.on("issue_comment:created", handleIssueComment);
+events.on("issue_comment:edited", handleIssueComment);
 
 events.on("exec", (e, p) => {
   Group.runAll([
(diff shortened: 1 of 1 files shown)


brigade.js
```diff
@@ -178,6 +180,29 @@ function runSuite(e, p) {
...
+// text and determining whether or not to trigger a corresponding action
+function handleIssueComment(e, p) {
+  payload = JSON.parse(e.payload);
+
+  // Extract the comment body and trim whitespace
+  comment = payload.body.comment.body.trim();
+
+  // Here we determine if a comment should provoke an action
+  switch(comment) {
+    case "/test":
```
vdice: It's totally up to us (or, in general, each `brigade.js` author) on what sort of comments we want to designate as official phrases/words to trigger actions.

In this example, we use `/test` to trigger the entire default test/build suite, but we could use a different phrase.

We might like to have cases for each particular check run _or_ we may decide that the one, do-all comment to kick off the entire suite is all we need (since, remember, once the suite is run, any failing checks can be re-run via the Check Suite UI on a given PR.)

carolynvs: Since we also have `/azp run` to trigger the azure pipeline, can we call it `/brig run` to help people remember?

vdice: Good call!  Updated.  Only have the `/brig run` handler for now... as you've mentioned, we can tweak/update after seeing how we use it day-to-day...

PR #362, opened 2019-05-29
Prevent using Dockerfile as the template name

### LINKED ISSUE #355: [Funny bug]: Porter build adds cnab copy everytime
As the title says, this is a funny bug 😄 
If a `porter.yaml` has the `dockerfile` set, then everytime we do a `porter build`, it will had a new instance for copying the `cnab` directory into the `dockerfile`:
image

I suddenly ended with 5 additional layers to my docker image and didn't understand why 😅 

Hope this helps

carolynvs:
Since Porter generates a Dockerfile during porter build, the template
used for the Dockerfile should not be named Dockerfile (or dockerfile
because case-insensitive file systems...) so that the template isn't
overwritten.

Fixes #355

### DIFF
--- a/pkg/config/manifest.go
+++ b/pkg/config/manifest.go
@@ -262,6 +262,10 @@ func (c *Config) LoadDependencies() error {
 func (m *Manifest) Validate() error {
 	var result error
 
+	if strings.ToLower(m.Dockerfile) == "dockerfile" {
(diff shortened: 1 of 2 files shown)


nunix:
Thank You ⭐️ 

This will prevent another 20 layers fresh container build 😲

PR #368, opened 2019-05-30
updated buildDockerfile to include execute permissions on run command

### LINKED ISSUE #369: Porter doesn't fix the execution bit on existing run scripts with it unset
If a run script\file contained in a porter bundle definition does not have x bit set then when an action is run the invocation image will fail

---

When a user modifies the run script created by porter and unsets or otherwise loses the execution bit on the run script, porter doesn't reapply it properly.

This is a bug in how we are using [ioutil.Writefile](https://golang.org/pkg/io/ioutil/#WriteFile) during porter build. The permissions that we are applying when writing files to contents of the local cnab directory aren't being applied when the files already exist. We may have other problems with how we are using that function elsewhere.

simongdavies:
Added a chmod 755 for /cnab/app/run to generated DockerFile fixes #369

### DIFF
--- a/pkg/porter/build.go
+++ b/pkg/porter/build.go
@@ -66,7 +66,7 @@ func (p *Porter) buildDockerfile() ([]string, error) {
 
 	mixinLines, err := p.buildMixinsSection()
 	if err != nil {
-		return nil, errors.Wrap(err, "error generating Dockefile content for mixins")
+		return nil, errors.Wrap(err, "error generating Dockerfile content for mixins")
 	}
 	lines = append(lines, mixinLines...)
 
@@ -76,7 +76,7 @@ func (p *Porter) buildDockerfile() ([]string, error) {
 		lines = append(lines, p.buildCNABSection()...)
 		lines = append(lines, p.buildPorterSection()...)
 	}
-	lines = append(lines, p.buildCMDSection())
+	lines = append(lines, p.buildCMDSection()...)
 
 	for _, line := range lines {
 		fmt.Fprintln(p.Out, line)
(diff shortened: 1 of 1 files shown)


carolynvs:
Hey Simon, when you open PRs that aren't linked to an issue would you please include the problem that you are solving? The integration tests are all passing, so I'm wondering what bug I'm unaware of.

simongdavies:
@carolynvs  Sorry - I've updated it and added a related issue.

carolynvs:
That issue just restates what I could deduce from your PR. Can you include some info on the regression that you are seeing that prompted you to submit the PR in the first place? Repro steps or a log of a failed run, the example bundle that caused it to not work, etc.

Giving you a hard time because I'm working off of tip of master and can't reproduce. 😀  So I want to make sure that we properly understand the problem before adding in a fix.

```
root@d47e00611957:/cnab/app# ls -la
total 49368
drwxr-xr-x 1 root root     4096 May 30 22:24 .
drwxr-xr-x 1 root root     4096 May 30 22:24 ..
-rw-r--r-- 1 root root      181 May 30 22:24 .dockerignore
-rw-r--r-- 1 root root       17 May 30 22:24 .gitignore
-rw-r--r-- 1 root root      296 May 30 22:24 Dockerfile
-rw-r--r-- 1 root root      564 May 30 22:24 Dockerfile.tmpl
-rw-r--r-- 1 root root     1522 May 30 22:24 README.md
drwxr-xr-x 3 root root     4096 May 30 22:24 mixins
-rwxr-xr-x 1 root root 50508400 May 30 22:24 porter-runtime
-rw-r--r-- 1 root root      988 May 30 22:24 porter.yaml
-rwxr-xr-x 1 root root       79 May 30 22:24 run
```

simongdavies:
OK, I found the issue by building a bundle that had a run script without the x bit set and then trying to run it in ACI , I was actually using duffle to run a bundle created by porter, I will work on a simple repro tomorrow

carolynvs:
Thanks for the repo steps. This is a bug in how we are using `ioutil.Writefile` during porter build. The permissions that we are applying when writing files to contents of the local cnab directory aren't being applied when the files already exist. We may have other problems with how we are using that function elsewhere. Let me chat with Jeremy on how we want to fix this quick.

carolynvs:
Closing this PR in favor of a different solution, see https://github.com/deislabs/porter/issues/369#issuecomment-499548859.

PR #375, opened 2019-06-01
feat(credentials): add credentials list command

vdice:
* adds a credentials list command for porter
* renames porter's credentials generator pkg from `credentials` to `credentialsgenerator` to avoid naming confusion with duffle's/cnab-go's `credentials` pkg.  ping @jeremyrickard on this change for thoughts, etc.

A few TODOs remain, but I wanted to get early feedback on this one.

(Note: planning on tackling the table printing revamp TODO either in https://github.com/deislabs/porter/issues/371 or a separate ticket, as this will affect multiple pre-existing commands.)

Closes https://github.com/deislabs/porter/issues/366

### DIFF
--- a/cmd/porter/credentials.go
+++ b/cmd/porter/credentials.go
@@ -9,7 +9,7 @@ import (
 func buildCredentialsCommand(p *porter.Porter) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:         "credentials",
-		Aliases:     []string{"cred"},
+		Aliases:     []string{"credential", "cred", "creds"},
 		Annotations: map[string]string{"group": "resource"},
 		Short:       "Credentials commands",
 	}
@@ -102,24 +102,31 @@ will then provide it to the bundle in the correct location. `,
 }
 
 func buildCredentialsListCommand(p *porter.Porter) *cobra.Command {
-	opts := struct {
-		rawFormat string
-		format    printer.Format
-	}{}
+	opts := porter.ListOptions{}
+
 	cmd := &cobra.Command{
-		Use:    "list",
-		Short:  "List credentials",
-		Hidden: true,
+		Use:     "list",
+		Aliases: []string{"ls"},
+		Short:   "List credentials",
+		Long: `List credentials available to Porter.
+
+A listing of credentials currently available to Porter will be provided, along with metadata such as modification time, etc.
+		
+Optional output formats include json and yaml.`,
+		Example: `  porter credentials list [-o table|json|yaml]`,
 		PreRunE: func(cmd *cobra.Command, args []string) error {
 			var err error
-			opts.format, err = printer.ParseFormat(opts.rawFormat)
+			opts.Format, err = printer.ParseFormat(opts.RawFormat)
 			return err
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
-			p.PrintVersion()
-			return nil
+			return p.ListCredentials(printer.PrintOptions{Format: opts.Format})
 		},
 	}
+
+	f := cmd.Flags()
+	f.StringVarP(&opts.RawFormat, "output", "o", "table",
+		"Specify an output format.  Allowed values: table, json, yaml")
 
 	return cmd
 }
--- a/cmd/porter/credentials.go
+++ b/cmd/porter/credentials.go
@@ -108,11 +108,7 @@ func buildCredentialsListCommand(p *porter.Porter) *cobra.Command {
 		Use:     "list",
 		Aliases: []string{"ls"},
 		Short:   "List credentials",
-		Long: `List credentials available to Porter.
-
-A listing of credentials currently available to Porter will be provided, along with metadata such as modification time, etc.
-		
-Optional output formats include json and yaml.`,
+		Long:    `List named sets of credentials defined by the user.`,
 		Example: `  porter credentials list [-o table|json|yaml]`,
 		PreRunE: func(cmd *cobra.Command, args []string) error {
 			var err error
--- a/pkg/porter/credentials.go
+++ b/pkg/porter/credentials.go
@@ -44,24 +44,27 @@ func (p *Porter) fetchCredentials() (*CredentialsFileList, error) {
 
 	credentialsFiles := CredentialsFileList{}
 	if ok, _ := p.Context.FileSystem.DirExists(credsDir); ok {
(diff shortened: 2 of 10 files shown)


nunix:
Thank you so much, you read my mind, I was in need to list the credentials created ⭐️ 

Looking for this one as I'm finalizing my next demo with porter (and yes, I did stay in the CNAB context this time 😈 )

pkg/porter/credentials.go
```diff
@@ -2,16 +2,100 @@ package porter
...
+func (p *Porter) ListCredentials(opts printer.PrintOptions) error {
+	dir, err := p.Config.GetCredentialsDir()
+	if err != nil {
+		return errors.Wrap(err, "unable to determine credentials directory")
+	}
+
+	// TODO: should this go somewhere else?
+	// i.e., where should logic for init-ing dirs needed by porter go?
+	// e.g., on 'porter creds list' with a fresh porter install/home
+	if ok, _ := p.Context.FileSystem.DirExists(dir); !ok {
```
carolynvs: I would prefer that a readonly command `list` not have a side-effect, creating a directory. One way to deal with this a bit easier is to have a function that does the walk and returns a slice of fileinfo's and a error.  Similar to how we have a function that gets the schema and another function that prints the schema. The retrieval isn't tied to the printing.

Then we could return an empty list when the directory doesn't exist.

What do you think?

pkg/porter/credentials.go
```diff
@@ -2,16 +2,100 @@ package porter
...
+	}
+	sort.Sort(sort.Reverse(credentialsFiles))
+
+	switch opts.Format {
+	case printer.FormatJson:
+		return printer.PrintJson(p.Out, credentialsFiles)
+	case printer.FormatYaml:
+		return printer.PrintYaml(p.Out, credentialsFiles)
+	case printer.FormatTable:
+		// TODO: update table printing
```
carolynvs: I'm not clear on what this TODO is talking about?

pkg/porter/credentials.go
carolynvs: If it's not clear, I'm suggesting that we break this function up into 2 functions. One that traverses the credentials directory that builds up a slice of the credentialsets and another than does the printing. 😀

pkg/porter/credentials.go
vdice: This TODO was a note to remind that we intend to move to a different table printing method, as mentioned in https://github.com/deislabs/porter/issues/371.  I've removed the note for now, since we're tracking in that separate issue.

pkg/porter/credentials.go
vdice: Good call; split into two functions.  Lemme know how this looks when convenient.

cmd/porter/credentials.go
```diff
@@ -102,25 +102,32 @@ will then provide it to the bundle in the correct location. `,
...
 	cmd := &cobra.Command{
-		Use:    "list",
-		Short:  "List credentials",
-		Hidden: true,
+		Use:     "list",
+		Aliases: []string{"ls"},
+		Short:   "List credentials",
+		Long: `List credentials available to Porter.
+
+A listing of credentials currently available to Porter will be provided, along with metadata such as modification time, etc.
```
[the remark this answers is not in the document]
vdice: Thank you; updated.

pkg/porter/credentials.go
```diff
@@ -35,48 +35,51 @@ func (l CredentialsFileList) Less(i, j int) bool {
...
-			}
-			if err = yaml.Unmarshal(data, credSet); err != nil {
-				return errors.Wrap(err, "unable to unmarshal credential set")
+	if ok, _ := p.Context.FileSystem.DirExists(credsDir); ok {
+		err = p.Context.FileSystem.Walk(credsDir, func(path string, info os.FileInfo, err error) error {
+			if !info.IsDir() {
+				credSet := &credentials.CredentialSet{}
+				data, err := p.Context.FileSystem.ReadFile(path)
+				if err != nil {
+					return errors.Wrap(err, fmt.Sprintf("unable to load credential set from %s:\n%s", path, err))
```
carolynvs: This is double printing the error `err`. The line should be 

```
return errors.Wrapf(err, "unable to load credential set from %s", path)
```

pkg/porter/credentials.go
```diff
@@ -35,48 +35,51 @@ func (l CredentialsFileList) Less(i, j int) bool {
...
+				}
+				if err = yaml.Unmarshal(data, credSet); err != nil {
+					return errors.Wrap(err, "unable to unmarshal credential set")
```
carolynvs: This error isn't saying which file can't be unmarshaled. I'd change it to include the credential set name in the error message.

pkg/porter/credentials.go
```diff
@@ -35,48 +35,51 @@ func (l CredentialsFileList) Less(i, j int) bool {
...
+				}
+				credentialsFiles = append(credentialsFiles,
+					CredentialsFile{Name: credSet.Name, Modified: info.ModTime()})
 			}
-			credentialsFiles = append(credentialsFiles,
-				CredentialsFile{Name: credSet.Name, Modified: info.ModTime()})
+			return nil
+		})
+		if err != nil {
+			return &CredentialsFileList{}, errors.Wrap(err, "encountered error while listing credentials")
```
carolynvs: It would be really frustrating if there was a problem with a single credential, and that preventing anything from being printed out.

I would lean towards (maybe in a follow up PR), is when it's an error reading or unmarshaling a credential, print the error to `p.Err`, potentially wrapped behind a p.Debug check. That way they could see the see the problem, and filtering it using 2>, and not have that prevent them from seeing their other credentials.

pkg/porter/credentials.go
vdice: If unmarshal-ing fails, `credSet.Name` never gets populated and so we don't have a name to print -- how about printing the filepath instead?

pkg/porter/credentials.go
vdice: Oops; good call; updated

pkg/porter/credentials.go
vdice: Good point.  I've updated the logic to log to `p.Err` if debug is set for both unmarshal and `ReadFile` failures and return `nil` in both cases, such that listing may proceed.  Added a test as well.  How does this look?

pkg/porter/credentials.go
carolynvs: We can print the name of the file, which is the same as `credSet.Name`.

Issue #322, opened 2019-05-06
Porter Bundle Cache


carolynvs:
Duffle calls this a local store but I plan on treating this a bit differently and encouraging users to think of it like a cache (to be ignored for the most part unless things go sideways or they want to micromanage stuff).

## Purpose
Cache of bundles that you have retrieved from registries as a side effect when you executed commands such as `porter install`.

### List
List all of the cached bundles
```
porter bundle cache list
```

### Show
Show the metadata for a cached bundle, in the specified output.
```
porter bundle cache show TAG [--output json|pretty|yaml]
```

### Clear
Removes all bundles from the cache.
```
porter bundle cache clear
```

### Add
Pulls a bundle by its tag from the registry, thick or thin, and unpacks it into the cache. 

```
porter bundle cache add [--tag]
```

NOTE: I am not using the term pull because it doesn't have a symmetrical verb for removing it from the cache, and you don't push from the cache either. I don't want to set up expectations that this command works like `docker pull`.

### Remove
Removes a single bundle from the cache by its tag.

```
porter bundle cache remove TAG
```

## Updated Commands
 
**This has been implemented**

### Install
Install a bundle using a definition from a registry.

```
porter install --tag TAG
```

### Upgrade
Upgrade a bundle using a definition from a registry.

```
porter upgrade --tag TAG
```

### Uninstall
Uninstall a bundle using a definition from a registry.

```
porter uninstall --tag TAG
```

### Create

Create a new bundle from an existing one, checking first that the current directory doesn't already have a bundle in it.

```
porter create --copy TAG
```

This seems useful? We need to work more on what this command should do when the source bundle isn't a porter bundle. Like just pop out the bundle.json?

---
# Deep Thoughts
* Should we always be identifying a bundle by tag? Is it ever appropriate to just operate by the bundle name (as defined in the bundle.json)?
* Duffle has more commands and flags to let you interact with the local store in more ways that outlined here. For now I want to start with just these. Mostly because these make the purpose clear: it's a cache that you don't directly use or interact with. The main mechanism for working with porter when iterating on a bundle is through the file system (i.e. in the directory of a porter bundle). The remaining time, you will use `--tag` and the cache is transparently used.

nano2007:
Somewhat unrelated, but since this mentions thick bundles: is there a way in porter to create a "thick" bundle (we are looking specifically for helm/k8s/docker)?

carolynvs:
[...]

PR #379, opened 2019-06-04
Publish bundles and Allow Install From Tag

jeremyrickard:
This PR starts down the road of implementing a bundle cache per #322. It doesn't fully implement that issue, but lays some groundwork.

The major addition here is that users can now `porter publish` a bundle to an OCI registry using [cnab-to-oci](https://github.com/docker/cnab-to-oci) and install a bundle from an OCI registry using a new `--tag` flag on `porter install`:

```
porter install --tag jeremyrickard/porter-kube-bundle:1.0 -c kool-kred
installing HELLO...
executing porter install configuration from /cnab/app/porter.yaml
Install Hello World App
```

When the `--tag` flag is provided, Porter will attempt to locate the specified bundle in the new ~/.porter/cache directory using the hashed tag name as a directory key. If it is not found, it is pulled with `cnab-to-oci`, stored in the cache, and the new path in the cache is returned and used to update the install opts. If it is found, the cached path is returned. 

Sorry this is rather large.


cmd/porter/bundle.go
```diff
@@ -149,7 +150,10 @@ For instance, the 'debug' driver may be specified, which simply logs the info gi
 		"Credential to use when installing the bundle. May be either a named set of credentials or a filepath, and specified multiple times.")
 	f.StringVarP(&opts.Driver, "driver", "d", "docker",
 		"Specify a driver to use. Allowed values: docker, debug")
-
+	f.StringVarP(&opts.Tag, "tag", "t", "",
+		"Install from a bundle in an OCI registry specified by the given tag")
+	f.BoolVar(&opts.InsecureRegistry, "insecure-registry", false,
+		"Treat the given bundle registry as insecure")
```
carolynvs: Can we improve this error message a bit? Based on what you told me, I think it means "Allow connecting to a registry that isn't secured with TLS certificates" or something like that, no?

cmd/porter/bundle.go
```diff
@@ -288,7 +291,7 @@ func buildBundlePublishCommand(p *porter.Porter) *cobra.Command {
 
 	f := cmd.Flags()
 	f.StringVarP(&opts.File, "file", "f", "", "Path to the Porter manifest. Defaults to `porter.yaml` in the current directory.")
-	f.StringVarP(&opts.Tag, "tag", "t", "", "Tag to apply to the CNAB bundle. Defaults to the `Tag` value currently in the Porter manifest.")
+	f.BoolVar(&opts.InsecureRegistry, "insecure", false, "Treat the registry as insecure.")
```
carolynvs: Just a note that whatever we change the flag description to above, let's change this one as well.

docs/content/distributing-bundles.md
```diff
@@ -3,11 +3,75 @@ title: Distributing Bundles
...
 
-This is a placeholder doc page. See our [contributing guide][contrib] 
-if you would like to add content for this page.
+Once you have built a bundle with Porter, the next step is to share the bundle and invocation image so others can use it. Porter uses OCI (Docker) registries to share both CNAB bundle manifest and invocation images.
 
-# TODO
+## Preparing For Bundle Publishing
 
-* This needs to wait until bundle registries is figured out
+Before you can publish your bundle, you must first run a `porter build` command. This will create the invocation image so it can be pushed to an OCI (Docker) registry along with your CNAB bundle manifest. It's a good idea to work with your bundle and test it locally before you decide to publish it to a registry.
```
carolynvs: nit: "before you decide to publish" -> "before you publish"

docs/content/distributing-bundles.md
```diff
@@ -3,11 +3,75 @@ title: Distributing Bundles
...
 
-[contrib]: https://github.com/deislabs/porter/blob/master/CONTRIBUTING.md#documentation
\ No newline at end of file
+## Bundle Publish
+
+Once you are satisfied with the bundle, the next step is to publish the bundle! Bundle publishing involves pushing the both the invocation image and the CNAB bundle manifest to an OCI registry. Porter uses [docker tags](https://docs.docker.com/engine/reference/commandline/tag/) for both nvocation images and CNAB bundle manifests. These are defined in your `porter.yaml` file:
```
carolynvs: spelling: "nvocation images" -> "invocation images"

docs/content/distributing-bundles.md
```diff
@@ -3,11 +3,75 @@ title: Distributing Bundles
...
+
+```yaml
+name: kube-example
+version: 0.1.0
+description: "An example Porter bundle using Kubernetes"
+invocationImage: deislabs/porter-kubernetes:latest
+tag: deislabs/porter-kube-bundle:1.0
+```
+
+This YAML snippet indicates that the invocation image will be built and tagged as `deislabs/porter-kubernetes:latest`. The first part of this reference, `deislabs` indicates the registry that the invocation image should eventually be published to. The `porter-kubernetes` segment identifies the iamge, while the `:latest` portion denotes a specific version. Much like the `invocationImage` attribute is used to control the name of resulting Docker invocation image, the `tag` attribute is used to specify the name and location of the resulting CNAB bundle. In both cases, when you are ready to publish your bundle, it would be a good idea to provide specific versions for both of these, such as `v1.0.0`. We recommend using [semantic versioning](https://semver.org/) for both the invocation image and the bundle. We also recommend specifying the same registry for both, in order to simplify access to your bundle and invocation image by end users.
```
carolynvs: spelling: "iamge"

docs/content/distributing-bundles.md
```diff
@@ -3,11 +3,75 @@ title: Distributing Bundles
...
+WARN[0005] reference for unknown type: application/vnd.cnab.config.v1+json
+Pushed successfully, with digest "sha256:57c34a53e84607562e396280563186759139454d1704c727180aac1819b75a4f"
+```
+
+Note: you can safely ignore the `WARN[0005] reference for unknown type: application/vnd.cnab.config.v1+json` message.
+
+When this command is complete, your CNAB bundle manifest and invocation image will have been successfully pushed to the specified OCI registry. It can then be installed with the `porter install` command:
+
+```
+$ porter install --tag deislabs/porter-kube-bundle:1.0 -c kool-kred
```
carolynvs: It would help for the documentation to use example names that have some meaning or hint at what people should use instead. I suggest using something either obviously fake like "foo" or better yet something like "--cred mycluster" or something like that for examples.

Also, I recommend using the full tag names, `--cred` and not the short flags in the examples, for consistency.

pkg/cache/cache.go
```diff
@@ -0,0 +1,86 @@
...
+
+}
+
+// StoreBundle will write a given bundle to the bundle cache, in a location derived
+// from the bundleTag.
+func (c *Cache) StoreBundle(bundleTag string, bun *bundle.Bundle) (string, error) {
+	bid := getBundleID(bundleTag)
+	bundleCnabDir, err := c.getCachedBundleCNABDir(bid)
+	cachedBundlePath := filepath.Join(bundleCnabDir, "bundle.json")
+	c.FileSystem.MkdirAll(bundleCnabDir, os.ModePerm)
```
carolynvs: This isn't checking for an error

pkg/cache/cache.go
```diff
@@ -0,0 +1,86 @@
...
+	bundleCNABPath := filepath.Join(bundleDir, "cnab")
+	return bundleCNABPath, nil
+
+}
+
+func getBundleID(bundleTag string) string {
+	// hash the tag, tags have characters that won't work as part of a path
+	// so hashing here to get a path friendly name
+	bid := md5.Sum([]byte(bundleTag))
+	return fmt.Sprintf("%x", bid)
```
carolynvs: You can also use `hex.EncodeToString`. Not sure if that's better or not, or if fmt uses that under the hood. It may be more clear though.

pkg/porter/install.go
```diff
@@ -46,6 +64,21 @@ func (o *InstallOptions) ToDuffleArgs() cnabprovider.InstallArguments {
 // InstallBundle accepts a set of pre-validated InstallOptions and uses
 // them to install a bundle.
 func (p *Porter) InstallBundle(opts InstallOptions) error {
+	// If opts.Tag is set, fetch the bundle
```
carolynvs: I see you had to put this here because there's a bug in applyDefaultOptions. It assumes there there is always a porter.yaml to load. Heads up that I'm fixing that as part of the work I'm doing today on rebuilding before actions when the hash is out of date.

pkg/porter/install_test.go
```diff
@@ -89,6 +90,7 @@ func TestInstallOptions_validateParams(t *testing.T) {
 		sharedOptions{
 			Params: []string{"A=1", "B=2"},
 		},
+		BundlePullOptions{},
```
carolynvs: Since you declared InstallOptions with BundlePullOptions as an embedded type (and not a pointer), you don't need to have this line here. Same for the other tests below. I

The only time you need stuff like this line here is if you defined it like this instead

```
type InstallOptions struct {
  *BundlePullOptions
}
```

pkg/porter/publish.go
```diff
@@ -1,55 +1,53 @@
...
-	if p.File == "" {
-		fs := porter.Context.FileSystem
+	fs := porter.Context.FileSystem
+	f := p.File
+	if f == "" {
 		pwd, err := os.Getwd()
 		if err != nil {
 			return errors.Wrap(err, "could not get current working directory")
 		}
+		f = filepath.Join(pwd, config.Name)
```
carolynvs: I notice that we are duplicating validation between sharedOptions and here. As part of my --cnab-file work, I'm making another struct that is going to extract out logic for --file and --cnab-file that will do the common validation for these flags across all the commands. Just a heads up.

pkg/porter/install_test.go
jeremyrickard: huh, the compiler complained until I added this.

PR #381, opened 2019-06-05
Add a call to MkdirAll for the credentials directory path.

### LINKED ISSUE #377: porter credentials generate fails with no directory found
On a fresh install of the latest porter release (via https://porter.sh/install/), when I go to generate credentials, say, for the `examples/wordpress` bundle in this repo, it fails with the following:

```
 $ porter credentials generate
Generating new credential wordpress from bundle wordpress
==> 1 credentials required for bundle wordpress
? How would you like to set credential "kubeconfig" environment variable
? Enter the environment variable that will be used to set credential "kubeconfig" foo
Saving credential to /Users/vdice/.porter/credentials/wordpress.yaml
Error: couldn't write credential file /Users/vdice/.porter/credentials/wordpress.yaml: open /Users/vdice/.porter/credentials/wordpress.yaml: no such file or directory

 $ ls -haltr ~/.porter
total 88M
drwxr-xr-x    7 vdice staff  224 May 31 16:35 mixins
drwxr-xr-x    5 vdice staff  160 Jun  4 13:11 bundles
drwxr-xr-x    3 vdice staff   96 Jun  4 13:11 claims
drwxr-xr-x    7 vdice staff  224 Jun  4 13:13 .
-rwxr-xr-x    1 vdice staff  41M Jun  4 13:26 porter
-rwxr-xr-x    1 vdice staff  46M Jun  4 13:26 porter-runtime
drwxr-xr-x+ 184 vdice staff 5.8K Jun  4 13:28 ..

 $ porter version
porter v0.7.0-ralpha.1+englishrose (83471fb)
```

It appears that the `credentials` sub-directory isn't auto-created... but may need to be.  Does this sound like the right diagnosis?

jeremyrickard:
This adds a call to MkdirAll with the credentials path before we attempt to write the credential file. This will ensure that the credential directory exists before we try and write the file. If the directory exists already, it's a NO-OP. 

Fixes #377

### DIFF
--- a/pkg/porter/credentials.go
+++ b/pkg/porter/credentials.go
@@ -178,9 +178,12 @@ func (p *Porter) GenerateCredentials(opts CredentialOptions) error {
 	// Make the credentials path if it doesn't exist. MkdirAll does nothing if it already exists
 	// Readable, writable only by the user
 	err = p.Config.FileSystem.MkdirAll(credentialsDir, 0700)
+	if err != nil {
+		return errors.Wrap(err, "unable to create credentials directory")
+	}
 	dest, err := p.Config.GetCredentialPath(genOpts.Name)
 	if err != nil {
-		return errors.Wrap(err, "unable to determine credentials directory")
+		return errors.Wrap(err, "unable to determine credentials path")
 	}
 
 	fmt.Fprintf(p.Out, "Saving credential to %s\n", dest)
--- a/pkg/porter/credentials.go
+++ b/pkg/porter/credentials.go
@@ -135,6 +135,9 @@ func (g *CredentialOptions) validateCredName(args []string) error {
 	return nil
 }
 
+// GenerateCredentials builds a new credential set based on the given options. This can be either
+// a silent build, based on the opts.Silent flag, or interactive using a survey. Returns an
+// error if unable to generate credentials
 func (p *Porter) GenerateCredentials(opts CredentialOptions) error {
 
 	//TODO make this work for either porter.yaml OR a bundle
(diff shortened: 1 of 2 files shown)


pkg/porter/credentials_test.go
```diff
@@ -231,3 +232,57 @@ good-creds   now`},
...
+	p := NewTestPorter(t)
+	p.TestConfig.SetupPorterHome()
+	p.CNAB = &TestCNABProvider{}
+
+	opts := CredentialOptions{
+		Silent: true,
+	}
+	opts.Name = "name"
+	credDir, err := p.Config.GetCredentialsDir()
+	require.NoError(t, err, "should have been able to get credentials directory path")
```
vdice: tiniest of nits on both these tests... perhaps separate logical blocks with empty lines  😬 ?  

But both added tests 💯 !

pkg/porter/credentials.go
```diff
@@ -168,13 +171,20 @@ func (p *Porter) GenerateCredentials(opts CredentialOptions) error {
...
 		return nil
 	}
-
+	credentialsDir, err := p.Config.GetCredentialsDir()
+	if err != nil {
+		return errors.Wrap(err, "unable to get credentials directory")
+	}
+	// Make the credentials path if it doesn't exist. MkdirAll does nothing if it already exists
+	// Readable, writable only by the user
+	err = p.Config.FileSystem.MkdirAll(credentialsDir, 0700)
```
carolynvs: This isn't handling the error, can you add that quick?

jeremyrickard: yeahhhh i have that in a commit i noticed it wasn't there after i pusheddddd

PR #385, opened 2019-06-05
Smart rebuild

carolynvs:
Introduce --cnab-file flag

All commands that supported a --file flag now have a --cnab-file flag,
and --file is now just for specifying the porter.yaml file. When they
need to work with a non-porter bundle, they use the --cnab-file flag to
pass in a bundle.json file.

---

Check hash and rebuild bundle if necessary

When the hash of the porter manifest doesn't match the hash found in the
bundle.json, rebuild the bundle before performing an action that uses
the bundle.json file.

---

Add programmatic integration test for rebuild

This is the start of testing with Go instead of Bash. What do you think?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,10 +77,13 @@ ifndef HAS_DEP
 endif
 	dep version
 
-test: clean-last-testrun test-unit test-cli
+test: clean-last-testrun test-unit test-integration test-cli
 
 test-unit: build
 	go test ./...
+
+test-integration: build
+	go test -tags=integration ./...
 
 test-cli: clean-last-testrun build init-porter-home-for-ci
 	REGISTRY=$(REGISTRY) KUBECONFIG=$(KUBECONFIG) ./scripts/test/test-cli.sh
--- a/brigade.js
+++ b/brigade.js
@@ -169,9 +169,9 @@ checks = {
   "verify": { runFunc: verify, description: "Verify" },
   "build": { runFunc: build, description: "Build" },
   "crossplatformbuild": { runFunc: xbuild, description: "Cross-Platform Build" },
-  "unitTest": { runFunc: testUnit, description: "Unit Test" },
-  "integrationTest": { runFunc: testIntegration, description: "Integration Test" },
-  "cliTest": { runFunc: testCLI, description: "CLI Test" }
+  "unittest": { runFunc: testUnit, description: "Unit Test" },
+  "integrationtest": { runFunc: testIntegration, description: "Integration Test" },
+  "clitest": { runFunc: testCLI, description: "CLI Test" }
 };
 
 // runCheck can be invoked to (re-)run an individual GitHub Check Run, as opposed to a full suite
--- a/brigade.js
+++ b/brigade.js
@@ -22,8 +22,9 @@ events.on("exec", (e, p) => {
     verify(e, p),
     build(e, p),
     xbuild(e, p),
-    test(e, p),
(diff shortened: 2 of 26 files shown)


carolynvs:
@vdice I updated the brigade.js file to have a new check "testIntegration" and "testCLI" but I'm not sure it worked. On the last run, neither reported back to github. Do you see any problems with my changes?

brigade.js
```diff
@@ -81,8 +84,8 @@ function xbuild(e, p) {
   return goBuild;
 }
 
-function test(e, p) {
-  var goTest = new GoJob(`${projectName}-test`);
+function testUnit(e, p) {
+  var goTest = new GoJob(`${projectName}-testUnit`);
```
vdice: re: the question posed on this PR on why CI seemed to stop working w/ these changes...  unfortunately, it's due to k8s resource [naming restrictions](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names) which only allow for _lowercase_ letters (we use the provided job name to construct the k8s Pod name).  

This isn't obvious and we didn't have a comment in this file mentioning this.  We could either add comment(s) around this _or_ add logic to lowercase all provided job names in our `GoJob` class constructor, etc.

brigade.js
```diff
@@ -164,8 +193,9 @@ function runCheck(e, p) {
 function runSuite(e, p) {
   if (e.revision.ref == "master" ) {
     Group.runEach([
-      checkRun(e, p, test, "Test"),
+      checkRun(e, p, testUnit(), "Unit Test"),
```
vdice: The referenced functions in this and line 198 should just be: `testUnit` and `testCLI` respectively (no parens).

brigade.js
vdice: (meaning, without further changes here, we'd need to rename the new jobs to `${projectName}-testunit`, `${projectName}-testintegration` and`${projectName}-testcli`, thus eliminating the problematic uppercase letters.)

Makefile
```diff
@@ -77,11 +77,14 @@ ifndef HAS_DEP
 endif
 	dep version
 
-test: clean-last-testrun test-unit test-cli
+test: clean-last-testrun test-unit test-integration test-cli
 
-test-unit: build
+test-unit:
```
vdice: Oh yea, re: the test unit Brigade Check failure... I guess previously we had called `build` which set up/populated the `bin` dir in a certain way that it appears the tests still expect.

I do like that this target doesn't invoke `build` anymore, though.  Maybe we add `make build` before the `test-unit` invocation to the `testUnit` function in the `brigade.js` file instead?

carolynvs: I think that I need to fix those tests in particular that rely on copying binaries from the bin to either not rely on that, or move them over to be this new integration tests type. In a follow up PR though since this is getting big.

pkg/porter/credentials.go
```diff
@@ -141,7 +142,7 @@ func (g *CredentialOptions) validateCredName(args []string) error {
 func (p *Porter) GenerateCredentials(opts CredentialOptions) error {
 
 	//TODO make this work for either porter.yaml OR a bundle
```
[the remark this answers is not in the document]
carolynvs: Oops, I didn't do a smart rebuild before `porter generate credentials`. Fixing

Makefile
carolynvs: https://github.com/deislabs/porter/issues/397

PR #390, opened 2019-06-05
Add a simple kubernetes example

jeremyrickard:


examples/kubernetes-mixin-example/porter.yaml
```diff
@@ -0,0 +1,39 @@
+# This is the configuration for Porter
+# You must define steps for each action, but the rest is optional
+# Uncomment out the sections below to take full advantage of what Porter can do!
```
carolynvs: nit: can you remove this comment section here? It doesn't really apply to your example

examples/kubernetes-mixin-example/README.md
```diff
@@ -0,0 +1,46 @@
+# My First Bundle
```
carolynvs: This file should be removed (since it's talking about my first bundle, and an empty bundle) or edited to explain the example

examples/kubernetes-mixin-example/Dockerfile.tmpl
```diff
@@ -0,0 +1,12 @@
+FROM quay.io/deis/lightweight-docker-go:v0.2.0
```
carolynvs: This file should be removed since it's not being used by the example

jeremyrickard:
@carolynvs updated!

PR #392, opened 2019-06-05
Add verbose flag to build

carolynvs:
The verbose flag hides all the extra output printed by the docker build
command, and also the contents of the dockerfile.

I tried to write unit tests for this but really build isn't easy to write tests for (the part that I modified, where we call out to docker). Not sure it's worth it, or if we do want to get tests for this, may require some more refactoring than I want to put into this PR.

### DIFF
--- a/cmd/porter/bundle.go
+++ b/cmd/porter/bundle.go
@@ -61,14 +61,20 @@ func buildCreateCommand(p *porter.Porter) *cobra.Command {
 }
 
 func buildBundleBuildCommand(p *porter.Porter) *cobra.Command {
-	return &cobra.Command{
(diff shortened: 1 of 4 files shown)


jeremyrickard:
💃 🔥

jeremyrickard:
Should we also add this to the publish command?

carolynvs:
> Should we also add this to the publish command?

Yeah I want to, just started small with build for this PR. I ran into a snag with install/upgrade/uninstall because the duffle docker driver writes directory stdout and filed an issue.

Also I'm working on writing better integration tests right now for rebuild, that don't use bash but Go instead.

PR #399, opened 2019-06-07
Fix credentials generate smart build

carolynvs:
I forgot to load the manifest first before calling the logic to do the
smart build. The integration test didn't catch it because the manifest
doesn't get nuked between calls. So I need to do more work to get the
integration tests to match how the CLI works.


jeremyrickard:
This has workshop content in it too, is that related to this change?

PR #400, opened 2019-06-07
Workshop: Helm mixin tutorial

carolynvs:


jeremyrickard:
It looks like your etcd operator workshop stuff is also in #399

workshop/etcd-operator/porter.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+credentials:
+- name: kubeconfig
+  path: /root/.kube/config
+
+install:
+  - helm:
+      description: "Install etcd-operator"
+      name: myetcd-operator
+      chart: stable/etcd-operator
```
jeremyrickard: We haven't implemented the "lock down" capability, do you think for the purpose of the workshop it's better to have the bundle pulled from the internet after in the beginning part we talk about adding everything (chart included) into the invocation image?

carolynvs: The purpose for this exercise was to go through making a bundle from scratch and using a new mixin (helm instead of exec). I don't think it's a good idea to introduce multiple of concepts in a single exercise, so if we want to cover how to bring in files from the local file system, that would be best in a follow-on exercise I think.

Though this one in particular highlights a gap in the helm mixin and porter that I'm not sure I want to go into in the workshop. The `porter build --embed (which porter publish is going to implicitly call for you)` functionality is going to take care of telling the mixins to do their best to embed resources like the charts.

Instead of highlighting the gap, I'd prefer to find an exercise that demonstrates how to copy their own unique files from the file system (like copying their own hand crafted bash scripts or fun gifs they found on the internet and serve them with nginx). Or another one would be to have a Dockerfile.tmpl and then use ADD https://.... and download from the canonical location as a best practice. I don't want to do an exercise that gets them thinking that they should copy the real chart into their local bundle, you know?

carolynvs: So maybe 2 more exercises:

1. Autocopying your own files and making a nginx that displayed them on a webpage
2. Dockerfile.tmpl and using a custom base image

carolynvs:
Oops, thanks I just removed the workshop files.

workshop/etcd-operator/porter.yaml
jeremyrickard: Yeah, that's a good idea.

PR #401, opened 2019-06-07
Implement porter credential show command

vdice:
* Implements the `porter credential show` command
* Uses a new [tablewriter](https://github.com/olekukonko/tablewriter) for this command

Sample:

```
 $ porter creds show porter-azure-wordpress
Name: porter-azure-wordpress

-----------------------------------------------------------------
  Name             Local Source                       Source Type
-----------------------------------------------------------------
  CLIENT_ID        AZURE_SP_ID                        EnvVar
  CLIENT_SECRET    AZURE_SP_SECRET                    EnvVar
  SUBSCRIPTION_ID  SUBSCRIPTION_ID                    EnvVar
  TENANT_ID        AZURE_SP_TENANT                    EnvVar
  kubeconfig       /Users/vdice/.kubes/mk/kubeconfig  Path
```

Question(s) for the reviewer:
 - Do we want to update the table writer in our `printer` pkg to use the tablewriter enlisted for credential show -- and therefore, apply same table formatting/config for all table printing -- or just keep it for this particular command?
 - Is the `reflect` craziness used for determining credential source type a sign I wildly missed an easier route?! 😅


pkg/porter/credentials.go
```diff
@@ -194,3 +206,102 @@ func (p *Porter) GenerateCredentials(opts CredentialOptions) error {
...
+
+// ShowCredential shows the credential set corresponding to the provided name, using
+// the provided printer.PrintOptions for display.
+func (p *Porter) ShowCredential(opts CredentialShowOptions) error {
+	credsDir, err := p.Config.GetCredentialsDir()
+	if err != nil {
+		return errors.Wrap(err, "unable to determine credentials directory")
+	}
+
+	credSet := &credentials.CredentialSet{}
```
carolynvs: Can you break this up into 2 functions: GetCredential and ShowCredential? It makes it easier for testing and automating porter that way.

pkg/porter/credentials.go
```diff
@@ -194,3 +206,102 @@ func (p *Porter) GenerateCredentials(opts CredentialOptions) error {
...
+		return printer.PrintJson(p.Out, credSet)
+	case printer.FormatYaml:
+		return printer.PrintYaml(p.Out, credSet)
+	case printer.FormatTable:
+		var data [][]string
+
+		// Iterate through all CredentialStrategies to build up our data set
+		for _, cs := range credSet.Credentials {
+			// Build a reflected Source of type reflectedStruct, for use below
+			reflectedSource := reflectedStruct{
```
carolynvs: I suggest making an exported function that accepts `Source` and returns the string representing the the source type. It will be easier to read and I bet others will want to use it too. This looks like it was hard to figure out! 👍

pkg/porter/credentials.go
vdice: Argh, you'd think I would've learned this best practice from the last credentials list PR!  Thank you for the reminder (again).  Updated.

pkg/porter/credentials.go
vdice: Good call; now broken out into exported function.

jeremyrickard:
@vdice let's get the changes from @carolynvs in a follow up, i merged this for the goodness

vdice:
@jeremyrickard @carolynvs I _think_ requested changes were all addressed prior to merge... let me know if not.

PR #402, opened 2019-06-07
Added --tag to upgrade and install

jeremyrickard:
This adds --tag to upgrade and install. Since the functionality is used in a couple
places and Install/Upgrade/Delete currently have their own Opts that embed the same things,
I actually just made a new struct called LifecycleOpts that has a logic to deal with tag
if it's present (fetch the bundle, update the opts to use it)

Also refactored cache a little bit to make it easier to use a Mock and not actually use the
cache during a unit test

### DIFF
--- a/cmd/porter/bundle.go
+++ b/cmd/porter/bundle.go
@@ -191,6 +191,8 @@ For instance, the 'debug' driver may be specified, which simply logs the info gi
   porter bundle upgrade --param-file base-values.txt --param-file dev-values.txt --param test-mode=true --param header-color=blue
   porter bundle upgrade --cred azure --cred kubernetes
   porter bundle upgrade --driver debug
+  porter bundle upgrade --driver debug
+  porter bundle upgrade MyAppFromTag --tag deislabs/porter-kube-bundle:v1.0
 `,
 		PreRunE: func(cmd *cobra.Command, args []string) error {
 			return opts.Validate(args, p.Context)
@@ -215,6 +217,10 @@ For instance, the 'debug' driver may be specified, which simply logs the info gi
 		"Credential to use when installing the bundle. May be either a named set of credentials or a filepath, and specified multiple times.")
 	f.StringVarP(&opts.Driver, "driver", "d", porter.DefaultDriver,
 		"Specify a driver to use. Allowed values: docker, debug")
+	f.StringVarP(&opts.Tag, "tag", "t", "",
+		"Install from a bundle in an OCI registry specified by the given tag")
+	f.BoolVar(&opts.InsecureRegistry, "insecure-registry", false,
+		"Don't require TLS for the registry")
 
 	return cmd
 }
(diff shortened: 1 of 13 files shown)


cmd/porter/bundle.go
```diff
@@ -191,6 +191,8 @@ For instance, the 'debug' driver may be specified, which simply logs the info gi
   porter bundle upgrade --param-file base-values.txt --param-file dev-values.txt --param test-mode=true --param header-color=blue
   porter bundle upgrade --cred azure --cred kubernetes
   porter bundle upgrade --driver debug
+  porter bundle upgrade --driver debug
```
carolynvs: looks like this line accidentally got duplicated

cmd/porter/bundle.go
```diff
@@ -245,6 +251,9 @@ For instance, the 'debug' driver may be specified, which simply logs the info gi
   porter bundle uninstall --param-file base-values.txt --param-file dev-values.txt --param test-mode=true --param header-color=blue
   porter bundle uninstall --cred azure --cred kubernetes
   porter bundle uninstall --driver debug
+  porter bundle uninstall --driver debug
```
carolynvs: same here, line got duped

pkg/porter/helpers.go
```diff
@@ -90,3 +93,22 @@ func (p *TestMixinProvider) GetVersion(m mixin.Metadata) (string, error) {
 func (p *TestMixinProvider) Install(o mixin.InstallOptions) (mixin.Metadata, error) {
 	return mixin.Metadata{Name: "exec", Dir: "~/.porter/mixins/exec"}, nil
 }
+
+// If you seek a mock cache for testing, use this
```
carolynvs: See and ye shall find

pkg/porter/helpers.go
```diff
@@ -90,3 +93,22 @@ func (p *TestMixinProvider) GetVersion(m mixin.Metadata) (string, error) {
...
+type mockCache struct {
+	findBundleMock        func(string) (string, bool, error)
```
carolynvs: You don't have to change this now. I just wanted to share that my plan for our test helpers was that they are usable for downstream consumers. So that they can write unit tests too.

That's why you can do things like `porter.NewTestPorter` from our integration tests in /tests/rebuild_test.go even though it's outside of the pkg/porter package. Same for TestContext and TestConfig, etc.

So eventually we may need something like TestCache? or exposing this mockCache struct to help them out too.

pkg/porter/install_test.go
```diff
@@ -55,9 +57,11 @@ func TestPorter_applyDefaultOptions_DebugOff(t *testing.T) {
 	require.NoError(t, err)
 
 	opts := &InstallOptions{
-		sharedOptions: sharedOptions{
-			bundleFileOptions: bundleFileOptions{
-				File: "porter.yaml",
+		BundleLifecycleOpts{
```
carolynvs: Just a suggestion, in some places I've switched to doing this because the deeply nested struct inits are getting hard to for me to manage

```go
opts := &InstallOptions{}
opts.File = "porter.yaml"
```

pkg/porter/lifecycle_test.go
```diff
@@ -0,0 +1,86 @@
...
+	fileExists, err := p.TestConfig.TestContext.FileSystem.Exists(fullPath)
+	require.True(t, fileExists, "this test requires that the file exist")
+
+	cache := mockCache{
+		findBundleMock: func(tag string) (string, bool, error) {
+			return fullPath, true, nil
+		},
+	}
+	p.Porter.Cache = &cache
+	_, ok, err := p.Cache.FindBundle("deislabs/kubekahn:1.0")
```
carolynvs: KHAAAAAAAAAAAAN

pkg/porter/helpers.go
jeremyrickard: Yeah, that sounds like a good idea. I'll do a follow up for that if that's cool.

PR #404, opened 2019-06-07
fix(brigade.js): misc Promise-related fixes

vdice:
A smattering of fixes re: recent learnings with @krancour while iterating on similar pipelines in other projects (https://github.com/brigadecore/brigade, https://github.com/brigadecore/kashti, etc.)


brigade.js
```diff
@@ -54,6 +54,8 @@ events.on("publish", (e, p) => {
 // Actions
 // **********************************************
 
+// Important: each Job name below must only consist of lowercase
+// alphanumeric characters and hyphens, per K8s resource name restrictions
```
carolynvs: 👍

PR #408, opened 2019-06-10
fix(brigade.js): update testIntegration function

vdice:
Removes kubeconfig/docker setup not needed for this make target

### DIFF
--- a/brigade.js
+++ b/brigade.js
@@ -96,28 +96,13 @@ function testUnit(e, p) {
   return goTest;
 }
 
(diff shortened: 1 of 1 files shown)


carolynvs:
Oops, I think I didn't explain the integration target very well. Sorry about that. It's a real live test of porter, just not through the cli. It will hit docker, kubernetes, etc. It just isn't automating Porter via the CLI (which is a bit harder to capture and manage) and is using the pkg/porter library instead.

So we do need docker and kubernetes config on the host. I'm going to close this PR, but let me know if you have questions!

PR #417, opened 2019-06-13
ci(*): publish test and example bundles

vdice:
Proposed implementation for https://github.com/deislabs/porter/issues/396

Note that I started off misinterpreting the original issue - I was thinking about the test bundles that we test in CI.  So, the first commit comprises changes to publish the test bundles while the second commit adds logic for publishing the example bundles.  We can either go with both or I can remove that first commit if we'd rather not publish the test variants.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -121,6 +121,35 @@ publish: prep-install-scripts
 	bin/porter mixins feed generate -d bin/mixins -f bin/atom.xml -t build/atom-template.xml
 	az storage blob upload -c porter -n atom.xml -f bin/atom.xml
 
+# all-bundles loops through all items under the dir provided by the first argument
+# and if the item is a sub-directory containing a porter.yaml file,
+# runs the make target(s) provided by the second argument
+define all-bundles
+	@for dir in $$(ls -1 $(1)); do \
+		if [[ -e "$(1)/$$dir/porter.yaml" ]]; then \
+			BUNDLE=$$dir make $(MAKE_OPTS) $(2) ; \
+		fi ; \
+	done
+endef
+
+EXAMPLES_DIR := examples
+
+.PHONY: build-bundle
+build-bundle:
+ifndef BUNDLE
+	$(call all-bundles,$(EXAMPLES_DIR),build-bundle)
+else
+	cd $(EXAMPLES_DIR)/$(BUNDLE) && porter build
+endif
+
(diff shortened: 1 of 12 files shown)


carolynvs:
I think that publishing our test bundles is almost part of our tests? Do we do test publishes right now? 🤔  Maybe during PRs and whatnot we should be publishing to a throw away CI registry somewhere? This is commentary for another issue that I should be making really.

I don't see a problem with publishing our test bundles, though I'm not sure how helpful they are either. I think of them as more boring example bundles. 😂

brigade.js
```diff
@@ -137,17 +143,36 @@ function testCLI(e, p) {
...
+    -p ${p.secrets.dockerhubPassword}`,
+    `REGISTRY=${p.secrets.dockerhubOrg} make test-cli`
   );
 
   return goTest;
 }
 
+function publishExamples(e, p) {
+  var examplePublisher = new GoJob(`${projectName}-publish-examples`);
+  // Enable docker so that the daemon can be used for duffle commands invoked by test-cli
```
carolynvs: nit: there is no duffle only zuul

Makefile
```diff
@@ -121,6 +121,35 @@ publish: prep-install-scripts
...
+			BUNDLE=$$dir make $(MAKE_OPTS) $(2) ; \
+		fi ; \
+	done
+endef
+
+EXAMPLES_DIR := examples
+
+.PHONY: build-bundle
+build-bundle:
+ifndef BUNDLE
```
carolynvs: I'm no make wizard, seems like there are two ways to call this:

`make build-bundle` which will do them all and also `make build-bundle BUNDLE=wordpress`. Is that right?

vdice: Yea!  Should I issue a follow-up PR mentioning this, maybe in a dev doc?  In case devs would like to push to their own registries?  (And actually, if we'd like to support pushing to non-deislabs registries, we'll have to add a `sed`-like string substitution on the inv image and bundle tag...)

brigade.js
vdice: Oh!  Good catch.  I thought I rm'd mention of `duffle` in these comments 😓 .   Lemme fix in a follow-up PR.

Makefile
carolynvs: No that's fine, I was just asking for my own book learnin'. 😀

PR #419, opened 2019-06-14
Add ImageMap

carolynvs:
This adds manual support for image map. The data can be defined in the porter.yaml and porter will populate it in the generated bundle.json file, so that it is made available at runtime per the spec.

However porter and the mixins do not have any explicit behaviors based on the presence of the data. When a user populates the image map, they must also add scripts to the bundle to use it at runtime.

### DIFF
--- a/pkg/porter/build.go
+++ b/pkg/porter/build.go
@@ -315,12 +315,13 @@ func (p *Porter) buildBundle(invocationImage string, digest string) error {
 		BaseImage: bundle.BaseImage{
 			Image:     invocationImage,
 			ImageType: "docker",
+			Digest:    digest,
 		},
 	}
-	if digest != "" {
-		image.Digest = digest
-	}
+
 	b.InvocationImages = []bundle.InvocationImage{image}
+
+	b.Images = p.generateBundleImages()
 	b.Parameters = p.generateBundleParameters()
 	b.Credentials = p.generateBundleCredentials()
 	b.Custom[config.CustomBundleKey] = p.GenerateStamp(p.Manifest)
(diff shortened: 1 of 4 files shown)


pkg/porter/build.go
```diff
@@ -338,6 +339,33 @@ func (p Porter) writeBundle(b bundle.Bundle) error {
 	return errors.Wrap(err, "error writing to cnab/bundle.json")
 }
 
+func (p *Porter) generateBundleImages() map[string]bundle.Image {
+	images := make(map[string]bundle.Image, len(p.Manifest.ImageMap))
+
+	for i, refImage := range p.Manifest.ImageMap {
+		img := bundle.Image{
+			Description: refImage.Description,
+			BaseImage: bundle.BaseImage{
```
vdice: Q: I see we are re-using `bundle.Image` and `BaseImage` structs to build up a corresponding mapped image.  Should we be using the `MappedImage` struct added to `manifest.go` instead?

pkg/porter/build.go
```diff
@@ -338,6 +339,33 @@ func (p Porter) writeBundle(b bundle.Bundle) error {
...
+				Image:         refImage.Image,
+				Digest:        refImage.Digest,
+				ImageType:     refImage.ImageType,
+				MediaType:     refImage.MediaType,
+				OriginalImage: refImage.OriginalImage,
+				Size:          refImage.Size,
+			},
+		}
+		if refImage.Platform != nil {
```
vdice: Do we want to add a quick unit test around a mappedImage that doesn't have `Platform` defined?

pkg/porter/build_test.go
```diff
@@ -233,3 +233,46 @@ func TestPorter_paramRequired(t *testing.T) {
 	require.True(t, ok)
 	require.True(t, p2.Required)
 }
+
+func TestPorter_generateImages(t *testing.T) {
```
vdice: Do we want test(s) that vet the required fields when defining an ImageMap in a porter.yaml vs fields that can be empty (`omitempty`)?  Or would this more be testing general yaml -> golang struct flow and thus perhaps not needed?

pkg/porter/build.go
carolynvs: This is going to be more clear in my next pull request #424 where I've split out the code for mapping a porter manifest to a bundle.json (I hope!). This code right here is making a bundle.json, so we are using the bundle structs from cnab-go. MappedImage is a struct for working with the porter manifest.

vdice: Ah!  I've got it now.  Thank you for spelling it out for me.

pkg/porter/build_test.go
carolynvs: Sorry, I'm not quite sure what you want to test here? Do you mean like the kind of test that I did for cnab-go here? To ensure that the proper set of fields have omit empty and force people to think about it each time a field is added?

https://github.com/deislabs/cnab-go/pull/24

vdice: Oh yup, something similar to what you've done in the link provided.  Can be a follow-up if we deem it necessary.

carolynvs: Yeah I can do that in a follow-up PR since it's a bit larger in scope than just image map, and we can adjust which ones do have omit empty on them at the same time.

carolynvs: I am not sure I'm going to get to this today so here's an issue to track it https://github.com/deislabs/porter/issues/425

PR #421, opened 2019-06-14
ref(brigade.js): update applicable jobs to run via DinD

vdice:
Proposed implementation for https://github.com/deislabs/porter/issues/409

Collabo with @krancour, who was of crucial assistance around some of the finer points of setting up the golang env in the alpine-based `docker:stable-dind` base image.

Uses https://github.com/vdice/go-dind currently, but we can amend/port/etc.  This image currently comes in at `814MB`, so at least chops some weight off of our golang image (the previous `deis/go-dev` image comes in at `1.86GB`)

### DIFF
--- a/brigade.js
+++ b/brigade.js
@@ -184,6 +184,10 @@ function publish(e, p) {
 
   goPublish.env.AZURE_STORAGE_CONNECTION_STRING = p.secrets.azureStorageConnectionString;
   goPublish.tasks.push(
+    "apk update && apk add py-pip && \
+      apk add --virtual=az-build libffi-dev musl-dev openssl-dev python-dev && \
+      pip install azure-cli && \
+      apk del --purge az-build",
     "make build xbuild-all publish"
   )
 
--- a/brigade.js
+++ b/brigade.js
@@ -106,8 +106,8 @@ function testUnit(e, p) {
 // to remove the need of re-building before running test-cli
 function testIntegration(e, p) {
   var goTest = new GoJob(`${projectName}-testintegration`);
-  // Enable docker so that the daemon can be used for duffle commands invoked by test-cli
-  goTest.privileged = true;
+  // Enable Docker-in-Docker
+  goTest.enableDind();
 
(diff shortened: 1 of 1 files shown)


vdice:
Hold please; I haven't checked the publish jobs based on this new image.

brigade.js
```diff
@@ -181,6 +181,10 @@ function publish(e, p) {
 
   goPublish.env.AZURE_STORAGE_CONNECTION_STRING = p.secrets.azureStorageConnectionString;
   goPublish.tasks.push(
+    "apk update && apk add py-pip && \
```
carolynvs: if we just need the azure cli for publish, then I recommend using 

https://github.com/carolynvs/az-cli

Though for some of our lightweight docker images I think it's installed by default.

vdice: I'd be down to switch up to using that cli.  It's not necessarily updated to the latest az cli codez routinely, right?  But perhaps for our purposes, it doesn't matter.

Alternatively, we can add `az` to the `go-dind` image used here.  Actually, if this `go-dind` image garners favor, we should probably move it to this `deislabs` org and potentially remove `kubectl`.  Then, we could have a porter CI-centric Dockerfile here based on `go-dind` but bringing in `kubectl`, `az` and any other dependencies that might come later on?  What do you think?

carolynvs: The az-cli repo I linked is a go implementation of just a few select commands from the full az cli which is written in python. It is a bit smaller and easier to install, so it's friendlier to get into a docker image is all. But because it's not actually using the az cli, it's not quite the same concern when it comes to using the latest code. More like, do you trust the code that I wrote and tested really.

I have been using it for a few open source projects, like service catalog, minibroker, etc to handle deployments to azure storage.

vdice: Good call!  Updated to use this `az`.  Saves so much time and just what we need here.  Thanks!

carolynvs: Sorry it was borked the first time you tried it. Thanks for showing me how to regex. 😀

PR #424, opened 2019-06-17
Populate Schema Version in the CNAB bundle.json

carolynvs:
* Extract the manifest to bundle.json conversion logic into its own pkg. The porter app is getting big and I think this will help.
* Populate `schemaVersion` on the generated bundle.json
* Bump cnab-go to latest
* Use a short-term branch of cnab-to-oci because of the cnab-go bump. After this goes in, I'll submit a PR to them with our branch (it just bumps cnab-go for them and fixes the breaking changes).
* Bump duffle to 0.2.0-beta.1

Closes #345

PR #433, opened 2019-06-20
Bump cnab-go to 0.1.1-beta.1 

carolynvs:
This bumps us to the first release of cnab-go! 🎉 It also gets us off of our short term fork of cnab-to-oci. 💃 

I'm going to do a follow-on PR that decides which fields we want to surface from the bump and how. This was just enough to get us building again against the latest version with the current set of functionality.


carolynvs:
omg y'all you won't believe this but in all my futzing, I lost my changes in #424 to actually populate schema version! So I'm fixing that now and adding the tests that got lost too. I think I stashed it and then the git gobblins ate...

carolynvs:
Ok finally ready for review. I added my schemaVersion changes that were supposed to be in #424 here because otherwise nothing passes because install now validates bundle.json before it runs it (and missing a schemaVersion is invalid now).

Issue #361, opened 2019-05-29
Pluggable Build Driver


carolynvs:
Right now we assume that build is going to be done on a local docker daemon. This won't work for certain contexts, such as Cloud Shell, hosted docker services or virtual machines. This issue is a placeholder to begin the bikeshedding on how to have porter support being able to have other build drivers that can be specified for a bundle or environment.

🚨  This may need coordination with duffle / cnab-go as it's not a porter unique problem.

carolynvs:
Closing now that we have experimental build drivers #1567. If we need more, we now have the framework to drop them in.

PR #435, opened 2019-06-24
Hide the local cnab directory and introduce $BUNDLE_ARG

### LINKED ISSUE #320: Improve abstraction over CNAB

### LINKED ISSUE #422: [Possible BUG] porter.yaml not copied when dependency exists
hi team,

I'm trying (really) hard to get my first package to work with a dependency.
The dependency is located in ./myBundle/bundles/myDependency/porter.yaml:

image

after a `porter build` here is the result:
image

Am I doing something wrong?

[EDIT] I just tried to move the dependency to `%HOME/.porter/bundles` and I do have the same result
image

SwissPorter 😄

### LINKED ISSUE #426: Remove generated files before build
Before `porter build`, any generated files leftover from a previous `porter build` should be deleted.

Files to delete:
* cnab/
* Dockerfile

carolynvs:
This hides the local cnab directory that porter generates during the
build phase from the user. The goal is to avoid giving the user
expectations about the directory and its contents.

In addition, we now delete and recreate the generated files, .cnab and
Dockerfile, before we do a build to avoid any odd bugs.

I've also tweaked our gitignore to take advantage of the new naming
rules.

Closes #320 
Closes #422 
Closes #426


pkg/build/build.go
```diff
@@ -0,0 +1,17 @@
+package build
```
carolynvs: I have plans to split out the build logic into this package in another PR. This is in prep for #361.

PR #438, opened 2019-07-01
Feat: bundle outputs

vdice:
This PR adds:

* functionality to parse a new `outputs` section in a porter manifest (Note: the current implementation requires a mixin-specific step/action output to be declared before it can be exposed/used as a bundle-level output.)
* functionality to place output values under `/cnab/app/outputs` in the execution environment, per the CNAB Spec (outputs are currently written only when added under the new `outputs` section and the optional `applyTo` specification doesn't conflict)
* functionality to place these same outputs under the local/user's `$PORTER_HOME/outputs/<bundle name>` directory, assuming the default Docker driver is being used
* cli command: `porter bundle show` to list a bundle's claim and outputs

TODO:

 - [x] add a cli command that users can invoke to print outputs from the corresponding local file (question around AC posed in https://github.com/deislabs/porter/issues/344#issuecomment-507460567)
 - [x] add unit tests for logic added in `run.go`
 - [x] docker driver bind mount refactor/units
 - [x] update the base porters schema template to include the outputs section
 - [x] add/update docs


vdice:
In the most recent [commit](https://github.com/deislabs/porter/pull/438/commits/80130a17c1b9b9a564048b2870004259fdcda7f0), I went ahead and added the cli commands mentioned/proposed in https://github.com/deislabs/porter/issues/344#issuecomment-507460567

vdice:
Putting into 'ready for review' while I finish up unit tests...

build/testdata/bundles/wordpress/porter.yaml
```diff
@@ -26,6 +26,10 @@ parameters:
 - name: wordpress-password
   type: string
   sensitive: true
+- name: mysql-password
```
carolynvs: I don't understand why you introduced this parameter and then later down flipped `externaldatabase.password` to go from being retrieved from an output to being supplied by a parameter. This really changes the intent of dependencies and what the bundle was demonstrating.

build/testdata/bundles/wordpress/porter.yaml
```diff
@@ -42,13 +46,34 @@ install:
...
         externalDatabase.user: "{{ bundle.dependencies.mysql.parameters.mysql-user }}"
-        externalDatabase.password: "{{ bundle.dependencies.mysql.outputs.mysql-password }}"
+        externalDatabase.password: "{{ bundle.parameters.mysql-password }}"
         externalDatabase.port: 3306
         mariadb.enabled: false
+      outputs:
+        - name: wordpress-password
+          secret: "{{ bundle.parameters.wordpress-name }}-wordpress"
+          key: wordpress-password
+        - name: mysql-password
```
carolynvs: Is this secret coming from the wordpress chart or from the mysql chart (i.e. the dependency)?

build/testdata/bundles/wordpress/porter.yaml
```diff
@@ -42,13 +46,34 @@ install:
...
         - "{{ bundle.parameters.wordpress-name }}"
+
+outputs:
+  - name: wordpress-password
+    description: "The Wordpress installation password"
+    type: string
+    applyTo:
+      - "install"
+      - "upgrade"
+  - name: mysql-password
```
carolynvs: This is an output of the mysql bundle, not the wordpress bundle. I'm really confused about why it's part of the wordpress bundle here.

cmd/porter/outputs.go
```diff
@@ -0,0 +1,72 @@
+package main
+
+import (
+	"github.com/deislabs/porter/pkg/porter"
+	"github.com/spf13/cobra"
+)
+
+func buildOutputsCommand(p *porter.Porter) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "outputs",
```
carolynvs: I think that this would be better not as a standalone command but as a task-based command that is used to inspect an installed bundle.

So instead of having `porter outputs -b bundle name` (which is off anyway because we are actually working with claim names right?), what if we had a command that let you look at an installed bundle and also had it dig deeper into additional available into (like outputs) if the bundle had it?

Maybe something like this:

```console
$ porter bundles list
NAME                  CREATED      MODIFIED     LAST ACTION   LAST STATUS
mysql                  2019-06-24   2019-06-24   install       success

$ porter bundle show mysql
Name          mysql
Created       2019-06-24
Modified      2019-06-24
Last Action  install
Last Status  success

Outputs
=====================================================================
Name           Type         Value
-----------+--------+--------------------------------------------------------+
hostName     String          wpmysql
port                Integer       5423
clientCert       String         /Users/carolynvs/.porter/outputs/.../mysql/clientCert
```

clientCert is writeOnly so we don't display it. Find some logic for not dumping sensitive or giant info

cmd/porter/outputs.go
```diff
@@ -0,0 +1,72 @@
...
+		Aliases: []string{"output"},
+		Short:   "Output commands",
+		Long:    "Commands for working with Bundle outputs. TODO",
```
carolynvs: typo: "Bundle/bundle"

cmd/porter/outputs.go
```diff
@@ -0,0 +1,72 @@
...
+	return cmd
+}
+
+func buildOutputListCommand(p *porter.Porter) *cobra.Command {
+	opts := porter.OutputListOptions{}
+
+	cmd := &cobra.Command{
+		Use:     "list",
+		Short:   "List outputs of a bundle",
+		Long:    "List bundle outputs. TODO",
```
carolynvs: Oops, still have a TODO here that needs to be filled out. Maybe

> List outputs generated by a bundle

It doesn't have to be terribly long and it's okay to be the same text as the short.

build/testdata/bundles/wordpress/porter.yaml
vdice: I did this as an exercise to explore how a dependency parameter might be overridden from a bundle that uses that dependency, but I can imagine that I didn't do it quite right and/or we don't wish to add for this particular bundle.

I'll remove my override of the dep chart param (`mysql-password`) from params/outputs and just use the `wordpress-password` for testing.

Though, in general, would it not be a valid use case to override arbitrary bundle dependency params via the bundle that utilizes these dependencies?  For this particular example, consider the case where the user wants to ensure the mysql/external db password is a certain value whilst installing the wordpress bundle that uses the mysql/external db bundle.

cmd/porter/outputs.go
vdice: Roger; I can switch tactics and implement `porter bundle show [NAME]` to show outputs for a bundle.

Should we leave this PR with that command or do we also want a solution/command for getting a particular output value?  (Perhaps to merely echo/print or perhaps to send to a file, etc.)

build/testdata/bundles/wordpress/porter.yaml
vdice: Oh, right, I see that any param overrides would go in the `dependencies` section under the appropriate dependency bundle 😅 .  So, perhaps, for the stated use case above, if a dependency param needed to be dynamically set/overridden (on `porter install`, etc.), we'd need to allow for this in the cli... (read: future feature, if valid)

carolynvs: We had our hands tied with dependency parameters originally (because they had to be merged ultimately into a single manifest/invocation image). So the way it originally worked doesn't matter too much, I'm in the process of deleting that code as I type. But yes, it was designed so that you would default a parameter on a dependency in the dependencies section like so:

```
dependencies:
- name: mysql
  parameters:
    database-name: wordpress
```

I'd like to be able to also pass through a parameter defined on the current bundle (something that wasn't possible with the original design). Something like this:

```
dependencies:
- name: debug
  parameters:
    debug: {{ porter.parameters.debug }}
```

The user will need a way to be able to control any parameter (from the main bundle or dependency). Anything set from the cli wins. Though there's a bit of design/thought to go into dealing with how to handle specifying params for deps when parameters can now have the same name between a deps and the main bundle, it's not as simple as it was before.

I think this is all out of scope of the current PR (?), it falls into the dependencies rewrite I'm doing now.

cmd/porter/outputs.go
carolynvs: If someone wants to just get a single output value, I would encourage them to use the --output json and then use jq or something to grab what they need. Maybe we can start with that for now and see if people want a more specialize command like `porter bundle output list/show` to work directly with outputs. We may never need to implement it, depending on lack of asks/feedback.

How does that sound?

vdice: That sounds good to me.  My only question revolves around the case when an actual output value is large (think: kubeconfig file) while listing.  Shall we just default to truncating all values at a certain character limit when printing via default/table and print the entire value if the printing option is otherwise (yaml or json)?

carolynvs: Let's start with truncating at certain number of characters and see how well that works.

vdice: Ok, updated to limit cli scope to just the `porter bundle show` command.  Please take a look when convenient.

pkg/cnab/provider/duffle.go
```diff
@@ -34,5 +43,41 @@ func (d *Duffle) newDriver(driverName string) (driver.Driver, error) {
...
+	// TODO: separate function/add tests
+	if dockerish, ok := driverImpl.(*duffledriver.DockerDriver); ok {
+		outputsDir, err := d.Config.GetOutputsDir()
+		if err != nil {
+			return nil, errors.Wrap(err, "unable to get outputs directory")
+		}
+
+		// Create outputs sub-directory using the bundle name
+		bundleOutputsDir := filepath.Join(outputsDir, claimName)
+		err = d.FileSystem.MkdirAll(bundleOutputsDir, 0755)
```
carolynvs: I think the permissions on that folder may be wrong, because it lets anyone on the machine read outputs created by the bundle. Let's do a sweep on permissions in a separate issue. I'll make one so that we can use a variable for this stuff going forward.

pkg/cnab/provider/duffle.go
```diff
@@ -34,5 +43,41 @@ func (d *Duffle) newDriver(driverName string) (driver.Driver, error) {
...
+			hostCfg.Mounts = append(hostCfg.Mounts, outputsMount)
+			return nil
+		}
+		dockerish.AddConfigurationOptions(cfgOpt)
+	}
+
 	return driverImpl, err
 }
+
+func (d *Duffle) FetchClaim(name string) (*claim.Claim, error) {
```
carolynvs: 👍

pkg/cnab/provider/helpers.go
```diff
@@ -0,0 +1,112 @@
...
+)
+
+type TestDuffle struct {
+	*Duffle
+	*config.TestConfig
+	ClaimStore TestStore
+}
+
+type TestStore struct {
+	backingStore map[string][]byte
```
carolynvs: I feel bad because you implemented all of this but I'm pretty sure we could have also used afero here and not written a test backing store at all. It would be an afero in memory file system from the test context (which we got from the constructor of the TestDuffle) and then the existing crud claim store.

I think that would be preferable to reimplementing another claim store? Let me know i you'd like to collaborate quick on how to use afero to do that.

pkg/config/manifest.go
```diff
@@ -109,6 +110,26 @@ type Dependency struct {
...
+
+	Schema `yaml:",inline"`
+}
+
+func (od *OutputDefinition) Validate() error {
+	if od.Name == "" {
+		return errors.New("output name is required")
+	}
+
+	// TODO: Validate inline Schema
```
carolynvs: Haha, yeah that's a big task on our plate in general for CNAB 1.0 no need to tackle it as part of this PR. 😀

pkg/porter/build.go
```diff
@@ -240,6 +240,17 @@ func (p *Porter) prepareDockerFilesystem() error {
 		}
 	}
 
+	// Create the local outputs directory, if it doesn't already exist
+	outputsDir, err := p.Config.GetOutputsDir()
+	if err != nil {
+		return errors.Wrap(err, "unable to get outputs directory")
+	}
+
+	err = p.FileSystem.MkdirAll(outputsDir, 0755)
```
carolynvs: I'm a bit unsure of the code's intent right here? This function is meant to create files on the local file system in pwd/.cnab in preparation before building the invocation image.

If we need to make the outputs directory ahead of time, then the code should be doing this:

```
p.FileSystem.MkdirAll(filepath.Join(build.LOCAL_APP, "outputs"), 0755)
```

Otherwise if your intent was to make sure the outputs directory in PORTER_HOME exists before the bundle is run (which is what this code is doing), then it's in the wrong spot. It should be located in maybe `pkg/porter/options.go applyDefaultOptions` or really a new function that is executed in each function before calling the cnab provider functions for install/upgrade/etc.

pkg/porter/list.go
```diff
@@ -40,6 +40,8 @@ func (l CondensedClaimList) Less(i, j int) bool {
 
 // ListBundles lists installed bundles using the printer.Format provided
 func (p *Porter) ListBundles(opts printer.PrintOptions) error {
+	// TODO: supply cnab.Provider interface as second arg, use to access ClaimStore
```
carolynvs: Once you move `FetchClaim` like I suggested above, then you will immediate have access to it, because CNABProvider is already on the Porter receiver here.

pkg/porter/build.go
vdice: You know, I think we're okay with not ensuring the outputs directory exists before a bundle is run.  The logic around setting up the volume mount (if driver is the default Docker driver) is sufficient (will create `${PORTER_HOME}/outputs` if it doesn't already exist, in addition to the `<claim name>` subdir.)

So, I've removed for now.

pkg/cnab/provider/helpers.go
vdice: Oops!  Good call; using afero was the way to go.  This file has since been removed.

pkg/cnab/provider/duffle.go
vdice: I've changed to `os.ModePerm` for now... w/ intentions to follow-up across codebase in issue you've mentioned...

pkg/cnab/provider/duffle.go
```diff
@@ -34,5 +43,41 @@ func (d *Duffle) newDriver(driverName string) (driver.Driver, error) {
 		configurable.SetConfig(driverCfg)
 	}
 
+	// If docker driver, setup host bind mount for outputs
+	// TODO: separate function/add tests
+	if dockerish, ok := driverImpl.(*duffledriver.DockerDriver); ok {
```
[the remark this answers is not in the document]
vdice: Separated, though unfortunately we can't yet test as thoroughly as we'd wish (no access yet to the docker driver configuration options... TODO note added.)

vdice:
Ok, latest commits are up intending to address all feedback.

Lemme know if there are any outstanding items; otherwise I/we can squash commits and merge when ready.

PR #443, opened 2019-07-09
Populate dependencies in bundle.json

### LINKED ISSUE #429: Write dependencies to bundle.json on build
Now that the cnab spec supports dependencies, when we generate the bundle.json we should populate the dependencies section of the bundle.json file

https://github.com/deislabs/cnab-spec/blob/master/500-CNAB-dependencies.md#dependencies-metadata

This relies on https://github.com/deislabs/cnab-go/issues/35 which we can tackle after the cnab-go 1.0 release.

This is going to require changes to the porter.yaml manifest to support specifying the bundle version for a dependency.

```yaml
dependencies:
- name: mysql
  tag: deislabs/azure-mysql:5.7
```

```yaml
dependencies:
- name: mysql
  tag: deislabs/azure-mysql
  versions:
   - 5.7.x - 6.3.x
   - 6.5.x - 6.8.x
   prereleases: true
```

We will handle smooshing version and versions into version.ranges and prereleases under version on the bundle.json representation.

Other fields on the dependency, such as parameters and connections, are not written to bundle.json.

carolynvs:
Writes any dependencies in porter.yaml to bundle.json when the bundle is built. 

There are three forms of specifying a dependency:

1. You can just say exactly which bundle, including the tag, that you depend upon, `deislabs/azure-mysql:5.7`
2. You can rely on a named bundle, but not say which version, `deislabs/azure-mysql`. Basically any version of the bundle is ok. You can specify if prereleases are allowed or not (default is false).
3. You can specify a set of version ranges to allow `deisalbs/azure-mysql` + `5.7 - 6.x, 7.1 - 7.5.x`

Closes #429

### DIFF
--- a/pkg/cnab/config_adapter/adapter_test.go
+++ b/pkg/cnab/config_adapter/adapter_test.go
@@ -5,7 +5,7 @@ import (
 
 	"github.com/deislabs/cnab-go/bundle"
 	"github.com/deislabs/cnab-go/bundle/definition"
-
+	"github.com/deislabs/porter/pkg/cnab/extensions"
 	"github.com/deislabs/porter/pkg/config"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -13,7 +13,7 @@ import (
 
 func TestPorter_ToBundle(t *testing.T) {
 	c := config.NewTestConfig(t)
-	c.TestContext.AddTestFile("../../config/testdata/simple.porter.yaml", config.Name)
+	c.TestContext.AddTestFile("testdata/porter.yaml", config.Name)
 
 	err := c.LoadManifest()
 	require.NoError(t, err)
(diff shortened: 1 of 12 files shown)


carolynvs:
The code's been updated to move dependencies into the custom extensions section.

pkg/cnab/config_adapter/adapter_test.go
```diff
@@ -53,6 +36,9 @@ func TestPorter_ToBundle(t *testing.T) {
 
 	assert.Contains(t, bun.Parameters.Fields, "porter-debug", "porter-debug parameter was not defined")
 	assert.Contains(t, bun.Definitions, "porter-debug", "porter-debug definition was not defined")
+
+	assert.Contains(t, bun.Custom, config.CustomBundleKey, "Dependencies was not populated")
```
vdice: Failure message here not accurate/needs updating

carolynvs:
This is a milestone/epic (see https://github.com/deislabs/porter/milestone/8) so it's going to take a number of PRs to get this to the point where anything is working. 😀  I was going to hold off on any documentation until I can actually run a bundle with dependencies. Then I'll add docs on how to define a dependency and what to expect. Until then things are still a bit in flux and may still change depending on what I run into as I implement this.

PR #446, opened 2019-07-11
feat(mixin/step outputs): place each mixin/step output into its own file

vdice:
- place each mixin output into its own file (this enables multiline outputs)


carolynvs:
I have a general question about how far we wanted to go with this. I was expecting a larger change, where the mixins themselves would write multiple files, instead of continuing to write key/value pairs.

It may make things easier/more natural for porter because we will be treating outputs the same throughout. But it may also make it easier for the mixins as well. Now that we allow very large complex outputs (like entire json documents and kubeconfig files, etc), writing each one to its own file, seems like it should be less confusing to look at for everyone involved when debugging.

Since this would be a big change to how mixins work, perhaps we should merge this first, then do a sweep on mixins to get them writing to separate files, then remove the logic to open the files (parsing for KEY=VALUE)? What do you think?

vdice:
@carolynvs Good point.  I'd say we just go all the way (mixins place each output into sep file, as you've mentioned).  I'll refactor this changeset and work on the necessary PRs for the various mixins...

carolynvs:
Sounds great, thanks! 💯

docs/content/mixin-commands.md
```diff
@@ -159,43 +169,43 @@ upgrade:
...
-host=127.0.0.1
+**/cnab/app/porter/outputs/mysql-password**
+```
+updatedtopsecret
 ```
 
 # uninstall
 
-The upgrade command (required) is called from inside the invocation image during
+The uninstall command (required) is called from inside the invocation image during
```
carolynvs: oh gosh, thanks for fixing this! 🤦‍♀

docs/content/mixin-commands.md
```diff
@@ -159,43 +169,43 @@ upgrade:
 - helm:
     description: "Upgrade MySQL"
     name: porter-ci-mysql
+    replace: true
+    set:
+      mysqlPassword: "updatedtopsecret"
```
carolynvs: Maybe in a quick follow-up we should make this a parameter instead of hard-coded so that people don't get confused about why we are bothering to output it?

carolynvs: Also would we get any outputs from upgrade if we don't have both a step level outputs that emits `mysql-root-password` and `mysql-password` like install does, plus a a global outputs section with applyTo which includes upgrade?

examples/kubernetes-mixin-example/porter.yaml
```diff
@@ -39,5 +39,5 @@ uninstall:
       wait: true
 
 outputs:
-  - name: ip_address
+  - name: IP_ADDRESS
```
carolynvs: 👍

pkg/config/manifest.go
```diff
@@ -422,24 +422,17 @@ func (m *Manifest) GetSteps(action Action) (Steps, error) {
...
 	}
 
-	for _, assignment := range assignments {
-		parts := strings.SplitN(assignment, "=", 2)
-		if len(parts) != 2 {
-			return errors.Errorf("invalid output assignment %v", assignment)
-		}
-		outvar := parts[0]
-		outval := parts[1]
+	for outvar, outval := range assignments {
```
carolynvs: Oh boy this code gets so much nicer this way! ✨

carolynvs:
Oh, looks like something is wrong with the CLI tests but it wasn't immediately obvious what. Maybe something about printing the values of the outputs in the table?

pkg/context/context.go
```diff
@@ -117,35 +120,20 @@ func (c *Context) CopyFile(src, dest string) error {
 	return errors.WithStack(err)
 }
 
-// WriteOutput writes the given lines to a file in the
-// output directory
-func (c *Context) WriteOutput(lines []string) error {
-	exists, err := c.FileSystem.DirExists("/cnab/app/porter/outputs")
+// WriteOutput writes the provided bytes to a file named by the provided filename
+// to Porter's outputs directory
+func (c *Context) WriteOutput(filename string, bytes []byte) error {
```
carolynvs: I was just reviewing the mixin code for your porter-azure PR and realized that this function name is really confusing. When people see `p.Context.WriteOutput` it could easily be construed as "write output/logs to stdout". Which is my fault.

If you agree that the name is really confusing, now would be the best time to fix it. What do you think? Maybe we could change it to `WriteMixinOutputToFile` or something much more explicit.

vdice: Sounds good to me; I'll update this name here and in the mixin PRs to be `WriteMixinOutputToFile`

vdice:
> Oh, looks like something is wrong with the CLI tests but it wasn't immediately obvious what. Maybe something about printing the values of the outputs in the table?

Alas, the cli test is failing on the terraform test bundle as the PR to update the porter-terraform mixin is still outstanding (https://github.com/deislabs/porter-terraform/pull/17).  I've tested locally (and will re-confirm after the latest updates).  I think we may have to merge without the cli tests passing in CI...

carolynvs:
Gotcha, thanks how quickly I had forgotten. 😝

docs/content/mixin-commands.md
vdice: I can add a params section to the example snippet, certainly.

As this doc appeared to just focus on mixin-level outputs (and the corresponding mixin output files), I'm omitting mention of global outputs section here.  Indeed, they would need to be declared in order to be retrieved by `porter bundle outputs list`.

PR #448, opened 2019-07-12
Support Custom Actions with porter invoke --action ACTION

### LINKED ISSUE #413: Support custom actions
We should support custom actions. We need to consider what this experience might look like. Maybe:

> porter invoke —action <action-name> (other args like install).

Porter would need to validate that the specific action is in the bundle and handle errors when mixins used by the bundle don’t support said mixin.

carolynvs:
- Implements custom actions in porter and its runtime
- Implements custom actions in the exec mixin
- Documents custom actions at https://deploy-preview-448--porter.netlify.com/authoring-bundles/#custom-actions
- Updates the json schema to allow for custom actions to be defined in the manifest. Mixins need to define steps named invokeStep to work with custom actions. If they don't have a schema property named .* then they are assumed to not work with custom actions.

Closes #413


pkg/cnab/config_adapter/adapter_test.go
```diff
@@ -340,3 +341,69 @@ func TestPorter_generateDependencies(t *testing.T) {
...
+}
+
+func TestManifestConverter_generateDefaultAction(t *testing.T) {
+	a := ManifestConverter{}
+
+	testcases := []struct {
+		action     string
+		wantAction bundle.Action
+	}{
+		{"dry-run", bundle.Action{
```
vdice: Should we add a test case testing the default (meaning, the `default` case of the `switch` statement).  Perhaps with an action named... `zombies`?!

pkg/cnab/config_adapter/testdata/porter-with-custom-action.yaml
```diff
@@ -0,0 +1,54 @@
...
+        - -c
+        - echo The world is on fire
+
+zombies:
+  - exec:
+      description: "Trigger zombie apocalypse"
+      command: bash
+      arguments:
+        - -c
+        - echo oh noes my brains
```
vdice: 😋

pkg/config/manifest.go
```diff
@@ -202,6 +213,55 @@ type BundleConnection struct {
...
+	}
+
+	// Identify data from the manifest that wasn't mapped, those are custom actions
+	unmappedData := make(map[string]interface{})
+	err = yaml.Unmarshal(manifestData, &unmappedData)
+	if err != nil {
+		return nil, errors.Wrap(err, "error unmarshaling the untyped manifest")
+	}
+
+	objValue := reflect.ValueOf(manifest).Elem()
```
vdice: Can a comment be added here describing this section in layman's terms as well, like the others?

vdice: Wondering if we should split out the untyped unmarshaling and custom action marshaling into a helper func?  Like, `DeriveCustomActions` or some such... idk; just thinking on how quick comprehension might be improved...

pkg/porter/testdata/porter.yaml
```diff
@@ -1,22 +1,40 @@
 mixins:
-- exec
+  - exec
```
vdice: I see we also have a new `bundle-with-custom-action.yaml` bundle in testdata, which this now appears to be equivalent to.  Just wanted to check that this is intended.

pkg/config/manifest.go
carolynvs: I went with adding comments for now.

pkg/porter/testdata/porter.yaml
carolynvs: Yeah I don't know what our testdata rules are? Personally I like to keep the testdata close to the test, and not try to keep mega reusable test files.  They do have different data in them (one is missing the customActions section).

If we wanted to reduce the amount of duplicate testdata, I'd vote for a set of rules to follow and sweep in a separate issue because I know that I've done this throughout Porter. 😊

PR #451, opened 2019-07-12
feat(*): add bundle outputs commands

vdice:
Closes https://github.com/deislabs/porter/issues/445

### DIFF
--- a/pkg/porter/outputs.go
+++ b/pkg/porter/outputs.go
@@ -2,13 +2,30 @@ package porter
 
 import (
 	"encoding/json"
+	"fmt"
 	"os"
 	"path/filepath"
 	"sort"
 	"strings"
 
+	"github.com/deislabs/porter/pkg/context"
+	"github.com/deislabs/porter/pkg/printer"
+	tablewriter "github.com/olekukonko/tablewriter"
 	"github.com/pkg/errors"
 )
+
+// OutputShowOptions represent options for a bundle output show command
+type OutputShowOptions struct {
+	sharedOptions
+	Output string
+}
+
+// OutputListOptions represent options for a bundle output list command
+type OutputListOptions struct {
+	sharedOptions
+	RawFormat string
+	Format    printer.Format
+}
 
 // Output represents a bundle output
 type Output struct {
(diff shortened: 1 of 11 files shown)


pkg/porter/outputs.go
```diff
@@ -2,14 +2,31 @@ package porter
...
+// OutputShowOptions represent options for a bundle output show command
+type OutputShowOptions struct {
+	sharedOptions
+	Output string
+}
+
+// OutputListOptions represent options for a bundle output list command
+type OutputListOptions struct {
+	sharedOptions
+	RawFormat string
```
carolynvs: This is a question about something we could look at in a follow-up, not in this PR since I think it may touch a couple commands. 

Do we have these (raw format and format) in a common struct? Would it help if we did or would it just make another struct but not help consolidate validation logic?

pkg/porter/outputs.go
```diff
@@ -38,30 +55,125 @@ func (l Outputs) Less(i, j int) bool {
...
+// setting attributes of OutputListOptions as applicable
+func (o *OutputListOptions) Validate(args []string, cxt *context.Context) error {
+	// Ensure only one argument exists (claim name) if args length non-zero
+	err := o.sharedOptions.validateClaimName(args)
+	if err != nil {
+		return err
+	}
+
+	// If args length zero, attempt to derive claim name from context
+	err = o.sharedOptions.defaultBundleFiles(cxt)
```
carolynvs: Based on the comment, I expected an if statement here.

vdice: Good call; comment wording a bit confusing.  Updated.

pkg/porter/outputs.go
vdice: I don't think we do yet... but I wonder if we just add the `RawFormat` field to https://github.com/deislabs/porter/blob/4d5fbb8636186575bc856a98997929f8bce17b00/pkg/printer/printer.go#L24-L26 and we'd have what we need to remove duplication?

carolynvs: I like it! Yeah that would be a nice to have at some later point, but not this PR. I have an itchy trigger finger. 😀

PR #452, opened 2019-07-15
ref(*): DRY print options via updated PrintOptions struct

vdice:
Refactor stemming from https://github.com/deislabs/porter/pull/451#discussion_r303459986

* DRYs out applicable commands to use an updated `PrintOptions` struct now containing both `RawFormat` and `Format` values

### DIFF
--- a/cmd/porter/bundle.go
+++ b/cmd/porter/bundle.go
@@ -6,7 +6,6 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/deislabs/porter/pkg/porter"
-	"github.com/deislabs/porter/pkg/printer"
 )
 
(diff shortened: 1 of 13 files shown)


cmd/porter/bundle.go
```diff
@@ -106,12 +105,10 @@ Optional output formats include json and yaml.`,
 		Example: `  porter bundle list
   porter bundle list -o json`,
 		PreRunE: func(cmd *cobra.Command, args []string) error {
-			var err error
-			opts.Format, err = printer.ParseFormat(opts.RawFormat)
-			return err
+			return opts.ParseFormat()
```
carolynvs: 😍  This looks so much better!

PR #462, opened 2019-07-18
feat(schema.json): update manifest schema

vdice:
* Updates Porter's manifest schema to add credentials, outputs and parameters and their corresponding definitions

Closes https://github.com/deislabs/porter/issues/449

### DIFF
--- a/pkg/porter/templates/schema.json
+++ b/pkg/porter/templates/schema.json
@@ -1,15 +1,134 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
   "type": "object",
+  "definitions": {
+    "credential": {
+      "description": "Credential defines a particular credential, and where it should be placed in the invocation image",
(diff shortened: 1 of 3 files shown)


pkg/porter/templates/schema.json
```diff
@@ -1,15 +1,134 @@
...
-    "name": {
-      "type": "string"
+    ".*": {
+      "type": "array",
+      "items": {
+        "anyOf": []
+      }
+    },
+    "credentials": {
+      "additionalProperties": {
```
carolynvs: Would you switch to the pattern that we have been using to reference json schema definitions? I'd like to be consistent instead of introducing that other way the cnab-spec went with.

```json
"credentials": { "$ref": "#/definitions/credential" },
```

This applies to `credentials`, `outputs`, and the other new entries added to the top level properties in schema.json

PR #463, opened 2019-07-18
fix(config.go): ensure porter home path is absolute

vdice:
- Ensures GetPorterHome() returns the absolute path to porter home

Fixes https://github.com/deislabs/porter/issues/460

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -48,7 +48,7 @@ func New() *Config {
 	}
 }
 
-// GetHomeDir determines the path to the porter home directory.
+// GetHomeDir determines the absolute path to the porter home directory.
 func (c *Config) GetHomeDir() (string, error) {
 	if c.porterHome != "" {
 		return c.porterHome, nil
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -76,7 +76,14 @@ func (c *Config) GetHomeDir() (string, error) {
 		home = filepath.Dir(porterPath)
 	}
 
-	c.porterHome = home
+	// As a relative path may be supplied via EnvHOME,
```
carolynvs: Can you update the comment to explicitly call out that we need this for the volume mounts for outputs? I think it will help to know at least one case of why it needs to be absolute.

That would have been another fix btw. Instead of making this function return an absolute, would be to have the volume mount for outputs do the work there. 

I think this fix is fine and doesn't have any unwanted side effects 🤞, however I wanted to call out that we may add more volume mounts to Porter (probably for dependencies) and this problem may come up again that we need to check and convert to absolute paths. So I wanted to keep that in our heads. 😀

vdice: Comment updated.  Indeed, my fix started as only updating the outputs dir logic to return an absolute path but then I realized there may be other consumers of the this top-level "give me the porter home directory" (perhaps soon for dependencies, as you've mentioned) and so went this route.

Issue #465, opened 2019-07-19
How to sometimes uninstall everything, and sometimes just uninstall the app


carolynvs:
With dependencies, it may be desirable to uninstall just the app, but not the underlying dependencies (because they may still be in use, or you want to keep it around for backups, etc). How do we want to allow someone to do this in Porter?

One way would be a new flag that is just for uninstall:
```
porter uninstall wordpress --keep-deps
```

brandonjbjelland:
To me, the flag you're pitching feels like the right UX. A more flexible approach would allow an executor to list all deps to exclude from uninstall while any NOT listed would be uninstalled:

```bash
porter uninstall wordpress --keep-deps dependent-anycloud-core-infra,dependent-anycloud-mysql-db # if not listed, perhaps a CDN bundle would be uninstalled
# alternatively throwing the flag multiple times
porter uninstall wordpress --keep-deps dependent-anycloud-core-infra --keep-deps dependent-anycloud-mysql-db
```

carolynvs:
Closing. This is planned but will be tracked under a new PEP for dependency lifecycle management.

Issue #432, opened 2019-06-19
Wire up a bundle's outputs into bundle.dependencies.NAME.outputs


carolynvs:
Once we have #344 implemented and #430, then we can collect output from bundle dependencies, and inject them into the bundle as `bundle.dependencies.NAME.outputs`

Issue #69, opened 2018-12-03
Design: Dependency Graph Traversal


carolynvs:
Currently we only resolve dependencies down a single level. Eventually more complicated bundles will want to have dependencies that have dependencies too. Resolve and apply them from the bottom up.

glyn:
By _nested_ dependencies do you mean anything other than transitive dependencies? The use of a different term made me wonder.

I haven't discovered how dependencies are specified in Porter (I spent at least 5 minutes scanning the code ;-) ), but have you any thoughts about the approach you might take on dependency resolution? See [this](https://thefeedbackloop.xyz/thoughts-on-dependency-hell-is-np-complete/) and [this](https://research.swtch.com/version-sat) for example.

carolynvs:
D'oh! You caught me out, I didn't use the official term. 👀  I've updated the title to reflect what I meant (transitive dependencies, aka dependencies of my dependencies). Right now porter doesn't traverse the dependency graph and only resolves direct dependencies.

Yup, I've seen from maintaining [dep](https://github.com/golang/dep) the many ways that dependency management and resolution are hard which is why we stopped at direct dependencies for the moment while we are focused on the POC and developer experience for authoring bundles.

Here is an example of dependencies in porter:

https://github.com/deislabs/porter/blob/55e259ce57aabb7138c28668210f6dbfcc0243a0/build/testdata/bundles/wordpress/porter.yaml#L8-L12

glyn:
Thanks @carolynvs. I've raised https://github.com/deislabs/porter/issues/147 to discuss dependencies further (still waiting for slack access).

carolynvs:
I dunno if it's a glitch in the matrix or what? I got immediate access through this form:  https://slack.cncf.io

carolynvs:
Closing. This will be implemented through [PEP 003](https://github.com/getporter/proposals/blob/main/pep/003-dependency-namespaces-and-labels.md).

PR #466, opened 2019-07-19
Naive implementation of dependencies

carolynvs:
This is a naive implementation of dependencies to get us trying it out. We can fine tune from here. Some things are out of scope because this is a large PR and I had to stop somewhere for the first iteration.

A few key things that will probably get bikeshed that you'll want to keep in mind:
* We always run dependencies first, except for uninstall. When uninstalling we do the main bundle first, then the dependencies (i.e. we uninstall backwards). 
* There is another issue to think about allowing skipping running deps at the user's discretion (see #465).
* You can set parameters for dependencies by doing `--param DEP#PARAM=VALUE`
* Deps aren't run in any particular order at the moment (because we don't traverse a graph yet).

- [x] Define it
- [x] Put it in a bundle.json
- [x] Install it
- [x] Set params on it
- [x] Integration test it
- [x] Document it (halfway there, livin' on a prayer)
- [x] Update schema for it

Out of Scope:
* Solving a version range: That's gonna use https://github.com/google/go-containerregistry but will come in a follow-on PR. This only supports exact dependency requirements for now, i.e. the dependency tag includes the version it needs and it doesn't use a range. 
* Putting anything in cnab-go/duffle.
* Templating, such as rewiring back up `bundle.dependencies.NAME.outputs` will happen in #432
* dependencies of dependencies isn't implemented (see #69)


build/testdata/bundles/wordpress/porter.yaml
```diff
@@ -33,23 +34,37 @@ parameters:
...
       description: "Install Wordpress"
-      name: "{{bundle.parameters.wordpress-name }}"
+      name: "{{ bundle.parameters.wordpress-name }}"
       chart: stable/wordpress
       namespace: "{{ bundle.parameters.namespace }}"
       replace: true
       set:
         wordpressPassword: "{{ bundle.parameters.wordpress-password }}"
-        externalDatabase.database: "{{ bundle.dependencies.mysql.parameters.database-name }}"
+        #externalDatabase.database: "{{ bundle.dependencies.mysql.parameters.database-name }}"
```
carolynvs: These are just commented out temporarily until #432 puts it back into place. PR was getting huge.

pkg/cnab/extensions/solver.go
```diff
@@ -0,0 +1,52 @@
...
+			Name: name,
+			Tag:  fmt.Sprintf("%s:%s", bundle, version),
+		}
+		q = append(q, lock)
+	}
+
+	return q, nil
+}
+
+func (s *DependencySolver) ResolveVersion(name string, dep Dependency) (string, error) {
```
vdice: Should we handle the case where a dep has no tag (i.e., no `:mytag` value, as in `mycloud/mysql`)?  Would this be valid at this stage?  Or, perhaps it would've already been caught earlier in bundle validation, etc... but maybe as a safe-guard?

pkg/config/config.go
```diff
@@ -16,6 +16,12 @@ const (
 	// EnvHOME is the name of the environment variable containing the porter home directory path.
 	EnvHOME = "PORTER_HOME"
 
+	// EnvBundleName is the name of the environment variable containing the name of the bundle.
+	EnvBundleName = "CNAB_ACTION"
```
vdice: value of `"CNAB_ACTION"` seems incorrect here?

pkg/porter/dependencies.go
```diff
@@ -0,0 +1,204 @@
...
+		err := e.prepareDependency(dep)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (e *dependencyExecutioner) Execute() error {
```
vdice: Should we add a check here to see if `Prepare` needs calling?  (To avoid persishing!)  Perhaps return an error if `e.parentOpts` or `e.action` ain't valid?

pkg/porter/install.go
```diff
@@ -30,6 +30,17 @@ func (p *Porter) InstallBundle(opts InstallOptions) error {
 		return err
 	}
 
+	deperator := newDependencyExecutioner(p)
```
vdice: `dependerator`?  `depsecutioner`? so many choices!

pkg/cnab/extensions/solver.go
carolynvs: When no tag is specified, then the resolver is going to pick a version based on the available tags in the registry. I already had an unimplemented error but I see that I missed the case when the tag wasn't specified. I'll update it to print the unimplemented error there too. 

I'm a follow up PR I'm going to implement hitting the registry to resolve all the available tags, cache them, and solve for the best tag based on the constraints (in this case when there's no tag and no range, any tag will do and we can grab the highest version according to semver).

pkg/porter/install.go
carolynvs: I know right?! I had no clue what to call this. 😀

PR #477, opened 2019-07-24
removed iex (invoke-expression) from windows install script

brwilkinson:
updates the Windows/PowerShell install script to remove iex (invoke-expression).

Replace with the call operator &.

iex is considered a code smell, with some well known security implications.

### DIFF
--- a/scripts/install/install-windows.ps1
+++ b/scripts/install/install-windows.ps1
@@ -7,13 +7,13 @@ mkdir -f $PORTER_HOME
 
 (new-object System.Net.WebClient).DownloadFile("$PORTER_URL/$PORTER_VERSION/porter-windows-amd64.exe", "$PORTER_HOME\porter.exe")
 (new-object System.Net.WebClient).DownloadFile("$PORTER_URL/$PORTER_VERSION/porter-linux-amd64", "$PORTER_HOME\porter-runtime")
-echo "Installed $(iex "$PORTER_HOME\porter.exe version")"
+echo "Installed $(& $PORTER_HOME\porter.exe version)"
 
 $FEED_URL="$PORTER_URL/atom.xml"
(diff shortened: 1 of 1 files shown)


scripts/install/install-windows.ps1
```diff
@@ -7,13 +7,13 @@ mkdir -f $PORTER_HOME
...
 (new-object System.Net.WebClient).DownloadFile("$PORTER_URL/$PORTER_VERSION/porter-linux-amd64", "$PORTER_HOME\porter-runtime")
-echo "Installed $(iex "$PORTER_HOME\porter.exe version")"
+echo "Installed $(& $PORTER_HOME\porter.exe version)"
 
 $FEED_URL="$PORTER_URL/atom.xml"
-iex "$PORTER_HOME/porter mixin install exec --version $PORTER_VERSION --feed-url $FEED_URL"
-iex "$PORTER_HOME/porter mixin install kubernetes --version $PORTER_VERSION --feed-url $FEED_URL"
-iex "$PORTER_HOME/porter mixin install helm --version $PORTER_VERSION --feed-url $FEED_URL"
-iex "$PORTER_HOME/porter mixin install azure --version $PORTER_VERSION --feed-url $FEED_URL"
+& $PORTER_HOME/porter mixin install exec --version $PORTER_VERSION --feed-url $FEED_URL
```
carolynvs: If there is a whitespace in someone's USERPROFILE (which is where PORTER_HOME is derived), wouldn't we need quotes around `$PORTER_HOME/porter`?

brwilkinson: @carolynvs 
I believe the call operator will take care of that.
e.g.
$mypath = "D:\Source\abc def"
mkdir $mypath
copy-item -path C:\WINDOWS\system32\curl.exe -Destination $mypath
ls $mypath
& $mypath\curl --help

I also tested the install using: $PORTER_HOME="D:\Source\abc def\.porter"
That installs just fine.

PR #495, opened 2019-08-05
Update the workshop examples:

jeremyrickard:
* Add a bundle output to porter-tf and porter-tf-aci
* Update the update/uninstall steps to not print anything
* Updated porter-tf-aci to deploy a trivial webapp to ACI

Still need to update the porter-tf-aci README so people can use it.

### DIFF
--- a/workshop/porter-tf-aci/azure/bundle/porter.yaml
+++ b/workshop/porter-tf-aci/azure/bundle/porter.yaml
@@ -48,6 +48,17 @@ parameters:
   - name: database-name
     type: string
 
+outputs:
+  - name: IP_ADDRESS
+    type: string
+
+imageMap:
+  websvc:
+      description: "A simple web service"
+      imageType: "docker"
+      image: "jeremyrickard/devops-days-msp@sha256:85b1a9b4b60a4cf73a23517dad677e64edf467107fa7d58fce9c50e6a3e4c914"
+      digest: "sha256:85b1a9b4b60a4cf73a23517dad677e64edf467107fa7d58fce9c50e6a3e4c914"
+
 install:
    - azure:
       description: "Create an Azure Storage Account"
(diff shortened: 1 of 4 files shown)


workshop/porter-tf-aci/azure/bundle/porter.yaml
```diff
@@ -48,6 +48,17 @@ parameters:
   - name: database-name
     type: string
 
+outputs:
+  - name: IP_ADDRESS
+    type: string
+
+imageMap:
```
carolynvs: omg we are using image map? just for fun or?

workshop/porter-tf-aci/azure/bundle/porter.yaml
```diff
@@ -89,66 +100,25 @@ install:
...
-        tenant_id: "{{bundle.credentials.tenant_id}}"
-        client_id: "{{bundle.credentials.client_id}}"
-        client_secret: "{{bundle.credentials.client_secret}}" 
-        server-name: "{{bundle.parameters.server-name}}"
+  - exec: 
+      description: "Perform an upgrade"
+      command: "bash"
+      arguments:
+      - "-c"
+      - "echo Not implemented yet"
```
carolynvs: 👍

workshop/porter-tf-aci/azure/bundle/porter.yaml
jeremyrickard: ITS FOR REALZ! I mean for completeness.....

I was thinking we should add the ability to wire up images though! like in this ACI thing, it would be cool if i could jsut say ${bundle.images.WHATEVER} and we could figure out how to deal with that crazy relocation stuff later if needed

carolynvs: I demand an issue to 🚲 🏠

PR #504, opened 2019-08-06
Ref(*): use cnab go schema

vdice:
Updates Porter to allow full support for JSON Schema (via cnab-go) in parameter definitions.

Also,
- removes Porter's own `Required` field for a Parameter.  This was motivated due to a conflict with cnab-go's `Schema` struct, which has a `Required` field of type `[]string`.  We embed/in-line the `Schema` struct, so there may be only one of this key.  Porter still defaults `Required` to `true` on the generated cnab-go `bundle.Parameter` struct, for any parameter that has no default specified, but we no longer expose the ability for a manifest author to supply this value.

Depends on ~~https://github.com/deislabs/cnab-go/pull/90 and https://github.com/docker/cnab-to-oci/pull/57 (Need to remove overrides/fork revisions from `Gopkg.toml`)~~ https://github.com/deislabs/porter/pull/505

Closes https://github.com/deislabs/porter/issues/444

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -222,7 +222,7 @@
   version = "v1.1.1"
 
 [[projects]]
-  digest = "1:9f60acd7d2ca0060ebd9064d616b4a09c8906aff099897273fffb2ec81f04815"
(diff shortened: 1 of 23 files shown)


vdice:
Closed in favor of https://github.com/deislabs/porter/pull/511

PR #506, opened 2019-08-10
feat(ci): add bundle validation

vdice:
Passing validation depends on getting Porter's cnab-go use (and resulting generated bundle.json files) up to spec in https://github.com/deislabs/porter/pull/505

Closes https://github.com/deislabs/porter/issues/423


vdice:
Commented out the added azure pipelines script and added a TODO per https://github.com/deislabs/porter/issues/513

PR #507, opened 2019-08-10
fix(duffle.go): add check for output's applyTo before attempting to read

vdice:
Fixes bug where Porter would attempt to read/write an output without checking to see if the output applied to the current action.  Now, if the output does not apply to the current action, Porter does not attempt to read/write.

Note some TODOs exist around https://github.com/deislabs/cnab-go/pull/99

### DIFF
--- a/pkg/cnab/provider/upgrade.go
+++ b/pkg/cnab/provider/upgrade.go
@@ -69,7 +69,7 @@ func (d *Duffle) Upgrade(args ActionArguments) error {
 	runErr := i.Run(&claim, creds, d.Out)
 
 	// Add/update the outputs section of a claim and capture error
-	err = d.WriteClaimOutputs(&claim)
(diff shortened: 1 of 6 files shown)


pkg/cnab/provider/upgrade.go
```diff
@@ -66,7 +66,7 @@ func (d *Duffle) Upgrade(args ActionArguments) error {
 	runErr := i.Run(&claim, creds, d.Out)
 
 	// Add/update the outputs section of a claim and capture error
-	err = d.WriteClaimOutputs(&claim)
+	err = d.WriteClaimOutputs(&claim, string(config.ActionInstall))
```
carolynvs: This should be ActionUpgrade

vdice: Good catch, thank you.  Fixed.

carolynvs:
🔥 AT WILL

Issue #207, opened 2019-03-06
Define an intermediate representation of a porter manifest


carolynvs:
Currently porter only consumes a yaml file, then converts that to a go struct, loads the dependencies, starts executing the bundles and then keeps an in-memory representation of the bundle as it executes.

Having the input representation tied so closely to the in-memory/live representation of the bundle makes it hard to make changes to the porter behavior going forward. It also limits the languages and extensibility of porter (see #173).

If we had an intermediate format that others could provide to porter, not necessarily the current yaml structure, then that would allow stuff like [moopo](https://github.com/jdolitsky/porter-moon-demo), or [lupo](https://github.com/jdolitsky/porter-moon-demo), or viper provided markup languages like toml, hcl, etc.

carolynvs:
cc @jdolitsky

jdolitsky:
wow so fast!

carolynvs:
Your demo inspired me. 😍

carolynvs:
Suggestion from @simongdavies

> Make sure that the templating (such as mustache/handlebars) doesn't get embedded into the runtime representation so that other tools can generate the intermediate form without having to muck(?) with the templating language too

jdolitsky:
How might jsonschema help us with this?

carolynvs:
> How might jsonschema help us with this?

I'm not sure if it would? Honestly I don't have a design yet in my head about what this would need to look like as a technical solution. What I do have are requirements coming in from a few sides:

* Feelings from myself and @jeremyrickard that working with our current representation which is just a single Go struct that yaml tags that we use both to load in the porter.yaml, and then keep a live representation of the manifest as we work through the bundle at runtime is getting awkward.
* Comments from @simongdavies about how the Azure ARM team has run into trouble with how they represented their data structures, and trying to convey lessons learned about having multiple representations of the manifest so that it's not all glommed into a single data structure.
* Your feedback on wanting to use something other than yaml to feed into into Porter, potentially talking to porter programmatically instead through the mechanisms made for the cli.

carolynvs:
Decided to finally take this on as part of dependency evaluation. 💪

carolynvs:
Closing since using a language other than YAML is no longer on our roadmap.

PR #509, opened 2019-08-13
Split manifest in twain: porter.yaml and the runtime

carolynvs:
Manifest: The struct representing porter.yaml along with the logic to validate it.

RuntimeManifest: The logic and extra state for keeping the running represetantion of the manifest as we execute the bundle.

This is just the first step to see how this works for us without causing extra work or refactoring for the sake of refactoring. I decided to take this on before redoing templating/outputs for dependencies. If this doesn't look useful, we can ditch and move on though.

I'm not sure if this "closes" #207, or is just a first effort?


carolynvs:
As I'm building on top of this I've realized that I called the new manifest the wrong name. It's not really a "runtime" manifest because it's used when porter is executing any command (like build, install, etc) not just when it's running inside a bundle.

LivingManifest? ManifestConCarne?

pkg/config/manifest.go
```diff
@@ -139,72 +186,84 @@ func (od *OutputDefinition) Validate() error {
...
 
-	outputs := make(map[string]interface{})
-	depVals["outputs"] = outputs
-	// TODO: Populate dependency output lookups
-	// TODO: Add outputs onto sensitive stuff
+// GetDescription returns a description of the step.
+// Every step must have this property.
+func (s *Step) GetDescription() (string, error) {
+	if s.Data == nil {
+		return "", errors.New("empty step data")
```
vdice: I'm assuming this wasn't changed here, but I had the thought:  I wonder if we want to append `"Must minimally provide a step description"` or some such here to this error message?

pkg/porter/dependencies.go
```diff
@@ -16,7 +16,7 @@ type dependencyExecutioner struct {
 	*context.Context
 	Resolver BundleResolver
 	CNAB     CNABProvider
-	Manifest *config.Manifest
+	Manifest *config.RuntimeManifest
```
vdice: I wonder if we'd want to change this `Manifest` field name to `RuntimeManifest`?

pkg/config/manifest.go
carolynvs: No code was changed, just moved into a separate file.

pkg/porter/dependencies.go
carolynvs: I didn't rename this field just yet because it would have touched a _bunch_ of files and made it hard to really tell what I was thinking about changing.

carolynvs:
> A new and/or updated example bundle would help me see its effects

This isn't a change that will affect the bundle format, so the example bundles are untouched. Take a look at https://github.com/deislabs/porter/issues/207 for the reasoning behind the requested change. The really short answer is to have different representations of the manifest, one just for the file format on disk and a different one that represents what we need for executing a bundle and holding all the various state necessary to run the bundle.

carolynvs:
Ok, I've reworked it so that `porter.Manifest` is the on-disk porter.yaml and does not need to be renamed eventually. The `porter run` command instantiates a `RuntimeManifest` when it's in a CNAB environment and needs to execute a bundle, so that the name "runtime manifest" really matches what it is now.

@vdice With that in mind, what do you think?

* Manifest = the config from the porter.yaml file plus the validate functions
* RuntimeManifest = the state and the logic for executing that manifest at runtime inside of a bundle.

vdice:
@carolynvs I like it!  👍 The separation makes sense.

PR #505, opened 2019-08-09
WIP chore(*): bump cnab-go dep

vdice:
Bumps to latest cnab-go [release](https://github.com/deislabs/cnab-go/releases/tag/v0.3.0-beta1)

Currently pending on corresponding duffle and cnab-to-oci bumps...

For the former duffle dep, https://github.com/deislabs/cnab-go/pull/97 and https://github.com/deislabs/cnab-go/pull/98 should represent the last few ports needed to remove the duffle dep from this repo...

PR #511, opened 2019-08-13
Ref(*): cnab go updates

vdice:
Merger of https://github.com/deislabs/porter/pull/504 and https://github.com/deislabs/porter/pull/505

From #504 : 

Updates Porter to allow full support for JSON Schema (via cnab-go) in parameter definitions.

Also,
- removes Porter's own `Required` field for a Parameter.  This was motivated due to a conflict with cnab-go's `Schema` struct, which has a `Required` field of type `[]string`.  We embed/in-line the `Schema` struct, so there may be only one of this key.  Porter still defaults `Required` to `true` on the generated cnab-go `bundle.Parameter` struct, for any parameter that has no default specified, but we no longer expose the ability for a manifest author to supply this value.

From #505 : 

Bumps to latest cnab-go [release](https://github.com/deislabs/cnab-go/releases/tag/v0.3.0-beta1)

Also:
- drops dependency on `deislabs/duffle`

TODO:
 - [x] update cnab-go dep once more to get the duffle ports needed to remove the `deislabs/duffle` dependency.


Gopkg.toml
```diff
@@ -1,16 +1,14 @@
-# Stay exactly in step with a specific duffle release
-[[constraint]]
-  name = "github.com/deislabs/duffle"
-  version = "=0.2.0-beta.1"
-
 [[constraint]]
   name = "github.com/deislabs/cnab-go"
-  version = "0.2.0-beta.1"
+  branch = "master"
+  # version = "0.3.0-beta.1"
```
carolynvs: Can we do a release of cnab-go today and then not rely on master? That doesn't need to hold up this PR, just asking if there's a reason we can't do another release.

pkg/cnab/config_adapter/testdata/porter-with-parameters.yaml
```diff
@@ -25,15 +23,12 @@ parameters:
       - red
       - purple
       - pink
-    required: true
```
carolynvs: I don't understand why we can't make a parameter required anymore. Can you walk me through that?

If it's because of a technical decision (and not because of a limitation of the spec), let's brainstorm how to get that functionality back.

pkg/cnab/extensions/testdata/bundle.json
```diff
@@ -42,9 +42,9 @@
       "env": "equux"
     }
   },
-  "requiredExtensions": {
-    "dependencies": "https://cnab.io/specs/v1/dependencies.schema.json"
-  },
+  "requiredExtensions": [
+    "https://cnab.io/specs/v1/dependencies.schema.json"
```
carolynvs: Looks like required extensions changed when I wasn't around? This isn't something you need to change. I'll follow up on it

https://github.com/deislabs/cnab-spec/blob/master/101-bundle-json.md#required-extensions

```json
"requiredExtensions":[ 
      "io.cnab.dependencies"
   ],
```

pkg/cnab/provider/claims.go
```diff
@@ -1,17 +1,23 @@
...
+	// ClaimsDirectory represents the name of the directory where claims are stored
+	ClaimsDirectory = "claims"
+)
+
 func (d *Duffle) NewClaimStore() claim.Store {
 	// TODO: I'm going to submit a PR after this so that GetHomeDir doesn't return an error
 	homepath, _ := d.GetHomeDir()
-	h := home.Home(homepath)
-	return claim.NewClaimStore(crud.NewFileSystemStore(h.Claims(), "json"))
+	claimsPath := filepath.Join(homepath, ClaimsDirectory)
```
carolynvs: Is this something that should be moved into a function on the Config struct like 

https://github.com/deislabs/porter/blob/94a5452744191574909e5b5249c39b4fd2958f7b/pkg/config/config.go#L121

pkg/cnab/provider/credentials.go
```diff
@@ -6,7 +6,11 @@ import (
 
 	"github.com/deislabs/cnab-go/bundle"
 	"github.com/deislabs/cnab-go/credentials"
-	"github.com/deislabs/duffle/pkg/duffle/home"
+)
+
+const (
+	// CredentialsDirectory represents the name of the directory where credentials are stored
+	CredentialsDirectory = "credentials"
```
carolynvs: Same comment as above, seems like this should be a function on Config?

pkg/cnab/provider/parameters.go
```diff
@@ -51,20 +50,21 @@ func (d *Duffle) loadParameters(claim *claim.Claim, rawOverrides map[string]stri
 	// *but* should not apply to this action.
 	// When this occurs, we set an override to the current value in the claim such that
 	// bundle.ValuesOrDefaults does not return an error
-	// TODO: will have to refactor loop once required is back as a bool on the ParameterDefinition
-	for _, required := range bun.Parameters.Required {
-		if _, exists := rawOverrides[required]; !exists {
-			if !appliesToAction(action, bun.Parameters.Fields[required]) {
-				overrides[required] = claim.Parameters[required]
+	for name, param := range bun.Parameters {
```
carolynvs: This logic makes my head hurt...  Am I reading it right that we want to populate overrides with the value from the claim when the parameter is required but the user didn't supply a value?

I'm worried that we may not have adequate test coverage for this code here. But that was the case before this code was changed so cest le vie! 😂

pkg/config/manifest.go
```diff
@@ -89,6 +78,8 @@ type CredentialDefinition struct {
 	Location `yaml:",inline"`
 }
 
+// TODO: use cnab-go's bundle.Location instead, once yaml tags have been added
```
carolynvs: Have we been making cnab-go add yaml tags so that we can straight up reuse their structs? Just checked and I guess we have!

I'm surprised that no one complained at us. 😀

Gopkg.toml
vdice: @carolynvs Yup, only intending to merge after this is updated with a new release 👍

Gopkg.lock
```diff
@@ -1217,6 +1268,25 @@
   pruneopts = "NUT"
   revision = "508f945e1a9bd8b5e3a9444bb2cdd13d0896132e"
 
+[[projects]]
+  digest = "1:a4824d8df1fd1f63c6b3690bf4801d6ff1722adcb3e13c0489196a7e248d868a"
+  name = "google.golang.org/appengine"
```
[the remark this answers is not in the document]
vdice: 🤔 Indeed!  I don't see any direct use of, say, `appengine`, in cnab-go... a dep of a dep (of a dep?!)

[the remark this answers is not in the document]
carolynvs: IT'S A MYSTERY! 🤷‍♀

pkg/cnab/config_adapter/testdata/porter-with-parameters.yaml
vdice: I dropped this due to a conflict of the embedded [Schema struct](https://github.com/deislabs/cnab-go/blob/master/bundle/definition/schema.go#L55) from deislabs/cnab-go.  I though it might be a candidate to drop as I didn't see its usefulness: if a parameter has no default, it would need to be required (and still is, since we do set it so in the resulting `bundle.json` in this case), but if a parameter does have a default, I didn't understand the logic of being able to deem it as required (it made better sense to be optional/non-required).

Nevertheless, opinions aside, let's brainstorm on keeping this functionality assuming we want it to stay put.  Since, as mentioned above, a `required` field is already being used when unmarshaling (by the embedded Schema definition), we could rename it to `isRequired` for an easy re-addition?

pkg/cnab/provider/parameters.go
vdice: > Am I reading it right that we want to populate overrides with the value from the claim when the parameter is required but the user didn't supply a value?

That is indeed the case.  Agreed, it's not very understandable logic.  This came in with https://github.com/deislabs/porter/pull/476 and included a test in https://github.com/deislabs/porter/blob/2a1d699b49b628fbc68a2bfdd52428b9a5434ed3/pkg/cnab/provider/parameters_test.go#L157-L191

However, perhaps we can create a work item to revisit?

pkg/config/manifest.go
vdice: Yes.  I wanted to prioritize using the canonical structs defined in cnab-go as much as possible/applicable...

pkg/config/manifest.go
```diff
@@ -59,26 +61,13 @@ type ParameterDefinition struct {
 	Name      string `yaml:"name"`
 	Sensitive bool   `yaml:"sensitive"`
 
-	// These could be swapped out with an inline bundle.ParameterDefinition
-	// from cnab-go, e.g. bundle.ParameterDefinition `yaml:",inline"`
-	Description string    `yaml:"description,omitempty"`
-	Destination *Location `yaml:"destination,omitempty"`
+	// These fields represent a subset of bundle.Parameter as defined in deislabs/cnab-go,
+	// minus the 'Description' field (definition.Schema's will be used)
+	Definition  string    `yaml:"definition"`
```
carolynvs: I'm not sure that this field is needed? It's not used in our code anywhere (the definition in the bundle.json is set to the parameter name currently).

Was the intent to allow them to override it with this? I thought at one point we were going to set it to "parameter_foo", "output_foo".

pkg/cnab/config_adapter/testdata/porter-with-parameters.yaml
carolynvs: Ah okay, no I wasn't trying to fight to keep it in. 😁 I just wasn't sure if we needed to keep it or not. Your explanation makes sense for why you removed it (default does all the work for us).

pkg/cnab/provider/parameters.go
carolynvs: Let's just bury the body and back away slowly. Thanks for digging up the test to make me feel better! 🙌

vdice: :joy:

Gopkg.toml
vdice: Version now updated to latest `v0.3.1-beta1`

pkg/cnab/provider/claims.go
vdice: Good call; added in most recent commit

pkg/cnab/provider/credentials.go
vdice: Good call; added in most recent commit (function already existed 😅 )

pkg/config/manifest.go
vdice: Good catch; removed!

vdice:
https://dev.azure.com/deislabs/porter/_build/results?buildId=2610 has passed but hasn't been updated here...

PR #514, opened 2019-08-14
Write claim outputs on invoke

carolynvs:
This is a follow-on to #507. I noticed during the review that it was missing writing the claim outputs during the new invoke action.

### DIFF
--- a/pkg/cnab/provider/invoke.go
+++ b/pkg/cnab/provider/invoke.go
@@ -62,10 +62,18 @@ func (d *Duffle) Invoke(action string, args ActionArguments) error {
 	}
 
 	// Run the action and ALWAYS write out a claim, even if the action fails
-	err = i.Run(&claim, creds, d.Out)
+	runErr := i.Run(&claim, creds, d.Out)
+
+	// Add/update the outputs section of a claim and capture error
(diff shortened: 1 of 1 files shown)


carolynvs:
What should take precedent? A custom action declaring that it's stateless or it actually producing an ouput? 🤔

carolynvs:
For now I'm going to say that if it makes an output, we should write it, even if the action did say that it's stateless. I think this is a good question for bundle validation, something that we should try to catch before we start executing the bundle. I'll add an issue to think about it further.

carolynvs:
I started writing up the issue but now I'm not sure we need it:

Stateless: Should the execution of the action be recorded? For example 'dry run' is stateless. This shouldn't affect this PR.

Modifies: Can the action modify resources managed by the bundle? Status is an example of a "modifies: false, stateless: false" action. I can see an action that is modfies false potentially generating an output though. For example, one that spits out your kubeconfig for you when you lose it. 😀 

So I think what we have here is fine.

vdice:
@carolynvs ah, thank you for the clarification on the two attributes.  Agreed, I don't think there is any follow-up needed here.  👍

PR #517, opened 2019-08-14
Use debian only for base image

carolynvs:
Install ca-certificates using apt instead of copying from another base image to not have to download multiple base images.


jeremyrickard:
YES

PR #538, opened 2019-08-20
Default dependency tag when not provided 

### LINKED ISSUE #534: Select an appropriate tag when none is specified for a dependency
When a dependency doesn't have a tag specified, and no range either:

1. List the tags for the bundle
2. Order the tags by semver rules in descending order. Use allow prerelease to filter out tags for prerelease versions.
3. Select the highest version.
4. If no semver versions are defined but a latest tag exists, select latest.

carolynvs:
Pick the highest semver tag available, respecting the allow prereleases flag, and when no semver tags are defined, try to fallback to 'latest'.

Closes #534


pkg/cnab/extensions/solver.go
```diff
@@ -42,16 +45,57 @@ func (s *DependencySolver) ResolveDependencies(bun *bundle.Bundle) ([]Dependency
...
+			return s.determineDefaultTag(dep)
 		}
 	}
 
-	return "", errors.Errorf("not implemented: dependency version range specified for %s", name)
+	return "", errors.Errorf("not implemented: dependency version range specified for %s", alias)
+}
+
+func (s *DependencySolver) determineDefaultTag(dep Dependency) (string, error) {
+	tags, err := crane.ListTags(dep.Bundle)
```
vdice: crane ftw!

carolynvs: Yes it is doing a lot of heavy lifting under the hood. You have no idea!

PR #541, opened 2019-08-21
Split the exec arguments into flags and args

carolynvs:
See the mixin doc for the new syntax https://deploy-preview-541--porter.netlify.com/mixins/exec/#mixin-syntax

Review this commit by commit, most of the files is me just tweaking existing usage of the mixin to be consistent with this change.


pkg/exec/flags_test.go
```diff
@@ -0,0 +1,54 @@
+package exec
+
+import (
+	"sort"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
```
vdice: Should we add a repeated flag test?

pkg/exec/flags.go
```diff
@@ -0,0 +1,136 @@
...
+
+	return result
+}
+
+// UnmarshalYAML takes input like this
+// flags:
+//   flag1: value
+//   flag2: value
+//   flag3:
+//	 - value1
```
vdice: nit: this line and 110 seems to have a bit of extra spacing?

PR #547, opened 2019-08-22
Rename claim to instance

### LINKED ISSUE #330: Porter bundles vs claims
A `porter bundle list` command was added in https://github.com/deislabs/porter/pull/329 which lists _installed_ bundles.  Technically, it is just listing the [claims](https://github.com/deislabs/cnab-spec/blob/master/400-claims.md) that have been generated/updated via `porter {install,upgrade,uninstall}` commands.  Therefore, the command would arguably be better categorized via `porter claim list`... and then, `porter claim show` could output a specific claim, etc.

Contrast this with the accepted interpretation of a `bundle`, which is a representation of a CNAB bundle, e.g., a [bundle.json file](https://github.com/deislabs/cnab-spec/blob/master/101-bundle-json.md).  One could/should feasibly be able to see a listing of bundles, if accessible (via local store, etc.), outside of needing to install them.  I believe Porter's current direction for listing these bundles is in https://github.com/deislabs/porter/issues/322.

In other words, a `bundle` is a representation of a CNAB bundle that may or may not have been installed and a `claim` represents an installation of a `bundle` in varying states.

Thoughts on how to more accurately portray a listing of bundle installations (claims) in Porter?  Simply update to `porter claim list`, opt for a different noun or other?

carolynvs:
* porter bundle list -> porter instances list
* alias porter instances list to porter list
* porter bundle outputs list -> porter instance outputs list
* porter bundle outputs show -> porter instance outputs show
* Made positional argument consistent for specifying the instance name
when listing outputs for an instance. I made it a flag since outputs
show uses a flag. Positional arguments should be used when it is the
name of the resource. In this case the resource is output, so instance
should have been a flag to begin with.
* Updated documentation to match.

Closes #330

### DIFF
--- a/cmd/porter/bundle.go
+++ b/cmd/porter/bundle.go
@@ -1,14 +1,12 @@
 package main
 
 import (
-	"strings"
-
 	"github.com/spf13/cobra"
 
 	"github.com/deislabs/porter/pkg/porter"
 )
 
-func buildBundlesCommand(p *porter.Porter) *cobra.Command {
+func buildBundleCommands(p *porter.Porter) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:     "bundles",
 		Aliases: []string{"bundle"},
@@ -21,28 +19,12 @@ func buildBundlesCommand(p *porter.Porter) *cobra.Command {
 
 	cmd.AddCommand(buildBundleCreateCommand(p))
 	cmd.AddCommand(buildBundleBuildCommand(p))
-	cmd.AddCommand(buildBundleListCommand(p))
 	cmd.AddCommand(buildBundleInstallCommand(p))
 	cmd.AddCommand(buildBundleUpgradeCommand(p))
 	cmd.AddCommand(buildBundleInvokeCommand(p))
 	cmd.AddCommand(buildBundleUninstallCommand(p))
-	cmd.AddCommand(buildBundleShowCommand(p))
-	cmd.AddCommand(buildBundleOutputCommand(p))
 
 	return cmd
-}
-
-func buildBundleAliasCommands(p *porter.Porter) []*cobra.Command {
-	return []*cobra.Command{
-		buildCreateCommand(p),
-		buildBuildCommand(p),
-		buildInstallCommand(p),
-		buildUpgradeCommand(p),
-		buildUninstallCommand(p),
-		buildInvokeCommand(p),
-		buildPublishCommand(p),
-		buildShowCommand(p),
-	}
 }
 
 func buildBundleCreateCommand(p *porter.Porter) *cobra.Command {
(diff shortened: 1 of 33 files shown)


docs/content/cli.md
```diff
@@ -250,10 +250,10 @@ Global Flags:
 
 ### Bundle List
 
-This command is available both as `porter bundle list` and `porter bundles list`.
+This command is available both as `porter instance list` and `porter bundles list`.
```
vdice: `s/bundles/instances/`, right?

docs/content/cli.md
```diff
@@ -304,17 +304,17 @@ Global Flags:
 
 ### Bundle Outputs List
 
-Outputs from a bundle claim can be listed via `porter bundle outputs list`.
+Outputs from a bundle claim can be listed via `porter instance outputs list`.
 
 ```console
- $ porter bundle outputs list --help
+ $ porter instance outputs list --help
```
vdice: line 311 below: `s/bundle/instance`?

docs/content/cli.md
```diff
@@ -304,17 +304,17 @@ Global Flags:
...
 Displays a listing of bundle outputs.
 
 Usage:
-  porter bundles output list [CLAIM] [flags]
+  porter bundles output list [INSTANCE] [flags]
```
vdice: `s/bundles/instances`?

cmd/porter/outputs.go
```diff
@@ -0,0 +1,73 @@
...
+
+	return &cmd
+}
+
+func buildBundleOutputShowCommand(p *porter.Porter) *cobra.Command {
+	opts := porter.OutputShowOptions{}
+
+	cmd := cobra.Command{
+		Use:   "show NAME [--instance|-i INSTANCE]",
+		Short: "Show a bundle output",
```
vdice: Perhaps `Show a bundle instance output` or `Show an output of a bundle instance`?

cmd/porter/bundle.go
```diff
@@ -168,23 +104,14 @@ For instance, the 'debug' driver may be specified, which simply logs the info gi
...
-		"group": "alias",
-	}
-	return cmd
-}
-
 func buildBundleUpgradeCommand(p *porter.Porter) *cobra.Command {
 	opts := porter.UpgradeOptions{}
 	cmd := &cobra.Command{
-		Use:   "upgrade [CLAIM]",
+		Use:   "upgrade [INSTANCE]",
```
vdice: line 111 below: perhaps update to `Upgrade a bundle instance`?  (Would apply similarly across this file)

docs/content/cli.md
carolynvs: Oops, find/replace strikes again. Fixing. I have next up on my list to autogenerate this page using cobra so I'm not too worried about what's on there though.

PR #570, opened 2019-08-29
Automatically scrape mixin outputs

carolynvs:
I had all the wiring but forgot to call process for each output type to see if a mixin had implemented these. Oops!

While I was in there I realized that maybe iterating multiple times over a buffer was bad(?) so I switched to keeping the command output as a string.


carolynvs:
I'm adding integration test bundles for this, will merge when I'm done.

tests/outputs_test.go
```diff
@@ -0,0 +1,76 @@
+package tests
```
vdice: Did this test need `// +build integration` at the top?  Or is it runnin' right w/o it?

PR #579, opened 2019-09-04
feat(*): enable parameters of type file in Porter manifest

vdice:
Enables Porter manifest parameters of type `file`.  Ref https://github.com/deislabs/porter/issues/480

TODO:
 - [x] remove `vendor` mods:  PR changes to `cnab-go` and update Gopkg w/ temp commit, etc.... or something else entirely (see https://github.com/deislabs/cnab-go/issues/117)

~~Note: validation broken pending https://github.com/deislabs/porter/pull/581~~

### DIFF
--- a/pkg/cnab/provider/parameters.go
+++ b/pkg/cnab/provider/parameters.go
@@ -71,17 +71,17 @@ func (d *Duffle) loadParameters(claim *claim.Claim, rawOverrides map[string]stri
 }
 
 func (d *Duffle) getUnconvertedValueFromRaw(def *definition.Schema, key, rawValue string) (string, error) {
-		// the parameter value (via rawValue) may represent a file on the local filesystem
-		if def.Type == "string" && def.ContentEncoding == "base64" {
-			if _, err := d.FileSystem.Stat(rawValue); err == nil {
-				bytes, err := d.FileSystem.ReadFile(rawValue)
-				if err != nil {
-					return "", errors.Wrapf(err, "unable to read file parameter %s", key)
-				}
-				return base64.StdEncoding.EncodeToString(bytes), nil
+	// the parameter value (via rawValue) may represent a file on the local filesystem
+	if def.Type == "string" && def.ContentEncoding == "base64" {
+		if _, err := d.FileSystem.Stat(rawValue); err == nil {
+			bytes, err := d.FileSystem.ReadFile(rawValue)
+			if err != nil {
+				return "", errors.Wrapf(err, "unable to read file parameter %s", key)
 			}
+			return base64.StdEncoding.EncodeToString(bytes), nil
 		}
-		return rawValue, nil
+	}
+	return rawValue, nil
 }
 
 // TODO: remove in favor of cnab-go logic: https://github.com/deislabs/cnab-go/pull/99
--- a/pkg/cnab/provider/parameters.go
+++ b/pkg/cnab/provider/parameters.go
@@ -1,6 +1,7 @@
 package cnabprovider
 
 import (
+	"encoding/base64"
 	"fmt"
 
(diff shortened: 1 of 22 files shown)


vdice:
> I think you are all set to rebase your cnab-go branch off of v0.4.0-beta1 and then we can reference your branch here and merge.

@carolynvs Oops, I rebased my cnab-go branch off of master as opposed to `v0.4.0-beta1`, but the only extra change appears to be:
```
--- a/vendor/github.com/deislabs/cnab-go/claim/claim.go
+++ b/vendor/github.com/deislabs/cnab-go/claim/claim.go
@@ -48,7 +48,7 @@ type Claim struct {
 }

 // ValidName is a regular expression that indicates whether a name is a valid claim name.
-var ValidName = regexp.MustCompile("^[a-zA-Z0-9_-]+$")
+var ValidName = regexp.MustCompile("^[a-zA-Z0-9._-]+$")
```

carolynvs:
master is fine, we just needed to not move backwards is all 👍

pkg/config/manifest.go
```diff
@@ -170,7 +178,42 @@ func (od *OutputDefinition) Validate() error {
...
+func (pd *ParameterDefinition) Validate() error {
+	if pd.Name == "" {
+		return errors.New("parameter name is required")
+	}
+
+	schemaValidationErrs, err := pd.Schema.Validate(pd)
+	if err != nil {
+		// Porter supports declaring a parameter of type: "file",
+		// which we will convert to the appropriate bundle.Parameter type in adapter.go
+		if err.Error() != `unable to build schema: error unmarshaling type from json: "file" is not a valid type` {
```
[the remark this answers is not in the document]
vdice: I believe it previously would have stopped on the first error.  However, checking the error string did seem fragile/error-prone, so I've updated to checking the modified copy as suggested.  Let me know how it looks, when convenient: https://github.com/deislabs/porter/pull/579/commits/56d52a12818202cd625a2a8d2d9176d5d292db21

pkg/cnab/provider/parameters.go
```diff
@@ -25,7 +26,19 @@ func (d *Duffle) loadParameters(claim *claim.Claim, rawOverrides map[string]stri
 			return nil, fmt.Errorf("definition %s not defined in bundle", param.Definition)
 		}
 
-		value, err := def.ConvertValue(rawValue)
+		unconverted := rawValue
```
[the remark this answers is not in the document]
vdice: Good call; updated; PTAL when convenient: https://github.com/deislabs/porter/pull/579/commits/56d52a12818202cd625a2a8d2d9176d5d292db21

pkg/config/manifest.go
```diff
@@ -112,6 +112,47 @@ type ParameterDefinition struct {
...
+
+	return result.ErrorOrNil()
+}
+
+// DeepCopy copies a ParameterDefinition and returns the copy
+func (pd *ParameterDefinition) DeepCopy() *ParameterDefinition {
+	return &ParameterDefinition{
+		Name:        pd.Name,
+		Sensitive:   pd.Sensitive,
+		ApplyTo:     pd.ApplyTo,
```
carolynvs: This is keeping the same slice and backing array for Apply to. So if you modified the copy of ApplyTo, it would modify the original.

What you can do instead to make a copy is rely on Go copy semanitcs (which copies the contents of all the fields for you automatically) and then copy the slice yourself.

https://play.golang.org/p/jVOtUyQ0rXj

vdice: Ah, thank you!  Updated.

carolynvs: STANDUP MERGE IN PROGRESS!

vdice:
(last force-pushed commit represents rebase off master)

PR #581, opened 2019-09-04
fix(Makefile): fix bundle validation by fetching/supplying definitions schema

vdice:
Now that the bundle schema [references](https://github.com/deislabs/cnab-spec/blob/master/schema/bundle.schema.json#L219) the new [definitions.schema.json](https://github.com/deislabs/cnab-spec/blob/master/schema/definitions.schema.json), we need to fetch this referenced schema and supply it to the `ajv` cli for validation.  (This manual step is needed until https://github.com/deislabs/cnab.io/issues/15 is closed.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -157,19 +157,30 @@ else
 	cd $(EXAMPLES_DIR)/$(BUNDLE) && porter publish
 endif
 
-BUNDLE_SCHEMA_PATH ?= /tmp/bundle.schema.json
+BUNDLE_SCHEMA      := bundle.schema.json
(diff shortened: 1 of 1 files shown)


carolynvs:
This looks good but I was hoping to see publishexamplebundles pass on the PR. It doesn't seem to have run though?

vdice:
@carolynvs previously/currently we only publish bundles on merges to master.  Perhaps the thinking was we might not want to publish them from some/all PRs.  We can update the logic to run that task on PRs, though.  WDYT?

carolynvs:
Thanks! I was looking for validate. 👍

PR #582, opened 2019-09-04
Default the mixin feed url to DeisLabs

carolynvs:
* Unless the user specifies `--url` or `--feed-url`, use the atom feed for deislabs
* Update all the install instructions for the mixins to not specify `--feed-url`
* Update stray references of --feed-url to not use it unless strictly necessary

### DIFF
--- a/cmd/porter/mixins.go
+++ b/cmd/porter/mixins.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"fmt"
+
 	"github.com/deislabs/porter/pkg/mixin"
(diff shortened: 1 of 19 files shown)


carolynvs:
The install check is failing because it is using `latest` of porter which doesn't have my changes and I changed the install script for Porter. I'm not sure there is a good way to test the install script inside of a PR, short of having it do a publish to a test blob and pulling from it. Not sure it's worth the work though to test a change like this? It happens to rarely. Maybe I'm wondering instead if the install check should only happen on master since it can't detect problems with a PR anyway, it's a lagging test. 🤔

PR #606, opened 2019-09-13
Add --tag to porter credentials generate

### LINKED ISSUE #492: Support --tag for porter credentials generate
```
$ porter credentials generate --tag TAG
```

This should pull the bundle first if necessary (just like install does) and then do the generation. The full user flow we want to support is this:

```
$ porter credentials generate --tag TAG
$ porter install --tag TAG
```

jeremyrickard:
This PR adds the --tag option to credentials generate so people can install from a tag without any unpleasant work arounds. It reuses the lifecycle methods, maybe we should rename? Although I feel like generating a credential is kind of related to the lifecycle of the bundle. Ideally, we can link this in automagically to the `invoke` actions next. No credential specified? GENERATE THE HECK OUT OF ONE!!!

This also does a small fix to the help text. 

Fixes #492

### DIFF
--- a/docs/content/cli/credentials_generate.md
+++ b/docs/content/cli/credentials_generate.md
@@ -36,8 +36,9 @@ porter credentials generate [NAME] [flags]
 ```
   porter credential generate
   porter bundle credential generate kubecred --insecure
-  porter bundle credential generate kubecred --file myapp/bundle.json
-  porter bundle credential generate kubecred --file myapp/bundle.json --dry-run
+  porter bundle credential generate kubecred --file myapp/porter.yaml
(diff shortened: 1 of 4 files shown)


docs/content/cli/credentials_generate.md
```diff
@@ -36,8 +36,9 @@ porter credentials generate [NAME] [flags]
 ```
   porter credential generate
   porter bundle credential generate kubecred --insecure
-  porter bundle credential generate kubecred --file myapp/bundle.json
-  porter bundle credential generate kubecred --file myapp/bundle.json --dry-run
+  porter bundle credential generate kubecred --file myapp/porter.yaml
```
carolynvs: BEST FEATURE EVER 😀

PR #621, opened 2019-09-16
Pass build context to mixins

### LINKED ISSUE #262: Provide mixin steps to mixin during build
When a mixin is run, we provide the step mixin to it knows what to do. But when we ask the mixin to provide the lines for the Dockerfile, we don't give it any context from the steps it will be asked to run.

Depending on the mixin, it would really help it to know what it's going to need to do at runtime. So we should pass in a list of all the steps associated with it on STDIN and let the mixin use that information to maybe do extra work during build.

Here is an example of what the data will look like that is passed to the mixin on STDIN:

```yaml
install:
- helm:
    description: install mysql
    chart: stable/mysql
    ...
- helm:
    description: install wordpress
    ...
uninstall:
- helm:
    description: uninstall wordpress
    ...
```

### LINKED ISSUE #583: Build time configuration of mixins
Allow mixins to accept a blob of yaml that they can use to configure the invocation image at build time.

```yaml
mixins:
  - az:
     extensions:
       - custom-extension1
```

alternatively

```yaml
mixins:
  - az

az:
  extensions:
    - custom-extension1
```

carolynvs:
# What does this change
Passes context to mixins when we ask them to build Dockerfile lines to inject into the invocation image. The mixin always gets the usage of the mixin in the manifest. Also if the mixin declaration contains any configuration data, then that is passed along as well.

The context is sent via stdin when we call `mixin build` just like we do with the other mixin commands like `mixin install`.

```yaml
mixins:
  - exec
  - az:
      extensions:
        - iot

install:
  - az:
      description: Login
      arguments:
      - login

uninstall:
  - exec:
      description: "Uninstall stuff"
      command: "bash"
      arguments:
      - echo bye
```

The `az build` command would receive this on stdin:

```yaml
config:
  extensions:
  - iot
actions:
  install:
  - az:
      arguments:
      - login
      description: Login
  uninstall: []
  upgrade: []
```

The `exec build` command would receive this on stdin:

```yaml
actions:
  install: []
  uninstall:
  - exec:
      arguments:
      - echo bye
      command: bash
      description: Uninstall stuff
  upgrade: []
```
# What issue does it fix
Closes #262 
Closes #583

# Notes for the reviewer
N/A

# Checklist
- [x] Unit Tests
- [x] Documentation

### DIFF
--- a/docs/content/authoring-bundles.md
+++ b/docs/content/authoring-bundles.md
@@ -62,6 +62,16 @@ mixins:
 - helm
 ```
 
+Some mixins allow you to specify configuration data that is provided to the mixin during `porter build`. Each mixin
+has its own format for the configuration. For example, the az mixin allows you to specify extensions to install:
+
+```yaml
+mixins:
+- az:
+    extensions:
+    - azure-cli-iot-ext
+```
+
 See [Using Mixins](/using-mixins) to learn more about how mixins work.
 
 ## Parameters
--- a/docs/content/mixin-commands.md
+++ b/docs/content/mixin-commands.md
@@ -24,19 +24,34 @@ recommended so that your mixin has a good user experience.
 # build
 
 The build command (required) is called on the local machine during the `porter
-build` command. The mixin should return lines for the Dockerfile on stdout.
+build` command. Any mixin configuration and all usages of the mixin are passed
+on stdin. The mixin should return lines for the Dockerfile on stdout.
 
 Example:
 
+**stdin**
+
+```yaml
+config:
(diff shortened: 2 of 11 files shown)


carolynvs:
I don't think that more than 1 mixin specified in a declaration is possible in yaml. At least I'm not sure how to represent it. But I was covering my bases in case it happened somehow. But I do know how to break yaml! Let me add a case for that. 👍

carolynvs:
I was wrong! Just pushed up a test case for more than 1 mixin specified

PR #624, opened 2019-09-17
Add image map wiring

### LINKED ISSUE #501: [proposal] Introduce imageMap wiring
We have the imageMap section in the `porter.yaml` and it ends up looking something like:

```
imageMap:
  websvc:
      description: "A simple web service"
      imageType: "docker"
      image: "jeremyrickard/devops-days-msp@sha256:85b1a9b4b60a4cf73a23517dad677e64edf467107fa7d58fce9c50e6a3e4c914"
      digest: "sha256:85b1a9b4b60a4cf73a23517dad677e64edf467107fa7d58fce9c50e6a3e4c914"
```

Later on, someone might want to define a step that takes that as a param:

```
 azure:
      description: "Create an ACI Instance"
      type: arm
      template: "arm/aci.json"
      name: "{{ bundle.parameters.server-name }}-aci"
      resourceGroup: "{{ bundle.parameters.backend_storage_resource_group }}"
      parameters:
        containerName: "{{ bundle.parameters.server-name }}-aci-go"
        location: "{{ bundle.parameters.location }}"
        imageName: "jeremyrickard/devops-days-msp@sha256:85b1a9b4b60a4cf73a23517dad677e64edf467107fa7d58fce9c50e6a3e4c914"
        mysqlFQDN: "{{bundle.outputs.mysql_fqdn}}"
        port: "8080"
```

Instead of having to retype/copy paste it, it might be cool if we could reference it like `{{bundle.imageMap.websvc}}` or something along those lines. This might allow us to support the relocation stuff that duffle is doing as well. Porter could take the relocation file and change those references as needed. 

Nothing stops someone from _not_ using that mechanism when it's introduced, but this would be a nice feature to have and we could write some best practice stuff up and encourage people to follow that pattern.

jeremyrickard:
# What does this change

This PR adds image map wiring / interpolation into Porter's bag of tricks. It changes the `imageMap` part of the porter.yaml to `images` to match the issue comments and does some various cleanup. 

# What issue does it fix
Closes #501

# Notes
Since mustache template interpolation is case sensitive, I decided to use `reflect` to iterate the MappedImage struct and `strings.ToLower` the name of the field, that way it matches up against the yaml. Otherwise, we ended up with references like `{{bundle.images.thing.Repository}}`, which bugged me. 

# Checklist
- [x] Unit Tests
- [x] Documentation

### DIFF
--- a/docs/content/wiring.md
+++ b/docs/content/wiring.md
@@ -3,9 +3,9 @@ title: Using Parameters, Credentials and Outputs
 description: How to wire parameters, credentials and outputs into steps
 ---
 
-# Parameters, Credentials and Outputs in Porter
+# Parameters, Credentials, Outputs, and Images in Porter
 
-In the Porter manifest, you can declare both parameters and credentials. In addition to providing a mechanism for declaring parameters and credentials at the bundle level, Porter provides a way to declare how each of these are provided to [mixins](/mixin-architecture). This mechanism is also applicable to declaring how output from one mixin can be passed to another, as well as how to consume parameters, credentials and outputs from bundle dependencies.
+In the Porter manifest, you can declare both parameters and credentials. In addition to providing a mechanism for declaring parameters and credentials at the bundle level, Porter provides a way to declare how each of these are provided to [mixins](/mixin-architecture). This mechanism is also applicable to declaring how output from one mixin can be passed to another, as well as how to consume parameters, credentials and outputs from bundle dependencies. Finally, you can also use this technique to reference images defined in the `images` section of the manifest.
 
 ## Parameters
 
@@ -152,6 +152,32 @@ For example, given the install step above, we can use the `MYSQL_URL` with the h
 ```
 
 Just like in the case of credentials and parameters, the value of the `bundle.outputs.MYSQL_URL` reference will be rewritten in the YAML before the helm mixin is invoked.
+
+## Wiring Images
+
+In the `porter.yaml`, you can define what images will be used within the bundle with the `images` section:
+
+```yaml
+images:
+  ALIAS:
+    description: A very useful image
+    imageType: docker # porter.yaml can default this to docker if we aren't already
+    repository: gcr.io/mcguffin-co/mcguffin
+    digest: sha256:85b1a9 # this is what bundle.json allows
+    tag: v1.1.0 # we can collect this and make it available but it won't land into bundle.json
+```
+
+These images will be used to build the `bundle.json` images section, but can also be referenced using the same syntax you would use for referencing `parameters`, `credentials`, and `outputs`.
+
+```yaml
+- description: "Helm Install Wordpress"
+    helm:
+      name: porter-ci-wordpress
+      chart: stable/wordpress
+      set:
+        image.repository: "{{ bundle.parameters.ALIAS.repository }}"
+        image.tag: "{{ bundle.parameters.ALIAS.tag }}"
+```
 
 ## Using Parameters, Credentials, and Outputs from Bundle Dependencies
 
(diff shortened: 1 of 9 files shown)


docs/content/wiring.md
```diff
@@ -3,9 +3,9 @@ title: Using Parameters, Credentials and Outputs
 description: How to wire parameters, credentials and outputs into steps
 ---
 
-# Parameters, Credentials and Outputs in Porter
+# Parameters, Credentials, Outputs, and Images in Porter
```
carolynvs: Can you update this to be the title of the article instead in the front matter and remove this duplicate heading?

docs/content/wiring.md
```diff
@@ -153,6 +153,32 @@ For example, given the install step above, we can use the `MYSQL_URL` with the h
...
 
+## Wiring Images
+
+In the `porter.yaml`, you can define what images will be used within the bundle with the `images` section:
+
+```yaml
+images:
+  ALIAS:
+    description: A very useful image
+    imageType: docker # porter.yaml can default this to docker if we aren't already
```
vdice: Should the comment change to say porter is defaulting it to docker (if that's the case)?

docs/content/wiring.md
```diff
@@ -153,6 +153,32 @@ For example, given the install step above, we can use the `MYSQL_URL` with the h
...
+    repository: gcr.io/mcguffin-co/mcguffin
+    digest: sha256:85b1a9 # this is what bundle.json allows
```
vdice: Should we reference the cnab spec around the bundle.json file here?

docs/content/wiring.md
```diff
@@ -153,6 +153,32 @@ For example, given the install step above, we can use the `MYSQL_URL` with the h
...
+    tag: v1.1.0 # we can collect this and make it available but it won't land into bundle.json
```
vdice: It won't land into bundle.json b/c the digest sha is provided?  Also, is the digest key/value mandatory or optional?  (Or more generally, is it worth explicitly mentioning which fields are mandatory?)

pkg/config/manifest.go
```diff
@@ -41,10 +41,11 @@ type Manifest struct {
 	Dependencies map[string]Dependency  `yaml:"dependencies,omitempty"`
 	Outputs      []OutputDefinition     `yaml:"outputs,omitempty"`
 
-	// ImageMap is a map of images referenced in the bundle. The mappings are mounted as a file at runtime to
-	// /cnab/app/image-map.json. This data is not used by porter or any of the deislabs mixins, so only populate when you
-	// plan on manually using this data in your own scripts.
-	ImageMap map[string]MappedImage `yaml:"imageMap,omitempty"`
+	// ImageMap is a map of images referenced in the bundle. If image relocation mapping occurs, that
+	// file will be mounted at as a file at runtime to /cnab/app/relocation-mapping.json.
```
vdice: Extra `at` in `...file will be mounted at as a file...`

docs/content/wiring.md
jeremyrickard: I think this is probably better in a separate doc about images? What say you @carolynvs. We added that but never really documented it (there was a comment that we did nothing with it before). I was going to open a more general "document how to use `images`" issue that we could give best practice info in and explain how to use for Helm charts, etc.

docs/content/wiring.md
jeremyrickard: yes indeed. it will require a change. thanks for catching that! 

I'll update the headline in wiring.md too. Good suggestion.

jeremyrickard:
Updated the docs and the workshop, however, I couldn't verify the bundle since there is an 
Azure outage related to MySQL. Feel free to review, but hold off on merging until I verify the bundle.

jeremyrickard:
@carolynvs @vdice I ran this locally again and it works. LMK if it's good to merge or if you want to take another look!

PR #625, opened 2019-09-17
Add scaffolding for new porter archive command

jeremyrickard:
This adds the scaffolding for the new Porter archive command. It's currently hidden, but I wanted to open a PR that introduced it so we could figure out if these options/command experience are correct.

### DIFF
--- a/pkg/porter/archive.go
+++ b/pkg/porter/archive.go
@@ -1,0 +1,23 @@
+package porter
+
+import (
+	"github.com/deislabs/porter/pkg/context"
+)
+
+// ArchiveOptions defines the valid options for performing an archive operation
+type ArchiveOptions struct {
+	bundleFileOptions
+	File string
+}
+
+// Validate performs validation on the publish options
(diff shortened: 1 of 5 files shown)


pkg/porter/archive.go
```diff
@@ -0,0 +1,23 @@
...
+// Validate performs validation on the publish options
+func (o *ArchiveOptions) Validate(args []string, cxt *context.Context) error {
+	return nil
+}
+
+// Archive is a composite function that generates a CNAB thick bundle. It will pull the invocation image, and
+// any referenced images locally (if needed), export them to individual layers, generate a bundle.json and
+// then generate a gzipped tar archive containing the bundle.json and the images
+func (p *Porter) Archive(opts ArchiveOptions) error {
+	return nil
```
carolynvs: Would you return an error saying "not implemented yet" just so that someone playing with canary doesn't wonder?

jeremyrickard: Yep! I set the cmds themselves to `Hidden: true` but I think this is also a good idea. Happy to write up some contributing guidelines if you'd like after this.

PR #632, opened 2019-09-19
Add in delete mixin command

### LINKED ISSUE #613: Support deleting a mixin with the porter mixin delete command
Support deleting a mixin with a new command

```console
$ porter mixin delete NAME
deleted NAME
```

The following changes need to be made:
* `cmd/porter/mixins.go` Add the delete command
* `cmd/porter/mixins.go` Add a function to Porter that handles the mixin command
* `pkg/mixins//mixin.go` Add the Delete function to the MixinProvider
* `pkg/mixins/provider/NEWFILE` Add delete.go and implement the delete functionality and add tests
* Link to the new command from the [doc nav](https://github.com/deislabs/porter/blob/master/docs/config.toml#L43)

Mixins are installed into PORTER_HOME/mixins/NAME. Delete a mixin by deleting the directory named after the mixin. If #610 has been implemented, you will also need to remove the entry from the mixin cache file as well.

bradcypert:
# What does this change
This PR adds in the ability to delete a mixin by using the `porter mixin delete NAME` command, where `NAME` is a mixin name like "helm".

```
➜ porter mixin delete
Error: no mixin name was specified

➜ porter mixin delete helm
Error: Could not find helm in the mixin directory.

➜ porter mixin install helm
installed helm mixin
helm v0.8.0-beta.1 (8325d71) by DeisLabs

➜ porter mixin delete helm
Deleted helm mixin
```

# What issue does it fix
Closes #613 

# Notes for the reviewer
This is my first PR and I tried to follow as many existing patterns and practices as I could, but if something seems off, just let me know and I'll try to update it :)

# Checklist
- [x] Unit Tests
- [x] Documentation

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -376,7 +376,7 @@ dependency injection and testing strategies.
 Example:
 
 ```golang
-fmt.Fprintln(p.Out, "Initiating battlestar protocol"
+fmt.Fprintln(p.Out, "Initiating battlestar protocol")
 fmt.Fprintln(p.Err, "DEBUG: loading plans from r2d2...")
 ```
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,27 +1,26 @@
 # Contributing Guide
 
 ---
-
-- [How to help](#how-to-help)
-  - [Find an issue](#find-an-issue)
-  - [When to open a pull request](#when-to-open-a-pull-request)
-  - [How to get your pull request reviewed fast](#how-to-get-your-pull-request-reviewed-fast)
-  - [The life of a pull request](#the-life-of-a-pull-request)
-- [Contribution Ladder](#contribution-ladder)
-  - [Contributor](#contributor)
-  - [How to become a contributor](#how-to-become-a-contributor)
-  - [Maintainer](#maintainer)
-  - [How to become a maintainer](#how-to-become-a-maintainer)
-  - [Admin](#admin)
-  - [How to become an admin](#admin)
-- [Developer Tasks](#developer-tasks)
-  - [Initial setup](#initial-setup)
-  - [Makefile explained](#makefile-explained)
-  - [Install mixins](#install-mixins)
-  - [Preview documentation](#preview-documentation)
-- [Code structure and practices](#code-structure-and-practices)
-  - [What is the general code layout?](#what-is-the-general-code-layout)
-  - [Logging](#logging)
+* [How to help](#how-to-help)
+  * [Find an issue](#find-an-issue)
+  * [When to open a pull request](#when-to-open-a-pull-request)
+  * [How to get your pull request reviewed fast](#how-to-get-your-pull-request-reviewed-fast)
+  * [The life of a pull request](#the-life-of-a-pull-request)
+* [Contribution Ladder](#contribution-ladder)
+  * [Contributor](#contributor)
+  * [How to become a contributor](#how-to-become-a-contributor)
+  * [Maintainer](#maintainer)
+  * [How to become a maintainer](#how-to-become-a-maintainer)
+  * [Admin](#admin)
+  * [How to become an admin](#admin)
+* [Developer Tasks](#developer-tasks)
+  * [Initial setup](#initial-setup)
+  * [Makefile explained](#makefile-explained)
+  * [Install mixins](#install-mixins)
+  * [Preview documentation](#preview-documentation)
+* [Code structure and practices](#code-structure-and-practices)
+  * [What is the general code layout?](#what-is-the-general-code-layout)
+  * [Logging](#logging)
 
 ---
 
@@ -42,8 +41,8 @@ slack, interactions on social media, project meetings, conferences and meetups.
 We have [good first issues][good-first-issue] for new contributors and [help
 wanted][help-wanted] issues for our other contributors.
 
-- `good first issue` has extra information to help you make your first contribution.
-- `help wanted` are issues suitable for someone who isn't a core maintainer.
+* `good first issue` has extra information to help you make your first contribution.
+* `help wanted` are issues suitable for someone who isn't a core maintainer.
 
 Maintainers will do our best regularly make new issues for you to solve and then
 help out as you work on them. 💖
@@ -70,7 +69,7 @@ If there isn't an issue for your PR, please make an issue first and explain the
 problem or motivation for the change you are proposing. When the solution isn't
 straightforward, for example "Implement missing command X", then also outline
 your proposed solution. Your PR will go smoother if the solution is agreed upon
-before you've spent a lot of time implementing it.
+before you've spent a lot of time implementing it. 
 
 Since Porter is a CLI, the "solution" will usually look like this:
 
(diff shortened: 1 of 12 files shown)


bradcypert:
@carolynvs if you have a moment, can you checkout this branch and see if builds work locally for you? I haven't done anything with `dep` before, and I'm not 100% sure it's setup properly. I'm getting errors when trying to build, but it looks like CI is fine?

```
➜ make build
go generate ./...
/Library/Developer/CommandLineTools/usr/bin/make --no-print-directory build MIXIN=porter -f mixin.mk BINDIR=bin
mkdir -p bin
go build -ldflags '-w -X github.com/deislabs/porter/pkg.Version=v0.16.1-beta.1-2-gcee6ef7 -X github.com/deislabs/porter/pkg.Commit=cee6ef7' -o bin/porter ./cmd/porter
# github.com/dieslabs/porter/cmd/porter
cmd/porter/mixins.go:77:10: undefined: mixin.DeleteOptions
cmd/porter/mixins.go:86:12: p.DeleteMixin undefined (type *porter.Porter has no field or method DeleteMixin)
make[1]: *** [build-client] Error 2
make: *** [build-porter] Error 2
```

carolynvs:
Oh! Not all of the CI runs for new contributors until we have reviewed and said it's safe to run. I'll make sure to mention that in the contributing guide.

Here let me kick it off for you quick so you can see the full CI run in azure pipelines.

carolynvs:
Oh boy, I will never be good at helping people find spelling mistakes! 😂

Gopkg.lock
```diff
@@ -1694,6 +1694,7 @@
     "github.com/docker/go-metrics",
     "github.com/ghodss/yaml",
     "github.com/gobuffalo/packr/v2",
+    "github.com/gobuffalo/packr/v2/file/resolver",
```
carolynvs: We are still improving our makefile targets to help prevent this package from sneaking in here but in the meantime, it's going to cause the CI build to fail when it runs `make verify`. To avoid triggering this in the future, before running `dep ensure`, run `make clean-packr`. Sorry about that!

Once you you can reset this file, you should be fine, and I'll kick off the build for you again.

pkg/mixin/mixin.go
```diff
@@ -6,7 +6,7 @@ import (
 
 func IsCoreMixinCommand(value string) bool {
 	switch value {
-	case "install", "upgrade", "uninstall", "build", "schema", "version":
+	case "install", "upgrade", "uninstall", "build", "schema", "version", "delete":
```
carolynvs: This function `IsCoreMixinCommand` should really have been documented! 😱 It is the list of commands that all mixins implement (see https://porter.sh/mixin-commands). 

It isn't the same thing as what you are implementing here `porter mixin delete` which is allowing a user to delete a specific mixin. So please remove `delete` from the case switch.

pkg/mixin/delete.go
```diff
@@ -0,0 +1,5 @@
+package mixin
```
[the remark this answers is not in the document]
carolynvs: Looks like you got it all figured out!

pkg/mixin/delete.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+type DeleteOptions struct {
+	Name string
+}
+
+func (o *DeleteOptions) Validate(args []string) error {
+	switch len(args) {
+	case 0:
+		// later on this may be okay (porter mixin install --feed deislabs) but not now
```
carolynvs: Looks like this comment was copied over from the InstallOptions Validation function but doesn't really apply here. Would you please remove it?

pkg/mixin/provider/delete.go
```diff
@@ -0,0 +1,39 @@
...
+	mixinsDir, err := fs.GetMixinsDir()
+	if err != nil {
+		return nil, err
+	}
+	mixinDir := filepath.Join(mixinsDir, name)
+	exists, _ := fs.FileSystem.Exists(mixinDir)
+	if exists == true {
+		err = fs.FileSystem.RemoveAll(mixinDir)
+		if err != nil {
+			return nil, err
```
carolynvs: Anytime we are returning an error from a non-porter library, the custom in Porter is to wrap it using `errors.Wrapf` since the error is going to end up being displayed directly to the user. 

This way we can show the original error, and provide a bit of context about what we were trying to do that caused it. This will help us when people open bugs later and give us back the error message.

Would you please update this to use the errors package?

```suggestion
			return nil, errors.Wrapf(err, "could not remove mixin directory %q", mixinDir)
```

pkg/mixin/provider/delete.go
```diff
@@ -0,0 +1,39 @@
...
+		}
+
+		m := mixin.Metadata{
+			Name: name,
+		}
+		return &m, nil
+	}
+
+	return nil, errors.New(fmt.Sprintf("Could not find %s in the mixin directory.", name))
```
carolynvs: So in this case, where they asked to delete a mixin that isn't installed, instead of having that be an error, let's instead print a debug message that the mixin was not installed and allow the command to succeed. 

Here is the [doc](https://github.com/deislabs/porter/blob/master/CONTRIBUTING.md#logging) how to do debug logging. You can tell if we are in debug mode by checking `fs.Debug`.

pkg/mixin/provider/delete_test.go
```diff
@@ -0,0 +1,28 @@
...
+	p := NewFileSystem(c.Config)
+
+	mixinsDir, _ := p.GetMixinsDir()
+	mixinDir := path.Join(mixinsDir, "mixxin")
+
+	opts := mixin.DeleteOptions{
+		Name: "mixxin",
+	}
+
+	p.Delete(opts)
```
carolynvs: nit: The test can also check that Delete didn't return an error

pkg/porter/mixins.go
```diff
@@ -82,6 +82,21 @@ func (p *Porter) InstallMixin(opts mixin.InstallOptions) error {
...
 }
 
+func (p *Porter) DeleteMixin(opts mixin.DeleteOptions) error {
+	m, err := p.Mixins.Delete(opts)
+	if err != nil {
+		return err
+	}
+
+	if p.Debug {
+		fmt.Fprintf(p.Out, "Deleted %s mixin from %s", m.Name, m.Dir)
```
carolynvs: Looks like `m.Dir` is always empty because it's not being set in delete.go

bradcypert:
@carolynvs I made some changes around the areas that you mentioned above. If anything else looks off, please don't hesitate to lemme know. I'm just thrilled to have the opportunity to learn from a large go project :)

CONTRIBUTING.md
```diff
@@ -1,26 +1,27 @@
...
-* [Developer Tasks](#developer-tasks)
-  * [Initial setup](#initial-setup)
-  * [Makefile explained](#makefile-explained)
-  * [Install mixins](#install-mixins)
-  * [Preview documentation](#preview-documentation)
-* [Code structure and practices](#code-structure-and-practices)
-  * [What is the general code layout?](#what-is-the-general-code-layout)
-  * [Logging](#logging)
+
+- [How to help](#how-to-help)
```
carolynvs: It helps to not include unrelated changes in a pull request. Fixing a single bad link is fine to sneak in though, and it sounds like from your commit message that was the intent. Let's revert the changes to this file other than the missing `)` and this should be ready to merge. 🚀

carolynvs:
I'm going to fix our CI so that the bulk of it runs automatically for external contributors. Thanks for you patience with us having to run them manually for the next few days! 😊

CONTRIBUTING.md
bradcypert: Sorry about that! I wanted to fix the missing `)` and I think my markdown plugin ran a format-on-save. I should have checked the diff before committing it! I'll get this fixed.

PR #639, opened 2019-09-23
Update cnab-to-oci to v0.2.0-beta1

radu-matei:
This PR updates [`cnab-to-oci` to v0.2.0-beta1](https://github.com/docker/cnab-to-oci/releases/tag/v0.2.0-beta1), which stores the entire bundle in the OCI registry.

The behaviour of `porter publish` is not changed.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -327,15 +327,16 @@
   revision = "f95ca8e1ba6c22c9abcdbf65e8dcc39c53958bba"
 
 [[projects]]
-  branch = "master"
-  digest = "1:a9d1406d7380d50675fddc88442cecdbfbdd069bc24191171ae901719cd4c79b"
+  digest = "1:51fef110ff737b08bebfa091471714db028ebcbeda6a3cad8ca4512b39b34581"
   name = "github.com/docker/cnab-to-oci"
   packages = [
     "converter",
+    "relocation",
     "remotes",
   ]
   pruneopts = "NUT"
-  revision = "85061934a54630aec0dd4adbe6427c10bd09a658"
+  revision = "c33e316902ce0e44adc6e14731c7241286d17d8e"
+  version = "v0.2.0-beta1"
 
 [[projects]]
   digest = "1:95f17095fc25ae7c9de6adbe34c048ea6102b0a3999c6fcf03fc663ab536c602"
--- a/pkg/cnab/cnab-to-oci/registry.go
+++ b/pkg/cnab/cnab-to-oci/registry.go
@@ -43,7 +43,7 @@ func (r *Registry) PullBundle(tag string, insecureRegistry bool) (*bundle.Bundle
 		insecureRegistries = append(insecureRegistries, reg)
 	}
 
-	bun, err := remotes.Pull(context.Background(), ref, r.createResolver(insecureRegistries))
+	bun, _, err := remotes.Pull(context.Background(), ref, r.createResolver(insecureRegistries))
 	if err != nil {
 		return nil, errors.Wrap(err, "unable to pull remote bundle")
 	}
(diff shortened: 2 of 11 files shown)


pkg/cnab/cnab-to-oci/registry.go
```diff
@@ -43,7 +43,7 @@ func (r *Registry) PullBundle(tag string, insecureRegistry bool) (*bundle.Bundle
 		insecureRegistries = append(insecureRegistries, reg)
 	}
 
-	bun, err := remotes.Pull(context.Background(), ref, r.createResolver(insecureRegistries))
+	bun, _, err := remotes.Pull(context.Background(), ref, r.createResolver(insecureRegistries))
```
radu-matei: A relocation map is generated by `Pull` - right now we are not using it (so the behaviour of `porter pull` is identical).

pkg/cnab/cnab-to-oci/registry.go
```diff
@@ -63,11 +63,11 @@ func (r *Registry) PushBundle(bun *bundle.Bundle, tag string, insecureRegistry b
 
 	resolver := r.createResolver(insecureRegistries)
 
-	err = remotes.FixupBundle(context.Background(), bun, ref, resolver, remotes.WithEventCallback(r.displayEvent))
+	rm, err := remotes.FixupBundle(context.Background(), bun, ref, resolver, remotes.WithEventCallback(r.displayEvent), remotes.WithAutoBundleUpdate())
```
radu-matei: `cnab-to-oci` checks if the digest of any image from the bundle is different after pushed in the registry, and by default does _not_ mutate the bundle with the new image digests.

`remotes.WithAUtoBundleUpdate()` ensures the new digest is added to the bundle and pushed to the registry.

The main question here is whether there should be a flag to enforce the digest check and error out if any digest does not match.

carolynvs: When @jeremyrickard and I have been discussing what happens during publish is that what is built and pushed is an immutable bundle by default (as best we can, scripts are gonna do what they do):

* The images referenced use digests, not tags
* Mixins have baked in their assets, e.g. the helm chart referenced is in the invocation image

https://github.com/deislabs/porter/issues/202

Given that, I vote that Porter always enforces digests. 

We can always make things more configurable later.  I don't want to encourage people to use a flag like that and  it will be confusing for most users. They'd have to dig into the docs to figure out what it even did, why they'd want to use it, etc. It's something that would be more suitable for duffle than Porter IMO.

Gopkg.lock
```diff
@@ -1694,6 +1695,7 @@
     "github.com/docker/go-metrics",
     "github.com/ghodss/yaml",
     "github.com/gobuffalo/packr/v2",
+    "github.com/gobuffalo/packr/v2/file/resolver",
```
carolynvs: The build is going to yell at you for this line. If you run `make clean-packr`, then run `dep ensure` again, the build will be happy. 

If you run `make verify` locally, you will find errors like this before the CI does. The reason why is that we don't check in packr generated files that don't need to be there, but dep picks them up if you have them present. To avoid this in the future run `make clean-packr` before running dep.

pkg/cnab/cnab-to-oci/registry.go
jeremyrickard: *yes gif*

carolynvs: @radu-matei Do we have the right plumbing in place to add that check about enforcing digests in this PR? I would prefer to not merge the "fixup" behavior, and instead merge the PR with the "check and return error" behavior instead.

vdice:
Checking in on this PR.  Hoping we can get this in soon!

jeremyrickard:
We'll handle @carolynvs's comment in a follow up issue: #709 709

jeremyrickard:
Drain the ocean!

PR #642, opened 2019-09-24
Extend version command to help troubleshoot porter

### LINKED ISSUE #614: Extend version command to help troubleshoot porter
I want a command to help me troubleshoot when someone submits a bug report. Instead of having to asking questions about "what os are you running", "what mixins are installed?" etc, I'd rather tell them "run this one command and send me the output".

I think we can add a `--system` flag that when present prints additional information about what is installed in the porter home directory.

```console
$ porter version --system
porter v0.14.1-beta.2 (72dd5df)

System
--------
os: darwin
arch: amd64

Mixins
------
Name         Version          Author
aws          v0.1.2-beta.1    DeisLabs
az           v0.2.1-beta.1    DeisLabs
```

The following will need to be changed:
* `cmd/porter/version.go` - add system flag
* `pkg/porter/version.go` - 
  * Add VersionOpts struct to hold what we use currently plus the system bool and have version take this new opts struct instead.
  * Add method to collect additional information when system flag is present (os, arch)
  * Use [porter.ListMixins](https://github.com/deislabs/porter/blob/b24cdd412147c8156942f6d8accb489e71e48e6f/pkg/porter/mixins.go#L42) to get a list of the installed mixins, and use similar table printing as [porter.PrintMixins](https://github.com/deislabs/porter/blob/b24cdd412147c8156942f6d8accb489e71e48e6f/pkg/porter/mixins.go#L26-L30)
  * Make a new struct to contain all the collected information and handle printing it according to the requested output format (json, table, etc).
  * Add tests to version_tests.go

phillipahereza:
# What does this change
Adds `--system` flag to the `version` command to print additional information about what the system on which porter is being run and any installed mixins

```
$ porter version --system
porter v0.14.1-beta.2 (72dd5df)

System
--------
os: darwin
arch: amd64

Mixins
------
Name         Version          Author
aws          v0.1.2-beta.1    DeisLabs
az           v0.2.1-beta.1    DeisLabs
```

# What issue does it fix
Closes #614

# Notes for the reviewer
This is a draft PR to get help

# Checklist
- [x] Unit Tests
- [ ] Documentation

### DIFF
--- a/cmd/porter/version.go
+++ b/cmd/porter/version.go
@@ -7,15 +7,18 @@ import (
 )
 
 func buildVersionCommand(p *porter.Porter) *cobra.Command {
-	opts := version.Options{}
+	opts := version.VersionOpts{}
 	cmd := &cobra.Command{
 		Use:   "version",
 		Short: "Print the application version",
 		PreRunE: func(cmd *cobra.Command, args []string) error {
-			return opts.Validate()
+			return opts.Options.Validate()
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return p.PrintVersion(opts)
+			if opts.System {
+				return p.PrintDebugInfo(opts.Options)
+			}
+			return p.PrintVersion(opts.Options)
 		},
 	}
 	cmd.Annotations = map[string]string{
@@ -25,6 +28,7 @@ func buildVersionCommand(p *porter.Porter) *cobra.Command {
 	f := cmd.Flags()
 	f.StringVarP(&opts.RawFormat, "output", "o", string(version.DefaultVersionFormat),
 		"Specify an output format.  Allowed values: json, plaintext")
+	f.BoolVarP(&opts.System, "system", "s", false, "Print system debug information")
 
 	return cmd
 }
--- a/cmd/porter/version.go
+++ b/cmd/porter/version.go
@@ -7,18 +7,15 @@ import (
 )
 
 func buildVersionCommand(p *porter.Porter) *cobra.Command {
-	opts := version.VersionOpts{}
+	opts := porter.VersionOpts{}
 	cmd := &cobra.Command{
 		Use:   "version",
 		Short: "Print the application version",
 		PreRunE: func(cmd *cobra.Command, args []string) error {
-			return opts.Options.Validate()
+			return opts.Validate()
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
-			if opts.System {
-				return p.PrintDebugInfo(opts.Options)
-			}
-			return p.PrintVersion(opts.Options)
+			return p.PrintVersion(opts)
 		},
 	}
 	cmd.Annotations = map[string]string{
--- a/pkg/porter/version.go
+++ b/pkg/porter/version.go
@@ -73,11 +73,15 @@ func getSystemInfo() *SystemInfo {
 
 func (p *Porter) PrintDebugInfo(ctx *context.Context, opts VersionOpts, versionMetadata mixin.Metadata) error {
 	opts.RawFormat = string(printer.FormatPlaintext)
-	mixins, err := p.ListMixins()
 	sysInfo := getSystemInfo()
-	if p.Debug {
-		fmt.Fprint(p.Err, err.Error())
+	mixins, err := p.ListMixins()
+	if err != nil {
+		if p.Debug {
+			fmt.Fprint(p.Err, err.Error())
+		}
+		return nil
 	}
+
 	sysDebugInfo := SystemDebugInfo{
 		Version: versionMetadata,
 		SysInfo: *sysInfo,
--- a/pkg/porter/version.go
+++ b/pkg/porter/version.go
@@ -8,6 +8,7 @@ import (
 	"github.com/deislabs/porter/pkg/mixin"
 	"github.com/deislabs/porter/pkg/porter/version"
 	"github.com/deislabs/porter/pkg/printer"
+	"github.com/pkg/errors"
 	"runtime"
 	"text/template"
 )
@@ -100,7 +101,7 @@ Mixins
 `
 		tmpl, err := template.New("systemDebugInfo").Parse(plaintextTmpl)
 		if err != nil {
-			return err
+			return errors.Wrap(err, "Failed to parse plaintext template")
 		}
 		err = tmpl.Execute(ctx.Out, sysDebugInfo)
 		return err
--- a/pkg/porter/version.go
+++ b/pkg/porter/version.go
@@ -1,71 +1,38 @@
 package porter
 
 import (
+	"bytes"
 	"fmt"
 	"github.com/deislabs/porter/pkg"
+	"github.com/deislabs/porter/pkg/context"
 	"github.com/deislabs/porter/pkg/mixin"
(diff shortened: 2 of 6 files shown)


phillipahereza:
@carolynvs am I on the right path

carolynvs:
@phillipahereza Yes, this is what I was looking for. 👍 Once you are ready for a review, just take the PR out of draft, by clicking the button "Ready for Review" or maybe it's called "Request Reviews". 

Then I'll go over it and provide detailed feedback. I just redid our CI pipeline, so I'll kick off a new test run for you right now.

pkg/porter/version.go
```diff
@@ -16,3 +26,55 @@ func (p *Porter) PrintVersion(opts version.Options) error {
...
+
+func (p *Porter) PrintDebugInfo(opts version.Options) error {
+	// force opts to print version as plaintext
+	opts.RawFormat = string(printer.FormatPlaintext)
+	err := p.PrintVersion(opts)
+	if err != nil {
+		return errors.Wrap(err, "Failed to print version")
+	}
+
+	printSectionHeader(p.Context.Out, "System")
```
dev-drprasad: You can achieve this functionality using go templates. I think it will more readable. Just a opinion

**example:**
```golang
const ts = `
version: {{.Version}}
{{if .Sys -}}

System
--------
os: {{.Sys.OS}}
arch: {{.Sys.Arch}}
{{- end}}
`
v := Version{
	Version: "1.0.0",
	Sys:     nil,
}
tmpl, _ := template.New("system").Parse(ts)
tmpl.Execute(p.Context.Out, v)
```

cmd/porter/version.go
```diff
@@ -7,15 +7,18 @@ import (
 )
 
 func buildVersionCommand(p *porter.Porter) *cobra.Command {
-	opts := version.Options{}
+	opts := version.VersionOpts{}
```
carolynvs: Would you please move `VersionOpts` into `pkg/porter/version.go`. The `version` package is intended for other mixins to reuse when they implement the version command. 

That will change this line to look like to `opts := porter.VersionOpts{}` but won't otherwise change much. It's just to help with the organization of the code.

cmd/porter/version.go
```diff
@@ -7,15 +7,18 @@ import (
...
 	cmd := &cobra.Command{
 		Use:   "version",
 		Short: "Print the application version",
 		PreRunE: func(cmd *cobra.Command, args []string) error {
-			return opts.Validate()
+			return opts.Options.Validate()
```
carolynvs: Please simplify the here back to: 

```suggestion
			return opts.Validate()
```

We can do that because `Options` is embedded in `VersionOpts` (it isn't a named field) so functions available on that type can be called without having to specify that it's coming from `Options`. Apologies if you knew about that, I'm just sharing Go tips in case they are helpful! 😀

cmd/porter/version.go
```diff
@@ -7,15 +7,18 @@ import (
...
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return p.PrintVersion(opts)
+			if opts.System {
```
carolynvs: The convention in Porter is that all logic should be in `pkg/porter`, and `cmd/porter` should just wire up cobra to functions defined in that package. 

Please move the check against `opts.System` into `PrintVersion`, so that the code in this file goes back to looking like:

```go
   return p.PrintVersion(opts)
}
```

If you are curious about why, I did a [talk about how Porter is designed at GopherCon](https://youtu.be/eMz0vni6PAw). The short answer is that it helps us test, and allows for people to build on top of porter's libraries.

pkg/porter/version.go
```diff
@@ -16,3 +26,55 @@ func (p *Porter) PrintVersion(opts version.Options) error {
...
+}
+
+func printSectionHeader(out io.Writer, header string){
+	underline := "-------"
+	_, _ = fmt.Fprintf(out, "\n%s\n%s\n", header, underline)
+
+}
+
+func (p *Porter) PrintDebugInfo(opts version.Options) error {
+	// force opts to print version as plaintext
```
carolynvs: Apologies for not catching this when I first checked the PR, but this new `--system` flag needs to work in conjunction with the `--output plaintext|json` flag. So `PrintDebugInfo` needs to support printing to json as well.

pkg/porter/version_test.go
```diff
@@ -51,3 +54,30 @@ func TestPrintJsonVersion(t *testing.T) {
...
+	pkg.Version = "v1.2.3"
+
+
+	p := NewTestPorter(t)
+
+	opts := version.Options{}
+	p.TestConfig.SetupPorterHome()
+	err := opts.Validate()
+	require.Nil(t, err)
+	err = p.PrintDebugInfo(opts)
```
carolynvs: After you move this method and the check for `opts.System` into `p.PrintVersion`, I recommend updating this unit test, to exercise your if statement as well. So it should create `VersionOptions {System: true}` and pass that to `p.PrintVersion(opts)`.

cmd/porter/version.go
phillipahereza: Thanks, this is all very helpful 🙏🏾

pkg/porter/version.go
```diff
@@ -1,80 +1,111 @@
...
-func (p *Porter) PrintDebugInfo(opts version.Options) error {
-	// force opts to print version as plaintext
+func (p *Porter) PrintDebugInfo(ctx *context.Context, opts VersionOpts, versionMetadata mixin.Metadata) error {
 	opts.RawFormat = string(printer.FormatPlaintext)
-	err := p.PrintVersion(opts)
+	mixins, err := p.ListMixins()
+	sysInfo := getSystemInfo()
 	if err != nil {
-		return errors.Wrap(err, "Failed to print version")
+		_ = errors.Wrap(err, "Failed to get list of mixins")
```
carolynvs: Instead of ignoring the error (like it's doing now), or returning it (which would make it very hard to use this command for troubleshooting problems 😅 ), we can have it print a debug message and then keep going:

The contributing guide has a section on how to handle debug logging:

https://github.com/deislabs/porter/blob/master/CONTRIBUTING.md#logging

You can check `p.Debug` and then print to `p.Err` following the example in the guide, then have the rest of the command continue. That way even if we can't print any mixin data, at least we are able to print the data we _do_ have.

pkg/porter/version.go
```diff
@@ -1,80 +1,111 @@
...
+os: {{.SysInfo.OS}}
+arch: {{.SysInfo.Arch}}
+{{if .Mixins}}
+Mixins
+-------
+{{.Mixins.PrintMixinsTable}}{{end}}
+`
+		tmpl, err := template.New("systemDebugInfo").Parse(plaintextTmpl)
+		if err != nil {
+			_ = errors.Wrap(err, "Failed to print system debug information")
```
carolynvs: At this point, if we can't parse the template, then we should return the error, and not ignore it with `_` because it's an unrecoverable error. 

If the template isn't parsed, then continuing on to call Execute on the template (which is nil when err is populated) will cause a nil reference exception.

phillipahereza:
> Thanks for adding this into Porter! It's going to really help out when people submit bug reports. 💯

My pleasure, thanks for guiding me through the whole process

PR #646, opened 2019-09-24
Add a porter explain command

### LINKED ISSUE #635: [enhancement] Let's add a `porter explain` or `porter inspect` command
**Background:**

Someone building a Porter bundle can `publish` it to an OCI registry and then other people can `install`  or `credential generate` from that tag. One thing that we don't really have a good story for is how do you discover the parameters and credentials that are required and how do you discover what outputs are produced  or what custom actions are available?  You could try and `install` the bundle over and over until you get it right or perhaps find documentation in a Github repo or README that someone shared with you out of band, but wouldn't it be better if we could do all of that from Porter?

**What this issue proposes:**

What if we add a new command to porter that is purpose built to help users learn about a bundle? This command could print out information about the bundle, such as parameter, output and credential definitions, as well as list the custom actions available, and we could support this from a tagged reference, just like `install` and `credentials generate`. 

I'm not 100% sold on which `subcommand` we want to surface (the two in my head right now are `explain` and `inspect`), but it might look something like this:

```
$ porter bundle explain --tag deislabs/tron:v1.0.2

name: Tron
description: "The classic game of light cycles and disc wars, deployed on your Nomad cluster" 
version: 1.2.2

==Parameters==
| Name             |   Type           |    Description   |       Default (Required)    | . Env or File  
<stuff>

==Credentials==
| Name |  Type | Description | Env or File
<stuff>

==Outputs== 
| Name             |   Type           |    Description   |       Default    | . Env or File  
<stuff>

==Actions==
| Name | Description | Stateless | Mutating | 
<stuff>
```

I'm not sure if we should print out all the JSON schema stuff for the params and outputs. Maybe we further have commands that drill into params and give you more info? 

```
$ porter bundle explain parameter --name name --tag <asdasdadadad>
```

We could also print out things in the custom section as we get to supporting other well known things. For example, we could determine that the bundle.json specifies dependencies and print those out as well. 

What do you think @carolynvs? This is partially inspired by `kubectl explain`:

```
kubectl explain pod
KIND:     Pod
VERSION:  v1

DESCRIPTION:
     Pod is a collection of containers that can run on a host. This resource is
     created by clients and scheduled onto hosts.

FIELDS:
   apiVersion	<string>
     APIVersion defines the versioned schema of this representation of an
     object. Servers should convert recognized schemas to the latest internal
     value, and may reject unrecognized values. More info:
     https://git.k8s.io/community/contributors/devel/api-conventions.md#resources

   kind	<string>
     Kind is a string value representing the REST resource this object
     represents. Servers may infer this from the endpoint the client submits
     requests to. Cannot be updated. In CamelCase. More info:
     https://git.k8s.io/community/contributors/devel/api-conventions.md#types-kinds

   metadata	<Object>
     Standard object's metadata. More info:
     https://git.k8s.io/community/contributors/devel/api-conventions.md#metadata

   spec	<Object>
     Specification of the desired behavior of the pod. More info:
     https://git.k8s.io/community/contributors/devel/api-conventions.md#spec-and-status

   status	<Object>
     Most recently observed status of the pod. This data may not be up to date.
     Populated by the system. Read-only. More info:
     https://git.k8s.io/community/contributors/devel/api-conventions.md#spec-and-status
```

I think this could really improve the experience working with remote bundles.

jeremyrickard:
Closes: 

# What does this change
This PR adds a new `explain` command that prints usage information for a bundle! Long gone are the days of trying to figure out how to use the bundle by running install until it works!

```
$ porter explain --tag jeremyrickard/porter-do-bundle:v0.4.1
Name: spring-music
Description: Run the Spring Music Service on Kubernetes and Digital Ocean PostgreSQL
Version: 0.4.1

Credentials:
+------------------+--------------------------------+----------+
|       NAME       |          DESCRIPTION           | REQUIRED |
+------------------+--------------------------------+----------+
| kubeconfig       | Kube config file with          | false    |
|                  | permissions to deploy app      |          |
| do_access_token  | Access Token for Digital Ocean | false    |
|                  | Account                        |          |
| do_spaces_key    | DO Spaces API Key              | false    |
| do_spaces_secret | DO Spaces API Secret           | false    |
+------------------+--------------------------------+----------+

Parameters:
+---------------+--------------------------------+---------+-------------------+----------+-------------+
|     NAME      |          DESCRIPTION           |  TYPE   |      DEFAULT      | REQUIRED | APPLIES TO  |
+---------------+--------------------------------+---------+-------------------+----------+-------------+
| node_count    | Number of database nodes       | integer | 1                 | false    | All Actions |
| porter-debug  | Print debug information from   | boolean | false             | false    | All Actions |
|               | Porter when executing the      |         |                   |          |             |
|               | bundle                         |         |                   |          |             |
| region        | Region to create Database and  | string  | nyc3              | false    | All Actions |
|               | DO Space                       |         |                   |          |             |
| space_name    | Name for DO Space              | string  | jrrportertest     | false    | All Actions |
| database_name | Name of database to create     | string  | jrrportertest     | false    | All Actions |
| helm_release  | Helm release name              | string  | spring-music-helm | false    | All Actions |
| namespace     | Namespace to install Spring    | string  | default           | false    | All Actions |
|               | Music app                      |         |                   |          |             |
+---------------+--------------------------------+---------+-------------------+----------+-------------+

Outputs:
+------------+--------------------------------+--------+-------------+
|    NAME    |          DESCRIPTION           |  TYPE  | APPLIES TO  |
+------------+--------------------------------+--------+-------------+
| service_ip | IP Address assigned to the     | string | All Actions |
|            | Load Balancer                  |        |             |
+------------+--------------------------------+--------+-------------+

No custom actions defined
```

This also provides support for `json` and `yaml` output. 

An implementation note: I defined "Printable" types for this that only expose _some_ of the fields of each type AND remove the omitempty on the json/yaml tags. I felt it was important to show values, even when they'd be omitted because they were "empty" (i.e. false...). 

# What issue does it fix
Closes #635

# Notes for the reviewer

* I added a new `explain.go` file, because bundle is getting rather large and I think I'd eventually like to explode this command more and support like an "explain" command on parameters or outputs themselves, like print the JSON schema in a useful format so you could leverage that info when crafting parameters. 

# Checklist
- [x] Unit Tests
- [x] Documentation


pkg/porter/explain.go
```diff
@@ -0,0 +1,291 @@
...
+	err := p.prepullBundleByTag(&o.BundleLifecycleOpts)
+	if err != nil {
+		return errors.Wrap(err, "unable to pull bundle before invoking credentials generate")
+	}
+
+	err = p.applyDefaultOptions(&o.sharedOptions)
+	if err != nil {
+		return err
+	}
+	err = p.ensureLocalBundleIsUpToDate(o.bundleFileOptions)
```
vdice: I'm thinking this call is not necessary, right?  (Would re-build if bundle differed/out-of-date with local manifest.)

pkg/porter/explain.go
```diff
@@ -0,0 +1,291 @@
...
+		pc := PrintableCredential{}
+		pc.Description = v.Description
+		pc.Required = v.Required
+
+		creds[c] = pc
+	}
+
+	params := map[string]PrintableParameter{}
+	for p, v := range bun.Parameters {
+		def, ok := bun.Definitions[v.Definition]
```
vdice: Idea that this reminded me of whilst writin' similar code in Porter: We should add a helper func in cnab-go, `GetDefinition(paramOrOutput)` or some such (may need a function for each type), that'd return a parameter's/output's definition or return the corresponding error (not found/empty/etc.)

pkg/porter/explain.go
vdice: Actually, even though perhaps a main use case would be to explain a remote bundle, I can also see a use case for explaining a local bundle... in that scenario, we might as well would want to ensure it'd be up-to-date.  Therefore, seems we'd want to keep this!

jeremyrickard: Yeah, we for sure want to be able to use this on canonical JSON bundle files that are supper hard to read or a porter.yaml that people might have shared with others.

pkg/porter/explain.go
jeremyrickard: That’s a good idea. Do you want to file an issue for it?

vdice: Yuuuup https://github.com/deislabs/cnab-go/issues/134

jeremyrickard: Thanks @vdice!

cmd/porter/explain.go
```diff
@@ -0,0 +1,38 @@
...
+	opts := porter.ExplainOpts{}
+	cmd := cobra.Command{
+		Use:   "explain",
+		Short: "Explain a bundle",
+		Long:  "Explain how to use a bundle by printing the parameters, credentials, outputs, actions and images.",
+		Example: `  porter bundle explain
+  porter bundle explain --file another/porter.yaml
+  porter bundle explain --cnab-file some/bundle.json
+  porter bundle explain --tag deislabs/porter-bundle:v0.1.0
+  porter bundle explain --cnab-file some/bundle.json -o json
```
carolynvs: nit: I don't think you need the last two examples as we don't provide examples for the --output flags. They are consistent across all the commands.

pkg/porter/explain.go
```diff
@@ -0,0 +1,291 @@
...
+)
+
+type ExplainOpts struct {
+	BundleLifecycleOpts
+	printer.PrintOptions
+}
+
+// PrintableBundle holds a subset of pertinent values to be explained from a bundle.Bundle
+type PrintableBundle struct {
+	Name        string                         `json:"name" yaml:"name"`
```
carolynvs: I was looking more closely at https://github.com/ghodss/yaml as a replacement for the standard yaml encoder and one nice perk is that you don't need to do double tag duty. It drives itself off of the json tags and marshalers which is pretty neat. Maybe something we could think about doing later.

pkg/porter/explain.go
```diff
@@ -0,0 +1,291 @@
...
+		return err
+	}
+	err = p.ensureLocalBundleIsUpToDate(o.bundleFileOptions)
+	if err != nil {
+		return err
+	}
+	bundle, err := p.CNAB.LoadBundle(o.CNABFile, o.Insecure)
+	// Print Bundle Details
+
+	pb, err := generatePrintable(bundle)
```
carolynvs: nit: I suggest splitting this function up into two. One that handles everything up to this point: pulling, handling the options, generating the dataset. Then another that handles printing it out. We have been using that pattern elsewhere, for `porter mixins list` is split into print mixins and list mixins:

https://github.com/deislabs/porter/blob/7da5cffdf3581c27dfab03af88df2df005baa02d/pkg/porter/mixins.go#L16

https://github.com/deislabs/porter/blob/7da5cffdf3581c27dfab03af88df2df005baa02d/pkg/porter/mixins.go#L42

I realize we aren't perfectly consistent. There are examples in the codebase of doing this differently, but I'm picking the one I think works best. This helps with testing (we aren't testing printing, just the data collected), and also ensures that we are making porter reusable for others to build upon.

pkg/porter/explain.go
```diff
@@ -0,0 +1,291 @@
...
+
+	pb.Actions = actions
+	pb.Credentials = creds
+	pb.Outputs = outputs
+	pb.Parameters = params
+
+	return &pb, nil
+}
+
+func (p *Porter) printTable(bun *PrintableBundle) error {
```
carolynvs: nit: Since this is a function off of Porter, it may help to name it a bit more specifically. 

```suggestion
func (p *Porter) printBundleTable(bun *PrintableBundle) error {
```

pkg/porter/explain.go
```diff
@@ -0,0 +1,291 @@
...
+	p.generateParametersBlock(bun)
+
+	p.generateOutputsBlock(bun)
+
+	p.generateActionsBlock(bun)
+
+	return nil
+}
+
+func (p *Porter) generateCredentialsBlock(bun *PrintableBundle) {
```
carolynvs: nit: These functions below should all be named print* since they do hit stdout and don't return anything

```suggestion
func (p *Porter) printCredentialsBlock(bun *PrintableBundle) {
```

pkg/porter/explain.go
```diff
@@ -0,0 +1,291 @@
...
+func (p *Porter) generateActionsTable(bun *PrintableBundle) {
+	h := []string{"Name", "Description", "Modifies Instance", "Stateless"}
+	d := [][]string{}
+	for k, v := range bun.Actions {
+		d = append(d, []string{k, v.Description, fmt.Sprintf("%v", v.Modifies), fmt.Sprintf("%v", v.Stateless)})
+	}
+	p.generateTable(h, d)
+}
+
+func (p *Porter) generateTable(h []string, d [][]string) {
```
carolynvs: I'd prefer that we use the same table look, and function for printing tabular data throughout Porter. It'll help a ton as people will be copy/pasting this code when they tackle new issues.

Is there a reason why you didn't use pkg/printer.PrintTable? Each function `generate*` could have called it and then we didn't need a new table printing function. 

e.g.

https://github.com/deislabs/porter/blob/7da5cffdf3581c27dfab03af88df2df005baa02d/pkg/porter/mixins.go#L23-L32

pkg/porter/testdata/explain/expected-table-output.txt
```diff
@@ -0,0 +1,20 @@
+Name: HELLO
```
carolynvs: 🤩

pkg/porter/explain.go
carolynvs: Sorry for the drive-by on your PR. I just thought you'd be interested in this after seeing the double tags. I've turned this into an issue for us to think about. https://github.com/deislabs/porter/issues/650

jeremyrickard: That sounds good!

pkg/porter/explain.go
jeremyrickard: Oops, didn't see that. I can go refactor.

pkg/porter/testdata/explain/expected-table-output.txt
jeremyrickard: Is there supposed to be an emoji here? I see a square :/

jeremyrickard:
Updated this to use our pkg/printer, new table format is thus:

```
$ porter explain --tag jeremyrickard/porter-do-bundle:v0.4.1
Name: spring-music
Description: Run the Spring Music Service on Kubernetes and Digital Ocean PostgreSQL
Version: 0.4.1

Credentials:
Name               Description                                       Required
do_access_token    Access Token for Digital Ocean Account            false
do_spaces_key      DO Spaces API Key                                 false
do_spaces_secret   DO Spaces API Secret                              false
kubeconfig         Kube config file with permissions to deploy app   false

Parameters:
Name            Description                                                     Type      Default             Required   Applies To
database_name   Name of database to create                                      string    jrrportertest       false      All Actions
helm_release    Helm release name                                               string    spring-music-helm   false      All Actions
namespace       Namespace to install Spring Music app                           string    default             false      All Actions
node_count      Number of database nodes                                        integer   1                   false      All Actions
porter-debug    Print debug information from Porter when executing the bundle   boolean   false               false      All Actions
region          Region to create Database and DO Space                          string    nyc3                false      All Actions
space_name      Name for DO Space                                               string    jrrportertest       false      All Actions

Outputs:
Name         Description                                Type     Applies To
service_ip   IP Address assigned to the Load Balancer   string   All Actions

No custom actions defined
```

pkg/porter/testdata/explain/expected-table-output.txt
carolynvs: Screen Shot 2019-09-26 at 2 13 01 PM

GITHUB Y U NO HAVE STARSTUCK EMOJI!

docs/config.toml
```diff
@@ -221,7 +221,11 @@ defaultContentLanguage = "en"
     identifier = "wiring"
     weight = 303
     parent = "authoring-bundles"
-
+[[menu.main]]
```
carolynvs: oh my gosh! I hope you didn't think that I was trying to say that you should write net new docs! I mean that's great that you did, I was just saying that you should link to the autogenerated docs for `porter explain` from the nav. Sorry about that if I was unclear. No worries, we can add the link in a follow-up.

Screen Shot 2019-09-26 at 2 17 14 PM

examples/kubernetes-mixin-example/porter.yaml
```diff
@@ -13,6 +13,11 @@ invocationImage: deislabs/porter-kubernetes:latest
 tag: deislabs/porter-kube-bundle:v0.1.0
 
 install:
+  - exec:
```
carolynvs: Did you intend to add this permanently to the example? Or was this accidentally included from our debugging session last night?

jeremyrickard: embarassed emoji.....

jeremyrickard: let me go clean that up :)

docs/config.toml
jeremyrickard: oh i thought i added that one too :) sorrrry.

PR #666, opened 2019-09-28
updated mixin install process to save info

### LINKED ISSUE #610: Save the source when a mixin is installed
When we install a mixin, we need to remember where it came from so that we can [upgrade](https://github.com/deislabs/porter/issues/30) it later without having to specify again where to get the new binaries.

Porter should save this to a new file located at **PORTER_HOME/mixins/cache.json**.

```json
{
  "mixins": [
    {
      "name": "azure",
      "feedURL": "https://cdn.deislabs.io/porter/atom.xml"
    },
    {
      "name": "mymixin",
      "url": "https://github.com/org/proj/releases/downloads"
    }
  ]
}
```

`feedURL` and `url` correspond to the original flags passed to `porter mixin install`, which can be either `--feedURL` or `--url`.

I suggest looking at implementing this functionality in this area:

https://github.com/deislabs/porter/blob/b24cdd412147c8156942f6d8accb489e71e48e6f/pkg/mixin/provider/install.go#L17-L23

It has a reference to a `Config` struct which has useful methods like `getMixinsDir()` which you will need in order to figure out where to write the file.

You can test this out by building and installing porter, then running `porter mixin install exec`. Then check the file in ~/.porter/mixins/cache.json.

abhirockzz:
# What does this change
This PR adds the ability to save mixin source when it is installed. It saves this info to a file located at `$PORTER_HOME/mixins/cache.json`

# What issue does it fix
Closes #610 

# Notes for the reviewer

First PR  :-)

# Checklist
- [ ] Unit Tests
- [x] Documentation

### DIFF
--- a/pkg/mixin/provider/install.go
+++ b/pkg/mixin/provider/install.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"net/url"
 	"path"
@@ -27,16 +26,19 @@ func (fs *FileSystem) Install(opts mixin.InstallOptions) (*mixin.Metadata, error
 	if err != nil {
 		return nil, err
 	}
-	fs.saveMixinInfo(opts)
+	err = fs.saveMixinInfo(opts)
+	if err != nil {
+		return nil, err
+	}
 	return metadata, nil
 }
 
 const MixinCacheJSON string = "cache.json"
 
 func (fs *FileSystem) saveMixinInfo(opts mixin.InstallOptions) error {
-	mixinsDir, err := fs.Config.GetMixinsDir()
+	mixinsDir, _ := fs.Config.GetMixinsDir()
 	cacheJSONPath := filepath.Join(mixinsDir, "/", MixinCacheJSON)
-	exists, err := fs.FileSystem.Exists(cacheJSONPath)
+	exists, _ := fs.FileSystem.Exists(cacheJSONPath)
 	if !exists {
 		_, err := fs.FileSystem.Create(cacheJSONPath)
 		if err != nil {
@@ -44,7 +46,7 @@ func (fs *FileSystem) saveMixinInfo(opts mixin.InstallOptions) error {
 		}
 	}
 
-	mixinCacheFileContents, err := ioutil.ReadFile(cacheJSONPath)
+	mixinCacheFileContents, err := fs.FileSystem.ReadFile(cacheJSONPath)
 	if err != nil {
 		return errors.Wrap(err, "error reading mixin cache.json")
 	}
@@ -56,7 +58,7 @@ func (fs *FileSystem) saveMixinInfo(opts mixin.InstallOptions) error {
 			return errors.Wrap(err, "error unmarshalling from mixin cache.json")
 		}
 	}
-	//if a mixin exists, skip. needs to be handled via "porter mixin update" ?
+	//if a mixin exists, skip.
 	for _, mixin := range mixinsDataJSON.Mixins {
 		if mixin.Name == opts.Name {
 			return nil
@@ -68,7 +70,7 @@ func (fs *FileSystem) saveMixinInfo(opts mixin.InstallOptions) error {
 	if err != nil {
 		return errors.Wrap(err, "error marshalling to mixin cache.json")
 	}
-	err = ioutil.WriteFile(cacheJSONPath, updatedMixinInfo, 0644)
+	err = fs.FileSystem.WriteFile(cacheJSONPath, updatedMixinInfo, 0644)
 
 	if err != nil {
 		return errors.Wrap(err, "error adding mixin info cache.json")
--- a/pkg/mixin/provider/install.go
+++ b/pkg/mixin/provider/install.go
@@ -56,11 +56,10 @@ func (fs *FileSystem) saveMixinInfo(opts mixin.InstallOptions) error {
 			return errors.Wrap(err, "error unmarshalling from mixin cache.json")
 		}
 	}
-	if len(mixinsDataJSON.Mixins) > 0 {
-		for _, mixin := range mixinsDataJSON.Mixins {
-			if mixin.Name == opts.Name {
-				return nil
-			}
+	//if a mixin exists, skip. needs to be handled via "porter mixin update" ?
+	for _, mixin := range mixinsDataJSON.Mixins {
+		if mixin.Name == opts.Name {
+			return nil
 		}
 	}
 	updatedMixinList := append(mixinsDataJSON.Mixins, mixinInfo{Name: opts.Name, FeedURL: opts.FeedURL, URL: opts.URL})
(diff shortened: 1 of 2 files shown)


pkg/mixin/provider/install.go
```diff
@@ -15,11 +17,73 @@ import (
...
 	if opts.FeedURL != "" {
-		return fs.InstallFromFeedURL(opts)
+		metadata, err = fs.InstallFromFeedURL(opts)
+	} else {
+		metadata, err = fs.InstallFromURL(opts)
 	}
+	if err != nil {
+		return nil, err
+	}
+	fs.saveMixinInfo(opts)
```
carolynvs: We are losing the error returned here by `saveMixinInfo`. This should check for an error and return it.

pkg/mixin/provider/install.go
```diff
@@ -15,11 +17,73 @@ import (
...
+	return metadata, nil
+}
+
+const MixinCacheJSON string = "cache.json"
+
+func (fs *FileSystem) saveMixinInfo(opts mixin.InstallOptions) error {
+	mixinsDir, err := fs.Config.GetMixinsDir()
```
carolynvs: Here you can use `mixinsDir, _` since we are ignoring the error returned by `GetMixinsDir`. Since at this point we are quite sure that method will succeed, it's okay to just ignore the error. 

I have had it on my todo list for quite a while to refactor all of these methods that get a directory relative to the porter home directory to not return an error. Now my inaction has come back to haunt others too! 👻 Sorry about that.

pkg/mixin/provider/install.go
```diff
@@ -15,11 +17,73 @@ import (
...
+	cacheJSONPath := filepath.Join(mixinsDir, "/", MixinCacheJSON)
+	exists, err := fs.FileSystem.Exists(cacheJSONPath)
```
carolynvs: Same on this line, since we only care about if it exists or not, and we aren't looking at `err`, please use `exists, _` so that people reading the code don't wonder about why we aren't checking the error coming back.

pkg/mixin/provider/install.go
```diff
@@ -15,11 +17,73 @@ import (
...
+	if !exists {
+		_, err := fs.FileSystem.Create(cacheJSONPath)
+		if err != nil {
+			return errors.Wrap(err, "error creating mixin cache.json")
+		}
+	}
+
+	mixinCacheFileContents, err := ioutil.ReadFile(cacheJSONPath)
```
carolynvs: Please keep using `fs.FileSystem` for all file system access. We use afero throughout porter to help with unit testing, and any switching back and forth to ioutil, will cause them to no work.
```suggestion
	mixinCacheFileContents, err := fs.FileSystem.ReadFile(cacheJSONPath)
```

pkg/mixin/provider/install.go
```diff
@@ -15,11 +17,73 @@ import (
...
+	}
+
+	mixinsDataJSON := &mixins{}
+	if len(mixinCacheFileContents) > 0 {
+		err = json.Unmarshal(mixinCacheFileContents, &mixinsDataJSON)
+		if err != nil {
+			return errors.Wrap(err, "error unmarshalling from mixin cache.json")
+		}
+	}
+	//if a mixin exists, skip. needs to be handled via "porter mixin update" ?
```
carolynvs: Yeah, we'll have to figure out in a later PR how we want to handle reinstalling a mixin using `porter mixin install` or when we finally have a `porter mixin update` command. That can be left for later though.

pkg/mixin/provider/install.go
```diff
@@ -15,11 +17,73 @@ import (
...
+			return nil
+		}
+	}
+	updatedMixinList := append(mixinsDataJSON.Mixins, mixinInfo{Name: opts.Name, FeedURL: opts.FeedURL, URL: opts.URL})
+	mixinsDataJSON.Mixins = updatedMixinList
+	updatedMixinInfo, err := json.MarshalIndent(&mixinsDataJSON, "", "  ")
+	if err != nil {
+		return errors.Wrap(err, "error marshalling to mixin cache.json")
+	}
+	err = ioutil.WriteFile(cacheJSONPath, updatedMixinInfo, 0644)
```
carolynvs: Again here please use `fs.FileSystem.WriteFile`

pkg/mixin/provider/install.go
abhirockzz: Updated

pkg/mixin/provider/install.go
abhirockzz: No problem! Maybe I can help clean things up in the near future!

pkg/mixin/provider/install.go
abhirockzz: Updated this as well

pkg/mixin/provider/install.go
abhirockzz: Yes, I did use `fs.FileSystem` for checking file existence. Updated now

pkg/mixin/provider/install.go
abhirockzz: Updated this one too

pkg/mixin/provider/install.go
abhirockzz: Yes that's what I thought. Removed the comment

abhirockzz:
> This is looking good! What else we need is some unit tests added to
> 
> https://github.com/deislabs/porter/blob/master/pkg/mixin/provider/install_test.go
> 
> The test should validate some of your edge conditions:
> 
> * no file exists
> * file already exists and we add the mixin entry to it and save
> * the mixin is already in the file
> 
> The other tests in that file show you how to setup a new `FileSystem` to test and use the https://github.com/spf13/afero library to test interactions with the file system. If you have questions, do what you can and I'll help out with the rest.

I have added one test for now. Will work on others based on feedback

pkg/mixin/provider/install_test.go
```diff
@@ -164,3 +165,49 @@ func TestFileSystem_Install_RollbackMissingRuntime(t *testing.T) {
...
 	assert.False(t, mixinDirExists)
 }
+
+func TestFileSystem_Install_MixinInfoSavedWhenNoFileExists(t *testing.T) {
+
+	c := config.NewTestConfig(t)
+	c.SetupPorterHome()
+	p := NewFileSystem(c.Config)
+
+	mixinURL := "https://cdn.deislabs.io/porter/mixins/helm"
```
carolynvs: The only change that should be made is to switch from downloading a real mixin to downloading a fake test mixin, because the size is much smaller. It helps our unit tests run _much_ faster.

Here's how the other test does this:

https://github.com/deislabs/porter/blob/bb837ade09a5624ca1bc9aa0fbf69bc7d28293d2/pkg/mixin/provider/install_test.go#L19-L32

abhirockzz: Thanks. I am aware of the fake test mixin, but did not use it since I did not trigger the installation in the unit test (no `p.Install(opts)`) - its just `p.saveInfo(opts)`. Do you think this is ok?

carolynvs: Oh! Silly me yes you are right. It isn't being downloaded so it is fine to use a real URL. 👍

PR #669, opened 2019-09-30
Set Default InvocationImage

### LINKED ISSUE #383: Make invocation image optional in Porter Manifest
Now that we are going to require a `tag` attribute in the porter manifest for the bundle publishing, we should be able to derive the invocation image name from the bundle. Invocation image is really a CNAB concept and if we don't require it, the experience will likely be enhanced. 

We should allow it to be specified, however, for people who wish to do so.

phillipahereza:
# What does this change
Adds a function `SetDefaults`, that is called after we load the manifest from the filesystem. It checks if `invocationImage` is empty and defaults it to `tag`-installer. For example, if `manifest.Tag` is `deislabs/porter-hello:latest`, then `manifest.InvocationImage` will be defaulted to `deislabs/porter-hello-installer:latest`.

# What issue does it fix
Closes #383

# Notes for the reviewer
N/A

# Checklist
- [x] Unit Tests
- [x] Documentation

### DIFF
--- a/pkg/config/manifest.go
+++ b/pkg/config/manifest.go
@@ -446,6 +446,7 @@ func (c *Config) readFromFile(path string) (*Manifest, error) {
 	if err != nil {
 		return nil, err
 	}
+	m.SetDefaults()
 
 	return m, nil
 }
@@ -499,4 +500,11 @@ func (c *Config) LoadManifestFrom(file string) error {
 	c.ManifestPath = file
 
 	return nil
+}
+
+func (m *Manifest) SetDefaults() {
+	if m.Image == "" {
+		registry_tag := strings.Split(m.BundleTag, ":")
+		m.Image = strings.Join([]string{registry_tag[0] + "-installer",registry_tag[1]}, ":")
+	}
 }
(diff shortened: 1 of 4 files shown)


pkg/config/manifest.go
```diff
@@ -446,6 +446,7 @@ func (c *Config) readFromFile(path string) (*Manifest, error) {
 	if err != nil {
 		return nil, err
 	}
+	m.SetDefaults()
```
vdice: I'm wondering if this call should be higher up (say, in `ReadManifest()`?)  This would set a default image regardless of if we are reading from a file or a URL, then.  Does that sound like a valid case as well?

phillipahereza: Yes it does, but does the version tag of the invocationImage default to the version tag of `manifest.Tag` or the `manifest.Version` or just `latest`?

vdice: Ah, great question!  Let's check with @carolynvs on precedence here.  I'm still unsure on which would be canonical to represent the bundle version.

vdice: @phillipahereza It looks like we want to default to `manifest.Version` for the version tag.

vdice:
(porter-integration failed due to a bug outside of this PR; fix merged in https://github.com/deislabs/porter/pull/696)

PR #671, opened 2019-09-30
fix(test-terraform.sh): remove fragile, redundant file contents check

vdice:
# What does this change

Removes a fragile check (based on literal text output which is subject to change) in the terraform portion of the cli test.  It is now redundant as we've since added `porter instance outputs list` checks to verify the expected file contents value.  

# What issue does it fix
Fixes the `test-cli` target in CI.

# Notes for the reviewer
N/A

# Checklist
- [ ] Unit Tests
- [ ] Documentation


carolynvs:
> fragile, redundant

Tell me how you really feel, @vdice 😂

vdice:
😂 I've allowed a bit of harshness since I'm responsible for its origins :)

PR #681, opened 2019-10-02
feat(pkg/exec): add overrideable Dash property to Flag

vdice:
# What does this change

Adds a `Dash` property to the `Flag` struct in the `exec/builder` package, for optional overriding of the dash appended to a given flag.  See https://github.com/deislabs/porter-terraform/pull/25 for example use/scenario.

# What issue does it fix
N/A

# Notes for the reviewer
N/A

# Checklist
- [x] Unit Tests
- [ ] Documentation

### DIFF
--- a/pkg/exec/builder/flags.go
+++ b/pkg/exec/builder/flags.go
@@ -11,6 +11,7 @@ import (
 type Flag struct {
 	Name   string
 	Values []string
+	Dash   string
 }
 
 // NewFlag creates an instance of a Flag.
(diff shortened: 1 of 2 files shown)


carolynvs:
I'm a bit reluctant to define dash per flag. It would seem to be more natural for a mixin to define it's preferred long and short flag prefix at the mixin level.

What if we added an interface that a mixin could implement if they want to override the flag dashes and then builder.ExecuteStep could check if that interface is implemented and pass the long and short flags to ToSlice to use?

Let me know if you think that's helpful or super java-esque. 😀

vdice:
@carolynvs let me know what you think of https://github.com/deislabs/porter/pull/714 when convenient.  If this approach looks better, I can close this PR.  Thanks!

vdice:
Closing in favor of https://github.com/deislabs/porter/pull/714

PR #686, opened 2019-10-03
feat(docs): Validate if destination exists before invoking command

### LINKED ISSUE #685: `porter docs` not verifying if destination exists
When executing `porter docs` with an invalid or without a `-dest` argument, the generator is invoked.

Instead, as part of the existing `Validate` function, the operation should be canceled if the destination doesn't exist.

ThorstenHans:
# What does this change

As part of validation, `porter docs` checks if the destination folder exists. If the destination doesn't exist, execution is canceled and a proper error message is presented to the user.

# What issue does it fix
Closes #685 

# Checklist
- [ ] Unit Tests
- [ ] Documentation
  - [ X ] Documentation Not Impacted

### DIFF
--- a/pkg/docs/generator.go
+++ b/pkg/docs/generator.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/deislabs/porter/pkg/context"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/cobra/doc"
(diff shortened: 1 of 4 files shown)


pkg/docs/generator.go
```diff
@@ -22,7 +22,12 @@ const DefaultDestination = "./docs/content/cli/"
 func (o *DocsOptions) Validate() error {
 	if o.Destination == "" {
 		o.Destination = DefaultDestination
-	}
+    }
+
+    _, err := os.Stat(o.Destination)
```
carolynvs: Porter uses the github.com/spf13/afero library, instead of using os.Stat or similar ioutil functions that hit the filesystem directly.

The Porter [context struct](https://github.com/deislabs/porter/blob/9e743f7afa082d6dafa011348b710a4fdabc4f96/pkg/context/context.go#L22) has an instance of an afero filesystem. The pattern that we use in the options validate function is to pass in the context struct as an argument to Validate like so

```golang
func (o *DocsOptions) Validate(cxt *context.Context) error {
```

and then you can do things like check if a file exists using afero

```golang
exists := cxt.FileSystem.Exists(o.Destination)
```

Now you are all set to write a unit test, here's an example test that you can follow: 

https://github.com/deislabs/porter/blob/master/pkg/porter/cnab_test.go

pkg/docs/generator.go
```diff
@@ -22,7 +22,12 @@ const DefaultDestination = "./docs/content/cli/"
...
+    if err != nil {
+        return errors.New(fmt.Sprintf("error generating the markdown documentation; desitination '%s' doesn't exist", o.Destination))
```
carolynvs: For the error message when we are validating a flag, we try to include the flag name so that the user can tell which flag they need to fix. In this case, changing the error message to "--destination %s doesn't exist" would work.

Also you can simplify this to `errors.Errorf("error message", o.Destination)` The errors package we use, `github.com/pkg/errors` has a nice function to combine those two methods into one.

ThorstenHans:
Hey @carolynvs thanks for the feedback, I’ll update the PR this evening according to your comments 👍🏼✌🏼

Issue #544, opened 2019-08-22
Porter Export


jeremyrickard:
CNAB [thick bundles](https://github.com/deislabs/cnab-spec/blob/master/104-bundle-formats.md#thick-and-thin-bundles) are used for transmitting across air-gapped networks. Duffle currently has some import/export capabilities, and I think we should add these in Porter as well.  This issue tackles the `export` side of the equation. 

The UX for export should probably be similar to the `install` type actions:

For a local `porter` based bundle: 

```
$ cd my-cool-bundle
$ porter export
```

We should also support the --cnab-file action:

```
$ porter export --cnab-file path-to-some-file
```

We should also probably support exporting a bundle that has a tag reference:

```
$ porter export --tag deislabs/my-cool-bundle:0.1.0
```

For reference, duffle's help info:

```
duffle export --help

Packages a bundle, and by default any images referenced by the bundle, within a single gzipped tarfile.

Unless --thin is specified, a thick bundle is exported. A thick bundle contains the bundle manifest and all images
(including invocation images) referenced by the bundle metadata. Images are saved as an OCI image layout in the
artifacts/layout/ directory.
```

All of these should produce a tgz file, including any referenced images.

vdice:
First iteration implemented in https://github.com/deislabs/porter/pull/691; closing.

PR #691, opened 2019-10-03
This introduces a WIP "export" using the logic from CNAB GO

jeremyrickard:
# What does this change

This introduces some initial plumbing for generating a "thick" bundle. 

# What issue does it fix
Part of #544

# Notes for the reviewer

Note: it doesn't use all of cnab-go yet, i copied some code over pending a refactor in
cnab-go. I'd like to merge this more or less as is so we can start experimenting with the flip side of the equation.

### DIFF
--- a/cmd/porter/bundle.go
+++ b/cmd/porter/bundle.go
@@ -292,6 +292,7 @@ func buildBundleArchiveCommand(p *porter.Porter) *cobra.Command {
 		Example: `  porter bundle archive [FILENAME]
   porter bundle archive --file another/porter.yaml [FILENAME]
   porter bundle archive --cnab-file some/bundle.json [FILENAME]
+  porter bundle archive --tag repo/bundle:tag [FILENAME]
 		  `,
 		PreRunE: func(cmd *cobra.Command, args []string) error {
 			return opts.Validate(args, p.Context)
--- a/cmd/porter/bundle.go
+++ b/cmd/porter/bundle.go
@@ -303,5 +303,7 @@ func buildBundleArchiveCommand(p *porter.Porter) *cobra.Command {
 	f := cmd.Flags()
 	f.StringVarP(&opts.File, "file", "f", "", "Path to the Porter manifest. Defaults to `porter.yaml` in the current directory.")
 	f.StringVar(&opts.CNABFile, "cnab-file", "", "Path to the CNAB bundle.json file.")
+	f.StringVarP(&opts.Tag, "tag", "t", "",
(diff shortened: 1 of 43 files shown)


carolynvs:
[fix.patch.txt](https://github.com/deislabs/porter/files/3691150/fix.patch.txt)

Here is the fix to the Makefile and lock that should get the build passing. Sorry about that, this was a break from another PR that I missed...

pkg/porter/archive.go
```diff
@@ -2,24 +2,172 @@ package porter
...
 
 // Archive is a composite function that generates a CNAB thick bundle. It will pull the invocation image, and
 // any referenced images locally (if needed), export them to individual layers, generate a bundle.json and
 // then generate a gzipped tar archive containing the bundle.json and the images
 func (p *Porter) Archive(opts ArchiveOptions) error {
-	return fmt.Errorf("archive is not yet implemented")
+
+	err := p.prepullBundleByTag(&opts.BundleLifecycleOpts)
+	if err != nil {
+		return errors.Wrap(err, "unable to pull bundle before invoking credentials generate")
```
vdice: Looks like the error message needs updating here

pkg/porter/archive.go
```diff
@@ -2,24 +2,172 @@ package porter
...
+		out:                   p.Config.Out,
+		logs:                  p.Config.Out,
+		bundle:                bun,
+		destination:           dest,
+		imageStoreConstructor: ctor,
+	}
+	if err := exp.export(); err != nil {
+		return err
+	}
+	// if ex.verbose {
```
vdice: Should we just remove this section?

cmd/porter/bundle.go
```diff
@@ -303,5 +303,7 @@ func buildBundleArchiveCommand(p *porter.Porter) *cobra.Command {
 	f := cmd.Flags()
 	f.StringVarP(&opts.File, "file", "f", "", "Path to the Porter manifest. Defaults to `porter.yaml` in the current directory.")
 	f.StringVar(&opts.CNABFile, "cnab-file", "", "Path to the CNAB bundle.json file.")
+	f.StringVarP(&opts.Tag, "tag", "t", "",
```
vdice: Shall we add an add'l example using this flag to the command examples text?

pkg/porter/archive.go
jeremyrickard: i was going to add it back, eventually. just needs to be updated to work like we want it to work,

Issue #655, opened 2019-09-26
Check for missing ARG BUNDLE_DIR declaration in dockerfile template


carolynvs:
When the dockerfile template doesn't have `ARG BUNDLE_DIR` in it, the docker build doesn't provide a useful error back to the user that hints what is wrong. Instead it just says this:

```
Starting Invocation Image Build =======>
Error: unable to build CNAB invocation image: failed to stream docker build output: cannot normalize nothing
```

Thankfully @clarenceb figured it out and fixed it in our azure workshop example, but it's an easy mistake to make and we should help provide a better error message when that happens.

Before a build, porter should look at the lines in the dockerfile template for `ARG BUILD_DIR` and if it's not present, return an explicit error

```
dockerfile.tmpl must declare the build argument BUILD_DIR. Add the following line to the file and re-run porter build:
ARG BUILD_DIR
```

This check should happen in:

https://github.com/deislabs/porter/blob/master/pkg/build/dockerfile-generater.go `getBaseDockerfile` which probably needs to be refactored into multiple functions now that we want to put more logic into it around validating the user supplied template.

AGMETEOR:
@carolynvs I am going to try this one out...

carolynvs:
Thanks!

AGMETEOR:
Started working on this. Apologies for the delay.

PR #693, opened 2019-10-04
Check dockerfile for ARG BUILD_DIR

AGMETEOR:
# What does this change
This change adds a method `readAndValidateDockerfile` to the `type` `*DockerfileGenerator` in order to check for `ARG BUILD_DIR` in the Dockerfile during the building of a bundle. If `ARG BUILD_DIR` is not found, an error message is returned.

# What issue does it fix
Closes [#655](https://github.com/deislabs/porter/issues/655)

# Checklist
- [x] Unit Tests
- [ ] Documentation

### DIFF
--- a/pkg/build/dockerfile-generater.go
+++ b/pkg/build/dockerfile-generater.go
@@ -82,6 +82,27 @@ func (g *DockerfileGenerator) buildDockerfile() ([]string, error) {
 	return lines, nil
 }
 
+func (g *DockerfileGenerator) readAndValidateDockerfile(s *bufio.Scanner) ([]string, error) {
+	buildArg := "ARG BUILD_DIR"
+	foundBuildArg := false
+	var lines []string
+	for s.Scan() {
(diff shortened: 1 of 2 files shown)


AGMETEOR:
Hello @carolynvs , wondering if you have any feedback for me on this...

carolynvs:
Thanks for the ping. I'll take a look right now. 👀

pkg/build/dockerfile-generater.go
```diff
@@ -22,6 +22,7 @@ type DockerfileGenerator struct {
 	*manifest.Manifest
 	*templates.Templates
 	mixin.MixinProvider
+	hasBuildArg bool
```
carolynvs: `hasBuildArg` can be defined as a local variable inside of `readAndValidateDockerfile` since it is set and used only in that function.

AGMETEOR: @carolynvs made the changes :)

Issue #678, opened 2019-10-02
porter maybe can cache old bundle versions forever


carolynvs:
## Describe the bug
I have seen this locally when working on the integration tests (for dependencies with the mysql and wordpress bundles). It was also just reported in Slack as well.

## To Reproduce
I'm not totally sure but I'm guessing it looks something like this:

1. Publish a bundle with under a version, e.g. 1.0
2. Run `porter install --tag mybuns:v1.0`
3. Make changes to that bundle and republish to that same version again
4. Run `porter install --tag mybuns:v1.0` again

## Expected behavior
The user expects it to use the latest version of that bundle.

I'm not sure how porter knows to expire the cache for that tag? Maybe it's just a problem for local developers who are pushing and repushing the same tag, so when we publish, we should nuke the cache for that bundle and tag combo?

## Version
v0.16.0-beta.1

vdice:
Ah, I've been running into this as well.  Indeed, currently, Porter simply caches using a hash of the bundle tag.  Therefore, if digest/contents change whilst still using the same tag, no new cache entry will be generated.

We probably need to add additional unique info (digest/etc.) when generated the cached bundle hash?  The only tricky bit is ensuring it is still easy to find a cached bundle purely on bundle tag (or, perhaps we can update this logic to incorporate the same unique info.)

PR #700, opened 2019-10-04
feat(*): add publishing from an archive

vdice:
# What does this change
- add `--archive` and `--tag` flags to `porter publish`
- defer to tag when publishing bundle (from archive or not)
- wire up archive importer/loader and then gen new bundle with updated images
- new/published bundle will have new images (invocation and other) with updated registry/orgs as derived from `--tag` value

TODO:
  - [x] address inline/code `TODO`s
  - [x] add test for general publishing from archive scenario (needs to be integration)?  units added and [integration test](https://github.com/deislabs/porter/pull/700/commits/7f46af633b2de864915d267607cd96bdb036c7ae)
  - [x] Currently, we don't attempt to change/update the invocation image on publishing from an archive; just the bundle tag.  Do so in this PR?  From `--tag` or add new `--registry` flag, etc.?

# What issue does it fix
Towards https://github.com/deislabs/porter/issues/545

# Notes for the reviewer
- `deislabs/cnab-go` dep currently using branch w/ DeepCopy() functionality in: https://github.com/deislabs/cnab-go/pull/143

# Checklist
- [x] Unit Tests
- [ ] Documentation

### DIFF
--- a/cmd/porter/bundle.go
+++ b/cmd/porter/bundle.go
@@ -266,7 +266,6 @@ func buildBundlePublishCommand(p *porter.Porter) *cobra.Command {
 		Example: `  porter bundle publish
   porter bundle publish --file myapp/porter.yaml
 	porter bundle publish --insecure
-	porter bundle publish --archive /tmp/mybuns.tgz
 	porter bundle publish --archive /tmp/mybuns.tgz --tag myrepo/my-buns:0.1.0
 		`,
 		PreRunE: func(cmd *cobra.Command, args []string) error {
@@ -281,7 +280,7 @@ func buildBundlePublishCommand(p *porter.Porter) *cobra.Command {
 	f.StringVarP(&opts.File, "file", "f", "", "Path to the Porter manifest. Defaults to `porter.yaml` in the current directory.")
 	f.BoolVar(&opts.InsecureRegistry, "insecure-registry", false, "Don't require TLS for the registry.")
 	f.StringVarP(&opts.ArchiveFile, "archive", "a", "", "Path to the bundle archive in .tgz format")
-	f.StringVarP(&opts.Tag, "tag", "t", "", "Bundle tag for newly published bundle")
+	f.StringVarP(&opts.Tag, "tag", "t", "", "Bundle tag for newly published bundle; required if --archive is supplied")
 
 	return &cmd
 }
--- a/cmd/porter/bundle.go
+++ b/cmd/porter/bundle.go
@@ -265,7 +265,9 @@ func buildBundlePublishCommand(p *porter.Porter) *cobra.Command {
 		Long:  "Publishes a bundle by pushing the invocation image and bundle to a registry.",
 		Example: `  porter bundle publish
   porter bundle publish --file myapp/porter.yaml
-  porter bundle publish --insecure
+	porter bundle publish --insecure
+	porter bundle publish --archive /tmp/mybuns.tgz
+	porter bundle publish --archive /tmp/mybuns.tgz --tag myrepo/my-buns:0.1.0
 		`,
 		PreRunE: func(cmd *cobra.Command, args []string) error {
 			return opts.Validate(p.Context)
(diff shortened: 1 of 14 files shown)


cmd/porter/bundle.go
```diff
@@ -265,7 +265,9 @@ func buildBundlePublishCommand(p *porter.Porter) *cobra.Command {
 		Long:  "Publishes a bundle by pushing the invocation image and bundle to a registry.",
 		Example: `  porter bundle publish
   porter bundle publish --file myapp/porter.yaml
-  porter bundle publish --insecure
+	porter bundle publish --insecure
+	porter bundle publish --archive /tmp/mybuns.tgz
```
carolynvs: mybuns FTW

cmd/porter/bundle.go
```diff
@@ -278,6 +280,9 @@ func buildBundlePublishCommand(p *porter.Porter) *cobra.Command {
 	f := cmd.Flags()
 	f.StringVarP(&opts.File, "file", "f", "", "Path to the Porter manifest. Defaults to `porter.yaml` in the current directory.")
 	f.BoolVar(&opts.InsecureRegistry, "insecure-registry", false, "Don't require TLS for the registry.")
+	f.StringVarP(&opts.ArchiveFile, "archive", "a", "", "Path to the bundle archive in .tgz format")
+	f.StringVarP(&opts.Tag, "tag", "t", "", "Bundle tag for newly published bundle")
```
carolynvs: We should mention that it is required when --archive is specified

pkg/porter/publish.go
```diff
@@ -61,13 +100,61 @@ func (p *Porter) Publish(opts PublishOptions) error {
 		return err
 	}
 
-	return p.Registry.PushBundle(bun, p.Manifest.BundleTag, opts.InsecureRegistry)
+	return p.Registry.PushBundle(bun, tag, opts.InsecureRegistry)
+}
+
+// TODO: tests!
+func (p *Porter) publishFromArchive(opts PublishOptions) error {
+	fmt.Fprintf(p.Out, "Extracting bundle from archive %s...\n", opts.ArchiveFile)
```
carolynvs: nit: I suggest making this conditional debug logging to `p.Err`

pkg/porter/publish.go
```diff
@@ -61,13 +100,61 @@ func (p *Porter) Publish(opts PublishOptions) error {
...
+	l := loader.NewLoader()
+	// as of writing, NewImporter always returns nil as error
+	imp, _ := packager.NewImporter(source, tmpDir, l, false)
+
+	err = imp.Import()
+	if err != nil {
+		return errors.Wrapf(err, "failed to extract bundle from archive %s", opts.ArchiveFile)
+	}
+
+	// TODO: do we want to import bundle into Porter's cache?
```
carolynvs: We have another open issue to have special handling of the cache and cache invalidation on publish, #678, so I suggest maybe leaving that for a follow on, or as part of that issue.

pkg/porter/publish.go
```diff
@@ -61,13 +101,151 @@ func (p *Porter) Publish(opts PublishOptions) error {
...
+
+// updateBundleWithNewImage updates a bundle with a new image at the provided index
+// constructed using the provided original image and new bundle tag
+func (p *Porter) updateBundleWithNewImage(bun *bundle.Bundle, img, tag string, index interface{}) error {
+	newImg, err := getNewImageNameFromBundleTag(img, tag)
+	if err != nil {
+		return err
+	}
+
+	digest, err := p.Registry.Copy(img, newImg)
```
carolynvs: So when we publish an archived bundle, the contents of the bundle don't really matter (the referenced images). We use the tags from where the images were originally published, and rely on being able to copy from there to the new location. We don't copy from the archive itself. Am I understanding that right?

vdice: Good clarification -- indeed, this is the current approach.  

As I understand it, the functionality for importing images from an OCI Image layout (i.e., images in the `/artifacts` dir of the archive) into the Docker daemon is not yet available (see https://github.com/deislabs/cnab-go/issues/136)

Should I add a note/comment in-line mentioning this?  Would it be prudent to add a ticket in our backlog to revisit this area, If/when the aforementioned functionality is available?  (To import images straight from the archive into the local daemon and then re-tag/push.)

jeremyrickard: So how does this work with the offline scenario where the bundle is on an air-gapped network?

I *think* we need to use the Pivotal relocation library here and copy *FROM* the archive up to a new registry. Loading things into the local docker daemon is nice, but for any case that involves a registry, it is kind of irrelevant isn't it?

vdice: Mmm good point.  I'll work on the layout relocation approach directly from the archive and update.

vdice: Ok @carolynvs & @jeremyrickard , updated logic to use the relocation lib to parse extracted OCI Layout and push renamed imgs directly from the archived images (same digests).  Did find one issue which I'll raise in the image-relocation repo, but I don't think it is a blocker (and I believe the fix would be in this dependency, anyhow).

Issue #701, opened 2019-10-05
Remote claims storage


carolynvs:
This is a larger epic that I'm using this issue to track. Porter will support plugins that store claims in remote systems, for example blob storage or in a database.

squillace:
Part of a larger state management issue, isn't it? #763 #770 and so on?

carolynvs:
This was implemented and released in v0.23.0-beta.1

PR #702, opened 2019-10-05
Move manifest into separate package from config

carolynvs:
# What does this change
This allows plugins to import porter config without also taking dependencies on everything related to the porter manifests (including our runtime manifest). They really are two separate concerns, porter's config and the bundle configuration, so this was a long time coming.

# What issue does it fix
This is preparatory refactoring for #701

# Notes for the reviewer
Sorry for the large change but config is used _everywhere_...

# Checklist
- [ ] Unit Tests
- [ ] Documentation


pkg/cnab/config_adapter/adapter.go
```diff
@@ -9,25 +9,28 @@ import (
...
 	return &ManifestConverter{
-		Config:       cfg,
+		Context:      cxt,
+		Manifest:     manifest,
 		ImageDigests: imageDigests,
 	}
 }
 
 func (c *ManifestConverter) ToBundle() *bundle.Bundle {
-	fmt.Fprintf(c.Out, "\nGenerating Bundle File with Invocation Image %s =======> \n", c.Manifest.Image)
```
carolynvs: I decided to remove these output lines, as doing a conversion from one data structure to another doesn't really warrant these log lines anymore. They maybe made sense last November for the prototype but now aren't necessary and are just noise.

pkg/manifest/manifest_test.go
```diff
@@ -12,50 +14,50 @@ import (
 )
 
 func TestReadManifest_URL(t *testing.T) {
-	c := NewTestConfig(t)
-	url := "https://raw.githubusercontent.com/deislabs/porter/master/pkg/config/testdata/simple.porter.yaml"
-	m, err := c.ReadManifest(url)
+	c := config.NewTestConfig(t)
+	url := "https://raw.githubusercontent.com/deislabs/porter/v0.17.0-beta.1/pkg/config/testdata/simple.porter.yaml"
```
carolynvs: This file moved in the refactoring so I am using a tag for now to prevent the build from breaking when the PR is merged.

PR #706, opened 2019-10-08
Make additional bundle metadata available for templating

### LINKED ISSUE #78: Make additional bundle metadata available to template
Right now the porter manifest lets a bundle author template out values for certain things, `bundle.parameters.NAME`, `bundle.credentials.NAME`, etc.

But metadata about the bundle can't be substituted in yet. I would like to be able to do something like this in my porter manifest:

```
bundle.name
```

We have a view pieces of information in our bundle metadata that would be nice to make available:

* Name
* Description
* Version
* Image

It may help you when you implement this to move those fields into a new embedded struct on the Manifest. If you need to do that, feel free.

This is the function that would need to be modified:
https://github.com/deislabs/porter/blob/9eed13889d855c5665c42b9b4b2ac8153a8feae4/pkg/config/manifest.go#L647

That would let someone use `bundle.*`

There is another function 

https://github.com/deislabs/porter/blob/9eed13889d855c5665c42b9b4b2ac8153a8feae4/pkg/config/manifest.go#L83

That lets someone use `bundle.dependencies.NAME.*`, for example to access metadata about one of their dependencies, `bundle.dependencies.mysql.version`.

It would be nice (though not required), if both were updated.

See https://porter.sh/authoring-bundles/ for more context on the porter manifest.

dev-drprasad:
# What does this change
Let authors to use bundle/dependency metadata (Name, Description, Version, Image) for templating in porter.yaml

# What issue does it fix
Closes #78

# Checklist
- [ ] Unit Tests
- [ ] Documentation


carolynvs:
@dev-drprasad Your PR can't be tested because it needs to be rebased with the latest changes from master. Let me know if you need help with that.

dev-drprasad:
@carolynvs i rebased onto master using `git rebase master`. Let me know if thats not the way to do it

I wrote a test case to test dependency metadata substitution. But test failing because there are no dependencies associated with manifest. In other words, `len(m.bundles)` is `0`. I am not sure how to load dependency to manifest. Can you please check test case function: `TestDependencyMetadataAvailableForTemplating` and confirm whether code is proper or not

pkg/manifest/runtime-manifest_test.go
```diff
@@ -211,6 +211,53 @@ func TestResolvePathParam(t *testing.T) {
...
+	assert.Equal(t, "echo \"name:HELLO version:0.1.0 description:An example Porter configuration image:jeremyrickard/porter-hello:latest\"", cmd)
+}
+
+func TestDependencyMetadataAvailableForTemplating(t *testing.T) {
+	cxt := context.NewTestContext(t)
+	cxt.AddTestFile("testdata/dep-metadata-substitution.yaml", config.Name)
+	cxt.AddTestDirectory("testdata/bundles", "bundles")
+
+	m, _ := LoadManifestFrom(cxt.Context, config.Name)
+	rm := NewRuntimeManifest(cxt.Context, ActionInstall, m)
```
carolynvs: After the runtime manifest is created, `rm.bundles` needs to be populated. Since this is a unit test, you can get away with setting it with just enough data needed for your test:

```go
rm.bundles := map[string]bundle.Bundle{
		"mysql": {
			Name:"Azure MySQL",
			Description: "Azure MySQL database as a service",
			Version: "v1.0.0",
		},
	}
```

Normally that field is populated by `rm.loadDependencyDefinitions()` which reads the bundle definitions from `/cnab/app/dependencies/ALIAS/bundle.json` from inside the running bundle. But that's way more than should be setup for this test.

dev-drprasad:
@carolynvs I am done from my side. Please review and let me know any changes required or if i missed somethings

dev-drprasad:
@carolynvs No problem. Health comes first

PR #708, opened 2019-10-08
Handle wiring images

### LINKED ISSUE #692: Handle updated image(s) with image wiring
We added support to reference bundle images with our mustache template {{ }} syntax. This is currently driven by the porter.yaml.

When we introduce thick bundles and bundle relocation, the porter.yaml within the invocation image will be out of date. We should “resolve” the real images using the bundle.json that gets mounted at runtime per the spec. We can later extend this to any relocation file that is provided per the spec, so we can ensure that the porter-runtime reflects the correct images.

jeremyrickard:
# What does this change

This change introduces additional run time wiring logic for the mustache like `bundle.images` functionality. When a bundle is executed, the `bundle.json` and an optional `relocation-mapping.json` file can be mounted into the invocation image. These can override the values present in the `porter.yaml` (which is inserted into the invocation image at build). The values provided by the `bundle.json` and `relocation-mapping.json` should override what is in the `porter.yaml` manifest. 

# What issue does it fix
Closes #692 

# Notes for the reviewer

👋 🌵 

# Checklist
- [x] Unit Tests


jeremyrickard:
@vdice nixed the Original Image and TODO comments

jeremyrickard:
I have an impending update of the relocation mapping. The format I had seen before isn't what is there for reals, so fixing that up at the moment.

jeremyrickard:
OK, this is updated now.

The image relocation map looks like:

```

File: pkg/porter/run.go [hunk 1]
```diff
@@ -102,6 +105,31 @@ func (p *Porter) Run(opts RunOptions) error {
 		return err
 	}
 
+	//Update the runtimeManifest images with the bundle.json and relocation mapping (if it's there)
```
carolynvs: nit: In a follow up, it would be nice to wrap this all up in a function in this same file so that `Run` continues to be pretty readable at a high level. Maybe `handleRelocationImages` or something.

jeremyrickard: good point, i'll do a follow up. I *think* i want to do a refactor of a bunch of the logic here

carolynvs: Yeah based on the touched file list you showed me, a refactor doesn't seem like a bad idea if you have ideas on how to improve the abstraction.

PR #711, opened 2019-10-09
Command driver support

simongdavies:
# What does this change
Adds support to Porter for Command Driver.
Updates cnab-go dependency to v0.7.0-beta1

### DIFF
--- a/pkg/porter/cnab.go
+++ b/pkg/porter/cnab.go
@@ -2,11 +2,13 @@ package porter
 
 import (
 	"bufio"
+	"fmt"
 	"os"
 	"path/filepath"
 
 	"github.com/deislabs/cnab-go/bundle"
 	"github.com/deislabs/cnab-go/claim"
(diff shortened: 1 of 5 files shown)


pkg/porter/cnab.go
```diff
@@ -304,9 +306,14 @@ func (o *sharedOptions) defaultDriver() {
 // validateDriver validates that the provided driver is supported by Porter
 func (o *sharedOptions) validateDriver() error {
 	switch o.Driver {
-	case "docker", "debug":
+	case "docker", "debug", "kubernetes":
```
carolynvs: I'm not sure why kubernetes was added in this PR? We haven't done any testing so I'm not sure if there's more work to be done on our side to make sure that we are properly loading and configuring that driver. Let's split that change into a separate PR and add an integration test for it.

simongdavies: OK, I removed the Kubernetes driver.

PR #714, opened 2019-10-09
feat(exec/builder): add GetDashes method to ExecutableStep interface

vdice:
# What does this change
Adds a `GetDashes()` method to the `ExecutableStep` interface such that implementations can modify the dashes appended to flags.

As proposed in https://github.com/deislabs/porter/pull/681#issuecomment-539102454

# What issue does it fix
N/A, but used by https://github.com/deislabs/porter-terraform/pull/25

# Notes for the reviewer
Is there a way to have this new method be "optional"? Or will we need to add implementations of the method in the other mixins that implement `ExecutableStep`?

# Checklist
- [x] Unit Tests
- [ ] Documentation

### DIFF
--- a/pkg/exec/action.go
+++ b/pkg/exec/action.go
@@ -38,7 +38,6 @@ func (a Action) GetSteps() []builder.ExecutableStep {
 }
 
 var _ builder.ExecutableStep = Step{}
-var _ builder.HasCustomDashes = Step{}
 var _ builder.StepWithOutputs = Step{}
 
 type Step struct {
@@ -63,10 +62,6 @@ func (s Step) GetArguments() []string {
 
 func (s Step) GetFlags() builder.Flags {
 	return s.Flags
-}
-
-func (s Step) GetDashes() builder.Dashes {
-	return builder.DefaultFlagDashes
 }
 
 func (s Step) GetOutputs() []builder.Output {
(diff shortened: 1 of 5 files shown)


carolynvs:
> Is there a way to have this new method be "optional"? Or will we need to add implementations of the method in the other mixins that implement ExecutableStep?

So what I was suggesting was that instead of adding it to the existing interface was to make a new interface. And only Steps that need to override the default `--` + `-` implement that interface. Then `ExecuteStep` can check if step implements that optional interface, if not it can supply the default `Dash` struct instance to `ToSlice()`.

It's very close to what you have now.

vdice:
@carolynvs ah I see.  Thanks so much for working with me on this.  I _think_ I have it now... PTAL when convenient.

pkg/exec/action.go
```diff
@@ -38,6 +38,7 @@ func (a Action) GetSteps() []builder.ExecutableStep {
 }
 
 var _ builder.ExecutableStep = Step{}
+var _ builder.HasCustomDashes = Step{}
```
carolynvs: I think that it would still work with this (line 41) and lines 68 -71 removed. The interface for HasCustomDashes is optional and since you are doing the default, there's no need to implement it, right?

vdice: Oh yes!  Thank you, good call; removed.

carolynvs:
💃 🕺

PR #719, opened 2019-10-10
Update Porter to handle relocation mapping on pull/operations

### LINKED ISSUE #543: Support image relocation
Images referenced by a CNAB bundle MAY be [relocated](https://github.com/deislabs/cnab-spec/blob/master/103-bundle-runtime.md#image-relocation) , for example by copying them to a private registry. This process results in a mapping file:

```
{  "gabrtv/microservice@sha256:cca460afa270d4c527981ef9ca4989346c56cf9b20217dcea37df1ece8120687": "my.registry/microservice@sha256:cca460afa270d4c527981ef9ca4989346c56cf9b20217dcea37df1ece8120687",
  "technosophos/helloworld:0.1.0": "my.registry/helloworld:0.1.0"
}
```

That mapping file _should_ be mounted into an invocation image at: `

The spec further says: 

```
The run tool MAY use this file to modify its behavior. For example, a run tool MAY substitute image references using the mapping in this file.
```

We should enhance porter to enable a relocation file to be provided and mounted into an invocation image,  then also modify the porter runtime so that anything in the relocation file can be used within the execution of a bundle operation. Ideally, if we implement #501, people that build bundles that reference images can simply refer to them like ``{{ bundle.images.SOMENAME}}``, like in a call to the `helm` mixin or some other mixin that needs to use images, and then Porter will helpfully provide them with the correct reference.

jeremyrickard:
# What does this change

This PR updates Porter to handle a relocation mapping if it is generated on a bundle pull. If generated, it is stored in the bundle cache and then passed along via an oper config to add the file to an operation before it is run. We previously only cached the bundle there. 

# What issue does it fix
Closes #543 

# Checklist
- [x] Unit Tests


pkg/cache/cache.go
```diff
@@ -30,41 +32,68 @@ func New(cfg *config.Config) BundleCache {
 // FindBundle looks for a given bundle tag in the Porter bundle cache and
 // returns the path to the bundle if it exists. If it is not found, an
 // empty string and the boolean false value are returned.
-func (c *cache) FindBundle(bundleTag string) (string, bool, error) {
+func (c *cache) FindBundle(bundleTag string) (string, string, bool, error) {
```
vdice: nit: perhaps update comment to mention that this function now also may return the relocation-mapping file path, if exists

pkg/cache/cache.go
```diff
@@ -30,41 +32,68 @@ func New(cfg *config.Config) BundleCache {
...
+		return cachedBundlePath, cachedReloPath, true, nil
+	}
+	return cachedBundlePath, "", true, nil
 
 }
 
 // StoreBundle will write a given bundle to the bundle cache, in a location derived
 // from the bundleTag.
-func (c *cache) StoreBundle(bundleTag string, bun *bundle.Bundle) (string, error) {
+func (c *cache) StoreBundle(bundleTag string, bun *bundle.Bundle, reloMap relocation.ImageRelocationMap) (string, string, error) {
```
vdice: nit: similar to `FindBundle`, I wonder if we should update the comment to mention a relocation mapping filepath may be returned?

pkg/cnab/cnab-to-oci/registry.go
```diff
@@ -31,10 +33,10 @@ func NewRegistry(c *portercontext.Context) *Registry {
 }
 
 // PullBundle pulls a bundle from an OCI registry.
-func (r *Registry) PullBundle(tag string, insecureRegistry bool) (*bundle.Bundle, error) {
+func (r *Registry) PullBundle(tag string, insecureRegistry bool) (*bundle.Bundle, relocation.ImageRelocationMap, error) {
```
vdice: nit: update comment to mention relo map will be returned as well?

pkg/manifest/runtime-manifest.go
```diff
@@ -311,8 +311,14 @@ func (m *RuntimeManifest) Prepare() error {
...
 // mounted bundle.json and relocation mapping
 func (m *RuntimeManifest) ResolveImages(bun *bundle.Bundle, reloMap relocation.ImageRelocationMap) error {
+	// It only makes sense to process this if the runtime manifest has images defined. If none are defined
+	// return early
+	if len(m.ImageMap) == 0 {
+		return nil
+	}
 	reverseLookup := make(map[string]string)
 	for alias, image := range bun.Images {
+		fmt.Fprintf(m.Out, "alias: %s, image: %s\n", alias, image.Image)
```
vdice: Double-checking to see if we want to keep this log?  If so, should we conditionalize on if debug is set?

pkg/cnab/provider/action.go
```diff
@@ -52,3 +60,23 @@ func (d *Runtime) AddFiles(args ActionArguments) action.OperationConfigFunc {
...
+	return func(op *driver.Operation) error {
+		if args.RelocationMapping != "" {
+			b, err := d.FileSystem.ReadFile(args.RelocationMapping)
+			if err != nil {
+				return errors.Wrap(err, "unable to add relocation mapping")
+			}
+			op.Files["/cnab/app/relocation-mapping.json"] = string(b)
+			var reloMap relocation.ImageRelocationMap
+			err = json.Unmarshal(b, &reloMap)
+			if mappedInvo, ok := reloMap[op.Image.Image]; ok {
```
vdice: Here we update the operation invocation image based on the value presented in the relo map, if exists, right?

pkg/cnab/provider/action.go
```diff
@@ -73,6 +73,10 @@ func (d *Runtime) AddRelocation(args ActionArguments) action.OperationConfigFunc
 			op.Files["/cnab/app/relocation-mapping.json"] = string(b)
 			var reloMap relocation.ImageRelocationMap
 			err = json.Unmarshal(b, &reloMap)
+			// If the invocation image is present in the relocation mapping, we need
+			// to update the operation and set the new iamge reference. Unfortunately,
```
vdice: `s/iamge/image` then good to go!  Thanks for the comment/description -- helped me :)

PR #722, opened 2019-10-11
Return relocation map for both push and pull

radu-matei:
This PR continues #639 and returns the relocation map for the push and pull operations.
It does not use the relocation yet - hence the draft status of the PR.

Up to you whether you think this should be implemented in this PR, or a follow-up.


jeremyrickard:
@radu-matei I'm doing this in #719

PR #727, opened 2019-10-14
Split release pipeline into stages

### LINKED ISSUE #673: CI: Parallelize (with stages) AZ release pipeline
Our release pipeline still currently runs in sequence.  See `build/azure-pipelines.release.yaml` , which is triggered by merges to master and tag pushes.

Now with support for [stages in Azure DevOps](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/stages?view=azure-devops&tabs=yaml), we can create a first stage for build/test targets - all to run in parallel - and a latter stage for publishing/releasing (which will, by default, only run if the first stage passes).  This should speed up the build a bit, similar to the big improvements in https://github.com/deislabs/porter/pull/648.

carolynvs:
# What does this change
Splits up the release pipeline into stages.

# What issue does it fix
Closes #673 

# Notes for the reviewer
🏎

# Checklist
- [ ] Unit Tests
- [x] Documentation

### DIFF
--- a/build/README.md
+++ b/build/README.md
@@ -4,7 +4,11 @@ These are assets used by the CI build.
 
 Our pipeline is broken into a few discrete builds so that we can control how and when they are triggered:
 
-* **azure-pipelines.release.yml**: Validates canary and tag releases.
+* **azure-pipelines.release.yml**: Validates canary and tag releases. This can be tested in a pull request 
+  using `/azp run porter-release` though steps that require credentials will fail.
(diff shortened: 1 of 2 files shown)


carolynvs:
You can see what the release build looks like here:

https://dev.azure.com/deislabs/porter/_build/results?buildId=3765

Once this is merged, then I can turn on being able to trigger it safely to test it in a PR.

PR #741, opened 2019-10-17
feat(publish): refresh local cache if newly published bundle exists

vdice:
# What does this change
Refreshes Porter's bundle cache with a newly published bundle, if it existed previously.

# What issue does it fix
Closes https://github.com/deislabs/porter/issues/678

# Notes for the reviewer
This is one approach to fixing https://github.com/deislabs/porter/issues/678.  Here we actually refresh the cache so that the next pull of the bundle with the same tag will use the refreshed cache.

Another approach would be to just remove the cached entry on publish.  Then, a subsequent pull would actually need to pull from remote.  The benefit here would be simplified logic, though again, would be forced to re-pull.

# Checklist
- [x] Unit Tests
- [ ] Documentation

### DIFF
--- a/pkg/porter/publish.go
+++ b/pkg/porter/publish.go
@@ -9,6 +9,7 @@ import (
 	"github.com/deislabs/cnab-go/bundle"
 	"github.com/deislabs/cnab-go/bundle/loader"
 	"github.com/deislabs/cnab-go/packager"
+	"github.com/docker/cnab-to-oci/relocation"
 	"github.com/docker/distribution/reference"
 	"github.com/pivotal/image-relocation/pkg/image"
 	"github.com/pivotal/image-relocation/pkg/registry"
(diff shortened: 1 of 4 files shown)


pkg/porter/publish.go
```diff
@@ -328,3 +343,14 @@ func (p *Porter) rewriteImageWithDigest(InvocationImage string, digest string) (
 	}
 	return fmt.Sprintf("%s@%s", named.Name(), digest), nil
 }
+
+// refreshCachedBundle will store a bundle anew, if a bundle with the same tag is found in the cache
+func (p *Porter) refreshCachedBundle(bun *bundle.Bundle, tag string, rm relocation.ImageRelocationMap) error {
+	if _, _, found, _ := p.Cache.FindBundle(tag); found {
+		_, _, err := p.Cache.StoreBundle(tag, bun, rm)
+		if err != nil {
```
carolynvs: i think this shouldn't be a fatal error? how about a warning?

vdice: Good call; agreed!  Will update.

vdice:
Had to update a test failure message to accurately reflect the test 😅 https://github.com/deislabs/porter/pull/741/commits/2d6a075772da88a9397649e940a065fc6fee3e67 -- otherwise the same as previous!

PR #742, opened 2019-10-17
feat(cmd): add --force flag to action commands

vdice:
# What does this change
This adds a `--force` flag to the standard action commands (install, upgrade, uninstall, --invoke) to enforce pulling a bundle (and all dependency bundles, if applicable).

A potential complement to https://github.com/deislabs/porter/pull/741

# What issue does it fix
N/A
# Notes for the reviewer

# Checklist
- [ ] Unit Tests
- [x] Documentation (auto-generated)

### DIFF
--- a/cmd/porter/bundle.go
+++ b/cmd/porter/bundle.go
@@ -103,6 +103,8 @@ For example, the 'debug' driver may be specified, which simply logs the info giv
 		"Use a bundle in an OCI registry specified by the given tag")
 	f.BoolVar(&opts.InsecureRegistry, "insecure-registry", false,
 		"Don't require TLS for the registry")
+	f.BoolVar(&opts.Force, "force", false,
+		"Force a fresh pull of the bundle and all dependencies")
 	return cmd
 }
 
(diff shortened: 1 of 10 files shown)


cmd/porter/bundle.go
```diff
@@ -103,6 +103,8 @@ For example, the 'debug' driver may be specified, which simply logs the info giv
 		"Use a bundle in an OCI registry specified by the given tag")
 	f.BoolVar(&opts.InsecureRegistry, "insecure-registry", false,
 		"Don't require TLS for the registry")
+	f.BoolVar(&opts.Force, "no-cache", false,
```
carolynvs: On the build command we were going to call this flag `--force`:

https://github.com/deislabs/porter/issues/690#issuecomment-538395784

How about we use the same flag for all our commands and go with `--force`? I liked that it's not exactly docker's command because it gives us room in the future to do more without people making assumptions that the flag will work the same way as docker's.

What do you think?

vdice: 👍 `--force` it is!

PR #755, opened 2019-10-18
chore(pkg/kubernetes): bump kubectl version for k8s mixin

vdice:
# What does this change
Bumps the kubectl version installed via the kubernetes mixin to the latest (v1.15.x) [v1.15.5 release](https://github.com/kubernetes/kubernetes/releases/tag/v1.15.5)

# What issue does it fix
N/A
# Notes for the reviewer
N/A
# Checklist
- [ ] Unit Tests
- [ ] Documentation

### DIFF
--- a/pkg/kubernetes/build.go
+++ b/pkg/kubernetes/build.go
@@ -4,7 +4,7 @@ import (
 	"fmt"
 )
 
-const kubeVersion = "v1.13.0"
+const kubeVersion = "v1.15.5"
 const dockerFileContents = `RUN apt-get update && \
 apt-get install -y apt-transport-https curl && \
 curl -o kubectl https://storage.googleapis.com/kubernetes-release/release/%s/bin/linux/amd64/kubectl && \


carolynvs:
I think this mixin is going to need the same magic that you added to the helm mixin. I just double checked that kubectl isn't backwards compatible indefinitely, just n-1/n+1.

https://kubernetes.io/docs/setup/release/version-skew-policy/#kubectl

So once we bump this, it will only with with 1.15 and 1.16 clusters. Knowings this, I think it may be better to bump to 1.15 of kubectl. As I understand it, that should support clusters versions 1.14-1.16 (theoretically). That should help us support 1.16 without losing support for as many older clusters until we can dynamically download the right version of the client.

What do you think?

carolynvs:
Also now that we can pass in values to mixins at build time, we can make this mixin smarter (and helm too). So that if they know the version of their cluster at build time, they can put it in the porter.yaml and we can build it in correctly the first time:

```yaml
mixins:
- kubernetes:
    version: 1.14.1
```

vdice:
Ah, great points @carolynvs .  I think restricting the bump to the latest 1.15.x version for this PR seems prudent, as you've pointed out.

For follow-ups, we have two potentials: dynamically installing a `kubectl` cli version based on the k8s cluster version and adding support for specifying the k8s version at build-time in manner mentioned in https://github.com/deislabs/porter/pull/755#issuecomment-545617318.

Do we want both follow-ups or put priority on the latter, perhaps negating the need for dynamic cli install?

carolynvs:
Sorry to attack your chore PR! All we need here is just 1.15. We can make the follow on issues later. 😀

MChorfa:
@vdice this issue is resolved

PR #760, opened 2019-10-21
Consider porter version and mixins version to generate digest

### LINKED ISSUE #572: porter hash doesn't take enough into account
When we stamp the bundle.json with a hash of the porter.yaml, that isn't enough to tell that we should rebuild the bundle because it is out out date. 

* One example is when the porter binary itself has since been updated, usually with bug fixes. In that case, the hash should be considered stale and the bundle rebuilt automatically. We should incorporate the porter version and the list of mixins used in the bundle, along with their version in the stamp.
* Another case would be when a mixin that is used by the bundle is updated.
* Another is when files in the current directory (or below) are changed, excluding those from `.dockerignore`

p.s. I was pressured into making the stamp under `sh.porter` and have always regretted it. Now would be a good time to fix that before it's too late! 😀

dev-drprasad:
# What does this change
Consider porter version and mixins version to generate digest

# What issue does it fix
Closes #572 

# Notes for the reviewer

# Checklist
- [ ] Unit Tests
- [ ] Documentation


dev-drprasad:
Hi @carolynvs @vdice, I imported `porter` package to `configadapter` to get porter version and mixins version. But i ended up with **circular dependencies problem**. 

* How can we resolve this issue  ?
* Is there any other way to get version info instead of my solution ?

vdice:
@dev-drprasad @carolynvs I wonder if we can instantiate a new MixinProvider, similar to how Porter does [here](https://github.com/deislabs/porter/blob/master/pkg/porter/porter.go#L45) and then list through this provider (so, only bringing in the `mixinprovider` pkg and thus avoiding the circular dep)?

carolynvs:
I like what @vdice is thinking. 👍  In `porter.buildBundle` we could have porter pass in its `MixinProvider` to `configadapter.NewManifestConverter`.

https://github.com/deislabs/porter/blob/c8b7b7a74a7429c18defd9516bb0392777bb16be/pkg/porter/build.go#L47-L52

carolynvs:
Oh actually, here's another idea on how to do this. Porter already has a function `PrintDebugInfo` that collects all the info we want to use. We can split that function into two functions `GetDebugInfo` and `PrintDebugInfo`

https://github.com/deislabs/porter/blob/bd404082eb41bc2fd836c15f70497c40ae120309/pkg/porter/version.go#L75

Then call `p.GetDebugInfo` and pass along the version and the mixin versions to `converter.ToBundle()` so that it can use some form of the data returned by that new function for the stamp. That way we aren't rebuilding the info twice in separate places.

What do you think?

dev-drprasad:
@carolynvs I am not sure about passing args to `ToBundle`. As it name suggest `ToBundle` function purpose suppose to only generate bundle in my opinion. Why don't we pass  output of `p.GetDebugInfo()` to `configadapter.NewManifestConverter`

Its your call as i don't have valid arguments to tell which is better.

carolynvs:
@dev-drprasad Feel free to pass parameters, or move some parameters from the constructor to other functions, as it makes sense as you work on the feature. 

I agree that the way the code is now, if you want to match the existing pattern would be to continue to put it all into the constructor. Another alternative would be to only pass the context, and debug info to the constructor, and pass the manifest and image digests to ToBundle. IMO it's a matter of preference which way to go, and you would be the one writing the code. 😀 

I'll leave it up to you to decide which you think is a better fit.

carolynvs:
@dev-drprasad Let me know if you are stuck on that build error. You should be fine to just update the digest in the test to match the new value to get the test to pass.

pkg/porter/build.go
```diff
@@ -46,7 +47,24 @@ func (p *Porter) Build(opts BuildOptions) error {
...
+
+	if err != nil {
+		return errors.Wrapf(err, "error while listing mixins")
+	}
+
+	mixins := []mixin.Metadata{}
+	for _, installedMixin := range installedMixins {
+		for _, m := range p.Manifest.Mixins {
+			if installedMixin.Name == m.Name {
+				fmt.Printf("%s", m.Name)
```
carolynvs: This debug line should be removed before we merge the PR.

pkg/porter/stamp.go
```diff
@@ -45,7 +46,23 @@ func (p *Porter) IsBundleUpToDate(opts bundleFileOptions) (bool, error) {
 			return false, errors.Wrapf(err, "could not load stamp from %s", opts.CNABFile)
 		}
 
-		converter := configadapter.NewManifestConverter(p.Context, p.Manifest, nil)
+		installedMixins, err := p.ListMixins()
```
carolynvs: This logic here would be good to move into a new function on `p`, such as `getUsedMixins`, so that we don't get drift between these two places where we use it (stamp and buildBundle) and would let us write a unit test for it.

pkg/porter/stamp.go
```diff
@@ -45,7 +46,23 @@ func (p *Porter) IsBundleUpToDate(opts bundleFileOptions) (bool, error) {
...
+
+		if err != nil {
+			return false, errors.Wrapf(err, "error while listing mixins")
+		}
+
+		mixins := []mixin.Metadata{}
+		for _, installedMixin := range installedMixins {
+			for _, m := range p.Manifest.Mixins {
+				if installedMixin.Name == m.Name {
+					fmt.Printf("%s", m.Name)
```
carolynvs: Debug line needs to go.

pkg/porter/stamp.go
dev-drprasad: Makes sense. I will move to separate function

dev-drprasad:
Hi @carolynvs,
How do we know files are changed in current directory ?
Do we need to calculate hash for content of all files  ?
If so, how do we exclude files in `.dockerignore` ?

My suggestion is steal some code from docker to do that
https://github.com/docker/docker-ce/blob/99ccbc45b354de39413ec001369f352a9f0bb090/components/cli/cli/command/image/build/context.go#L39

carolynvs:
@dev-drprasad This PR just needs to take into account the version and the mixins. If we want to get more advanced, I think your idea about using the same solution as Docker is a good one (since we build using Docker 😉 ) but it should go in a separate PR.

dev-drprasad:
@carolynvs I resolved merge conflicts

carolynvs:
Excellent! I've got my finger hovering over the big green button as soon as the builds go green. 👀

PR #776, opened 2019-10-25
Pass plugin config via stdin

carolynvs:
# What does this change
Reads optional configuration from the porter configuration file and passes it to the plugin on stdin as json. 

It is not sent as part of the plugin interface/contract because each plugin may or may not have config, it may look completely different or be loaded from other places, and I'd like to have the plugin contract mimic the cnab contracts. For example the interface for claims is the CrudStore.

```toml
default-instance-store = "porterci"

[[instance-store]]
  name = "porterci"
  plugin = "azure.blob"

  [instance-store.config]
    env = "PORTERCI_AZURE_STORAGE_CONNECTION_STRING"
```

# What issue does it fix
N/A

# Notes for the reviewer
This relies on a patch: https://github.com/hashicorp/go-plugin/pull/127

This is the first step in getting a general framework in place for plugins. Next steps after this will be to make the pluginProvider generic so it can load a plugin and config for any interface. Right now I'm just feeling out what works.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -786,15 +786,16 @@
   version = "v1.0.0"
 
 [[projects]]
-  digest = "1:446e0a4f73191887866ac789d1fd210e3408bb8eb8aeaf199d5f3a73ff63c108"
+  branch = "accept-stdin"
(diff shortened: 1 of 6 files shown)


carolynvs:
I was going to hold off until someone can install a plugin without having to compile it themselves.

PR #779, opened 2019-10-28
Fix whitespace in the copy help text

jeremyrickard:
Signed-off-by: Jeremy Rickard <jeremy.rickard@microsoft.com>

# What does this change

Fixes the whitespace in the long desc for the copy command. It was indented too much and looked bad during a demo last week.


jeremyrickard:
Doh, sorry I forgot that one!

PR #782, opened 2019-10-30
Clean some overridden dependencies

radu-matei:
This might also be needed - https://github.com/deislabs/duffle/pull/875/commits/e1ad5edfb7e6ad98c05e1f0169eca07713f891f0

And the override for `hcsshim` is also needed (this is why the build fails on Windows):

```
[[override]]
  name = "github.com/Microsoft/hcsshim"
  revision = "2226e083fc390003ae5aa8325c3c92789afa0e7a"
```


carolynvs:
Hey @radu-matei Can you redo this without bumping the version on all of our dependencies? 

1. Undo the changes to Gopkg.lock and vendor
2. Run `dep ensure` without any flags, so that it keeps our current versions and only does any updates necessary based on the constraints specified in Porter's Gopkg.toml and cnab-go's updated Gopkg.toml
3. Check in the updated Gopkg.lock and vendor.

I'm not saying that will fix the build. However I don't want to change our versions for everything and do want to limit the scope of this PR to just a single change.

carolynvs:
@radu-matei Let met know when you don't need this anymore and we can close it.

carolynvs:
With the move to go mods looming, I'm going to close this out.

PR #796, opened 2019-11-07
check return value of WriteMixinOutputToFile

JiangZhaoYang:


vdice:
Thank you for the PR @JiangZhaoYang !

PR #797, opened 2019-11-09
Add Kubernetes kustomize support

JiangZhaoYang:
## What does this change
Kubernetes mixin add kustomize support，so we can use porter in CI/CD or other various scenes.

## What issue does it fix
N/A

## Notes for the reviewer
The command line of kubectl add -k flag, and the mixin definition can use k8s manifest and kustomize both for deploy complex environment

### DIFF
--- a/examples/kubernetes-kustomize-example/README.md
+++ b/examples/kubernetes-kustomize-example/README.md
@@ -1,0 +1,10 @@
+# Kubernetes + Kustomize Example
+
+Sample for Porter bundle that makes use of Kubernetes Kustomize.
+
+```
+porter build
+porter credentials generate
+porter install -c PORTER_KUBE
+porter uninstall -c PORTER_KUBE
+```
--- a/examples/kubernetes-kustomize-example/kustomize/base/configMap.yaml
+++ b/examples/kubernetes-kustomize-example/kustomize/base/configMap.yaml
@@ -1,0 +1,7 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: the-map
+data:
+  altGreeting: "Good Morning!"
+  enableRisky: "false"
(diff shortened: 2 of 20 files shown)


JiangZhaoYang:
Since kubernetes mixin kubctl already up to the latest (v1.15.x) version, and kubctl version > 1.14.x support kustomize, so i add it. I think it can provide more scenes to be used for Porter.

vdice:
Thank you for the PR @JiangZhaoYang ! 

Before reviewing and testing out the logic in this PR, I wanted to mention that we'd prefer the Kustomize logic be contained in its own mixin.  We'd been meaning to rename the pre-existing Kubernetes mixin to `kubectl` to make clearer its intended scope.  With this in mind, having a separate `kustomize` mixin focused on the corresponding cli would be preferable.

Coincidentally, it looks like there is a pre-existing Kustomize mixin in https://github.com/donmstewart/porter-kustomize.  I wonder if it makes sense to see if we can build off of this mixin first (perhaps add some examples, add changes, etc.)?  WDYT?

Also, perhaps for future reference (if you hadn't already seen it), new Porter mixin repo creation and distribution steps exist in https://github.com/deislabs/porter-skeletor.  This should be the fastest track towards setting up a new mixin repo and using it in Porter bundles.

JiangZhaoYang:
Oops, I did not found kustomize mixin in this repo, so I add it. Thanks for detailed reply. I think whether to create a separate mixin for Kustomize depends on whether Kubectl supports it well. At this time, Kubectl is not fully compatible with Kustomize, so use separate Kustomize mixin maybe a good choice. I will check https://github.com/donmstewart/porter-kustomize to see if i can do something

vdice:
Great, thank you @JiangZhaoYang .  Indeed, we need to think about how to enable mixin discovery for those that are developed under organizations outside of `deislabs` -- otherwise they can/will be easily missed.

carolynvs:
Since it sounds like we've decided to have this in a separate repository I am going to close this PR. If you need help finding a home for it, just let us know!

PR #804, opened 2019-11-11
WIP: Tweak instance storage plugin naming

carolynvs:
# What does this change
Tweak naming used by the instance storage plugin before we move forward with it.

* Interface to communicate with the plugin <-> host: storage. This is a name that would be suitable for using in cnab-go.
* Package name: storage. Previously I had called it "instance-storage" but a) hyphens are hard in package names and b) once I started on credentials I realized I didn't need to distinguish between the two. Claim storage could just use storage.
* Configure default plugin for instance storage: storage-plugin.

Looking at credentials the names will be:

* plugin interface: credentials
* package: credentials
* config: credentials-plugin

So this will help us be consistent from the start before people use the plugins.

# What issue does it fix
N/A

# Notes for the reviewer
Once this is in, I'll tag a release and then reference it, and use it in porter-azure-plugins.

# Checklist
- [ ] Unit Tests
- [ ] Documentation


carolynvs:
I am thinking that I'm going to change how storage is used so that it's not just claims storage but storing other stuff from porter home too. So this may need more changes... Putting back into DRAFT.

carolynvs:
Closing in favor of #859

Issue #809, opened 2019-11-20
Document which registries CNAB works with


carolynvs:
* Docker Hub 👍 * but doesn't show tags for bundles
* ACR 👍 
* Quay 🛑
* Digital Ocean ❓ 
 
**quay** 
```
$ porter publish
Pushing CNAB invocation image...
The push refers to repository [quay.io/porter/hello-installer]
09f69eaf9bf9: Preparing
1ea4b434c5b0: Preparing
39421e9c894c: Preparing
13d2047f5c7d: Preparing
badfbcebf7f8: Preparing
09f69eaf9bf9: Pushed
13d2047f5c7d: Pushed
39421e9c894c: Pushed
1ea4b434c5b0: Pushed
badfbcebf7f8: Pushed
[DEPRECATION NOTICE] registry v2 schema1 support will be removed in an upcoming release. Please contact admins of the quay.io registry NOW to avoid future disruption.
0.1.0: digest: sha256:7e54b5c176def3890714bcc8523185466d8ae32e8d13feae797c1a74ea5e5f96 size: 5454

Rewriting CNAB bundle.json...
Starting to copy image quay.io/porter/hello-installer:0.1.0...
Failed to copy image quay.io/porter/hello-installer:0.1.0: failed commit on ref "unknown-sha256:7e54b5c176def3890714bcc8523185466d8ae32e8d13feae797c1a74ea5e5f96": unexpected status: 400 Bad Request
Error: failed commit on ref "unknown-sha256:7e54b5c176def3890714bcc8523185466d8ae32e8d13feae797c1a74ea5e5f96": unexpected status: 400 Bad Request
```

rumpl:
To complete your list because we have been testing different registries with docker app:

* Nexus 👎 - doesn't handle indexes
* Artifactory 👎 - has problems with the containerd client
* GCR 👍 
* ECR 👎

carolynvs:
This is super helpful thank you!

squillace:
It sure appears that you have done what this issue requested. Close?

carolynvs:
So say we all

PR #811, opened 2019-11-20
List which registries are compatible

carolynvs:
# What does this change
Document which registries are compatible with CNAB and which ones have been explicitly tested with porter.

Preview at https://deploy-preview-811--porter.netlify.com/compatible-registries/

# What issue does it fix
Part of #809 

# Notes for the reviewer
This is a bit of an ongoing issue that we will need to keep working on. I've documented what we know so far, but we should test porter with the others and figure out why it doesn't work with some, like quay, and make sure it's not something we can change on our end.

# Checklist
- [ ] Unit Tests
- [x] Documentation

### DIFF
--- a/docs/content/compatible-registries.md
+++ b/docs/content/compatible-registries.md
@@ -9,7 +9,7 @@ tested a bunch of registries for compatibility with CNAB and so far here is what
 we have found.
 
 There is an explicit verification using Porter because we use specific libraries,
-such as cnab-to-oci, and this helps us communicate confidently that we've tested
+such as [cnab-to-oci], and this helps us communicate confidently that we've tested
 out a particular registry and know that it will work for you.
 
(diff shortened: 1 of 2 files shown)


docs/content/compatible-registries.md
```diff
@@ -0,0 +1,28 @@
...
+description: Understanding which OCI registries work with CNAB 
+---
+
+Cloud Native Application Bundles are very new, and support for storing anything
+other than container images in a registry is inconsistent. The community has
+tested a bunch of registries for compatibility with CNAB and so far here is what
+we have found.
+
+There is an explicit verification using Porter because we use specific libraries,
+such as cnab-to-oci, and this helps us communicate confidently that we've tested
```
vdice: Do we want to link to https://github.com/docker/cnab-to-oci here?

PR #826, opened 2019-12-06
Deprecate the azure mixin for az+arm

carolynvs:
# What does this change
Keep the old page around just to tell people to use az+arm instead. I've added a placeholder page for the arm mixin. Once I have examples for the arm mixin, we can update it to match the other pages.

# What issue does it fix
None, just general housekeeping.

# Notes for the reviewer
I like lucky charms. ☘️

# Checklist
- [ ] Unit Tests
- [x] Documentation


carolynvs:
I kicked the build and it really passed. It's just showing up twice (once failed and once passing). So merging with my super powers. 💪

PR #827, opened 2019-12-09
fix(pkg/cnab): be sure output path uses unix conventions

vdice:
# What does this change
Set the output path using Unix path conventions.

# What issue does it fix
Closes https://github.com/deislabs/porter/issues/761

# Notes for the reviewer
I'm still trying to get a Windows VM set up for testing.  If a Windows user is able to test, this would be very helpful!

# Checklist
- [ ] Unit Tests
- [ ] Documentation

### DIFF
--- a/pkg/cnab/config-adapter/adapter.go
+++ b/pkg/cnab/config-adapter/adapter.go
@@ -2,7 +2,7 @@ package configadapter
 
 import (
 	"fmt"
-	"path/filepath"
+	unix_path "path"
 	"strings"
 
 	"get.porter.sh/porter/pkg/cnab/extensions"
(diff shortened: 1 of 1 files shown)


pkg/cnab/config-adapter/adapter.go
```diff
@@ -2,7 +2,7 @@ package configadapter
 
 import (
 	"fmt"
-	"path/filepath"
+	unix_path "path"
```
carolynvs: lol not sure we need to name the package unix_path 😉 Maybe just put a comment where we use it that we have to use unix paths because the path is used inside the bundle?

PR #832, opened 2019-12-16
fix(outputs.go): set displayOutput.Value

vdice:
# What does this change
Sets the `Value` field of the `DisplayOutput` object.

Previously, the displayed output in json or yaml would show:
```
 $ p instance outputs list basic-tf-example -o yaml
- name: a
  definition:
    type: string
  value: null
  displayvalue: |
    a
  type: string
```

Whereas now it will show:
```
 $ p instance outputs list basic-tf-example -o yaml
- name: a
  definition:
    type: string
  value: |
    a
  displayvalue: |
    a
  type: string
```

# What issue does it fix
Closes # _(issue)_

# Notes for the reviewer

# Checklist
- [x] Unit Tests
- [ ] Documentation

### DIFF
--- a/pkg/porter/outputs.go
+++ b/pkg/porter/outputs.go
@@ -83,16 +83,15 @@ func (p *Porter) ShowBundleOutput(opts *OutputShowOptions) error {
 }
 
 type DisplayOutput struct {
-	Name         string
-	Definition   definition.Schema
-	Value        interface{}
-	DisplayValue string
-	Type         string
+	Name       string
+	Definition definition.Schema
+	Value      string
+	Type       string
 }
 
 // ListBundleOutputs lists the outputs for a given bundle,
-// according to the provided options
-func (p *Porter) ListBundleOutputs(c claim.Claim) []DisplayOutput {
+// according to the provided claim and display format
+func (p *Porter) ListBundleOutputs(c claim.Claim, format printer.Format) []DisplayOutput {
 	// Get sorted keys for ordered printing
 	keys := make([]string, 0, len(c.Outputs))
 	for k := range c.Outputs {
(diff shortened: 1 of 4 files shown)


vdice:
> I'm not really sure what motivated this PR?

Oops, it was kinda subtle.  I noticed that `value` was always `null` for all display outputs when listing `-o yaml|json`, regardless of output type.  e.g.
```
 $ p instance outputs list -o yaml
- name: a
  definition:
    type: integer
  value: null
  displayvalue: |
    900
  type: integer
```
etc.

Indeed, if the type is string _and_ the value is less than 60 chars, they will be equivalent (if greater, the displayValue would be truncated).  Nevertheless, good call on adding tests for other types - will do!

vdice:
@carolynvs before going further here, I wonder if we should just consolidate to one value field for the [DisplayOutput](https://github.com/deislabs/porter/blob/master/pkg/porter/outputs.go#L85-L91) struct?  As mentioned, the original issue was confusion that the `Value` field was always `null`, hence setting it in this PR.  However, it's arguably redundant to have both `Value` and `DisplayValue` on this struct.  I wonder if we should just reduce to `Value` and it being the display value (truncated, perhaps, from the actual/original value), since the struct itself is only intended for display.  WDYT?

carolynvs:
Oh it doesn't even look like `Value` is being used at all! Yeah I like your idea, let's just have `Value` and set it to the truncated value and consolidate the two fields. 👍

vdice:
Thanks @carolynvs .  Okay, I pushed up a new commit to reduce to just `Value` on the `DisplayOutput` but modified the `ListBundleOutputs` function signature to now take in a `printer.Format`, for determining whether or not to truncate (for nice table view; otherwise no truncation).  What do you think?

vdice:
@carolynvs PTAL when convenient

PR #837, opened 2019-12-17
Standardize default and template dockerfiles

carolynvs:
# What does this change
This standardizes the contents of the default and template Dockerfile generated by Porter. It also has Porter treat both the same, instead of having conditional logic based on if the user has a template or not.

# What issue does it fix
None, I just noticed in another PR, https://github.com/vdice/porter-packer/pull/1/files/c9b3d6ef73a4212ae2fbfb5f267a6e831cbb2885#diff-39844b97085b014640182ce3ec1e1282, that we were still using an old base image that I had thought that I had removed. Turns out the default Dockerfile and the template are using different base images. It's confusing and I wanted to make it easier for us and users to understand what Porter is doing.

# Notes for the reviewer
We don't have to merge this, it's just a suggestion to think about.

# Checklist
- [x] Unit Tests
- [ ] Documentation

### DIFF
--- a/pkg/build/dockerfile-generater.go
+++ b/pkg/build/dockerfile-generater.go
@@ -151,8 +151,9 @@ func (g *DockerfileGenerator) buildPorterSection() []string {
 
 func (g *DockerfileGenerator) buildCNABSection() []string {
 	return []string{
-		`COPY .cnab /cnab`,
+		// Putting RUN before COPY here as a workaround for https://github.com/moby/moby/issues/37965, back to back COPY statements in the same directory (e.g. /cnab) _may_ result in an error from Docker depending on unpredictable factors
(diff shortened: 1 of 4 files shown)


carolynvs:
I just tried re-running and it failed again. Let me look at a bit more today and see if I can figure out why.

carolynvs:
This only reproduces on azure, which also has non-standard `docker version` numbers output. But it does seem similar to https://github.com/moby/moby/issues/37965. I just put in the workaround suggested there and it fixed the build.

carolynvs:
Integration is failing because it is failing on master...

carolynvs:
Ah okay the integration test intermittently fails when it goes over 20m, that's what happened on master. If I see that again I'll bump the timeout.

PR #839, opened 2019-12-19
fix(*): fixes case where error occurs when parameter not specified

### LINKED ISSUE #835: Parameters that only apply to some actions can trigger errors when not specified
When a parameter has `applyTo` set, and no `default`, it can cause errors when it isn't specified:

```
Error: invalid parameters: cannot use value: <nil> as parameter config: type should be string
```

For example a parameter may only apply to upgrade, so the user rightly doesn't set the parameter during install. But that triggers this error.

We should either default the param or find a way to avoid this error from being triggered in the first place.

vdice:
# What does this change
 - Adds logic to supply appropriate zero values for parameters in cases where they are required (e.g., have no declared default) but not applicable to the current action.  These zero values are to appease the json schema validation done on the parameter value, which is the culprit for the error reported in #835 

- Removes the check to see if there are more than zero user-supplied parameter overrides before loading and validating parameters for the `uninstall`, `upgrade` and `invoke` actions.  With these checks removed (meaning, we'll always load/validate parameters) we will ensure errors are thrown in appropriate cases (for instance, a parameter is required to run `uninstall`, but is not supplied by the user and has no default or value in a pre-existing claim -- in this case, validation will bubble up an error, whereas it did not prior.)

# What issue does it fix
Closes #835 

# Notes for the reviewer

I'd love some help/suggestions on better test matrix tools/setup for the added tests.  The matrix of test cases is quite large and most likely there are better ways to represent them than what I have here.  Technically, as I review the tests, an additional dimension of whether or not the parameter value exists in a pre-existing claim could/should be added 😬 .

Additionally, do the test case expectations look correct for each scenario?

# Checklist
- [x] Unit Tests
- [ ] Documentation


pkg/cnab/provider/parameters.go
```diff
@@ -37,31 +37,39 @@ func (d *Runtime) loadParameters(claim *claim.Claim, rawOverrides map[string]str
...
+	// rawOverrides (meaning, user-supplied overrides at time of action invocation)
+	// may supply no entry for a parameter designated as required *but* that does not apply to this action.
+	//
+	// When this occurs, we set an override to either the current value in the claim or the default value.
+	// If neither exists, the zero value according to the parameter type will be used.
+	// Otherwise, if unset/nil, json validation in bundle.ValuesOrDefaults would return an error
 	for name, param := range bun.Parameters {
+		def, ok := bun.Definitions[param.Definition]
+		if !ok {
+			return nil, fmt.Errorf("definition %s not defined in bundle", param.Definition)
```
carolynvs: nit: this error message is probably going to pop up in the middle of an action without enough context to understand that we are referring to a parameter's definition. I suggest saying "parameter definition %s..."

pkg/cnab/provider/parameters.go
```diff
@@ -83,3 +91,22 @@ func (d *Runtime) getUnconvertedValueFromRaw(def *definition.Schema, key, rawVal
 	}
 	return rawValue, nil
 }
+
+// getZeroValue returns the zero value for a parameter according to its type
+func getZeroValue(name string, def *definition.Schema, claim *claim.Claim) interface{} {
```
carolynvs: looks like the claim parameter isn't used?

pkg/cnab/provider/parameters.go
```diff
@@ -83,3 +91,22 @@ func (d *Runtime) getUnconvertedValueFromRaw(def *definition.Schema, key, rawVal
...
+	switch def.Type {
+	case "integer", "number":
```
carolynvs: It doesn't really matter but I don't think `number` is possible, because of how cnab uses canonical json, right?

pkg/cnab/provider/parameters_test.go
```diff
@@ -230,3 +282,181 @@ func Test_loadParameters_fileParameter(t *testing.T) {
...
+			testcases := []struct {
+				name            string
+				required        bool
+				provided        bool
+				defaultExists   bool
+				appliesToAction bool
+				expectedVal     interface{}
+				expectedErr     string
+			}{
+				// Are you ready to enter the Matrix?
```
carolynvs: 😂

pkg/cnab/provider/parameters_test.go
```diff
@@ -230,3 +282,181 @@ func Test_loadParameters_fileParameter(t *testing.T) {
...
+						err = d.Upgrade(args)
+					case "invoke":
+						err = d.Invoke("zombies", args)
+					case "uninstall":
+						err = d.Uninstall(args)
+					}
+
+					if tc.expectedErr != "" {
+						require.EqualError(t, err, tc.expectedErr)
+					} else if action != "uninstall" {
```
carolynvs: I don't understand why this `if action != "uninstall"` is here instead of only a bare `else`. Where are the asserts happening for the uninstall case? We should at least check that uninstall didn't return an error right? Even if the claim was deleted?

pkg/cnab/provider/parameters.go
vdice: Yea, I had the same thinking but wasn't sure on whether to support it here (just in case) or limit only to `integer`... could be convinced either way.  WDYT?

pkg/cnab/provider/parameters.go
vdice: Ah, thank you; indeed.

pkg/cnab/provider/parameters_test.go
vdice: 🕶

pkg/cnab/provider/parameters_test.go
vdice: Ah, indeed, it currently has that gap of checking the uninstall case (that no err occurred when not expected).  Thank you; updating!

pkg/cnab/provider/parameters.go
carolynvs: How about keep it but add a comment saying it is just in case?

PR #840, opened 2019-12-19
feat(*): add logic to resolve/interpolate output values in a Porter manifest

vdice:
# What does this change
Note: ~~**Do not merge yet**~~ - ~~Blocked by~~/Depends on https://github.com/cnabio/cnab-spec/pull/300

- Adds logic to enable using bundle output values in a Porter manifest via the supported templating language
- As an example, `{{ bundle.outputs.foo }}` can now be used in a step for action `next`, which will pull `foo`'s value from the claim, assuming the action that produced the current claim (say, `install`) produced an output of `foo`.

One caveat here is that currently, until work is done to address https://github.com/deislabs/porter/issues/836, interpolation of these values will only be possible if the previous action supplied the output needed.  Otherwise, outputs don't yet span more than one action _if the actions spanned don't also produce the same output_.

# What issue does it fix
Closes https://github.com/deislabs/porter/issues/834 

# Notes for the reviewer

# Checklist
- [x] Unit Tests
- [ ] Documentation

### DIFF
--- a/pkg/runtime/runtime-manifest.go
+++ b/pkg/runtime/runtime-manifest.go
@@ -7,10 +7,12 @@ import (
 	"reflect"
 	"strings"
 
+	"get.porter.sh/porter/pkg/config"
 	"get.porter.sh/porter/pkg/context"
 	"get.porter.sh/porter/pkg/manifest"
 	"github.com/cbroglie/mustache"
 	"github.com/cnabio/cnab-go/bundle"
+	"github.com/cnabio/cnab-go/claim"
 	"github.com/docker/cnab-to-oci/relocation"
 	"github.com/docker/distribution/reference"
 	"github.com/pkg/errors"
@@ -98,6 +100,33 @@ func resolveCredential(cd manifest.CredentialDefinition) (string, error) {
 	} else {
 		return "", fmt.Errorf("credential: %s is malformed", cd.Name)
 	}
+}
+
+// resolveOutput will attempt to resolve an output value if required by string interpolation
+// e.g. via {{ bundle.outputs.foo }}
+// It looks up the claim for the value written by the last action and, if exists, returns this value
+func (m *RuntimeManifest) resolveOutput(output manifest.OutputDefinition) (string, error) {
+	if m.Action != manifest.ActionInstall {
+		claimBytes, err := m.FileSystem.ReadFile(config.ClaimFilepath)
+		if err != nil {
+			return "", errors.Wrap(err, "unable to read claim")
+		}
+
+		claim := &claim.Claim{}
+		err = yaml.Unmarshal(claimBytes, claim)
+		if err != nil {
+			return "", errors.Wrap(err, "unable to unmarshal claim")
+		}
+
+		if err == nil {
+			if val := claim.Outputs[output.Name]; val != nil {
+				return fmt.Sprintf("%v", val), nil
+			}
+		}
+	}
+
+	// Action is Install, so no claim is expected to exist yet
+	return "", nil
 }
 
 func (m *RuntimeManifest) GetSensitiveValues() []string {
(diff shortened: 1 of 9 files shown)


carolynvs:
I haven't reviewed this yet but one thing that is important to note about this change is that it creates bundles that can only be run by Porter because it relies upon the client injecting claim data into the runtime at a specific point `/cnab/claim.json` which isn't part of the spec yet. I don't think that has changed yet upstream, right?

What is our thoughts/policy for changes like that? Is it okay as a prototype that we then use to get an upstream spec change in? But we make sure to document it as experimental which may be removed once the spec is finalized? Or do we want to avoid doing this like this and always do spec first so that we don't make "porter only bundles" even for experimental stuff? We kind of have this as well with dependencies, though I think the other clients respect the spec around required custom extensions, and refuse to try to run those bundles that need dependency support.

cc @jeremyrickard

vdice:
Good point @carolynvs .  While mounting the claim to `/cnab/claim.json` is proposed in the spec in https://github.com/cnabio/cnab-spec/pull/300, it hasn't yet gained consensus and/or merged.  For this particular case, should we just consider this PR blocked until 300 is merged?

carolynvs:
My vote is to follow/block on the spec and only make things that are porter only (like dependencies) when we are really prototyping something new and easy to warn people is experimental. This is so subtly different that no one would expect it to not work with other tools.

vdice:
@carolynvs now that https://github.com/cnabio/cnab-spec/pull/300 is merged, this should be ready for review.

pkg/runtime/runtime-manifest.go
```diff
@@ -100,6 +102,33 @@ func resolveCredential(cd manifest.CredentialDefinition) (string, error) {
 	}
 }
 
+// resolveOutput will attempt to resolve an output value if required by string interpolation
+// e.g. via {{ bundle.outputs.foo }}
+// It looks up the claim for the value written by the last action and, if exists, returns this value
+func (m *RuntimeManifest) resolveOutput(output manifest.OutputDefinition) (string, error) {
+	if m.Action != manifest.ActionInstall {
```
carolynvs: I believe we have upcoming work to support running stateless custom actions before install is executes, i.e. without a claim. I suggest changing this to check if the claim is present, and if not, skip without relying on the action.

vdice: Good call.  Will update to check if the claim is present, as suggested.

vdice:
> I was expecting to see resolveOutputs first look in the step outputs (bun.outputs) and then fallback to the output value from the claim. Did I miss this part in the code?

Indeed, I'm adding this now.   Working on integration test coverage for these two cases (step-level and bundle-level outputs).

vdice:
Ok @carolynvs , this is back and ready for re-review when convenient.  Thanks!

pkg/runtime/runtime-manifest.go
```diff
@@ -100,13 +102,51 @@ func resolveCredential(cd manifest.CredentialDefinition) (string, error) {
...
+// resolveBundleOutput will attempt to resolve a bundle output value if required
+// by string interpolation, e.g. via {{ bundle.outputs.foo }}
+// It first checks the outputs field on the RuntimeManifest struct and, if non-existent,
+// attempts to pull a value from the claim, if exists
+func (m *RuntimeManifest) resolveBundleOutput(output manifest.OutputDefinition) (string, error) {
+	if m.outputs[output.Name] != "" {
+		return m.outputs[output.Name], nil
+	}
+
+	if claimBytes, err := m.FileSystem.ReadFile(config.ClaimFilepath); err == nil {
```
carolynvs: nit: How about we make a new function `loadClaim` and then run it from `Validate` (which is called once before any steps are run). Then we don't reparse the claim for each output reference.

pkg/runtime/runtime-manifest.go
```diff
@@ -202,13 +239,31 @@ func (m *RuntimeManifest) buildSourceData() (map[string]interface{}, error) {
...
+		m.setSensitiveValue(val)
 		creds[pe] = val
 	}
 
 	bun["outputs"] = m.outputs
-	for _, output := range m.outputs {
-		m.sensitiveValues = append(m.sensitiveValues, output)
+	// Iterate through the runtime manifest's step outputs and mask by default
+	for _, stepOutput := range m.outputs {
+		// TODO: check if output declared sensitive or not; currently we default always sensitive
```
carolynvs: Do we need to create a follow-up issue to track this todo?

vdice: Good call; I created https://github.com/deislabs/porter/issues/855 and added a link below this comment.

pkg/runtime/runtime-manifest.go
vdice: Definitely, good thinking.  This logic has been revised appropriately.

PR #843, opened 2020-01-03
fix(archive.go): update mode on archive file

### LINKED ISSUE #838: Archives are created as executable files
## Describe the bug
The gzipped tarballs created by `porter bundle` are created as executable

## To Reproduce
1. Create a new bundle via `porter create`
2. Run `porter archive hello.tar.gz`
3. Run `ls -al` and observe that `hello.tar.gz` has `rwxr-x-r-x`

## Expected behavior
Archives are not marked as executable

## Version

porter v0.21.1-beta.1 (3c4c078)

## Notes

I believe this is because the temp folder is created with 755 and the final tarball gets the permissions of the root folder

https://github.com/deislabs/porter/blob/master/pkg/porter/archive.go#L96

vdice:
# What does this change
Sets file mode appropriately for archive file and parent directory.  (Neither should be executable.)

# What issue does it fix
Closes #838 

# Notes for the reviewer
N/A

# Checklist
- [x] Unit Tests
- [ ] Documentation

### DIFF
--- a/tests/archive_test.go
+++ b/tests/archive_test.go
@@ -3,6 +3,7 @@
 package tests
 
 import (
+	"os"
 	"path/filepath"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


tests/archive_test.go
```diff
@@ -38,6 +39,10 @@ func TestArchive(t *testing.T) {
 	err = p.Archive(archiveOpts)
 	require.NoError(p.T(), err, "archival of bundle failed")
 
+	info, err := p.FileSystem.Stat("mybuns.tgz")
```
carolynvs: ❤️

PR #845, opened 2020-01-03
Switch from dep to go modules

carolynvs:
# What does this change
Now that cnab-go has switched to go modules, we have to as well.

- [x] Changed from dep to modules.
- [x] Use the new versions referenced/vetted by cnab-go.
- [x] Updated our Makefile to set `GO111MODULE=on` before any `go` command.
- [x] Removed the vendor directory.
- [x] Fix broken unit tests
- [x] Fix broken integration tests

# What issue does it fix
I don't think we had an issue to track this. We just knew it was coming. Like jaws. _du duh_

# Notes for the reviewer
* ~I did not update us from 1.12 to 1.13 at the same time to minimize changes in a single PR. If we want to switch to 1.13, let's make a new issue for that.~ Scratch that we have to move to 1.13
* As a follow-up I'd like us to setup an Athens proxy and use it for our main cache of modules, especially before we move to 1.13 which starts going through the default google proxy.
* This impacts all of our mixins and skeletor. So once this is complete, those are next.

# Checklist
- [ ] Unit Tests
- [ ] Documentation


jeremyrickard:
🖖 dep

carolynvs:
This has been a journey of really bad error messages and pain @jeremyrickard. FEEL MY PAIN 😂

carolynvs:
```
go: get.porter.sh/porter/pkg/porter tested by
	get.porter.sh/porter/pkg/porter.test imports
	github.com/deislabs/cnab-go/bundle: github.com/deislabs/cnab-go@v0.8.0-beta1: parsing go.mod:
	module declares its path as: github.com/cnabio/cnab-go
	        but was required as: github.com/deislabs/cnab-go
```

This is killing me, I updated all of our code to use the new import so where is this coming from?

carolynvs:
I fixed the regression in cnab-go and that fixed the integration tests, https://github.com/cnabio/cnab-go/pull/171

PR #853, opened 2020-01-13
Doc that custom dockerfiles must have apt

### LINKED ISSUE #844: kubernetes mixin: allow portability across different base images
During an invocation image build, if the base image specified in `Dockerfile.tmpl` does not support `apt-get` the build process fails. In my case I'm looking to use an alpine image with Golang like [this](https://github.com/davidmontoyago/cnab-grpc-hello-world-svc/blob/master/Dockerfile.tmpl#L1).

```
Error: unable to build CNAB invocation image: failed to stream docker build output: The command '/bin/sh -c apt-get update && apt-get install -y apt-transport-https curl && curl -o kubectl https://storage.googleapis.com/kubernetes-release/release/v1.15.5/bin/linux/amd64/kubectl && mv kubectl /usr/local/bin && chmod a+x /usr/local/bin/kubectl' returned a non-zero code: 127
```

WRT to kubectl, have y'all considered using the kubernetes API client instead of having the hard dependency on the kubectl binary? That way the mixin could be built as a (Golang?) lib for the target platform and avoid having to figure out which kubectl binary to use and how to install it.

Apologies in advance is this issue is unclear or out of place as I'm still trying to wrap my head around the porter workflow.

## To Reproduce
1. Use a linux base image that does not support apt-get.
2. Configure kubernetes mixin in porter.yaml.
3. `porter build`

## Expected behavior
Kubernetes mixin is cross platform and works seamlessly across diff base images.

## Version
`porter v0.22.0-beta.1 (8f3e429)`

carolynvs:
# What does this change
Document requirements for the base image that you may use in your custom Dockerfile.

# What issue does it fix
Closes #844 

# Notes for the reviewer
🐱 

# Checklist
- [ ] Unit Tests
- [x] Documentation

### DIFF
--- a/docs/content/custom-dockerfile.md
+++ b/docs/content/custom-dockerfile.md
@@ -13,11 +13,11 @@ When you run `porter create` template Dockerfile is created for you
 in the current directory named **Dockerfile.tmpl**:
 
 ```Dockerfile
-FROM quay.io/deis/lightweight-docker-go:v0.2.0
 FROM debian:stretch
-COPY --from=0 /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
 
 ARG BUNDLE_DIR
+
+RUN apt-get update && apt-get install -y ca-certificates
 
 # This is a template Dockerfile for the bundle's invocation image
 # You can customize it to use different base images, install tools and copy configuration files.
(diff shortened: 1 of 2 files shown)


carolynvs:
It's a draft because it a PR that relies on another PR (#852). It has commits from that PR in it.

carolynvs:
Ok, I've merged the other PR and rebased this one.

Issue #824, opened 2019-12-04
Publish all assets to porter domains / orgs


carolynvs:
- [x] Publish mixins and plugins to porter.sh instead of deislabs.io
- [x] Use front door instead of verizon CDN because the CDN has been giving us problems and we are bypassing it at all times at this point.
- [x] Publish our remaining example bundles to getporter on docker hub.
    - [x] Update/revisit example bundles https://github.com/deislabs/porter/issues/856
- [x] Publish our test bundles to getporterci too (make a new org) - see https://github.com/deislabs/porter/pull/857

vdice:
Currently working on `Publish our remaining example bundles to getporter on docker hub.` and had a few Q's.

1. These example bundles are all the ones in the `examples` directory, correct?  Do we wish to publish any of the bundles in the `workshop` dir?  And/or from another location?

2. The current list of bundles in the `examples` dir is:

- aks-spring-music
- azure-ark
- azure-mysql-wordpress
- azure-tf-example
- exec-outputs
- gke-example
- kubernetes-mixin-example
- user-service-fabric-cli

Do we wish to continue maintaining/publishing all of these?  For instance, I know the `azure-ark` bundle is currently outdated (😬  still haven't gotten to https://github.com/deislabs/porter/issues/324).

The README in `examples` currently only mentions the top 3 in the list above, so we can add new entries depending on the final list.

vdice:
Also, as I'm going through all of the bundles (examples, test) to update the org (`deislabs` -> `getporter` or `getporterci`), I've been eliminating the `invocationImage` section since we now have the logic to create the appropriately named invocation image based on `tag`.  Wanted to check in here to see if now is a good time to establish that precedent?  (Only keeping the updated `tag` field)

jeremyrickard:
I think we should maybe limit the examples. The Ark example kind of points to having too many to make maintain.

vdice:
> I think we should maybe limit the examples. The Ark example kind of points to having too many to make maintain.

Ok, created https://github.com/deislabs/porter/issues/856 that I can work on as a follow-up

carolynvs:
I'm sorry for the slow replies! 😬 

> These example bundles are all the ones in the examples directory, correct? Do we wish to publish any of the bundles in the workshop dir? And/or from another location?

I was suggesting just the bundles in examples. I don't recommend publishing the workshop bundles.

> Do we wish to continue maintaining/publishing all of these? For instance, I know the azure-ark bundle is currently outdated

[...]

PR #858, opened 2020-01-17
chore(*): update cdn url and mixin author name

vdice:
# What does this change
- Updates the cdn url to `https://cdn.porter.sh`
- Updates the mixin author to `Porter Authors`

# What issue does it fix
Ref https://github.com/deislabs/porter/issues/824

# Notes for the reviewer
Similar to other PRs contributing towards #824, I avoided changes to the `docs/content/slides` and `workshop/` directories.

PRs in the associated mixin repos forthcoming, though this shouldn't necessarily need to block on them.

# Checklist
- [ ] Unit Tests
- [ ] Documentation

### DIFF
--- a/cmd/porter/mixins.go
+++ b/cmd/porter/mixins.go
@@ -52,10 +52,10 @@ func BuildMixinInstallCommand(p *porter.Porter) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "install NAME",
 		Short: "Install a mixin",
-		Example: `  porter mixin install helm --url https://cdn.deislabs.io/porter/mixins/helm
-  porter mixin install helm --feed-url https://cdn.deislabs.io/porter/atom.xml
-  porter mixin install azure --version v0.4.0-ralpha.1+dubonnet --url https://cdn.deislabs.io/porter/mixins/azure
-  porter mixin install kubernetes --version canary --url https://cdn.deislabs.io/porter/mixins/kubernetes`,
+		Example: `  porter mixin install helm --url https://cdn.porter.sh/mixins/helm
+  porter mixin install helm --feed-url https://cdn.porter.sh/atom.xml
+  porter mixin install azure --version v0.4.0-ralpha.1+dubonnet --url https://cdn.porter.sh/mixins/azure
+  porter mixin install kubernetes --version canary --url https://cdn.porter.sh/mixins/kubernetes`,
 		PreRunE: func(cmd *cobra.Command, args []string) error {
 			return opts.Validate(args)
 		},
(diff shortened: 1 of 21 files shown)


cmd/porter/mixins.go
```diff
@@ -52,10 +52,10 @@ func BuildMixinInstallCommand(p *porter.Porter) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "install NAME",
 		Short: "Install a mixin",
-		Example: `  porter mixin install helm --url https://cdn.deislabs.io/porter/mixins/helm
-  porter mixin install helm --feed-url https://cdn.deislabs.io/porter/atom.xml
-  porter mixin install azure --version v0.4.0-ralpha.1+dubonnet --url https://cdn.deislabs.io/porter/mixins/azure
-  porter mixin install kubernetes --version canary --url https://cdn.deislabs.io/porter/mixins/kubernetes`,
+		Example: `  porter mixin install helm --url https://cdn.porter.sh/mixins/helm
+  porter mixin install helm --feed-url https://cdn.porter.sh/atom.xml
```
carolynvs: So now that we have the CDN rules all fixed, the /atom.xml URL is going to become legacy just to support clients until we can get a PR in place to update Porter to start using /mixins/atom.xml. I think we should stop mentioning this old URL anywhere, and instead with this PR start using the new path.

What do you think?

docs/content/mixin-dev-guide/distribution.md
```diff
@@ -64,7 +64,7 @@ When `--version` is not specified, it is defaulted to `latest` which should
 represent the most recent version of the mixin.
 
 You may also choose to publish `canary` versions of the mixin, which are
-unpublished builds from the master branch. Porter and the official DeisLabs
+unpublished builds from the master branch. Porter and the official Porter Authors
```
carolynvs: Seems like this can be just `the official Porter mixins`

pkg/mixin/install.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 const (
-	DefaultFeedUrl = "https://cdn.deislabs.io/porter/atom.xml"
+	DefaultFeedUrl = "https://cdn.porter.sh/atom.xml"
```
carolynvs: This should be /mixins/atom.xml right? Or are we not ready to have that be our default feed URL yet for installing mixins?

cmd/porter/mixins.go
vdice: Definitely, thanks for catching.  Updated.

docs/content/mixin-dev-guide/distribution.md
vdice: Updated.

pkg/mixin/install.go
vdice: You're right; updated.

PR #860, opened 2020-01-21
ref(mysql/porter.yaml): use latest chart; set mysql-user default

vdice:
# What does this change
- Updates the mysql test/sample bundle to install the latest `stable/mysql` chart (by removing a version override)

- Sets a default value for the `mysql-user` parameter

# What issue does it fix
When testing the bundle (prior to this change) on a k8s cluster of v1.17.0, I was greeted with the following error:

```
...
/usr/local/bin/helm helm install --name porter-ci-mysql stable/mysql --version 0.10.2 --replace --set mysqlDatabase=mydb --set mysqlUser=admin
Error: validation failed: unable to recognize "": no matches for kind "Deployment" in version "extensions/v1beta1"
Error: exit status 1
Error: mixin execution failed: exit status 1
```

# Notes for the reviewer
Would also be open to just bumping the version to a semver release that fixes the issue mentioned above, if we are hesitant to default to the latest version.  WDYT?

# Checklist
- [ ] Unit Tests
- [ ] Documentation

### DIFF
--- a/build/testdata/bundles/mysql/porter.yaml
+++ b/build/testdata/bundles/mysql/porter.yaml
@@ -17,6 +17,7 @@ parameters:
   env: DATABASE_NAME
 - name: mysql-user
   type: string
+  default: mysql-admin
   env: MYSQL_USER
 - name: namespace
   type: string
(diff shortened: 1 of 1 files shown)


carolynvs:
We had originally set to a fixed version to avoid rando breaking changes from creeping into our tests without us intending it. It was "decided" in an earlier PR

https://github.com/deislabs/porter/pull/696/files#r331581290

My vote is still for picking a newer version of mysql and staying with it vs being on latest. I think that will result in less churn but let me know if you see it differently.

PR #867, opened 2020-01-27
Show value instead of paths for sensitve outputs

### LINKED ISSUE #863: Rethinking display of sensitive outputs in table display
Now that outputs aren't stored in porter home anymore, displaying the path _inside the bundle runtime_ to the output doesn't seem helpful for sensitive outputs.

Here is what someone sees when they list outputs in table format:

```
-----------------------------------------------
  Name                 Type    Value (Path if sensitive)
-----------------------------------------------
  IP_ADDRESS           string  40.88.49.175
  STORAGE_ACCOUNT_KEY  string  /cnab/app/outputs/STORAGE_ACCOUNT_KEY
```

Back when the path was a local path to porter home, e.g. ~/.porter/outputs/BUNDLE/STORAGE_ACCOUNT_KEY, it made more sense. They could choose to go look and see if they had permission to view that file. But now that we display a path that is inaccessible, /cnab/app/... isn't available after a bundle has been run, I'd prefer that we think about what else to show.

## Options

A. Just show the value, regardless of sensitive or not (just like you would see if it's json/yaml). Why are tables special? **WINNER**

B. Show *** and only show the full value when they request the named output with `porter instance output show NAME`. 
What are we trying to avoid here? Someone looking over the shoulder? We *** in the logs to avoid sensitive values being written to logs and then to disk without people thinking about it. But when they explicitly ran "show me the outputs" I feel like we aren't being helpful.

C. Write in a better option below

I guess at the moment I'm leaning towards A but am looking for other people to say that's OK or give better options. 😀 

---

Option A was selected

---

1. In outputs.go remove the logic to switch out sensitive output values with the path. This is not obvious but an output is considered sensitive when its `WriteOnly` field is true.

https://github.com/deislabs/porter/blob/4401e74fd8933a892d99f9b4244a91efa55e03c2/pkg/porter/outputs.go#L127-L129

2. Update the tests in outputs_test.go to match the changes made.

3. Some of our documentation and text still has the old text in the header "Value (Path if sensitive)". It doesn't appear to be in the actual tables printed by the CLI anymore but let's do a sweep for it and clean up the rest of the repo to be consistent.

4. Good stuff to try after doing this change: 
  * `make build test-unit`
  * ` make install`
  * Use your local install of porter to create a new bundle, define a sensitive output, install it and then list the output

Our [CONTRIBUTING](https://github.com/deislabs/porter/blob/master/CONTRIBUTING.md#developer-tasks) explains how to get started with development with Porter but if you have questions, feel free to ask in Slack or here.

scriptonist:
# What does this change
Now that outputs aren't stored in porter home anymore, displaying the path inside the bundled runtime to the output doesn't seem helpful for sensitive outputs.

Here is what someone sees when they list outputs in table format:
```
-----------------------------------------------
  Name                 Type    Value (Path if sensitive)
-----------------------------------------------
  IP_ADDRESS           string  40.88.49.175
  STORAGE_ACCOUNT_KEY  string  /cnab/app/outputs/STORAGE_ACCOUNT_KEY
```
Back when the path was a local path to porter home, e.g. ~/.porter/outputs/BUNDLE/STORAGE_ACCOUNT_KEY, it made more sense. They could choose to go look and see if they had permission to view that file. But now that we display a path that is inaccessible, /cnab/app/... isn't available after a bundle has been run.

After discussion in #863 , the following is solution was accepted,

show the value, regardless of sensitive or not (just like you would see if it's json/yaml). 

# What issue does it fix
Closes #863 

# Checklist
- [x] Unit Tests
- [ ] Documentation

### DIFF
--- a/build/testdata/bundles/mysql/porter.yaml
+++ b/build/testdata/bundles/mysql/porter.yaml
@@ -17,6 +17,7 @@ parameters:
   env: DATABASE_NAME
 - name: mysql-user
   type: string
+  default: mysql-admin
   env: MYSQL_USER
 - name: namespace
   type: string
@@ -30,7 +31,7 @@ install:
     description: "Install MySQL"
     name: "{{ bundle.parameters.mysql-name }}"
     chart: stable/mysql
-    version: 0.10.2
+    version: 1.6.2
     namespace: "{{ bundle.parameters.namespace }}"
     replace: true
     set:
@@ -57,7 +58,7 @@ upgrade:
       name: "{{ bundle.parameters.mysql-name }}"
       namespace: "{{ bundle.parameters.namespace }}"
       chart: stable/mysql
-      version: 0.10.2
+      version: 1.6.2
       outputs:
       - name: mysql-root-password
         secret: "{{ bundle.parameters.mysql-name }}"
--- a/docs/content/slides/pack-your-bags-msp/index.md
+++ b/docs/content/slides/pack-your-bags-msp/index.md
@@ -1071,7 +1071,7 @@ $ porter bundle show
 
 Outputs:
 -------------------------------------------------------------------------------------------
-  Name                 Type    Value (Path if sensitive)
+  Name                 Type    Value
 -------------------------------------------------------------------------------------------
   STORAGE_ACCOUNT_KEY  string  JKb9C+J+nFtGrDyBW4Y0zaIK5hzIvi2gW3SfnmnkcunyXSYV3HucQGNIo...
 ```
--- a/pkg/porter/outputs.go
+++ b/pkg/porter/outputs.go
@@ -124,9 +124,9 @@ func (p *Porter) ListBundleOutputs(c claim.Claim, format printer.Format) []Displ
 		}
 		do.Definition = *def
 
-		if def.WriteOnly != nil && *def.WriteOnly {
-			valueStr = output.Path
-		}
+		// if def.WriteOnly != nil && *def.WriteOnly {
+		// 	valueStr = output.Path
+		// }
 
 		outputType, _, err := def.GetType()
 		if err != nil {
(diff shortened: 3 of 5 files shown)


scriptonist:
@carolynvs On the docs side, I'm getting some weird errors. So wasn't able to confirm the changes.

Following is a sample of the error
```
ERROR 2020/01/27 08:10:55 Error while rendering "page" in "cli/": template: theme/_default/single.html:1:3: executing "theme/_default/single.html" at <partial "header.html...>: error calling partial: template: theme/partials/header.html:4:3: executing "theme/partials/header.html" at <partial "head.html" ...>: error calling partial: template: "theme/partials/head.html" is an incomplete or empty template
```
If you can help me figure that out, I can confirm the changes in docs side as well.

vdice:
@scriptonist when convenient, can you address the merge conflict?  We'd just merged some amendments to some workshop and example bundles... the conflict appears to be in the README, so hopefully straightforward to address.

carolynvs:
@scriptonist The docs build without error on the CI server. I'm wondering if the problem is how Hugo is installed on your local computer? Perhaps we need to improve the instructions for that. On my local machine and on the CI server it's hugo `v0.53/extended`, what version is yours?

scriptonist:
@carolynvs thanks for reminding me about the version, should've checked it before! 
I was running `0.40` earlier :sweat_smile: I've upgraded to `v0.63.2-934EE21F`, now it seems I've encountered another problem.
```
porter|disable-masking-sesitive-outputs ⇒ make docs-preview                                                         (minikube/default)
hugo serve --source docs/
Error: add site dependencies: create deps: create PathSpec: build filesystems: create main fs: symlinks not allowed in this filesystem
Makefile:101: recipe for target 'docs-preview' failed
make: *** [docs-preview] Error 255

```
Do folks have any clue?

carolynvs:
Ah okay, looks like Hugo has altered their support for symlinks which we use in our theme:

https://gohugo.io/news/0.56.0-relnotes#notes

If you are able to install exactly v0.53 I think that would let you build locally but in the meantime, there's nothing wrong with using netlify's PR preview (look in the PR checks to see it) to do your doc checking until we fix our doc to not use symlinks anymore.

Issue #775, opened 2019-10-24
Common Plugin Version Command


carolynvs:
All plugins should implement a common version command that tells you not only what version it is but what implementations they contain:

```json
$ azure version --output json
{
  "name": "azure",
  "version": "v0.1.0",
  "commit": "abc123",
  "plugins": [
     {
       "type": "instance-storage",
        "implementation": "blob"
     }
  ] 
}
```

Right now the azure plugin reused the version struct from the porter mixins package but that will need to change to accommodate this.

https://github.com/deislabs/porter-azure-plugins/blob/master/pkg/azure/version.go

The new struct should probably go here: https://github.com/deislabs/porter/tree/master/pkg/plugins

carolynvs:
Closed by #868 and #874

PR #868, opened 2020-01-27
Implement command to get plugin list

### LINKED ISSUE #774: List plugins command
🚨 This relies on plugins reporting the types and implementations they contain (#775). Right now they just report version and author.

Add a command similar to `porter mixins list` that lists installed plugins.

```
$ porter plugins list
Name        Type                 Implementation    Version                 Author
azure       instance-storage     blob              v0.1.2-beta.1           DeisLabs
aws         instance-storage     s3                v0.1.0                  Amazon
```

dev-drprasad:
# What does this change
Implement command to get plugin list

```bash
$ porter plugins list
Name        Type                 Implementation    Version                 Author
azure       instance-storage     blob              v0.1.2-beta.1           DeisLabs
azure       instance-storage     mongo             v0.1.2-beta.1           DeisLabs
aws         instance-storage     s3                v0.1.0                  Amazon
```

# What issue does it fix
Closes #774 

# Notes for the reviewer

# Checklist
- [x] Unit Tests
- [x] Documentation


dev-drprasad:
Hi @carolynvs ,

I see you referenced issue #775 with this and you mentioned `plugin version --output json` contains implementations which is an array.
Does it mean each plugin has multiple implementations ?
If so, then what should porter plugin list return ? list of plugins or list of implementations ?

carolynvs:
Good question! I've been pretty loose with the terms as I have been feeling it out as I developed things. 

Now that #859 is nearly complete I think it would be more clear to stick with calling "azure" or "aws" the plugin and then each plugin can have multiple implementations of a type.

Here's an extended example:

```
$ azure version --output json
{
  "name": "azure",
  "version": "v0.1.0",
  "commit": "abc123",
  "plugins": [
     {
       "type": "instance-storage",
        "implementation": "blob"
     },
     {
       "type": "instance-storage",
        "implementation": "mongo"
     },
  ] 
}
```

So perhaps in another PR we should change `plugins` above to `implementations` or something to avoid confusion, though I am not sure it is a problem.

```console
$ porter plugins list
Name        Type                 Implementation    Version                 Author
azure       instance-storage     blob              v0.1.2-beta.1           DeisLabs
azure       instance-storage     mongo             v0.1.2-beta.1           DeisLabs
aws         instance-storage     s3                v0.1.0                  Amazon
```

After #859 in your porter configuration file ~/.porter/config.toml you will be able to specify which plugin implementation to use when storing data:

```toml
default-storage = "dev"

[[storage]]
  name = "dev"
  plugin = "azure.blob"
  
  [storage.config]
    env = "DEV_AZURE_STORAGE_CONNECTION_STRING"
```

pkg/plugins/plugins.go
```diff
@@ -22,6 +29,124 @@ type PluginKey struct {
 	IsInternal     bool
 }
 
+type implementaion struct {
```
carolynvs: Let's export this so that plugins can use the type when implementing the version command.

pkg/porter/porter.go
```diff
@@ -26,6 +27,7 @@ type Porter struct {
 	Builder         BuildProvider
 	Manifest        *manifest.Manifest
 	Mixins          mixin.MixinProvider
+	Plugins         plugins.PluginProvider
```
carolynvs: When we add interfaces to the Porter app, at the same time it really helps to add a test struct so that we can write unit tests and not end up hitting the file system and expect real plugins to be there.

So let's follow the same pattern that we used for mixins:

1. Add a TestPluginProvider using [TestMixinProvider](https://github.com/deislabs/porter/blob/d904e39007954fbec5e4d870e13ad16b04c355e2/pkg/mixin/helpers.go#L11) as an example. Though we don't need the run assertions stuff that it does, just whatever stubbing is necessary so that we could call `PrintPlugins` for example and have it print out a hard coded plugin like 'myplugin'.
1. Update the [TestPorter](https://github.com/deislabs/porter/blob/d904e39007954fbec5e4d870e13ad16b04c355e2/pkg/porter/helpers.go#L39) struct to use the new test stub by default.
2. Add a test for PrintPlugins that uses the new test plugin provider, using [TestPrintMixins](https://github.com/deislabs/porter/blob/d904e39007954fbec5e4d870e13ad16b04c355e2/pkg/porter/mixins_test.go#L12) as an example.

pkg/plugins/plugins.go
```diff
@@ -22,6 +29,113 @@ type PluginKey struct {
...
+
+	// make json.Marshal return `[]` instead of `nil` for not having implementations
+	if len(metadata.Implementations) == 0 {
+		metadata.Implementations = make([]Implementaion, 0)
+	}
+	return &metadata, nil
+}
+
+// PrintMetadata writes plugin metadata to `ctx.Out` as plaintext|json|yaml format based on `opts`
+func PrintMetadata(ctx *context.Context, opts version.Options, metadata PluginMetadata) error {
```
dev-drprasad: I named method as `PrintMetadata` instead of `PrintVersion` as it doesn't only print version. If you agree with that we can change `mixins.PrintVersion` to `mixins.PrintMetadata` to make it consistent in other PR

pkg/plugins/runner.go
```diff
@@ -0,0 +1,81 @@
...
+		return errors.Errorf("Plugin %s doesn't exist in filesystem with path %s", r.pluginName, pluginPath)
+	}
+
+	return nil
+}
+
+func (r *PluginRunner) Run(commandOpts CommandOptions) error {
+	if r.Debug {
+		fmt.Fprintln(r.Err, "DEBUG Plugin Name: ", r.pluginName)
+		fmt.Fprintln(r.Err, "DEBUG Plugin Command: ", commandOpts.Command)
```
dev-drprasad: I dont understand why we are using `r.Out` instead `r.Err` for debugging and why writing anything to `r.Out` not working. can you please explain ?

carolynvs: Some of our commands takes stdout from the command's execution and parses it as a document. This is mainly (entirely?) due to the `schema` command which has the plugin output a json schema document to stdout which porter then parses and uses. If debug information was printed by either the mixin or porter, the document would be malformed. By writing debug lines to stderr, it helps us keep the two separate. 

I tried to document this a bit in our contributing guide under [Logging](https://github.com/deislabs/porter/blob/master/CONTRIBUTING.md#logging). If it's not clear, would you please let me know so that I can improve it?

docs/config.toml
```diff
@@ -146,6 +146,12 @@ defaultContentLanguage = "en"
     identifier = "cli-schema"
     weight = 198
     parent = "cli"
+  [[menu.main]]
```
carolynvs: Thanks for adding this to the docs nav! 👍

pkg/plugins/helpers.go
```diff
@@ -0,0 +1,27 @@
...
+		impl = []Implementaion{
+			{Type: "instance-storage", Name: "blob"},
+			{Type: "instance-storage", Name: "mongo"},
+		}
+	}
+	return &PluginMetadata{
+		Name:            pluginName,
+		ClientPath:      fmt.Sprintf("/home/porter/.porter/plugins/%s", pluginName),
+		Implementations: impl,
+		VersionInfo:     VersionInfo{Version: "v1.0", Commit: "abc123", Author: "Deis Labs"},
```
carolynvs: Just a heads up that we now set the author information to "Porter Authors" for any mixins we write to be more inclusive. This doesn't impact your tests, but I wanted you to know.

pkg/plugins/plugins.go
```diff
@@ -22,6 +29,113 @@ type PluginKey struct {
 	IsInternal     bool
 }
 
+// Implementaion stores implementation type (e.g. instance-storage) and its name (e.g. s3, mongo)
+type Implementaion struct {
```
carolynvs: Not a huge deal but this is misspelled, should be `Implementation` (missing the t)

pkg/plugins/plugins.go
```diff
@@ -22,6 +29,113 @@ type PluginKey struct {
...
+	Commit  string `json:"commit"`
+	Author  string `json:"author,omitempty"`
+}
+
+type PluginProvider interface {
+	List() ([]string, error)
+	GetMetadata(string) (*PluginMetadata, error)
+}
+
+func NewFileSystem(config *config.Config) *fileSystem {
```
carolynvs: I realize that parts of your PR conflict with mine (#874). Let's merge yours first and I will rationalize the differences in mine.

pkg/plugins/plugins.go
carolynvs: That's fine, the current method name made more sense when all it did was literally print "v0.1.0" 😀

carolynvs:
@dev-drprasad Are you in the #porter channel on the CNCF slack? Just curious if you do slack or not (it's okay if you don't!), I just couldn't find your name. If you would like an invite here's the info:

https://porter.sh/community#slack

pkg/plugins/plugins.go
dev-drprasad: 🤦

pkg/plugins/plugins.go
dev-drprasad: 🤦

pkg/plugins/runner.go
dev-drprasad: Got it. Actually i haven't read doc

dev-drprasad:
@carolynvs I resolved conflicts.

PR #882, opened 2020-02-05
feat: porter mixin search

vdice:
# What does this change
- adds a `porter mixin search` command
- introduces list of remote mixins to search
- adds docs on how to add a mixin to the list

# What issue does it fix
Closes https://github.com/deislabs/porter/issues/808

# Notes for the reviewer
This may need refactoring depending on timing of https://github.com/deislabs/porter/pull/874.  We'll want to generalize listing/searching for both mixins and plugins.  

# Checklist
- [x] Unit Tests
- [x] Documentation


pkg/mixin/search.go
```diff
@@ -0,0 +1,82 @@
...
+	copy(results, rml)
+
+	if opts.Name != "" {
+		results = RemoteMixinList{}
+		for _, rmi := range rml {
+			if strings.Contains(rmi.Name, opts.Name) {
+				results = append(results, rmi)
+			}
+		}
+	}
```
dev-drprasad: **nit:** I felt this block is hard to understand as why `rml` copied to `results` first and again became empty. Can't we just do something like

```
if opts.name == "" {
  return rml
}
results := RemoteMixinList{}
for _, rmi := range rml {
  if strings.Contains(rmi.Name, opts.Name) {
    results = append(results, rmi)
  }
}
sort.Sort(results)
return results
```

pkg/mixin/search.go
```diff
@@ -0,0 +1,82 @@
...
+
+// Search searches for mixins matching the optional provided name,
+// returning the full list if none is provided
+func (s *Searcher) Search(opts SearchOptions) (RemoteMixinList, error) {
+	data, err := s.Box.Find("index.json")
+	if err != nil {
+		return nil, errors.Wrap(err, "error loading remote mixin list")
+	}
+
+	var rml RemoteMixinList
```
dev-drprasad: Its not huge deal but i see two identifier names `rml` and `rms` used to represent `RemoteMixinList`. Maybe, maybe one is better

pkg/mixin/search.go
```diff
@@ -0,0 +1,82 @@
...
+	err = json.Unmarshal(data, &rml)
+	if err != nil {
+		return nil, errors.Wrapf(err, "could not parse remote mixin list")
+	}
+
+	results := make(RemoteMixinList, len(rml))
```
dev-drprasad: **nit:** I see `RemoteMixinList ` has  `Len` method.

pkg/porter/mixins.go
```diff
@@ -62,6 +63,41 @@ func (p *Porter) ListMixins() ([]mixin.Metadata, error) {
...
+	box := packr.New("get.porter.sh/porter/pkg/mixin/remote-mixins", "../mixin/remote-mixins")
+	mixinSearcher := mixin.NewSearcher(box)
+
+	remoteMixins, err := mixinSearcher.Search(opts)
+	if err != nil {
+		return err
+	}
+
+	if len(remoteMixins) == 0 && opts.Name != "" {
+		fmt.Fprintf(p.Out, "No mixins found for %s\n", opts.Name)
```
dev-drprasad: Should we write errors to `p.Out` instead `p.Err` ? other party who is calling `search --output json` will get parse errors

vdice: Great call, thank you.  Updated.

pkg/mixin/search.go
vdice: Indeed; updated.

pkg/mixin/search.go
vdice: Good catch; all were meant to be updated to `rms`.  This should now be done.

pkg/mixin/search.go
vdice: Much simpler and easier to understand, agreed.  Thank you!  Updated.

pkg/mixin/metadata.go
```diff
@@ -12,9 +12,31 @@ type Metadata struct {
...
+
+func (rms RemoteMixinList) Len() int {
+	return len(rms)
+}
+func (rms RemoteMixinList) Swap(i, j int) {
+	rms[i], rms[j] = rms[j], rms[i]
+}
+func (rms RemoteMixinList) Less(i, j int) bool {
+	return rms[i].Name < rms[j].Name
+}
```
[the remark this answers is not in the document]
vdice: I added a comment to signify that these methods are to implement the `sort.Interface`... hopefully this helps?

vdice:
>1. users and authors need to wait until porter next version released to get  latest updates
>2. users must update porter to get latest updates or else they endup with stale list
>3. For whatever reasons, if mixin author  change anything related to mixin, then users with old versions of porter will endup with broken data

Great points all.  Indeed, building the list into the binary is unsustainable/unsatisfactory for the reasons you've noted above.  

> Can't we use seperate github repo and fetch data from there. ?

This could be one solution.  Another might be that we/Porter hosts the `index.json` file alongside the main `atom.xml` mixin feed.  So, the index could be fetchable via `https://cdn.porter.sh/mixins/index.json` or some such.  This is nice since we already have the domain/SSL/etc. infra set up.

Then, we can add CI to update this list on merges to master to ensure it is up-to-date.  We'd also adjust the logic here/in the cli to pull results from this external list.  This would de-couple the list from the binary.  WDYT?

carolynvs:
I haven't done a full review yet, sorry. I just saw @dev-drprasad 's comment about the index.json being in the main porter repo and wanted to echo that I thought it would get it's own dedicated repo. Would that be helpful, annoying or neutral? 😀

vdice:
> wanted to echo that I thought it would get it's own dedicated repo

Although I don't think this is strictly necessary to decouple the tracked list with the `porter` CLI, I'd be open to go this route.

The (minor) cons I see with this option is: a bit more overhead tracking an add'l repo, both for maintainers and for users to discover/PR against, add'l CI pipeline(s).  That being said, a clear separation of repo concerns may be nice.

One further thought: when we wish to build support for searching plugins, would this mean we'd want another repo for the plugins list?  Or would want to combine lists/maintenance/CI for both mixins and plugins the same separate repo?

carolynvs:
I would say that both a plugins list and the mixin list would live in the same repo

vdice:
Another thought around the separate repo path: When/if the remote package (mixin or plugin) listing structure changes, code may need to change in two repos: updating the entries in the list in one repo and updating the parsing/searching logic in another.  (Ideally, changes can be in a backwards-compatible way, so older cli binaries can parse updated lists with differing structure.)

We could move the parse/search logic into this separate repo as well, so that the aforementioned alterations would be consolidated into one, at the cost of further overhead (this new repo now would contain a golang api that the porter cli can consume, etc.  Naturally, build/test pipelines would need to be added.)

Perhaps, if we're agreed that it is a good idea to host the package (mixin/plugin) lists on the domain we already manage (`https://cdn.porter.sh`) and we're comfortable with the logic in this PR, the next actionable item could be to wire up the CI/infra for hosting said lists, all still in this main repo.  Once finished, we could then make the decision on splitting out applicable code/CI into a separate repo.

Thoughts?

pkg/mixin/metadata.go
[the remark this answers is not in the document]
dev-drprasad: I didn't know that we should implement those methods to make sort work. Comment is indeed helpful

pkg/mixin/search.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	var rml RemoteMixinList
+	err = json.Unmarshal(data, &rml)
+	if err != nil {
+		return nil, errors.Wrapf(err, "could not parse remote mixin list")
+	}
+
+	if opts.Name == "" {
+		return rml, nil
```
dev-drprasad: Sorry, last time i forget mention in suggestion. As we are sorting `results` then we have to sort `rml` also

dev-drprasad:
> Another thought around the separate repo path: When/if the remote package (mixin or plugin) listing structure changes, code may need to change in two repos: updating the entries in the list in one repo and updating the parsing/searching logic in another. (Ideally, changes can be in a backwards-compatible way, so older cli binaries can parse updated lists with differing structure.)
> 
> We could move the parse/search logic into this separate repo as well, so that the aforementioned alterations would be consolidated into one, at the cost of further overhead (this new repo now would contain a golang api that the porter cli can consume, etc. Naturally, build/test pipelines would need to be added.)
> 
> Perhaps, if we're agreed that it is a good idea to host the package (mixin/plugin) lists on the domain we already manage (`https://cdn.porter.sh`) and we're comfortable with the logic in this PR, the next actionable item could be to wire up the CI/infra for hosting said lists, all still in this main repo. Once finished, we could then make the decision on splitting out applicable code/CI into a separate repo.
> 
> Thoughts?

I realised creating new repo is extra overhead. I agree that putting everything in main repo is better and serve both mixin/plugin list through CDN

pkg/mixin/remote-mixins/index.json
```diff
@@ -0,0 +1,79 @@
+[
```
carolynvs: I know you've been kicking around having less duplicate information and just the entry for the feed or source url here. I'm not sure how much we need to optimize out of the gate. For now this works great, and the only burden is on us to list new mixins.

Longer term, it would be nice if we could just list our atom.xml and not list each mixin for example. Even longer term, I would assume if we are wildly sucessful beyond our dreams, support for localization would come up.

I think this file format is just fine for now though.

pkg/mixin/remote-mixins/index.json
```diff
@@ -0,0 +1,79 @@
...
+  {
+    "name": "arm",
+    "author": "Porter Authors",
+    "description": "A mixin for deploying ARM templates on Azure",
+    "sourceURL": "https://cdn.porter.sh/mixins/arm",
```
carolynvs: This should match the format used by the install commands so that someone would know how to invoke an install command based on what they found from the search results. Source URL (--url) and Feed URL (--feed-url) are mutually exclusive, only one should be set. So for our official mixins, it's the feed url.

pkg/mixin/search.go
```diff
@@ -0,0 +1,81 @@
...
+	if err != nil {
+		return err
+	}
+
+	return o.ParseFormat()
+}
+
+// validateMixinName validates either no mixin name is provided or only one is
+func (o *SearchOptions) validateMixinName(args []string) error {
+	switch len(args) {
```
carolynvs: This []string{} name validation is common enough that I think we should make a resuable utility function out of it and use it everywhere. 😀  Not in your PR just saying I think I've seen it in at least 6 places.

pkg/mixin/search.go
```diff
@@ -0,0 +1,81 @@
...
+		return nil
+	default:
+		return errors.Errorf("only one positional argument may be specified, the mixin name, but multiple were received: %s", args)
+	}
+}
+
+// Search searches for mixins matching the optional provided name,
+// returning the full list if none is provided
+func (s *Searcher) Search(opts SearchOptions) (RemoteMixinList, error) {
+	data, err := s.Box.Find("index.json")
```
carolynvs: I am guessing that after your PR, in a follow-up we'd have CI publish this index.json to a well-known location and we would remove this box entirely?

pkg/mixin/testdata/remote-mixins/index.json
```diff
@@ -0,0 +1,23 @@
+[
+  {
+    "name": "az",
+    "author": "Porter Authors",
+    "description": "A mixin for using the az cli",
+    "sourceURL": "https://cdn.porter.sh/mixins/az",
```
carolynvs: Again here in the test data, only source url or feed url should be set, not both.

pkg/porter/mixins.go
```diff
@@ -62,6 +63,41 @@ func (p *Porter) ListMixins() ([]mixin.Metadata, error) {
...
+func (p *Porter) SearchMixins(opts mixin.SearchOptions) error {
+	box := packr.New("get.porter.sh/porter/pkg/mixin/remote-mixins", "../mixin/remote-mixins")
+	mixinSearcher := mixin.NewSearcher(box)
+
+	remoteMixins, err := mixinSearcher.Search(opts)
+	if err != nil {
+		return err
+	}
+
+	if len(remoteMixins) == 0 && opts.Name != "" {
```
carolynvs: I would move this logic into mixinSearcher.Search and return it as an error from Search. A couple reasons why:

1. When someone else uses this code, i.e. the tests or someone else in porter or a third party developer, they will get this logic for free. Right now only the CLI gets it.
2. By returning it as an error, it will be automatically handled by lines 72-74.
3. When someone searches for a mixin and nothing comes back they will get a non-zero exit code which will be helpful when they are scripting.

pkg/porter/mixins.go
```diff
@@ -62,6 +63,41 @@ func (p *Porter) ListMixins() ([]mixin.Metadata, error) {
 	return mixins, nil
 }
 
+// SearchMixins searches the internal remote mixins list according to the provided options
+func (p *Porter) SearchMixins(opts mixin.SearchOptions) error {
```
carolynvs: Can you split this function into two functions? One that handles doing the search and other than applies the printer options and does the printing?

As always I ask for that because it helps people reuse code (you can't reuse code that prints to the console) and test smaller units of work.

pkg/porter/mixins.go
```diff
@@ -62,6 +63,41 @@ func (p *Porter) ListMixins() ([]mixin.Metadata, error) {
...
+	case printer.FormatTable:
+		printMixinRow :=
+			func(v interface{}) []interface{} {
+				m, ok := v.(mixin.RemoteMixinListing)
+				if !ok {
+					return nil
+				}
+				return []interface{}{m.Name, m.Description, m.Author, m.SourceURL, m.FeedURL}
+			}
+		return printer.PrintTable(p.Out, remoteMixins, printMixinRow, "Name", "Description", "Author", "Source URL", "Feed URL")
```
carolynvs: I'm not sure where "Source URL" came from as a term? Is that following an existing precedent in our CLI that I'm brainfarting on? 😀  

If not, would you mind using "URL" and "Feed URL" to be consistent with `--url` and `--feed-url` that people will be familiar with from the `porter mixins install` command? I'd like to avoid introducing a new term that is an alias for something we already have.

carolynvs:
I still lean towards a separate repository. But we don't need to decide or do that in this PR. Here are my reasons:

* Easier for people to find, it's not nestled down inside go code files with a dedicated readme front and center.
* We can use finer grained permissions on the repo if we need.
* Our CI pipeline can be super fast and tight, without kicking stuff off that isn't related to the code. It's annoying when just a doc change triggers a long pipeline run and it's something we struggle to manage properly. Having a repo with it's own pipeline for just this will help avoid making that worse.
* We already have lots of focused repos, e.g. our vanity repo, a homebrew repo, mixin repos, plugin repos, skeletor. We in general avoid a monorepo.
* Less confusion about if it's tied to our release cycle or not.

docs/content/mixin-dev-guide/distribution.md
```diff
@@ -68,4 +68,43 @@ unpublished builds from the master branch. The official Porter mixins follow
...
+  {
+    "name": "mymixin",
+    "author": "My Name",
+    "description": "A mixin for doing great things",
+    "sourceURL": "https://github.com/org/project/releases/download",
+    "feedURL": "https://my.domain.com/mixins/atom.xml"
+  },
+```
+
+It is okay to leave `feedURL` empty if no atom feed yet exists.
```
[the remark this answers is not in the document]
vdice: Ah, true, in CLI use, one supplies one or the other, not both.  Initially, I was thinking if a mixin actually had both options, both would be listed in its listing.  However, I can see this might be confusing.

Should we gravitate to only displaying `feedURL` or `sourceURL` (this latter to change to just `URL` as you've suggested elsewhere), with priority given to the `feedURL`, for a mixin listing?

Nevertheless, indeed, it would be more useful to document each field individually here.  I can do so.

docs/content/mixin-dev-guide/distribution.md
```diff
@@ -68,4 +68,43 @@ unpublished builds from the master branch. The official Porter mixins follow
...
+Porter maintains a list of remote mixins that users can search via
+`porter mixins search [NAME]`. If no name is supplied, the full listing will be
+returned.  See the help menu for all command options: `porter mixins search -h`.
+
+For example, here we search for a `cowsay` mixin:
+
+```console
+$ porter mixins search cowsay
+Name     Description                        Author           Source URL                                                    Feed URL
+cowsay   A mixin for using the cowsay cli   Porter Authors   https://github.com/deislabs/porter-cowsay/releases/download
```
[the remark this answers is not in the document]
vdice: I'll update the output to, say, the `az` mixin for now.

pkg/mixin/metadata.go
```diff
@@ -12,9 +12,33 @@ type Metadata struct {
...
-// Information from running the version command against the mixin.
+// VersionInfo represents information from running the version command against the mixin.
 type VersionInfo struct {
 	Version string `json:"version"`
 	Commit  string `json:"commit"`
 	Author  string `json:"author,omitempty"`
 }
+
+// RemoteMixinListing represents an informational listing for a remote mixin
+type RemoteMixinListing struct {
```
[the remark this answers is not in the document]
vdice: Good call.  How about `MixinListing`?  `MixinDirectory` could then represent `[]MixinListing`?

Or, soon, `PackageListing` and `PackageDirectory` for generalized mixin and plugin support?

pkg/mixin/remote-mixins/index.json
vdice: Yes, as we've discussed, _if_ we decide to mandate that an atom feed URL be available for any porter mixin intended for broadcast (ideally we'd supply/demonstrate a quick/easy way to host said feed), all of the metadata we'd need could potentially be sourced from the feed XML.  If we go this route, we may want to add a few more fields to our atom feed template; for example, a `subtitle` or `summary` field which could hold a description of the plugin, etc.

pkg/mixin/search.go
vdice: Yea, definitely.  I noticed that when writing this code.  One subtle difference is we allow `args` to have length of 0 (no name supplied).  Perhaps this means one more consolidated helper function... or an option on the one consolidated function.

pkg/mixin/search.go
vdice: Indeed.

docs/content/mixin-dev-guide/distribution.md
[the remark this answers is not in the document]
carolynvs: Thanks, I just transferred cowsay to my user since it's in a general state of brokenness. It will be a good demo anyway of a personal mixin if i ever clean it up. :)

pkg/mixin/metadata.go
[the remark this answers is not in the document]
carolynvs: Ooh yeah PackageListing is great 👍

docs/content/mixin-dev-guide/distribution.md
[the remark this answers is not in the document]
carolynvs: Right now, there aren't any mixins that are distributed by both GH releases and an atom feed. The intention was that really motivated people with multiple mixins would bother to make a feed (us) and people who make a single mixin would use GH releases. So it would end up being one or the other. 

If someone did happen to supply both, we should prefer the atom feed.

pkg/mixin/remote-mixins/index.json
carolynvs: A summary would be really nice 👍

pkg/mixin/search.go
carolynvs: Search may be a one-off. Thanks for calling that out so that I didn't blindly think that it could use the same function as the others!

pkg/porter/mixins.go
vdice: Good call; should be split up now.

pkg/porter/mixins.go
vdice: Good call; updated.

pkg/porter/mixins.go
```diff
@@ -64,6 +66,49 @@ func (p *Porter) ListMixins() ([]mixin.Metadata, error) {
...
+	mixinSearcher := mixin.NewSearcher(box)
+
+	mixinList, err := mixinSearcher.Search(opts)
+	if err != nil {
+		return err
+	}
+	return p.PrintPackages(opts, mixinList)
+}
+
+// PrintPackages prints the provided package list according to the provided options
```
vdice: This should move to the `pkgmgmt` package, but I was thinking we'd do so in a future refactor PR when we intro a plugin directory and searching.  WYDT?

vdice:
@carolynvs this branch has been rebased and all review feedback addressed (I think 😅).  PTAL when convenient.

pkg/porter/mixins.go
carolynvs: If we follow our current structure (printing logic inside of the porter pkg), then SearchMixins would be renamed to SearchPackages, NewSearcher would move to pkgmgmt, but SearchPackages and PrintPackages would stay in pkg/porter. Yeah this can all happen in a folllow up

carolynvs:
Oh gosh! Sorry I missed your messages! SHIP IT! 🚢

PR #889, opened 2020-02-10
feat(creds): Add delete credentials command

### LINKED ISSUE #854: Support for deleting the credentials
What is the recommendation to delete the credentials created part of authoring the build?

**porter credentials** doesn't have an option to delete the existing credentials.

adamreese:
# What does this change
Add command to delete credentials

# What issue does it fix
Closes #854

# Checklist
- [x] Unit Tests
- [x] Documentation

### DIFF
--- a/cmd/porter/credentials.go
+++ b/cmd/porter/credentials.go
@@ -18,7 +18,7 @@ func buildCredentialsCommands(p *porter.Porter) *cobra.Command {
 	cmd.AddCommand(buildCredentialsEditCommand(p))
 	cmd.AddCommand(buildCredentialsGenerateCommand(p))
 	cmd.AddCommand(buildCredentialsListCommand(p))
-	cmd.AddCommand(buildCredentialsRemoveCommand(p))
+	cmd.AddCommand(buildCredentialsDeleteCommand(p))
 	cmd.AddCommand(buildCredentialsShowCommand(p))
 
 	return cmd
@@ -62,18 +62,18 @@ func buildCredentialsGenerateCommand(p *porter.Porter) *cobra.Command {
 		Long: `Generate a named set of credentials.
 
 The first argument is the name of credential set you wish to generate. If not
-provided, this will default to the bundle name. By default, Porter will 
+provided, this will default to the bundle name. By default, Porter will
 generate a credential set for the bundle in the current directory. You may also
 specify a bundle with --file.
 
-Bundles define 1 or more credential(s) that are required to interact with a 
+Bundles define 1 or more credential(s) that are required to interact with a
 bundle. The bundle definition defines where the credential should be delivered
-to the bundle, i.e. at /root/.kube. A credential set, on the other hand, 
-represents the source data that you wish to use when interacting with the 
-bundle. These will typically be environment variables or files on your local 
-file system. 
+to the bundle, i.e. at /root/.kube. A credential set, on the other hand,
+represents the source data that you wish to use when interacting with the
+bundle. These will typically be environment variables or files on your local
+file system.
 
-When you wish to install, upgrade or delete a bundle, Porter will use the 
+When you wish to install, upgrade or delete a bundle, Porter will use the
 credential set to determine where to read the necessary information from and
 will then provide it to the bundle in the correct location. `,
 		Example: `  porter credential generate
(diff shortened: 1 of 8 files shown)


cmd/porter/credentials.go
```diff
@@ -128,19 +128,20 @@ func buildCredentialsListCommand(p *porter.Porter) *cobra.Command {
...
+	return &cobra.Command{
+		Use:   "delete NAME",
+		Short: "Delete a Credential",
+		Long:  `Delete a named credential set.`,
+		PreRunE: func(_ *cobra.Command, args []string) error {
+			return opts.Validate(args)
 		},
-		RunE: func(cmd *cobra.Command, args []string) error {
-			return errors.New("Not implemented")
+		RunE: func(_ *cobra.Command, _ []string) error {
```
vdice: Neat, didn't know about this simplification.

pkg/porter/credentials.go
```diff
@@ -139,15 +140,10 @@ func (p *Porter) GenerateCredentials(opts CredentialOptions) error {
...
 // Validate validates the args provided Porter's credential show command
 func (o *CredentialShowOptions) Validate(args []string) error {
-	switch len(args) {
-	case 0:
-		return errors.Errorf("no credential name was specified")
-	case 1:
-		o.Name = strings.ToLower(args[0])
-	default:
-		return errors.Errorf("only one positional argument may be specified, the credential name, but multiple were received: %s", args)
+	if err := validateCredentialName(args); err != nil {
```
carolynvs: I was gonna comment on this not being a receiver function but it's the notorious validate name that I was just saying to Vaughn we need a common function for across everything. I'm gonna fix that today after your PR goes through because it's buggin' me. 😀

pkg/porter/credentials.go
```diff
@@ -139,15 +140,10 @@ func (p *Porter) GenerateCredentials(opts CredentialOptions) error {
...
+		return err
 	}
-
+	o.Name = strings.ToLower(args[0])
```
carolynvs: Credential set names are allowed by the spec to have upper case letters in the name. I'm not sure why we went with forcing `ToLower` here. 🤔 Seems to work but it's weird. I'm gonna chew on that for a while and figure out if it's a problem, a weird quirk that we can remove? @vdice if you have any context on this, I'm all ears 👂

pkg/porter/credentials.go
```diff
@@ -211,6 +207,43 @@ func (p *Porter) ShowCredential(opts CredentialShowOptions) error {
...
+type CredentialDeleteOptions struct {
+	Name string
+}
+
+// DeleteCredential deletes the credential set corresponding to the provided
+// names.
+func (p *Porter) DeleteCredential(opts CredentialDeleteOptions) error {
+	err := p.Credentials.Delete(opts.Name)
+	if err == crud.ErrRecordDoesNotExist {
+		fmt.Fprintln(p.Out, "credential set does not exist")
```
carolynvs: Currently when this situation happens (Delete something that's not there) we have been only [logging](https://github.com/deislabs/porter/blob/master/CONTRIBUTING.md#logging) this message when p.Debug is set, and then printing to p.Err.

If you have feels about that behavior, it's up for debate, but if you don't, let's change this here to be consistent with the current behavior.

carolynvs:
This looks great! 👍 

I just manually kicked the build because of a hiccup in our build system over the weekend. Going forward builds of `porter` should automatically kick off, and `porter-integration` is something that we would manually trigger only if it is affected by the PR.

pkg/porter/credentials.go
carolynvs: I just verified that we shouldn't be using `strings.ToLower`. It definitely blows up when using plugins (which care about case). Would you update this to take the arg as-is?

pkg/porter/credentials.go
```diff
@@ -210,6 +206,43 @@ func (p *Porter) ShowCredential(opts CredentialShowOptions) error {
...
+type CredentialDeleteOptions struct {
+	Name string
+}
+
+// DeleteCredential deletes the credential set corresponding to the provided
+// names.
+func (p *Porter) DeleteCredential(opts CredentialDeleteOptions) error {
+	err := p.Credentials.Delete(opts.Name)
+	if err == crud.ErrRecordDoesNotExist {
+		fmt.Fprintln(p.Err, "credential set does not exist")
```
carolynvs: Here is an example from another part of Porter that shows what I was trying to say with my previous comment. So we just need the check for `p.Debug` to be consistent.

https://github.com/deislabs/porter/blob/9c49056088dd07b954a0fb6238c6a60f5dfd4940/pkg/pkgmgmt/client/delete.go#L35-L37

PR #911, opened 2020-02-18
Detect documenation only builds

### LINKED ISSUE #623: Short circuit build for documentation changes
When the change is not related to code, we shouldn't have to wait for the entire porter build to go. It should short circuit and complete early.

* Changes to md files at the root of the directory, or to docs should not trigger a full azure pipelines run.

carolynvs:
# What does this change
When only documentation files have been modified, skip running jobs for code so that the build completes quickly.

# What issue does it fix
Closes #623

# Notes for the reviewer
* The porter and porter-integration pipelines will report back but all the jobs inside will have skipped. So they should come back pretty quickly after checking out the code and detecting that it's a doc only change.
* I originally wrote this to compare master..HEAD but azure pipelines checks out the repo in such a way that it doesn't find master. I spent too much time trying to figure that out so I fell back to comparing HEAD~..HEAD instead since on the CI server it's always going to be a merge commit. If you run the script locally it won't work the same. That is something we could improve later if it becomes a problem.
* Once this is merged we can test it out on the plugin tutorial PR.

# Checklist
- [ ] Unit Tests
- [x] Documentation

### DIFF
--- a/build/README.md
+++ b/build/README.md
@@ -9,11 +9,17 @@ See all Porter related [pipelines across all repositories](https://dev.azure.com
 
 Our pipeline is broken into a few discrete builds so that we can control how and when they are triggered:
 
-* **azure-pipelines.release.yml**: Validates canary and tag releases. This can be tested in a pull request 
+* **porter-release: azure-pipelines.release.yml**: Validates canary and tag releases. This can be tested in a pull request 
   using `/azp run porter-release` though steps that require credentials will fail. [View Latest Builds](https://dev.azure.com/deislabs/porter/_build?definitionId=23)
-* **azure-pipelines.install.yml**: Validates our install scripts against canary and tag releases. [View Latest Builds](https://dev.azure.com/deislabs/porter/_build?definitionId=16)
-* **azure-pipelines.pr-automatic.yml**: Validates everything we can without a live environment. [View Latest Builds](https://dev.azure.com/deislabs/porter/_build?definitionId=6)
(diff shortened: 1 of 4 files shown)


build/README.md
```diff
@@ -9,11 +9,17 @@ See all Porter related [pipelines across all repositories](https://dev.azure.com
...
   using `/azp run porter-release` though steps that require credentials will fail. [View Latest Builds](https://dev.azure.com/deislabs/porter/_build?definitionId=23)
-* **azure-pipelines.install.yml**: Validates our install scripts against canary and tag releases. [View Latest Builds](https://dev.azure.com/deislabs/porter/_build?definitionId=16)
-* **azure-pipelines.pr-automatic.yml**: Validates everything we can without a live environment. [View Latest Builds](https://dev.azure.com/deislabs/porter/_build?definitionId=6)
-* **azure-pipelines.pr-manual.yml**: Validates a pull request using a live environment. Requires manual triggering 
+* **porter-check-install: azure-pipelines.install.yml**: Validates our install scripts against canary and tag releases. [View Latest Builds](https://dev.azure.com/deislabs/porter/_build?definitionId=16)
+* **porter: azure-pipelines.pr-automatic.yml**: Validates everything we can without a live environment. [View Latest Builds](https://dev.azure.com/deislabs/porter/_build?definitionId=6)
+* **porter-integration: azure-pipelines.pr-manual.yml**: Validates a pull request using a live environment. Requires manual triggering
```
vdice: Yes!  Thanks for adding the names here... I seem to always forget when commenting `/azp run...`

PR #912, opened 2020-02-18
docs(wiring.md): add docs/example around file parameter usage

vdice:
# What does this change
Adds docs/usage around use of file parameters in a Porter bundle.

# What issue does it fix
N/A
# Notes for the reviewer
N/A
# Checklist
- [ ] Unit Tests
- [x] Documentation

### DIFF
--- a/docs/content/wiring.md
+++ b/docs/content/wiring.md
@@ -25,6 +25,35 @@ You can also provide any other attributes, as specified by the CNAB [parameters]
   default: "wordpress"
 ```
 
+### File Parameters
+
+Porter also enables the use of file parameters in a bundle.
+
+For instance, a bundle might declare a parameter `mytar` of type `file`, to exist at `/root/mytar` in the runtime container:
+
+```yaml
+- name: mytar
+  type: file
+  path: /root/mytar
+```
+
+which can be used in a step like `install`:
+
+```yaml
+install:
+  - exec:
+      description: "Install"
+      command: bash
+      flags:
+        c: tar zxvf /root/mytar
+```
+
+Passing in the local file representing the `mytar` parameter is done similarly to any other parameter:
+
+```console
+$ porter install --param mytar=./my.tar.gz
+```
+
 ## Wiring Parameters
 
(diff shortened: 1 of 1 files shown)


docs/content/wiring.md
```diff
@@ -25,6 +25,35 @@ You can also provide any other attributes, as specified by the CNAB [parameters]
   default: "wordpress"
 ```
 
+### File Parameters
+
+Porter also enables the use of file parameters in a bundle.
+
+For instance, a bundle might declare a parameter `mytar` of type `file`, to exist at `/root/mytar` in the runtime container:
```
carolynvs: I think we should avoid saying "container" when we are referring to the bundle's execution environment at runtime. It's probably not just "bundle runtime" since that refers to what executes the bundle (in spec talk).

I'm not 100% sure if we have been consistent in our docs, or if there is an official term from the CNAB world. Just that avoiding conflating containers and the bundle runtime would be a good idea.

If you have an idea for what to call it, go for it. I know that when I'm talking about it I've said "bundle runtime environment" and "execution environment" and all sorts of things. 🤷 It would be nice to know what's the right term! 😅

docs/content/wiring.md
```diff
@@ -25,6 +25,35 @@ You can also provide any other attributes, as specified by the CNAB [parameters]
...
+```yaml
+install:
+  - exec:
+      description: "Install"
+      command: bash
+      flags:
+        c: tar zxvf /root/mytar
+```
+
+Passing in the local file representing the `mytar` parameter is done similarly to any other parameter:
```
carolynvs: Suggestion: "The syntax to pass a parameter to `porter` is the same for both regular and file parameters:"

carolynvs:
This will pass once #914 is merged. Sorry about that!

carolynvs:
Because the integration test ran and accidentally failed once (due to a bug in my doc build short circuit), and now it's not running and reporting a pass, I need to turn off "enforce for admins" then manually merge, then turn that setting back on.

This is a side-effect of the changes we put into place to prevent direct pushes to master.

PR #916, opened 2020-02-18
Explain why we need credential sets

carolynvs:
# What does this change
* Add explanation for why we can define credential sources in porter.yaml from Slack to a new Q & A section on the Credentials page.
* Rename the page to just "Credentials" to match how other pages are named and update references to the page, adding a redirect for the old name.
* Fix whitespace on inline `code` blocks so they don't wrap

# What issue does it fix
Get answers out of slack and into our docs

# Notes for the reviewer
Make sure Carolyn didn't screw up please. 🙇‍♀ 

# Checklist
- [ ] Unit Tests
- [x] Documentation


docs/content/credentials.md
```diff
@@ -32,3 +34,20 @@ before executing the step.
 Once the bundle finishes executing, the credentials are NOT recorded in the
 bundle instance (claim). Parameters are recorded there so that you can view them
 later using `porter instances show NAME --output json`.
+
+## Q & A
+
+### Why can't the credential source be defined in porter.yaml?
+
+The source of the credentials is specific to each installation of the bundle. An
```
vdice: Perhaps `The source of a credential is specific...` or `The sources of credentials are specific`?

PR #919, opened 2020-02-19
ref(*): generalized package search functionality

vdice:
# What does this change
* Refactor to support generic package search functionality

# What issue does it fix
Closes https://github.com/deislabs/porter/issues/906
# Notes for the reviewer

# Checklist
- [x] Unit Tests
- [ ] Documentation


pkg/mixin/pkgmgmt.go
```diff
@@ -77,3 +78,9 @@ var _ pkgmgmt.PackageMetadata = Metadata{}
 
 // Metadata about an installed mixin.
 type Metadata = pkgmgmt.Metadata
+
+// GetMixinDirectory returns a directory/list of mixins available to install
+func GetMixinDirectory() *packr.Box {
```
carolynvs: suggestion: Since we very recently had a function named this, to help people find the directory where a mixin was installed, do you mind renaming this function to `GetMixinDirectoryListings` or something like that. Just to help avoid mixing up the old function and this new one when reading code?

pkg/pkgmgmt/search.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	if name == "" {
+		sort.Sort(pl)
+		return pl, nil
+	}
+
+	results := PackageList{}
+	for _, p := range pl {
+		if strings.HasPrefix(p.Name, name) {
```
carolynvs: Sorry for not asking the first time around. Neither of these are requests, just thinking out loud about how the search command should behave and am looking for your opinion.

1. Are we sure we want to do a case-sensitive comparison? It may be helpful to use ToLower first and compare when doing a search.

2. Would `string.Contains` be more intuitive for a search than prefix?

pkg/mixin/pkgmgmt.go
vdice: Good call, definitely.

pkg/pkgmgmt/search.go
vdice: 1. Had not thought of this.  I like your idea on sanitizing the name via toLower before search. 👍 
2. Good detail to call out. I had waffled on this; first Contains then HasPrefix.  Thinking on it more, indeed, we wouldn't want to miss, say, a `hashicorp-vagrant` mixin if searching `vagrant`.  I'll switch back to `Contains`.

pkg/pkgmgmt/search_test.go
```diff
@@ -42,6 +42,10 @@ func TestSearch_TestBox(t *testing.T) {
 		name:      "package name single match",
 		pkg:       "az",
 		wantItems: PackageList{fullList[0]},
+	}, {
```
carolynvs: 💯

PR #920, opened 2020-02-19
Feat/add plugin search

vdice:
# What does this change
* Adds plugin search functionality

Docs at https://deploy-preview-920--porter.netlify.com/
# What issue does it fix
Closes https://github.com/deislabs/porter/issues/907
# Notes for the reviewer
Depends on https://github.com/deislabs/porter/pull/919

# Checklist
- [ ] Unit Tests
- [x] Documentation


docs/content/plugins/_index.md
```diff
@@ -8,5 +8,9 @@ Porter. [Plugins are very different from mixins][vs], which give you building
 blocks for authoring bundles. There are a couple [types of plugins][types] and
 each plugin binary may contain multiple implementations.
 
+See the [Discoverability Guide][discoverability] on how to search for available plugins and/or
```
carolynvs: This is great. How about we add a "Find Plugins" link under Plugins in the Nav too?

docs/content/plugins/_index.md
```diff
@@ -8,5 +8,9 @@ Porter. [Plugins are very different from mixins][vs], which give you building
...
+add your own to the list.
+
 [vs]: /mixins-vs-plugins/
 [types]: /plugins/types/
+[discoverability]: /plugins/discoverability.md
```
carolynvs: The link is broken 
```suggestion
[discoverability]: /plugins/discoverability/
```

docs/content/plugins/discoverability.md
```diff
@@ -0,0 +1,45 @@
+---
+title: Plugin Discoverability
```
carolynvs: I feel like "discoverability" is a word that makes sense to us, as developers of Porter for this feature. However it may not be the word that people think about when they are 1) users looking for something to install or 2) creators looking to promote their creation.

How do you feel about calling this page (and the links to it) "Plugin Search" (/plugins/search/? Then we can point people to it by saying "Find a plugin" or "Search for a plugin" and to creators "Promote your plugin" or "List your plugin". Something like that.

I was going to suggest marketplace then hit myself with a frying pan, so I'm better now, thanks. 😀

docs/content/plugins/discoverability.md
```diff
@@ -0,0 +1,45 @@
...
+description: Learn more about searching and broadcasting available plugins for Porter
```
carolynvs: Suggestion: "Find a plugin for Porter and list your own plugin in our search results"

Again just trying to use words that people will use when searching on the web.

docs/content/plugins/discoverability.md
```diff
@@ -0,0 +1,45 @@
...
+---
+
+## Search
+
+Porter maintains a list of plugins that users can search via
+`porter plugin search [NAME]`. If no name is supplied, the full listing will be
```
carolynvs: If you decide to change the help text (above), this should be updated as well to reflect whatever it is called, e.g. query or term. `If no query is supplied`.

docs/content/plugins/discoverability.md
```diff
@@ -0,0 +1,45 @@
...
+returned.  See the help menu for all command options: `porter plugin search -h`.
```
carolynvs: How about instead of telling them to run the help command, let's link them to the help text that is generated and hosted on the site. So when we mention the command above, link it to `/cli/porter_plugin_search/`

docs/content/plugins/discoverability.md
```diff
@@ -0,0 +1,45 @@
...
+
+For example, here we search for an `azure` plugin:
```
carolynvs: "here we search for a plugin with the term 'azure' in the name", just to make it clear that it's not an exact match

docs/content/plugins/discoverability.md
```diff
@@ -0,0 +1,45 @@
...
+```console
+$ porter plugin search azure
+Name    Description                                    Author           URL                                      URL Type
+azure   A plugin for utilizing Azure cloud resources   Porter Authors   https://cdn.porter.sh/plugins/atom.xml   Atom Feed
+```
+
+## Broadcast
+
+To add your plugin to the list, add an entry to the plugin directory
+[list](https://github.com/deislabs/porter/blob/master/pkg/plugins/directory/index.json)
```
carolynvs: I think you could drop "with all the pertinent information fields filled out".

docs/content/plugins/discoverability.md
```diff
@@ -0,0 +1,45 @@
...
+with all the pertinent informational fields filled out.
+
+For instance, a new entry might look like:
```
carolynvs: "For example:"

docs/content/plugins/discoverability.md
```diff
@@ -0,0 +1,45 @@
...
+```json
+  {
+    "name": "myplugin",
+    "author": "My Name",
+    "description": "A plugin for doing great things",
+    "URL": "https://github.com/org/project/releases/download",
+  },
+```
+
+The `URL` field should either be an Atom Feed URL (for example, Porter uses
```
carolynvs: This is a bit hard to read with the parentheticals. I think people could take this in easier with some bullet points. ALL THE BULLET POINTS

The `URL` field should be one of the following:

* **Atom Feed URL:** Porter uses the following...
* **Download URL:** Directory where binaries are hosted, such as GitHub releases `https://github.com/org/project/releases/download`

docs/content/plugins/discoverability.md
```diff
@@ -0,0 +1,45 @@
...
+the following for its stable plugins: `https://cdn.porter.sh/plugin/atom.xml`) or
+a download URL (like the GitHub download URL shown above:
+`https://github.com/org/project/releases/download`)
+
+With this change pushed to a branch on your fork of this repo, you're
+now ready to open up a [Pull Request](https://github.com/deislabs/porter/pulls).
```
carolynvs: I don't think we need to link to our pull request page here since this isn't how they would open a pull request and clicking on it would just take them off of our doc site to an irrelvant page.

```suggestion
now ready to open up a pull request.
```

docs/content/plugins/discoverability.md
```diff
@@ -0,0 +1,45 @@
...
+After the changes are approved, merged and included in the next release, your
```
carolynvs: Does it need to be included in the next release? I thought we were talking about a follow-up PR to have us stop using a box and grab the live data? I think it's okay to document what we want to do since this is going to happen soon.

pkg/plugins/directory/index.json
```diff
@@ -0,0 +1,8 @@
+[
+  {
+    "name": "azure",
+    "author": "Porter Authors",
+    "description": "A plugin for utilizing Azure cloud resources",
```
carolynvs: Let's change the description to make this a bit more clear on how the plugin helps. If you copied that from somewhere, can you let me know so that I can fix it there too?

```suggestion
    "description": "Integrate Porter with Azure. Store Porter's data in Azure Cloud and secure your bundle's secrets in Azure Key Vault.",
```

cmd/porter/plugins.go
```diff
@@ -48,6 +49,31 @@ func buildPluginsListCommand(p *porter.Porter) *cobra.Command {
 	return cmd
 }
 
+func buildPluginSearchCommand(p *porter.Porter) *cobra.Command {
+	opts := porter.SearchOptions{}
+
+	cmd := &cobra.Command{
+		Use:   "search [NAME]",
+		Short: "Search available plugins",
```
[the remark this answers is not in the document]
vdice: 👍 I believe I've now done this for plugin and mixin search commands.

docs/content/mixin-dev-guide/distribution.md
```diff
@@ -80,7 +80,7 @@ Porter maintains a list of remote mixins that users can search via
 `porter mixins search [NAME]`. If no name is supplied, the full listing will be
 returned.  See the help menu for all command options: `porter mixins search -h`.
 
-For example, here we search for a `cowsay` mixin:
+For example, here we search for an `az` mixin:
```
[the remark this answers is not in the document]
vdice: 🐄 🗣 will return stronger than ever before!

docs/content/plugins/_index.md
vdice: Oops, good reminder. Should be added now.

docs/content/plugins/discoverability.md
vdice: I consolidated the docs for search/broadcast of mixins and plugins into one doc, that both relevant mixin/plugin docs can link to.  What do you think?

I have yet to update terminology from `broadcast`... shall I change to `Promote` or `List` as mentioned above?

docs/content/plugins/discoverability.md
vdice: * Good call
* Definitely reads better
* Now featuring bullet points!

docs/content/plugins/discoverability.md
vdice: True; section removed/wording updated.

docs/content/plugins/discoverability.md
carolynvs: "promote" feels icky. Let's go with list.

cowsay we all

vdice:
Thank you for all the great improvements @carolynvs !

I *think* I've addressed them all.  PTAL when convenient.

docs/content/plugins/discoverability.md
carolynvs: I took a technical writing class in college to meet a writing requirement to graduate. While everyone else wrote 20pg papers on great english literature, I learned that people don't read, use moar bullet points. 😀

PR #921, opened 2020-02-20
feat(creds): Add edit credentials command

### LINKED ISSUE #890: Support porter credentials edit command
Now that credentials are stored in remote places, the cloud! maybe a database!, not always on your file system, we need a command to help people easily locate and edit their credential sets.

```console
$ porter credentials edit NAME
```

* Should read the credential from the plugin system
* Should open the credential using $EDITOR. We may need to be sneaky and save it to a temp file and then have the editor open the temp file. I'm not sure what needs to happen here, so let's look at other tools to see what they do. EDITOR doesn't have to be a text based UI like emacs, nano or vim. It could be VS Code or Notepad.
* I have no idea how to tell when they are done editing it. How does kubectl do this magic? Let's look at other tools again and get inspired. 😀 
* When they are done editing, the credential set should be saved through the plugin system again.

Porter has a field, `Credentials`, that you should use to read and save credentials. This handles the plugin system for you automatically.

https://github.com/deislabs/porter/blob/a533aa84264ccf5cc490df096e72726084fc063d/pkg/credentials/credentialProvider.go#L8-L20

halkyon:
# What does this change
Adds `porter credentials edit` command allowing the user to edit their credentials.

# What issue does it fix
#890

# Checklist
- [ ] Unit Tests
- [ ] Documentation

### DIFF
--- a/pkg/porter/credentials.go
+++ b/pkg/porter/credentials.go
@@ -181,12 +181,12 @@ func (p *Porter) EditCredential(opts CredentialEditOptions) error {
 
 	err = yaml.Unmarshal(output, &credSet)
 	if err != nil {
-		return errors.Wrap(err, "unable to load credentials")
+		return errors.Wrap(err, "unable to process credentials")
 	}
 
 	err = p.Credentials.Validate(credSet)
 	if err != nil {
-		return err
+		return errors.Wrap(err, "credentials are invalid")
 	}
 
 	credSet.Modified = time.Now()
--- a/pkg/porter/credentials.go
+++ b/pkg/porter/credentials.go
@@ -184,6 +184,11 @@ func (p *Porter) EditCredential(opts CredentialEditOptions) error {
 		return errors.Wrap(err, "unable to load credentials")
 	}
 
+	err = p.Credentials.Validate(credSet)
+	if err != nil {
+		return err
+	}
+
 	credSet.Modified = time.Now()
 	err = p.Credentials.Save(credSet)
 	if err != nil {
--- a/pkg/porter/credentials.go
+++ b/pkg/porter/credentials.go
@@ -7,6 +7,7 @@ import (
 
 	"get.porter.sh/porter/pkg/context"
 	"get.porter.sh/porter/pkg/credentialsgenerator"
+	"get.porter.sh/porter/pkg/editor"
 	"get.porter.sh/porter/pkg/printer"
 	"gopkg.in/yaml.v2"
 
@@ -171,61 +172,26 @@ func (p *Porter) EditCredential(opts CredentialEditOptions) error {
 		return errors.New("no credentials exist in this set")
 	}
 
-	// use a flattened data structure for the credentials, so when we marshal
-	// the structure to yaml it's easier for the user to edit.
-	type tempStrategy struct {
-		Name        string `yaml:"name"`
-		SourceType  string `yaml:"type"`
-		SourceValue string `yaml:"value"`
-	}
-	type tempCredSet struct {
-		Credentials []tempStrategy `yaml:"credentials"`
-	}
-	tempSet := &tempCredSet{}
-	for _, cs := range credSet.Credentials {
-		tempSet.Credentials = append(tempSet.Credentials, tempStrategy{
-			Name:        cs.Name,
-			SourceType:  cs.Source.Key,
-			SourceValue: cs.Source.Value,
-		})
-	}
-
-	data, err := yaml.Marshal(tempSet)
+	data, err := yaml.Marshal(credSet)
 	if err != nil {
-		return err
+		return errors.Wrapf(err, "unable to load credentials")
 	}
 
-	editor := &Editor{
-		Data:     []byte(data),
-		Filename: fmt.Sprintf("porter-%s", credSet.Name),
-		Context:  p.Context,
-	}
-
+	editor := editor.New(p.Context, fmt.Sprintf("porter-%s", credSet.Name), []byte(data))
 	output, err := editor.Run()
 	if err != nil {
-		return err
+		return errors.Wrapf(err, "unable to open editor to edit credentials")
 	}
 
-	err = yaml.Unmarshal(output, tempSet)
+	err = yaml.Unmarshal(output, &credSet)
 	if err != nil {
-		return err
+		return errors.Wrapf(err, "unable to load credentials")
 	}
 
-	newCredentials := []credentials.CredentialStrategy{}
-	// todo: validate for correct source types
-	for _, tcs := range tempSet.Credentials {
-		cs := credentials.CredentialStrategy{}
-		cs.Name = tcs.Name
-		cs.Source.Key = tcs.SourceType
-		cs.Source.Value = tcs.SourceValue
-		newCredentials = append(newCredentials, cs)
-	}
-
-	credSet.Credentials = newCredentials
 	credSet.Modified = time.Now()
 	err = p.Credentials.Save(credSet)
 	if err != nil {
-		return err
+		return errors.Wrapf(err, "unable to save credentials")
 	}
 
 	return nil
--- a/pkg/porter/credentials.go
+++ b/pkg/porter/credentials.go
@@ -3,13 +3,12 @@ package porter
 import (
 	"encoding/json"
 	"fmt"
-	"strings"
 	"time"
 
 	"get.porter.sh/porter/pkg/context"
 	"get.porter.sh/porter/pkg/credentialsgenerator"
 	"get.porter.sh/porter/pkg/printer"
-	"gopkg.in/AlecAivazis/survey.v1"
+	"gopkg.in/yaml.v2"
 
 	dtprinter "github.com/carolynvs/datetime-printer"
 	credentials "github.com/cnabio/cnab-go/credentials"
@@ -162,8 +161,6 @@ func (o *CredentialEditOptions) Validate(args []string) error {
 }
 
 // EditCredential edits the credentials of the provided name.
-// TODO: Allow piping in a credential from stdin?
-// TODO: Allow passing another flag for the specific credential in the set instead of surveying?
 func (p *Porter) EditCredential(opts CredentialEditOptions) error {
 	credSet, err := p.Credentials.Read(opts.Name)
 	if err != nil {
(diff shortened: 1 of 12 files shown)


pkg/porter/editor.go
```diff
@@ -0,0 +1,65 @@
...
+	_, err = tempFile.Write(data)
+	if err != nil {
+		return nil, err
+	}
+
+	// close here without defer so the next command can grab the file
+	tempFile.Close()
+
+	// todo: ensure editor command with spaces works here
+	// todo: other editors like Visual Studio Code don't seem to work with this, and don't block execution until the
```
carolynvs: Looks like vs code has a `--wait` flag to help out using it as an editor for git commit messages and we can use that for our purposes too. So if someone was using vs code as their editor, it would look like this `export EDITOR="code --wait"` or they would have this in their ~/.gitconfig 

```
[core]
        editor = code --wait
```

https://stackoverflow.com/questions/30024353/how-to-use-visual-studio-code-as-default-editor-for-git

carolynvs:
Oh one thing I just realized after watching the gif (thanks for that, very cool 👍) is that this doesn't let the user change where the credential is found. For example, when they generated the credential set, they said "find my database password in an environment variable Then later they need to edit to to find it in another source, for example in a secret store, so they need to change the CredentialSource key from `env` to `secret`.

It would be preferable if either they can edit the entire CredentialSource (both key and value, right now this only lets them set the value) or even the entire CredentialSet at once as an option.

halkyon:
> 
> This is looking exactly like what I was hoping for!
> 
> I think this one may be difficult to unit test, due to the editor part. But I do have some tricks that I can help out with that we have used in other parts.
> 
> We use a helper method that needs to be present in the package (it gives us a test main that mocks out the executable being called on the file system)
> 
> https://github.com/deislabs/porter/blob/master/pkg/exec/main_test.go
> 
> Here's a test that uses it to verify that a command as called and that it received the expected arguments.
> 
> https://github.com/deislabs/porter/blob/master/pkg/exec/exec_test.go#L42
> 
> I see you still have this PR in draft still. Are you ready for a full review or do you still have more work/questions that you have first?

Thanks for the feedback Carolyn!

I still have some work to do, and to give it a good test before submitting it for review :) thanks for the hints on the unit tests, that's something I still need to do as well, and some manual testing on Windows. Speaking of which, I was having some troubles compiling porter on Windows, do you have a few pointers on doing that? Maybe there was a doc I missed, I'm not sure!

Good point on editing the entire credential set at once including the source. Do you think editing the entire set at once in a text representation, sort of like `git rebase -i` for interactively editing commits would be suitable, or am I overlooking a much simpler way of doing it?

pkg/porter/editor.go
halkyon: This is great, I didn't know about `--wait`. Thanks!

carolynvs:
Dang we really need to [support compiling on Windows](https://github.com/deislabs/porter/issues/115). Right now the way to test on Windows is to run `make xbuild-all` and then copy the resulting windows binary from your bin to a Windows machine.

What you just showed, opening the entire credential set in yaml is what I had originally imagined how the feature would work. 👍

halkyon:
Ah, gotcha! I did find `make xbuild-all` after poking around the CI to see how it builds for Windows :)

Good to see we're thinking the same things so far! I'll take the PR out of draft so you can have a review if you like? I've tested on Linux and Windows, as well as setting Visual Studio code as an editor with the `--wait` option. All seems to work pretty well. Some feedback would be good if you don't mind having a look :)

pkg/porter/credentials.go
```diff
@@ -146,6 +151,86 @@ func (o *CredentialShowOptions) Validate(args []string) error {
...
+		return err
+	}
+
+	if len(credSet.Credentials) == 0 {
+		return errors.New("no credentials exist in this set")
+	}
+
+	// use a flattened data structure for the credentials, so when we marshal
+	// the structure to yaml it's easier for the user to edit.
+	type tempStrategy struct {
```
carolynvs: Is the intent behind the custom types to remove the extra level of indentation? e.g. 

```yaml
source:
  env:...
```

Or is it to not have they use the key value pair with a dynamic key?

```yaml
env: MY_ENV_VAR
```

We could alter the format of the credential set, though anyone who looks at the data on disk or in the cloud would be confused about why it looks different. I'd be okay with that if we had a clear reason on why we needed to hide/alter that for the user in this editing mode.

I am concerned that the words we have chosen may not map well to how people think about the credential based on how they made it, e.g. "type". They aren't different types of credentials, and when we asked them to create it we didnt' use that word. But before bikeshedding, I'd like to make sure we really want to alter the structure from the user in the first place.

pkg/porter/credentials.go
```diff
@@ -146,6 +151,86 @@ func (o *CredentialShowOptions) Validate(args []string) error {
...
+		return err
+	}
+
+	err = yaml.Unmarshal(output, tempSet)
+	if err != nil {
+		return err
+	}
+
+	newCredentials := []credentials.CredentialStrategy{}
+	// todo: validate for correct source types
```
carolynvs: Right now we don't have a way to dynamically validate, so you will need to compare against a known list, if you look in the credential generator you can find it.

pkg/porter/editor.go
```diff
@@ -0,0 +1,86 @@
...
+	"path/filepath"
+	"runtime"
+	"strings"
+
+	"get.porter.sh/porter/pkg/context"
+)
+
+const (
+	defaultEditor        = "vi"
+	defaultEditorWindows = "notepad"
```
carolynvs: Here's another way to do this that doesn't require a runtime check.

I use build tags at the top of the file to make sure that binaries built for windows get the right constant and binaries for other OS's (I'm calling that *nix) get the other definition of the same constant.

https://github.com/deislabs/porter/blob/041c93a33f0bd3eeb1b203f950bb8feb6a481ce7/pkg/pkgmgmt/pkgmgmt_windows.go

https://github.com/deislabs/porter/blob/041c93a33f0bd3eeb1b203f950bb8feb6a481ce7/pkg/pkgmgmt/pkgmgmt_nix.go

Then in the rest of your code, you can reference `defaultEditor` and not need to check at runtime which OS you are on.

pkg/porter/editor.go
```diff
@@ -0,0 +1,86 @@
...
+)
+
+// Editor displays content to a user using an external text editor, like vi or notepad.
+// The content is captured and returned.
+//
+// The `EDITOR` environment variable is checked to find an editor.
```
carolynvs: As a follow-up issue, we can allow the user to set `editor` in our [config file](https://porter.sh/configuration/) and use that before checking the environment variable. I can make an issue to do that once your PR is merged.

pkg/porter/editor.go
```diff
@@ -0,0 +1,86 @@
...
+	editor := defaultEditor
+	if runtime.GOOS == "windows" {
+		editor = defaultEditorWindows
+	}
+	if os.Getenv("EDITOR") != "" {
+		editor = os.Getenv("EDITOR")
+	}
+
+	args := []string{}
+	if strings.ContainsAny(editor, "\"'\\") {
```
carolynvs: Can you change this to a raw string so that it's easier to tell what characters you are testing for? A comment explaining what is going on with an example of the args and why you are looking for those characters and then splitting on space would really help too.

I think you are looking for `\`, `"` and `'`
```suggestion
	if strings.ContainsAny(editor, `"'\`) {
```

pkg/porter/editor.go
```diff
@@ -0,0 +1,86 @@
...
+// The `EDITOR` environment variable is checked to find an editor.
+// Failing that, use some sensible default depending on the operating system.
+//
+// This is useful for editing things like configuration files, especially those
+// that might be stored on a remote server. For example: the content could be retrieved
+// from the remote store, edited locally, then saved back.
+type Editor struct {
+	Data     []byte
+	Filename string
+	Context  *context.Context
```
carolynvs: The pattern we use elsewhere is to embed the context struct so that we can access its members without having to do `x.Context.FileSystem` or `x.Context.Out`, and can instead do `x.Filesystem` and `x.Out`, `x.Debug`, etc. So let's change this to be consistent.

```suggestion
	*context.Context
```

pkg/porter/editor.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+// Run opens the editor, displaying the content through a temporary file.
+// The content is returned once the editor closes.
+func (e *Editor) Run() ([]byte, error) {
+	tempFile, err := os.OpenFile(filepath.Join(os.TempDir(), e.Filename), os.O_RDWR|os.O_CREATE|os.O_EXCL, 0600)
+	if err != nil {
+		return nil, err
+	}
+	defer os.Remove(tempFile.Name())
```
carolynvs:
```suggestion
	defer e.FileSystem.Remove(tempFile.Name())
```

pkg/porter/editor.go
```diff
@@ -0,0 +1,86 @@
...
+	cmd := e.Context.NewCommand(args[0], args[1:]...)
+	cmd.Stdout = e.Context.Out
+	cmd.Stderr = e.Context.Err
+	cmd.Stdin = e.Context.In
+	err = cmd.Run()
+	if err != nil {
+		return nil, err
+	}
+
+	data, err := ioutil.ReadFile(tempFile.Name())
```
carolynvs:
```suggestion
	data, err := e.FileSystem.ReadFile(tempFile.Name())
```

pkg/porter/editor.go
```diff
@@ -0,0 +1,86 @@
...
+
+	args := editorArgs()
+	args = append(args, tempFile.Name())
+	cmd := e.Context.NewCommand(args[0], args[1:]...)
+	cmd.Stdout = e.Context.Out
+	cmd.Stderr = e.Context.Err
+	cmd.Stdin = e.Context.In
+	err = cmd.Run()
+	if err != nil {
+		return nil, err
```
carolynvs: Another pattern we use is that anytime an error occurs that wasn't generated by Porter, we wrap it using `github.com/pkg/errors` and give it a user friendly error message providing context about the current operation since it will be displayed directly to the user and is often our only lead when people report errors.

Something like `"error opening credential set %s in editor %s"`

pkg/porter/credentials_test.go
```diff
@@ -427,6 +429,38 @@ func TestGetCredentialSourceValueAndType(t *testing.T) {
 	}
 }
 
+func TestCredentialsEdit(t *testing.T) {
+	os.Setenv("EDITOR", "vi")
+	os.Setenv(test.ExpectedCommandEnv, "vi "+filepath.Join(os.TempDir(), "porter-kool-kreds"))
```
carolynvs: 💯

pkg/porter/credentials.go
halkyon: Primary reason was to simplify the structure, a bit of both of those reasons. But as you say, altering this will be confusing to users, and I totally agree, I was a bit unsure of this in the first place.

pkg/porter/credentials.go
halkyon: Sounds good, I'll use that for the validation.

pkg/porter/editor.go
halkyon: Good call, that would make this a bit cleaner. Will fix this up. Thanks!

pkg/porter/editor.go
halkyon: Sounds good.

pkg/porter/editor.go
halkyon: Will do.

pkg/porter/editor.go
halkyon: Sounds good, will do that!

pkg/porter/editor.go
halkyon: I was wondering what those were about! Thanks for the heads up, I'll sort those errors out.

pkg/porter/credentials.go
carolynvs: How about for now let's directly expose the underlying credential set structure and then pay attention to how well that is working for people. If we get feedback that the language (yaml) or the structure is making it easy to make mistakes, we can iterate and provide a simplified structure for them to edit, or consider tweaking the language (indenting is hard, no joke).

pkg/porter/credentials.go
halkyon: Question around this: should credentials.EditCredentials be responsible for validating credentials, or should that go somewhere upstream, like the credential provider? I was looking at potentially calling Resolve from here to validate the sources, but I'm not 100% sure. Thoughts?

carolynvs: Calling Resolve wouldn't do what you are hoping for, since that would actually try to find a value that matches the credential, for example an environment variable named "AZURE_CLIENT_ID", and if one can't be found, it would fail.

Yes, this could be a new function on the CredentialProvider interface, `Validate`, and then implemented on CredentialStorage. Then for now we can check against the set of constants that we know are valid source keys.

halkyon:
@carolynvs Thanks for all the feedback, should be sorted out now. Would you mind doing another review? Hopefully it's getting close to merge 😄

Issue #325, opened 2019-05-08
Credentials UX Improvements


jeremyrickard:
There is a WIP effort to add the ability to generate, list and otherwise manage credentials with Porter. We should take this a step further and enhance the UX even more by enhancing the install / upgrade/ delete commands to allow the Porter user to select from an existing credential if one is not specified, something along the lines of:

```
$ porter install
? Choose a set of credentials to use while installing this bundle
   jerrycar-test
> jerrycar-admin
  jerrycar-readonly
```

If no credentials exist, Porter should help you out there as well

```
$ porter install
INFO: No credentials found, let's generate one! 
For more information about credentials, see https://porter.sh/credentials
? Choose a source for "kubeconfig"  [Use arrows to move, type to filter]
  specific value
> environment variable
  file path
  shell command
```

The examples above use the existing survey text from Duffle. We can start with that and iterate to improve the messaging. 

Porter should continue to allow you to specify a credential as well:

```
$ porter install -c jerrycar-admin 
```

carolynvs:
Here were some notes from our chat yesterday:
* When we provide a list of creds, only show ones that match the shape of the creds in the bundle
* Prompt to generate creds when there aren't any existing credentials in PORTER_HOME that match the shape of the creds in the bundle
* Don't do any prompts when they do specify a credential set and allow them to specify multiple sets to build up something that meets what the bundle needs (e.g. the azure creds come from one credential set and the kubeconfig may come from another).

vdice:
Note that it sounds like https://github.com/deislabs/porter/issues/226 will be covered by the implementation mentioned above.

carolynvs:
We are no longer going in the direction of more interactive commands, and this isn't on our roadmap anymore. Closing

Issue #877, opened 2020-02-04
Don't require credentials when applyTo makes them unnecessary


carolynvs:
For this porter.yaml, it would be really useful if credentials had applyTo, just like parameters and outputs. Then I could define my credentials as only needed for install:

```yaml
credentials:
- name: db-password
  env: DB_PASSWORD
  type: string
  applyTo:
  - install
```

Then porter could inspec the credentials when deciding if they are required and take `applyTo` and the current action into account when enforcing if credentials are required and if the specified credential set satisfies them.

This would require changes to the cnab-spec and cnab-go before we could implement it in porter.

carolynvs:
Following the code, it appears that the fix could be done in cnab-go.

First this is the code that is executed when we try to load credentials before executing a bundle and none were provided:

https://github.com/deislabs/porter/blob/f92186310727f3d039ea72c82ada6eeaf63e58b1/pkg/cnab/provider/credentials.go#L11-L14

and here's the cnab-go code that is failing to check `applyTo`:

https://github.com/cnabio/cnab-go/blob/d9edf60d03325f5afeef5f5f7a3285a9497c40d9/credentials/credentialset.go#L90-L97

PR #922, opened 2020-02-20
Prompt credential select while performing action if not provided using -c

### LINKED ISSUE #226: porter install with interactive drop into credentials generate

dev-drprasad:
# What does this change

Prompt credential select while performing an action if not provided using -c

# What issue does it fix
Closes #226
Related #325 

# Notes for the reviewer

Initial menu
Screenshot 2020-03-26 at 10 16 07 PM

if some creds presents (default to bundle name)
Screenshot 2020-03-26 at 10 15 15 PM

if some creds presents (no cred name with bundle name)
Screenshot 2020-03-26 at 10 15 56 PM

generating new cred set
Screenshot 2020-03-26 at 10 17 46 PM

if no credentials present
Screenshot 2020-03-26 at 10 30 21 PM

on quit
Screenshot 2020-03-26 at 10 33 26 PM

on ctrl+c in select cred menu
Screenshot 2020-03-26 at 10 34 36 PM

# Checklist
- [ ] Unit Tests
- [ ] Documentation


dev-drprasad:
I need few days to complete this PR as i have some personal work. I will get back to you in march 1st week

pkg/porter/install.go
```diff
@@ -8,6 +8,9 @@ import (
 	"github.com/pkg/errors"
 )
 
+const generateCredCode = "generate new credential set"
+const quitGenerateCode = "quit"
```
dev-drprasad: What do you think of variable names ?

dev-drprasad:
Unfortunately, `porter credentials generate` without `-c` don't prompt for credential name. Because of following logic
https://github.com/deislabs/porter/blob/6da074e0114723c050ce84b6025cba883718f702/pkg/porter/options.go#L6-L20

What should we do about it ?

vdice:
> Unfortunately, porter credentials generate without -c don't prompt for credential name.

@dev-drprasad It looks like the credential name defaults to the bundle name?  Does this prove problematic?  Perhaps for this PR we should see if a credential set with the bundle name exists and if not, we run through the credential select logic?

go.mod
```diff
@@ -37,6 +37,7 @@ require (
 	github.com/xeipuuv/gojsonschema v1.2.0
 	gopkg.in/AlecAivazis/survey.v1 v1.8.7
 	gopkg.in/yaml.v2 v2.2.4
+	gotest.tools v2.2.0+incompatible
```
carolynvs: I'm pretty sure this line shouldn't be here, I bet you meant to use testify assert instead somewhere and instead the IDE pulled in this. This happens to me all the time now with go mods, I usually can just remove this line and commit the file and everything is fine.

pkg/porter/install.go
```diff
@@ -32,6 +35,14 @@ func (p *Porter) InstallBundle(opts InstallOptions) error {
 		return err
 	}
 
+	if len(opts.CredentialIdentifiers) == 0 {
```
vdice: I'd vote to add this same logic to the other commands (upgrade, uninstall, custom action)... I think @carolynvs was mentioning the same?

dev-drprasad:
> > Unfortunately, porter credentials generate without -c don't prompt for credential name.
> 
> @dev-drprasad It looks like the credential name defaults to the bundle name? Does this prove problematic? Perhaps for this PR we should see if a credential set with the bundle name exists and if not, we run through the credential select logic?

`porter credentials generate` use bundle name if no cred name passed explicitly. But `porter install` ask for cred name and if user `enter` with nothing then it will use bundle name. As you can see there is inconsistency. I just want to point out that. To get same behaviour,`applyDefaultOptions` has to be modified but that will be a big refactor as it is used in many places

dev-drprasad:
> Sorry I don't have time tonight for a full review. I'll review this tomorrow for sure!
> 
> One thing that I realized is that the original issue was ... missing a lot of detail. So thanks for filling in the gaps. If we don't put it into this PR, we should have a follow-up issue to have this behavior added to all commands upgrade/uninstall/invoke as well. People forget all the time to specify `-c` for those as well and this will be very helpful for that.

Lets add that logic for all commands in this PR only. I don't see why not

dev-drprasad:
I just realized there can be multiple `--cred`/`-c` args. Of course, we can handle it as survey supports multi select. Let me add that logic

carolynvs:
> porter credentials generate use bundle name if no cred name passed explicitly. But porter install ask for cred name and if user enter with nothing then it will use bundle name. As you can see there is inconsistency.

I'd like to explain why these two commands act differently. It has to do with security and what data is exposed by the default behavior. 

With `porter credentials generate`, when the default behavior names the credential after the bundle, nothing sensitive is exposed. Someone's lack of awareness of the default behavior can't accidentally cause a leak of sensitive information.

With `porter install`, the current default behavior is to require the user to explicitly pass in the credentials sets for use by the bundle. If they forget to specify a credential, porter doesn't execute it and they are able to correct the mistake and rerun the command.

If we modified the default behavior of `porter install` to act like `porter credentials generate`, where if no `-c` flag was specified, this can cause unintended exposure of sensitive data (credentials). If a user, unaware of the default behavior or accidentally omitting the `-c` flag, triggers this new default behavior, then credentials that they didn't explicitly specify (and say say are okay to give to the bundle) are retrieved and injected into the bundle. 

For example, if the bundle was named very generic, maybe after a cloud provider like `aws` or `azure`, or perhaps it's named `dev` or `staging` (remember the name of the bundle doesn't have to match the tag) it could end up being given someone's cloud credentials that they created to pass to some bundles, but did not intend to pass to THAT bundle. It's a contrived example but should help demonstrate how the name of the bundle should not mean that the credentials are meant for that bundle.

That isn't a default behavior that we consider to be safe. 😀

pkg/porter/credentials.go
```diff
@@ -106,22 +107,91 @@ func (p *Porter) GenerateCredentials(opts CredentialOptions) error {
...
+	var selectedOption string
+	if len(credSets) > 0 {
+		for _, credSet := range credSets {
+			credSetNames = append(credSetNames, credSet.Name)
+		}
+
+		selectCredPrompt := &survey.Select{
+			Message: "Choose a set of credentials to use while installing this bundle",
+			Options: append(credSetNames, generateCredCode, quitGenerateCode),
+			Default: credSetNames[0],
```
carolynvs: In a follow-up, we should consider checking if the any of the credential sets has the same name as the bundle and suggest that as the default here.

I realize that I gave a very long reply about security using it as the _automatic_ default when no credential is provided but I believe it is a safe default in interactive mode and will get people closer to what they most likely wanted when there is a match.

pkg/porter/credentials.go
```diff
@@ -106,22 +107,91 @@ func (p *Porter) GenerateCredentials(opts CredentialOptions) error {
...
+		}
+	} else {
+		shouldGenerateCredPrompt := &survey.Confirm{
+			Message: "No credential identifier given. Generate one ?",
+		}
+		survey.AskOne(shouldGenerateCredPrompt, &shouldGenerateCred, nil)
+	}
+
+	if !shouldGenerateCred {
+		return "", errors.New("Credentials are mandatory to install this bundle")
```
carolynvs:
```suggestion
		return "", errors.New("Credentials are mandatory to install this bundle but none were provided with the `--cred` flag")
```

I think it would help to tell them what to do to fix the command when they run it again. Both here and above.

pkg/porter/credentials.go
```diff
@@ -106,22 +107,91 @@ func (p *Porter) GenerateCredentials(opts CredentialOptions) error {
...
+}
+
+func (p *Porter) generateNewCredentialSet(CNABFile string, credIdentifierName string, silent bool, dryRun bool) (string, error) {
+	bundle, err := p.CNAB.LoadBundle(CNABFile)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to load bundle while generating credentials")
 	}
+
+	if credIdentifierName == "" {
+		if !silent {
```
carolynvs: Just a suggestion but it's usually easier to read the positive case `if silent` than the inverse, especially when there is an else clause. Since you do have an action for when silent is true, it would be easier to follow if this was `if silent {  // logic for silent case } else { // logic for not silent }`.

Up to you if you want to switch it up.

pkg/porter/credentials.go
```diff
@@ -106,22 +107,91 @@ func (p *Porter) GenerateCredentials(opts CredentialOptions) error {
...
-		Name:        name,
+		Name:        credIdentifierName,
 		Credentials: bundle.Credentials,
-		Silent:      opts.Silent,
+		Silent:      silent,
+		DryRun:      dryRun,
+	}
+
+	fmt.Fprintf(p.Out, "  Generating new credential %s from bundle %s\n", genOpts.Name, bundle.Name)
+	fmt.Fprintf(p.Out, "  %d credentials required for bundle %s\n", len(genOpts.Credentials), bundle.Name)
```
carolynvs: How come there is now a two space indent for these lines?

pkg/porter/credentials.go
```diff
@@ -106,22 +107,91 @@ func (p *Porter) GenerateCredentials(opts CredentialOptions) error {
...
+	err = p.generateAndSaveCredentialSet(genOpts)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to generate credentials")
 	}
-	fmt.Fprintf(p.Out, "Generating new credential %s from bundle %s\n", genOpts.Name, bundle.Name)
-	fmt.Fprintf(p.Out, "==> %d credentials required for bundle %s\n", len(genOpts.Credentials), bundle.Name)
+	fmt.Fprintln(p.Out, "Credentials generated and saved successfully for future use")
```
carolynvs: We don't usually write such verbose success messages (or at all really). I'd either remove it or keep it short with

```suggestion
	fmt.Fprintf(p.Out, "Credentials saved as %s\n", credIdentifierName)
```

pkg/porter/credentials.go
```diff
@@ -140,7 +210,11 @@ func (p *Porter) GenerateCredentials(opts CredentialOptions) error {
 		return nil
 	}
 	err = p.Credentials.Save(*cs)
-	return errors.Wrapf(err, "unable to save credentials")
+	if err != nil {
```
carolynvs: I'm not sure why you changed this line? `Wrapf` returns nil when err is nil.

pkg/porter/install.go
carolynvs: One critical part of this check that is missing right now is that we don't want to prompt for credentials unless credentials are needed by the bundle for this action. Otherwise we would bring up this prompt when a credential isn't required.

I don't want to compound the problems that we are already facing in #877.

For example, the hello world bundle that we have people run doesn't have credentials and this change as-is would force them to create credentials. So as a litmus test, let's make sure that the solution allows someone to run the [quickstart](https://porter.sh/quickstart/#install-from-the-registry) without modification to use credentials.

pkg/porter/resolver.go
```diff
@@ -10,8 +10,8 @@ type BundleResolver struct {
 	Registry Registry
 }
 
-// Resolves a bundle from the cache, or pulls it and caches it
-// Returns the location of the bundle or an error
+// Resolve resolves a bundle from the cache, or pulls it and caches it
```
carolynvs:
```suggestion
// Resolve a bundle from the cache, or pulls it and caches it
```

carolynvs:
This is looking really good! 💯 I think we just have 1-2 things to figure out first before we can merge.

pkg/porter/credentials.go
dev-drprasad: My Bad

pkg/porter/credentials.go
dev-drprasad: couldn't agree more

pkg/porter/credentials.go
dev-drprasad: 👍

pkg/porter/credentials.go
dev-drprasad: Those extra two spaces to align text with survey prompt questions. You know, I work as UI Engineer :)

Screenshot 2020-03-12 at 4 45 09 PM

I will also add comment if you agree to keep those spaces.

carolynvs: I apologize, this is super nit-picky. I only asked at first because I really didn't understand where the spaces were coming from.

I believe when the user generates credentials and does specify the name for the credential set, e.g. `porter credentials generate aws` then the extra indent may not be what you desired:

```
$ porter porter credentials generate aws
  Generating new credential aws...
  4 credentials required for bundle mybuns...
? How would you like to set credential 1?
```

To me, the indent only makes sense if it's aligning text that always go with that question. But since they are independent, the alignment is more confusing and disjointed. What do you think?

dev-drprasad: I forgot about `porter credentials generate aws`. That's good catch. Will remove those spaces

dev-drprasad:
@carolynvs @vdice Before doing code review, I would suggest you to go through latest screenshots and give feedback on user experience

test cases are missing as i am not sure how to test survey's select/multiselect menus. For sure, `survey` is not writing anything to `p.Out` or `p.Err`. I think its directly writing text to `os.Stdin`, i am not sure. Any idea how to test this feature ?

pkg/porter/credentials.go
```diff
@@ -106,22 +112,112 @@ func (p *Porter) GenerateCredentials(opts CredentialOptions) error {
...
+	switch chooseOrGenerate {
+	case generateCode:
+		newCredSet, err := p.generateNewCredentialSet(bundle, "", false, false)
+		return []string{newCredSet}, err
+	case quitCode:
+		return []string{}, errors.New("Credentials are mandatory to install this bundle but none were provided with the `--cred` flag")
+	case chooseCode:
+		return p.chooseCredentialSet(bundle, credSetNames)
+	default:
+		return []string{}, errors.New("Unknow option")
```
vdice, line 160:
```suggestion
		return []string{}, errors.New("Unknown option")
```

pkg/porter/credentials.go
```diff
@@ -106,22 +112,112 @@ func (p *Porter) GenerateCredentials(opts CredentialOptions) error {
...
+
+	selectCredPrompt := &survey.MultiSelect{
+		Message: "Choose a set of credentials to use while installing this bundle",
+		Options: credSetNames,
+		Default: []string{bundle.Name},
+	}
+
+	err := survey.AskOne(selectCredPrompt, &selectedCredSets, nil)
+	if err != nil {
+		return []string{}, errors.New("no credential set selected")
```
vdice, line 176: Should we update this line to print out the `err` that occurs and perhaps update the message here?  `errors.Wrapf(err, ...)`, etc.

pkg/porter/credentials.go
```diff
@@ -106,22 +112,112 @@ func (p *Porter) GenerateCredentials(opts CredentialOptions) error {
...
 	genOpts := credentialsgenerator.GenerateOptions{
-		Name:        name,
+		Name:        credIdentifierName,
 		Credentials: bundle.Credentials,
-		Silent:      opts.Silent,
+		Silent:      silent,
+		DryRun:      dryRun,
 	}
+
+	// two extra spaces to align with survey prompts
```
vdice, line 208: Is this comment still applicable/needed?

dev-drprasad: Somehow i missed it.

pkg/porter/credentials.go
dev-drprasad: I agree

dev-drprasad:
> Re: testing, have you looked at the tests in the survey library repo? (https://github.com/AlecAivazis/survey#testing) It looks like there are some examples to emulate there...

Let me look into it

carolynvs:
I feel just awful about how I have failed to handle this PR properly. I apologize, @dev-drprasad. I will strive to treat everyone more respectfully in the future.

However in light of wanting to get away from interactive CLI prompts, #1080, what would we like to do with this in particular?

My personal preference is to not have interactive prompts in porter. From a maintenance perspective, I've seen that we have trouble testing them and depending on screen size, etc there are a lot of edge cases that make it more difficult than it first appears. From a user perspective, having a cli that normally is driven via arguments and flags to suddenly jump into interactive mode, requiring stdin and needing you to pass a flag to prevent it from doing that is often unwelcome.

@vdice, @dev-drprasad, what are your thoughts?

vdice:
> @vdice, @dev-drprasad, what are your thoughts?

I also support moving away from the interactive prompts/generation of cred and param sets, for reasons stated above and as mentioned in https://github.com/getporter/porter/issues/1080#issuecomment-642938806.  Also, a good idea was broached by @squillace in https://github.com/getporter/porter/issues/1080#issuecomment-648454726: even though Porter may remove interactive generation, we could get creative with a separate tool/approach to adding this functionality, both for cred and param sets, as well as other areas (Porter's config, a Porter manifest for a new bundle, etc.)

dev-drprasad:
Agree with interactive prompts are complicated and there are lot of edge cases to consider. Better to go with alternatives suggested in #1080 . So, closing this PR 😔

PR #931, opened 2020-02-27
Split mixin command args by whitespace

carolynvs:
# What does this change
Some mixin commands have whitespace and right now we are passing them in as packed into a single array element when executing the command

`exec.Cmd("aws", []string{"ec2", "run-instances", "--security-group-ids group1 group2"})`

When the debug command is printed, we can't tell that this has happened, and why the command failed. What needs to happen is that when there is a space, and it's not enclosed in double quotes, it should be split into its own array element

`exec.Cmd("aws", []string{"ec2", "run-instances", "--security-group-ids", "group1", "group2"})`

# What issue does it fix
I think this will help with https://github.com/deislabs/porter-aws/issues/15

Here is how it would work
```yaml
install:
  - aws:
      description: "Provision VM"
      service: ec2
      operation: run-instances
      flags:
        image-id: ami-xxxxxxxxxxxxxxxxx
        instance-type: t2.micro
        region: eu-west-1
        subnet-id: subnet-xxxxxx
        security-group-ids: sg-1 sg2
```

If you really needed the flag to get both arguments passed together then you would use this:

```yaml
install:
  - aws:
      description: "Provision VM"
      service: ec2
      operation: run-instances
      flags:
        image-id: ami-xxxxxxxxxxxxxxxxx
        instance-type: t2.micro
        region: eu-west-1
        subnet-id: subnet-xxxxxx
        security-group-ids: '"sg-1 sg2"'
```

# Notes for the reviewer
Since the command looks the same when printed, I wasn't sure how to test this out other than just testing the function I added. 

# Checklist
- [x] Unit Tests
- [ ] Documentation

### DIFF
--- a/pkg/exec/builder/execute.go
+++ b/pkg/exec/builder/execute.go
@@ -2,6 +2,7 @@ package builder
 
 import (
 	"bytes"
+	"encoding/csv"
 	"fmt"
 	"io"
 	"strings"
@@ -77,6 +78,11 @@ func ExecuteStep(cxt *context.Context, step ExecutableStep) (string, error) {
 	}
 	args = append(args, flags.ToSlice(dashes)...)
 
+	// Split up any arguments or flags that have spaces so that we pass them as separate array elements
+	// It doesn't show up any differently in the printed command, but it matters to how the command
+	// it executed against the system.
+	args = expandOnWhitespace(args)
+
 	cmd := cxt.NewCommand(step.GetCommand(), args...)
 	output := &bytes.Buffer{}
 	cmd.Stdout = io.MultiWriter(cxt.Out, output)
@@ -98,4 +104,28 @@ func ExecuteStep(cxt *context.Context, step ExecutableStep) (string, error) {
 	}
 
 	return output.String(), nil
+}
+
+// expandOnWhitespace finds elements with multiple words that are not "glued" together with quotes
+// and splits them into separate elements in the slice
+func expandOnWhitespace(slice []string) []string {
+	expandedSlice := make([]string, 0, len(slice))
+	for _, chunk := range slice {
+		r := csv.NewReader(strings.NewReader(chunk))
+		r.Comma = ' '
+		r.LazyQuotes = true
+		group, err := r.Read()
+		if err != nil {
+			if err == io.EOF {
+				break
+			}
+			expandedSlice = append(expandedSlice, chunk)
+		} else {
+			for _, chunkette := range group {
+				expandedSlice = append(expandedSlice, chunkette)
+			}
+		}
+	}
+
+	return expandedSlice
 }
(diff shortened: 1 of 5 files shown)


pkg/exec/builder/execute.go
```diff
@@ -77,6 +78,11 @@ func ExecuteStep(cxt *context.Context, step ExecutableStep) (string, error) {
 	}
 	args = append(args, flags.ToSlice(dashes)...)
 
+	// Split up any arguments or flags that have spaces so that we pass them as separate array elements
+	// It doesn't show up any differently in the printed command, but it matters to how the command
+	// it executed against the system.
+	args = expandOnWhitespace(args)
```
halkyon: Do you think passing the command off to the shell (which parses this for us) be something that would work here instead of parsing the command ourselves? Looks similar to the problem I ran into here: https://github.com/deislabs/porter/pull/921/files#diff-9cde54dd12e8d18ebea48c996aa54c92R45

carolynvs: @halkyon Do you have a cross platform way of doing that? I was looking for ideas yesterday on how to make the shell deal with this instead of that kludge function (that doesn't work anyway). I would love some help fixing this problem! 😂

carolynvs: It is so tempting to be like `bash -c "dump all their stuff here"` but actually their command may also be a bash -c "..." command too. Or it could be calling to a binary that is in a scratch based invocation image.

halkyon: Ah, right, I didn’t realise the environments commands could be running in, still getting my head around porter. Sorry for butting in a bit on this PR by the way!

carolynvs: Oh hey no worries, I liked the suggestion and would have used it in a heartbeat if I could!

carolynvs:
I don't think I'm going to get away with my cheap trick with the csv parser. 😉 I just added a failing test case for enclosing escaped double quotes and will work on writing an artisanal small batch parser for our own needs.

carolynvs:
Okay, artisanal small batch parser works. I want to hold off on merging this until I get some more pieces in place however since this is a disruptive change. It removes a bug that people relied upon that allowed them to call `bash -c foo bar baz` without the quotes around "foo bar baz".

Here are some things that need to be done before we do a release:
* Add a best practice around embedding bash in porter.yaml
* Scan for us using anything that relied upon the bug in this repo
* Add a build hard validation for anyone calling bash -c without wrapping quotes because the quotes are necessary now.
* Add a build soft validation about using bash -c with the exec mixin, and point to the best practice.

We don't need all of these done before this can be merged, I just want to get a plan in place to do this in a timely fashion so that we don't have canary borked for a while.

PR #951, opened 2020-03-15
Add ability to access Docker host

jdolitsky:
Opening this PR mostly to start a discussion and get some advice. I'm looking to use Porter for all the things - including building and pushing container images from source as part of a full system installation.

The problem is that Porter itself runs "porter install" commands within the context of a "docker run". In order to run something like "docker build", this requires certain access to the underlying OS, which is disabled by default.

With Docker specifically, you need to run a docker daemon (i.e. "dind"), and that does not work without the --privileged flag. Even alternative tools like [buildah ](https://github.com/containers/buildah) and [img](https://github.com/genuinetools/img) appear to require extra privileges/flags in order to work.

This PR in its current state will set the `--privileged` flag on `docker run`, if `PRIVILEGED=1` is set in the environment:
```
PRIVILEGED=1 porter install --debug
```

The correct place for this code is probably the underlying cnab-go library, so I've also opened a PR there. See https://github.com/cnabio/cnab-go/pull/199

Is there something I'm missing in regards to this? Are other people using Porter to build images? Thanks in advance. Related to https://github.com/deislabs/porter/issues/442

### DIFF
--- a/pkg/cnab/provider/runtime.go
+++ b/pkg/cnab/provider/runtime.go
@@ -10,6 +10,7 @@ import (
 	"github.com/cnabio/cnab-go/driver/docker"
 	"github.com/cnabio/cnab-go/driver/lookup"
 	"github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/api/types/mount"
 )
 
 type Runtime struct {
@@ -29,18 +30,11 @@ func NewRuntime(c *config.Config, claims claims.ClaimProvider, credentials crede
 func (d *Runtime) newDriver(driverName string, claimName string, args ActionArguments) (driver.Driver, error) {
 	var driverImpl driver.Driver
 
-	// TODO: Remove once this PR is merged: https://github.com/cnabio/cnab-go/pull/199
-	if driverName == "docker" {
-		dockerDriver := &docker.Driver{}
-		if val, ok := os.LookupEnv("PRIVILEGED"); ok && val == "1" {
-			dockerDriver.AddConfigurationOptions(func(cfg *container.Config, hostCfg *container.HostConfig) error {
-				hostCfg.Privileged = true
-				return nil
-			})
-		}
-		driverImpl = driver.Driver(dockerDriver)
+	if d.Dind && driverName == "docker" {
+		driverImpl = dockerDindDriver()
 	} else {
-		driverImpl, err := lookup.Lookup(driverName)
+		var err error
+		driverImpl, err = lookup.Lookup(driverName)
 		if err != nil {
 			return driverImpl, err
 		}
@@ -59,4 +53,27 @@ func (d *Runtime) newDriver(driverName string, claimName string, args ActionArgu
 	}
 
 	return driverImpl, nil
+}
+
+// If Dind support is enabled ("Docker-in-Docker"), set
+// the equivalent of the following "docker run" options:
+// 1.) -v /var/run/docker.sock:/var/run/docker.sock
+// 2.) --privileged
+func dockerDindDriver() driver.Driver {
+	dockerDriver := &docker.Driver{}
+	dockerDriver.AddConfigurationOptions(func(cfg *container.Config, hostCfg *container.HostConfig) error {
+		dockerSockMount := mount.Mount{
+			Source:   "/var/run/docker.sock",
+			Target:   "/var/run/docker.sock",
+			Type:     "bind",
+			ReadOnly: false,
+		}
+		if hostCfg.Mounts == nil {
+			hostCfg.Mounts = []mount.Mount{}
+		}
+		hostCfg.Mounts = append(hostCfg.Mounts, dockerSockMount)
+		hostCfg.Privileged = true
+		return nil
+	})
+	return driver.Driver(dockerDriver)
 }
--- a/pkg/cnab/provider/runtime.go
+++ b/pkg/cnab/provider/runtime.go
@@ -1,16 +1,9 @@
 package cnabprovider
 
 import (
-	"os"
-
 	"get.porter.sh/porter/pkg/claims"
 	"get.porter.sh/porter/pkg/config"
 	"get.porter.sh/porter/pkg/credentials"
-	"github.com/cnabio/cnab-go/driver"
-	"github.com/cnabio/cnab-go/driver/docker"
-	"github.com/cnabio/cnab-go/driver/lookup"
-	"github.com/docker/docker/api/types/container"
-	"github.com/docker/docker/api/types/mount"
 )
 
 type Runtime struct {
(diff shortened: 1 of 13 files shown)


radu-matei:
I'm wondering if the build functionality would be better suited for a plugin, rather than a Docker in Docker build.

The very first version of Duffle had the concept of _components_, and `duffle build` would read all components from the config file and build the container images associated with those components, which ended up as part of the bundle.

The reason I'm leaning towards such a plugin, besides avoiding running privileged containers (which we specifically removed from `cnab-go` / Duffle a while back), is to re-use this potential plugin.

Ultimately, the difficulty with such component is defining the application model - do you have some examples of applications?

ref https://github.com/cnabio/duffle/pull/158

jdolitsky:
What I'm hoping for is not to have to change my porter.yaml definition in any way. It would be nice for Porter to use the same scripts I use locally, some of them which may contain a "docker build" within.

In terms of the application - simple app, 3 custom services (1 frontend, 2 backends). All running as containers on Kubernetes. 

I want for Porter to be given info such as GIT_SSH_KEY, GIT_REMOTE_URL, GIT_COMMIT, and convert source code into a full blown production environment. For example - instantiate a container registry resource on cloud X, and in a later step, build and push images to it.

In the case of img, looks like there are some other flags we can set at docker runtime that do not necessarily require the use of --privileged:
```
docker run --rm -it \
    --name img \
    --volume $(pwd):/home/user/src:ro \ # for the build context and dockerfile, can be read-only since we won't modify it
    --workdir /home/user/src \ # set the builder working directory
    --volume "${HOME}/.docker:/root/.docker:ro" \ # for credentials to push to docker hub or a registry
    --security-opt seccomp=unconfined --security-opt apparmor=unconfined \ # required by runc
    r.j3ss.co/img build -t user/myimage .
```

jdolitsky:
I'll also point out that I'm currently using a forked release of Porter based on this patch as part of my CI process, and it does exactly what I'm trying to accomplish like I described above.

carolynvs:
Right now people are not accessing Docker from inside a bundle. Early on in duffle, people decided that privileged containers by default was a security problem and it was closed off. They replaced it by making the docker driver configurable. So Porter, Docker App or Duffle could pass in additional configuration options to Docker. That's the background on the enhancement that everyone wants and why we haven't just turned on privileged containers.

We have an open issue to consider how we would like to pass configuration to the docker driver to support bundles building images, pushing, using docker-compose, etc.

carolynvs:
@radu-matei The problem isn't limited to building application images, people also want access to using docker for other things, like `docker-compose`.

jdolitsky:
> We have an open issue to consider how we would like to pass configuration to the docker driver to support bundles building images

@carolynvs - this seems that it's already supported by cnab-go using `dockerDriver.AddConfigurationOptions` ?

If I can make this work with img using `--security-opt` flags vs. `--privileged` would that be more likely to be considered? For example:
```
porter install --debug \
  --docker-security-opt seccomp=unconfined \
  --docker-security-opt apparmor=unconfined
```

I'm interested in eventual full-blown docker-in-docker support (would like to use Porter with [Tilt](https://tilt.dev/) or example), but using img works for what I'm trying to do despite lack of layer caching etc. without the docker daemon.

carolynvs:
> this seems that it's already supported by cnab-go using dockerDriver.AddConfigurationOptions

Yup, that is what I was trying to point out. It's already there, we just don't use it yet because we haven't decided how we want to expose configuring it and we don't want it on by default.

The approach you have right now, an ambient environment variable only would need to be tweaked to also include in the bundle metadata any non-standard docker configuration required to run. First so that people can understand that when evaluating if they are okay with running a bundle using `porter explain`, and also so porter which settings to pass the flags properly to cnab-go. We probably will have a gating [configuration](https://porter.sh/configuration/) setting so the user can approve elevated privileges but it should follow our existing configuration system.

jdolitsky:
Ok cool. So is it safe to say I should close the PR on cnab-go then (https://github.com/cnabio/cnab-go/pull/199)? That this setup would be a porter-specific configuration option?

carolynvs:
As long as all the settings you need to make your bundles work can be done through `AddConfigurationOptions`, then I recommend closing the other PR on cnab-go. Then we can get this done all through just Porter changes and use that.

jdolitsky:
I've extended this to be enabled via "dind" config option:
```
PORTER_DIND=true porter install
```

I do have trouble sourcing this setting from ~/.porter/config.toml .. what am I missing?

To build and and run docker containers within porter, the following flags are set on "docker run" command when dind is enabled:

1.) -v /var/run/docker.sock:/var/run/docker.sock
2.) --privileged

A follow up from this might be a docker/docker-compose/etc plugin(s) that are able to configure this for a single step? Let me know thoughts

jdolitsky:
Works as a driver for docker-compose! Lots of possibilities here. Example:

```yaml
# This is a porter.yaml that exhibits how to use PORTER_DIND=true
# to build and run Docker containers with Docker Compose.
#
# To build/run docker-compose containers (starts a custom nginx and a redis):
#   PORTER_DIND=true porter install --debug
#
# To stop all containers:
#   PORTER_DIND=true porter uninstall --debug
#
name: composerator
version: 0.1.0
description: My bundle
tag: some/bundle

# In custom Dockerfile (Dockerfile.tmpl), install Docker CLI + Docker Compose:
#
#   ENV DOCKER_VERSION="19.03.8"
#   RUN apt-get update && apt-get install -y python3-pip wget && pip3 install --upgrade pip && \
#         wget https://download.docker.com/linux/static/stable/x86_64/docker-${DOCKER_VERSION}.tgz && \
#         tar -xvf docker-${DOCKER_VERSION}.tgz && \
#         mv docker/docker /usr/bin/docker && \
#         chmod +x /usr/bin/docker && \
#         rm -rf docker/ docker-${DOCKER_VERSION}.tgz && \
#         pip3 install docker-compose
#
dockerfile: Dockerfile.tmpl

mixins:
  - exec

install:
  - exec:
      description: Docker compose up
      command: bash
      flags:
        c: |-
          echo "FROM nginx" > Dockerfile
          cat <<EOF > docker-compose.yml
          version: '3'
          services:
            web:
              build: .
              ports:
                - "8000:80"
            redis:
              image: "redis:alpine"
          EOF
          docker-compose up -d

uninstall:
  - exec:
      description: Docker compose down
      command: bash
      flags:
        c: |-
          docker-compose down --remove-orphans
```

and I'm able to access nginx locally at http://localhost:8000/

pkg/cnab/provider/runtime.go
```diff
@@ -42,5 +52,28 @@ func (d *Runtime) newDriver(driverName string, claimName string, args ActionArgu
...
+			Source:   "/var/run/docker.sock",
+			Target:   "/var/run/docker.sock",
+			Type:     "bind",
+			ReadOnly: false,
+		}
+		if hostCfg.Mounts == nil {
+			hostCfg.Mounts = []mount.Mount{}
+		}
+		hostCfg.Mounts = append(hostCfg.Mounts, dockerSockMount)
+		hostCfg.Privileged = true
```
vdice: FWIW, one item I've learned whilst configuring DinD or DooD in/for Brigade is this:

* Privileged is necessary (or alternate security setup that you'd mentioned prior) for running DinD, that is, running the docker daemon _inside_ of the container process.  (Therefore, image needs to be based on `docker:stable-dind` and/or can execute `dockerd-entrypoint.sh`)
* Mounting of the docker socket is necessary for running DooD, that is, Docker _outside_ of Docker via the host's docker socket.  No intra-container docker daemon process need be started, as it is using the host's.

See also https://docs.brigade.sh/topics/scripting/#docker-runtime and https://github.com/brigadecore/brigade/issues/1049#issuecomment-577157659

We might think about adding toggles for either/or (and both _could_ be toggled on, if for some reason this was desirable).

pkg/config/config.go
```diff
@@ -28,6 +28,9 @@ const (
 	// EnvDEBUG is a custom porter parameter that signals that --debug flag has been passed through from the client to the runtime.
 	EnvDEBUG = "PORTER_DEBUG"
 
+	// EnvDIND is a custom porter parameter that signals to enable "Docker-in-Docker" support.
```
carolynvs: This isn't the right location to set this flag (pretty much all the files from here down need to be reverted).

pkg/cnab/provider/runtime.go
carolynvs: Agreed, thanks for the good explanation Vaughn!

As far as the porter config goes, I think for now we can keep it as a single flag (i.e. do we want to give the bundle access to the host which both privileged and the docker sock provide). Later when we work on bundle level config, we will need to think about what the bundle can request and how Porter can creatively be providing given various environments it's installed on.

jdolitsky:
@carolynvs @vdice thanks for taking a look! The DooD vs. DinD thing is interesting..

Updated to address comments and tested it - works. I'm just missing where to check for `PORTER_ALLOW_DOCKER_HOST_ACCESS` .. any pointers? Didn't see anything relevant in pkg/config/datastore

jdolitsky:
@vdice in terms of examples, I agree. I like your compose example, maybe submit it after this is merged?

carolynvs:
Right now we don't have a magic place to check and set that value Suggestions welcome for improving this part of the config system for a later time! 😂  

So for now add an accessor function to DataStore and put in it any special logic for that field. Then instead of using the field directly in the driver, use the accessor function.

Here is an example that uses an accessor to fallback to a default plugin when none is specified:
https://github.com/deislabs/porter/blob/9110c71d4d65c5b1d2d0705b618b035efd93c672/pkg/config/datastore.go#L40-L46

Our function will want to prefer the configured value on the field, then the environment variable and at last default to false.

I hope that helps!

carolynvs:
Yeah adding an example as a follow-up is 👍 in my book.

jdolitsky:
Thanks for that explanation, done. Should be good to go now.

I also created new file at `pkg/cnab/provider/driver.go` containing constants for the CNAB drivers and moved all driver-related code there. The string "docker" started to show up everywhere so it seemed appropriate, but I'm more than happy to revert that stuff (the constants are probably more appropriate in cnab-go).

pkg/config/datastore_test.go
```diff
@@ -1,12 +1,53 @@
...
+	// Default should be false
+	allowDockerHostAccess, err := d.GetAllowDockerHostAccess()
+	require.NoError(t, err, "GetAllowDockerHostAccess failed")
+	assert.Equal(t, false, allowDockerHostAccess, "Allow docker host access should be false if config is missing")
+
+	// Unset the env var we are about to mess with after this test is done
+	defer os.Unsetenv(EnvAllowDockerHostAccess)
+
+	// Env var - invalid
+	os.Setenv(EnvAllowDockerHostAccess, "sureenableitplz")
```
vdice: 😂

vdice:
@carolynvs when convenient, can you sign off and merge this PR if all change requests have been addressed?

jdolitsky:
🙌

PR #953, opened 2020-03-20
Add porter lint command and run lint automatically during build

### LINKED ISSUE #944: Validate bash quoting in exec mixin during porter build
Do a hard validation during porter build when the exec mixin is used to call `bash -c "some command"` that there are double quotes embedded in the -c argument now that #931 is merged.

Make it a linter stage that is printed out by cowsay because people will listen to the cow.

Cow say we all.

carolynvs:
# What does this change
Add a `porter lint` command that analyzes porter.yaml for errors and best practices, calling out to the mixins to parse their own sections of the manifest. It is a separate command, but we also lint automatically during `porter build` to catch major errors and abort the build. You can use `porter build --no-lint` when you think it will work anyway.

I also split out a new package mixin/query that can apply a command, like build or lint, to a set of mixins and collect their replies. It's now used for both of those commands.

# What issue does it fix
Closes #944

# Notes for the reviewer
As part of fixing the tests, I had to update some of the test porter.yaml files to have them stop using embedded bash to make the linter happy. In a follow-up, I'm going to run the linter over all the porter.yaml files in the entire repo so that we can keep them in a state where a user could copy them and not be copying bad practices or errors.

# Checklist
- [x] Unit Tests
- [ ] Documentation


pkg/exec/action.go
```diff
@@ -7,23 +7,35 @@ import (
...
+}
+
 // UnmarshalYAML takes any yaml in this form
 // ACTION:
 // - exec: ...
 // and puts the steps into the Action.Steps field
 func (a *Action) UnmarshalYAML(unmarshal func(interface{}) error) error {
-	var steps []Step
-	results, err := builder.UnmarshalAction(unmarshal, &steps)
+	results, err := builder.UnmarshalAction(unmarshal, a)
```
carolynvs: This is a breaking change to that function signature. Mixins won't be affected unless they were using it, for example the kuberentes mixin doesn't but the az mixin does. They don't need to update and use this change unless they want to do linting or mixin configuration. This helps them unmarshal the action data for that.

carolynvs: I have created a follow-up issue in the skeletor repo to update and use this change: https://github.com/deislabs/porter-skeletor/issues/16

pkg/context/helpers.go
```diff
@@ -107,12 +112,14 @@ func (c *TestContext) AddTestDirectory(srcDir, destDir string) {
 	}
 }
 
+// GetOutput returns all text printed to stdout.
 func (c *TestContext) GetOutput() string {
-	return string(c.output.Bytes())
+	return string(c.capturedOut.Bytes())
```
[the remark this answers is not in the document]
carolynvs: Yeah it was a pain that we would get different output depending on if we had --debug on or not, but since we send debug to stderr, this should help our testing.

PR #962, opened 2020-03-30
Explain allow-docker-host-access setting

carolynvs:
# What does this change
Update the configuration page with how to use the
allow-docker-host-access setting as either an environment variable or
configuration setting. It is not yet available as a command flag.

Also note that based on how the Azure Cloud Shell driver is implemented
today (using ACI as the host environment for running the porter commands
such as porter install), this setting won't work yet in that
environment

https://deploy-preview-962--porter.netlify.com/configuration/#porter-allow-docker-host-access

# What issue does it fix
Tidying up after #951

# Notes for the reviewer
We will need some more follow-up issues to add best practices and examples. But this is limited to just how to use the new config.

# Checklist
- [ ] Unit Tests
- [x] Documentation


docs/content/configuration.md
```diff
@@ -40,9 +44,28 @@ value of the environment variable, when defined.
 For example, you can set `PORTER_DEBUG=true` and then all subsequent porter
 commands will act as though the `--debug` flag was passed.
 
+### PORTER_ALLOW_DOCKER_HOST_ACCESS
+
+The `PORTER_ALLOW_DOCKER_HOST_ACCESS` environment variable is not exposed as a
+flag yet. 
+
+It controls whether or not the Docker host should be available to
```
vdice:
```suggestion
It controls whether or not the Docker host should be available for
```

PR #970, opened 2020-04-02
Add --allow-docker-host-access to bundle actions

carolynvs:
# What does this change
Add the flag --allow-docker-host-access to the commands: install,
upgrade, invoke and uninstall. This automatically picks up and applies
the configuration file setting and environment variable. I messed up and
we should have used this from the start, sorry! 🤦‍♀ 

I've moved the doc for the flag to be the authoritative destination and
added a redirect for the old anchor link.

# What issue does it fix
Follow-up to #951 

# Notes for the reviewer
I am working on adding a declaration to the manifest saying that a bundle requires this setting to run and adding an example bundle showing how to use it. That is coming later in a follow-on PR.

# Checklist
- [x] Unit Tests
- [x] Documentation


carolynvs:
So I messed up when I gave guidance on how to do the configuration for this. When I originally wrote the configuration system, it was really smart! If you read the https://porter.sh/configuration/ page the way it works with a hierarchy is how all flags work by default. We don't need to write extra code to support it, all the plumbing is in place already for it, we just need to add the flag. The flag will "automatically" get its value set based on the flag > env var > config value.

I just pushed a fix for flags with dashes in them, that wasn't working thanks for the sanity check!

vdice:
Ah, I see now!  That is really neat.

PR #976, opened 2020-04-08
Stamp bundles with mixins and encoded manifest

### LINKED ISSUE #800: Re-enable defaulting dependency parameters from the manifest
We had to turn off defaulting dependency parameters from the manifest temporarily in #799. Once we have a solution for consistently having access to the manifest for both file based and tag based bundle execution, let's turn it back on.

### LINKED ISSUE #801: Make the porter manifest available to the client for tag based bundle execution
When we are working with a bundle that we pulled from a registry, we don't have access to the porter.yaml. It would be very helpful to know if it was a porter bundle and be able to take advantage of any porter specific features. For example, dependency overrides, "file" parameter types, etc.

Let's come up with how we'd like to figure out that a bundle was made by porter and get at the manifest.

1. Crack the invocation image and stuff it in the cache on pull
2. Stuff it in the custom metadata in the bundle.json
3. goat magic 🎩 ✨

### LINKED ISSUE #938: Populate mixins used in bundle.json
During `porter build` we create a bundle.json file from the porter.yaml. It would be very helpful to include the list of mixins used in the bundle in the custom bundle metadata.

* The CNAB Dashboard that is being created can use that to aggregate metadata about what type of bundles we are building and installing
* The explain command could use it to say what is used in the bundle, or people/other tools can use it to decide if they are okay with what the bundle is doing. Like they don't want to use bundles that use the exec mixin but are okay with the helm mixin.

```json
{
    "custom": {
        "sh.porter": {
            "manifestDigest": "2f9d84ae8f4d328a82a71df0c8b3ad2b5798e47a937113a923aff23c65bbda9d",
            "mixins": {
                "exec": {},
                "helm": {}
            }
        }
    }
}
```

We are giving each mixin a struct to store additional information but can't populate it yet because we don't have it. At a later date, we can also store the version of the mixin used for example.

* The code that populates the custom section is [here](https://github.com/deislabs/porter/blob/6da074e0114723c050ce84b6025cba883718f702/pkg/cnab/config-adapter/adapter.go#L64-L66)
* Ideally we should move all the code that populates the custom section into a separate function as it has grown over time.
* We should add a unit test that checks that it is populated correctly when there are no mixins, and mixins used.

carolynvs:
# What does this change
* Stamp bundles with mixins (#938) and encoded manifest (#801) 
* Extract porter.yaml and store in cache if available
* Refactor cache to give easy access to the decoded manifest
* Always write a manifest (even just an empty one) to avoid guard checks everywhere
* Add back in defaulting parameters of dependencies (#800)

# What issue does it fix
Closes #800 
Closes #801
Closes #938 

# Notes for the reviewer
* Apologies for the mega PR, this all just kind of went together and was hard to tease apart. I broke it down into commits, and recommend reviewing it commit by commit.
* This pulls in an unmerged PR to cnab-go, [improve matching of not found errors](https://github.com/cnabio/cnab-go/pull/201). Without it I was getting the wrong error messages back. We can either merge with this pointing to the branch, or wait until that PR is merged and a new release of cnab-go is tagged.

# Checklist
- [x] Unit Tests
- [x] Documentation
- [ ] Schema (porter.yaml)


pkg/porter/stamp.go
```diff
@@ -51,8 +51,14 @@ func (p *Porter) IsBundleUpToDate(opts bundleFileOptions) (bool, error) {
 		}
 
 		converter := configadapter.NewManifestConverter(p.Context, p.Manifest, nil, mixins)
-		newStamp := converter.GenerateStamp()
-		return oldStamp.ManifestDigest == newStamp.ManifestDigest, nil
+		newDigest, err := converter.DigestManifest()
+		if err != nil {
+			if p.Debug {
+				fmt.Fprintln(p.Err, errors.Wrap(err, "could to determine if the bundle is up-to-date so will rebuild just in case"))
```
vdice, line 57:
```suggestion
				fmt.Fprintln(p.Err, errors.Wrap(err, "could not determine if the bundle is up-to-date so will rebuild just in case"))
```

PR #977, opened 2020-04-09
Feat/required extensions

vdice:
# What does this change
* Adds ability to declare required extensions (and config) in a Porter manifest via `required`
* At runtime, adds functionality to parse required extensions (and config) from bundle
   * If a required extension is unknown/unsupported by Porter, error out
   * Processed/supported extension config is added to the `Runtime` struct for access by interested parties to inform setup (e.g. set up the Docker driver according to extension config)
* Adds the `DockerHostAccessExtension` to the list of known/supported required extensions

# What issue does it fix
Closes https://github.com/deislabs/porter/issues/973

# Notes for the reviewer
* Some TODOs linger... help in determining which should be addressed here and which should be tracked in follow-ups

# Checklist
- [x] Unit Tests
- [x] Documentation
- [x] Schema (porter.yaml)


pkg/cnab/provider/driver.go
```diff
@@ -23,12 +24,24 @@ func (r *Runtime) newDriver(driverName string, claimName string, args ActionArgu
...
-		driverImpl, err = r.dockerDriverWithHostAccess()
+		// Parse extension config to inform setup
+		config, ok := ext.(*extensions.DockerHostAccess)
+		if !ok && extensionIsRequired {
+			return nil, errors.Errorf("unable to parse extension config: %+v", config)
+		}
+		driverImpl, err = r.dockerDriverWithHostAccess(config)
 	} else {
+		if extensionIsRequired {
+			return nil, errors.Errorf("extension %q is required but allow-docker-host-access was not enabled",
```
carolynvs, line 42:
```suggestion
			return nil, errors.Errorf("extension %q is required but allow-docker-host-access was not enabled, see https://porter.sh/configuration/#allow-docker-host-access for more information",
```

pkg/cnab/provider/driver_test.go
```diff
@@ -51,4 +51,7 @@ func TestNewDriver_Docker(t *testing.T) {
 		_, err := d.newDriver(DriverNameDocker, "myclaim", args)
 		assert.EqualError(t, err, "allow-docker-host-access was specified but could not detect a local docker daemon running by checking for /var/run/docker.sock")
 	})
+
+	// TODO: add tests that check the Docker configuration options based on the provided required extension config.
```
carolynvs, line 55: I think this is fine to follow-up on later.

pkg/templates/templates/schema.json
```diff
@@ -233,6 +233,13 @@
       "description": "Dependent bundles that must be executed with the bundle",
       "type": "object"
     },
+    "required": {
+      "description": "Required extensions for this bundle",
+      "type": "array",
+      "items": {
+        "anyOf": []
```
carolynvs, line 240: Since they can only use extensions that we know, do we want to list the known extensions here with the config supported? This would give users autocomplete in VS Code.

vdice: Good idea.  For now, should we list the extensions and config _and also_ allow additionalProperties (so users may define their own).  Or lock it down w/ `additionalProperties: false`?

pkg/cnab/config-adapter/adapter.go
```diff
@@ -318,3 +309,48 @@ func toBool(value bool) *bool {
...
+	return requiredExtensions
+}
+
+// lookupExtensionKey is a helper method to return a full key matching a
+// supported extension, if applicable
+func lookupExtensionKey(name string) string {
+	key := name
+	// If an official supported extension, we grab the full key
+	// TODO: Do we want to error out here if unsupported?
+	if supportedExt, _ := extensions.GetSupportedExtension(name); supportedExt != nil {
```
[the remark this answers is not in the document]
vdice: Can't recall; but updated to prep for issuing a linter warning if `err != nil`.

docs/content/author-bundles.md
```diff
@@ -301,6 +303,52 @@ image: "{{bundle.images.websvc.repository}}@{{bundle.images.websvc.digest}}"
...
+Name: `docker-host-access`
+
+Configuration:
+  * `Privileged: boolean` - Whether or not the `--privileged` flag should be set on the container run invocation
+
+Example:
+
+```yaml
+required:
+  - docker-host-access:
```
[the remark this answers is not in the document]
vdice: Sounds good!  Done.

pkg/templates/templates/schema.json
carolynvs: I think that may be a good idea for now, help people with autocomplete but don't lock it down completely. 👍

pkg/cnab/config-adapter/adapter.go
```diff
@@ -358,8 +358,10 @@ func (c *ManifestConverter) generateRequiredExtensions() []string {
 func lookupExtensionKey(name string) string {
 	key := name
 	// If an official supported extension, we grab the full key
-	// TODO: Do we want to error out here if unsupported?
-	if supportedExt, _ := extensions.GetSupportedExtension(name); supportedExt != nil {
+	supportedExt, err := extensions.GetSupportedExtension(name)
+	if err != nil {
+		// TODO: Issue linter warning
```
carolynvs, line 363: When the porter manifest defines a required extension that is not supported by that version of porter, the linter should return a warning.

I'm using this issue to try out this feature in github where we can turn a pull request comment into an issue: https://help.github.com/en/github/managing-your-work-on-github/opening-an-issue-from-a-comment

PR #982, opened 2020-04-13
Fix regression in loading bundle from claim

### LINKED ISSUE #981: Working directly from the instance name is broken
## Describe the bug
After installing a bundle using a tag or file, when I then try to manage it using just the instance name, I get errors.

## To Reproduce
1. `porter create`
2. Edit the porter.yaml to publish to your repository.
3. `porter publish`
4. `porter install HELLO --tag YOURNAME/porter-hello:v0.1.0`
5. `porter upgrade HELLO`

The final step fails because it can't find the bundle for HELLO.

## Expected behavior
The `porter upgrade` to succeed like it used to.

## Porter Version
This is a regression in `porter v0.26.0-beta.1 (b15c4109)`

carolynvs:
# What does this change
v0.26.0-beta.1 (specifically #976) caused a regression where an error aborted actions that relied on the bundle instance name to identity and retrieve the bundle.

* Some porter bundles will have a stamp, but not an embedded manifest. Gracefully handle this when caching the bundle, and return hard errors when people try to read the manifest out of the cache.
* Previously we were accidentally ignoring errors when bundle load failed in the dependencies code. This helped cover up not handling when the bundle was supplied through just an instance name. In the latest release, I fixed the lack of error handling but didn't catch that the
code wasn't loading bundles from the claim name too (just from the file or tag). This adds in logic to also load the bundle from an instance name.

# What issue does it fix
Closes #981 

# Notes for the reviewer
There were two bugs in one with this, both had to be fixed for all bundles (those made by porter post v0.26.0-beta1, those made by porter earlier, and non-porter bundles) to load properly.

# Checklist
- [x] Unit Tests
- [ ] Documentation - not impacted
- [ ] Schema (porter.yaml) - not impacted

### DIFF
--- a/pkg/porter/dependencies.go
+++ b/pkg/porter/dependencies.go
@@ -116,19 +116,29 @@ func (e *dependencyExecutioner) ApplyDependencyMappings(args *cnabprovider.Actio
 func (e *dependencyExecutioner) identifyDependencies() error {
 	// Load parent CNAB bundle definition
 	var bun *bundle.Bundle
-	if e.parentOpts.Tag != "" {
+	if e.parentOpts.CNABFile != "" {
+		bundle, err := e.CNAB.LoadBundle(e.parentOpts.CNABFile)
+		if err != nil {
+			return err
+		}
+		bun = bundle
+	} else if e.parentOpts.Tag != "" {
 		cachedBundle, err := e.Resolver.Resolve(e.parentOpts.BundlePullOptions)
 		if err != nil {
 			return errors.Wrapf(err, "could not resolve bundle")
 		}
 
 		bun = &cachedBundle.Bundle
-	} else {
-		bundle, err := e.CNAB.LoadBundle(e.parentOpts.CNABFile)
+	} else if e.parentOpts.Name != "" {
+		c, err := e.Claims.Read(e.parentOpts.Name)
 		if err != nil {
 			return err
 		}
-		bun = bundle
+
+		bun = c.Bundle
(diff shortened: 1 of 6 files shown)


pkg/porter/dependencies.go
```diff
@@ -116,19 +116,29 @@ func (e *dependencyExecutioner) ApplyDependencyMappings(args *cnabprovider.Actio
...
+		c, err := e.Claims.Read(e.parentOpts.Name)
 		if err != nil {
 			return err
 		}
-		bun = bundle
+
+		bun = c.Bundle
+	} else {
+		// If we hit here, there is a bug somewhere
+		return errors.New("identifyDependencies failed to load the bundle because no bundle was specified. Please report this bug to https://github.com/deislabs/porter/issues/new/choose")
```
adamreese, line 141: if this condition only occurs if there is a bug, should it `panic()`

carolynvs: I think an error is fine in this case. The panic would give a stack trace but in this case the stack trace would not reveal anything, it's the same everytime. The real culprit would in in those pesky functions called at the beginning of InstallBundle, UpdateBundle, UninstallBundle where we initialize values on the Options struct, or allowed an invalid request to make it this far.

carolynvs:
Merged with my magic admin privileges because the intrepid vdice is on vacation to give the magic green checkmark. Thanks @adamreese for the review! 🙇‍♀️

PR #989, opened 2020-04-15
fix(query.go): actually propogate mixin runErr

vdice:
# What does this change
* updates logic in `query.go` to return mixin errors

It appears we hadn't actually been returning mixin errors if they occurred during mixin build invocations (e.g. when assembling the mixin section of the invocation image Dockerfile).  This makes sure we capture such errors and halt the bundle build appropriately.

# What issue does it fix
N/A
# Notes for the reviewer
Note the changes in order to cover under test.  Open to suggestions on a different way.

# Checklist
- [x] Unit Tests
- [x] Documentation - not impacted

### DIFF
--- a/pkg/mixin/query/query.go
+++ b/pkg/mixin/query/query.go
@@ -110,20 +110,21 @@ func (q *MixinQuery) Execute(cmd string, inputGenerator MixinInputGenerator) (ma
 		if response.runErr == nil {
 			results[response.mixinName] = response.output
 		} else {
-			runErr = multierror.Append(runErr, response.runErr)
+			runErr = multierror.Append(runErr,
+				errors.Wrapf(response.runErr, "error encountered from mixin %q", response.mixinName))
 		}
 	}
 
 	if runErr != nil {
 		if q.RequireAllMixinResponses {
-			return nil, err
+			return nil, runErr
 		}
 
 		// This is a debug because we expect not all mixins to implement some
 		// optional commands, like lint and don't want to print their error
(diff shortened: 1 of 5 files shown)


pkg/mixin/query/query.go
```diff
@@ -110,20 +110,21 @@ func (q *MixinQuery) Execute(cmd string, inputGenerator MixinInputGenerator) (ma
...
-			runErr = multierror.Append(runErr, response.runErr)
+			runErr = multierror.Append(runErr,
+				errors.Wrapf(response.runErr, "error encountered from mixin %q", response.mixinName))
 		}
 	}
 
 	if runErr != nil {
 		if q.RequireAllMixinResponses {
-			return nil, err
+			return nil, runErr
```
carolynvs, line 120: 🤦‍♀️

pkg/mixin/query/query.go
```diff
@@ -110,20 +110,21 @@ func (q *MixinQuery) Execute(cmd string, inputGenerator MixinInputGenerator) (ma
 		if response.runErr == nil {
 			results[response.mixinName] = response.output
 		} else {
-			runErr = multierror.Append(runErr, response.runErr)
+			runErr = multierror.Append(runErr,
```
carolynvs, line 113: 👍 Nice catch, this really needed to be wrapped.

vdice:
Looking into the CI failure (cli-test)...

vdice:
Unable to reproduce cli test error locally or via CI re-run 🤔 ... Blame Catalina?

PR #994, opened 2020-04-16
ref(parameters.go): remove zero value/default param logic

vdice:
# What does this change
* Removes custom logic to supply a zero/empty or default value for parameters when they don't apply to a given action.  This had been added to appease cnab-go validation.  However, said validation is proposed to be changed in https://github.com/cnabio/cnab-go/pull/204.

Depends on https://github.com/cnabio/cnab-go/pull/204

TODO:
 - [x] remove override in go.mod

# What issue does it fix
Unintended behavior may result from populating parameters with zero/empty values even if not applicable to a given action.

# Notes for the reviewer
One of the subtle effects of the previous behavior is parameter persistence.  Since zero or default values (the default may originate from a previous claim value) are no longer being provided for parameters that don't apply to a particular action, these values will no longer 'persist' across actions.  

For example, if a value of 'foo' was provided for parameter A during install, but upgrade is run and parameter A does not apply, steps executed during upgrade may not access A's parameter value (it will be `nil`).  If a future action is run, which parameter A _does_ apply to, and a value is not provided for parameter A, parameter A's default will be used (as declared in the bundle) _or_, if no default is provided, the action will fail, returning an error along the lines of "parameter A is required but no value is provided".

# Checklist
- [x] Unit Tests
- [x] Documentation - n/a
- [x] Schema (porter.yaml) - n/a

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -108,6 +108,8 @@ github.com/cnabio/cnab-go v0.9.0-beta1/go.mod h1:5c4uOP6ZppR4nUGtCMAElscRiYEUi44
 github.com/cnabio/cnab-go v0.10.0-beta1/go.mod h1:5c4uOP6ZppR4nUGtCMAElscRiYEUi44vNQwtSAvISXk=
 github.com/cnabio/cnab-go v0.10.1-beta1 h1:ZxeHSuasovGuL2LmyKR8WH/jGy+RbDun9kS482kRO30=
 github.com/cnabio/cnab-go v0.10.1-beta1/go.mod h1:5c4uOP6ZppR4nUGtCMAElscRiYEUi44vNQwtSAvISXk=
+github.com/cnabio/cnab-go v0.11.0-beta1 h1:S8M3sHzVC6kKcPBKDx/o1WdEUKsoWFGodPl8xp8RKHY=
+github.com/cnabio/cnab-go v0.11.0-beta1/go.mod h1:5c4uOP6ZppR4nUGtCMAElscRiYEUi44vNQwtSAvISXk=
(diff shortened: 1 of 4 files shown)


go.sum
```diff
@@ -672,6 +674,7 @@ gopkg.in/yaml.v2 v2.2.1/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
 gopkg.in/yaml.v2 v2.2.2/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
 gopkg.in/yaml.v2 v2.2.4 h1:/eiJrUcujPVeJ3xlSWaiNi3uSVmDGBK1pDHUHAnao1I=
 gopkg.in/yaml.v2 v2.2.4/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
+gotest.tools v2.2.0+incompatible h1:VsBPFP1AI068pPrMxtb/S8Zkgf9xEmTLJjfM+P5UIEo=
```
carolynvs, line 677: I would like the record to show that I dislike go mods because of stuff like this. I think we need to start using `go mods tidy` to get stuff like this that isn't in our code, out of this sum file... I'll create a follow-up issue.

vdice: Oh good eye; didn't think to check.  Indeed, if we can auto-tidy, that would be great.

carolynvs:
The cnab-go PR has merged. Do you want to tag a new version, and then we can reference it here and merge this PR too?

PR #1010, opened 2020-04-22
ref(linter.go): update Location

vdice:
# What does this change
* Update linting result location to contain line and column number of result culprit in manifest

# What issue does it fix
This came up whilst working on https://github.com/deislabs/porter/issues/978... during discussions with @carolynvs , it sounds like to support VS Code integration, the location should at least contain line and column number.  We voted to simplify down to just these two values during the process.

# Notes for the reviewer
Help with testing VS Code integration -- if we're ready.

# Checklist
- [x] Unit Tests
- [x] Documentation - N/A

### DIFF
--- a/pkg/exec/lint.go
+++ b/pkg/exec/lint.go
@@ -46,7 +46,7 @@ func (m *Mixin) Lint() (linter.Results, error) {
 	results := make(linter.Results, 0)
 
 	for _, action := range input.Actions {
-		for stepNumber, step := range action.Steps {
+		for _, step := range action.Steps {
 			if step.Command != "bash" {
 				continue
 			}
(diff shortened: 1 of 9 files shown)


carolynvs:
We aren't ready for the vs code integration. The line numbers was brought up as table stakes in order for VS Code to draw squiggles for us. So once we have that, then we can talk to them. 😀

pkg/exec/lint.go
```diff
@@ -63,17 +63,20 @@ func (m *Mixin) Lint() (linter.Results, error) {
 				continue
 			}
 
+			// Derive key to be used to locate coordinates in the manifest.
+			// The embedded bash flag, 'c' is not unique enough,
+			// so set to the unique step description, or, if values non-empty,
+			// the first value
+			key := step.Description
```
carolynvs, line 70: I'm concerned that this solution doesn't guarantee that the line number is correct. Given a non-unique description (porter doesn't require unique descriptions, and won't in the future), and the same lint error multiple times in the file, this could identify the same lines over and over again instead of different lines throughout the file.

Let's step back and see if we can find a way to identify the line number with confidence. This could involve a couple solutions, maybe you see more?

1. the mixin reports a relative line number based on the raw text input given. Then porter uses the context of the full manifest to turn the relative line number into an absolute line number.
2. Porter starts passing in different data to build, either the full manifest, or structured data with line numbers so that the mixin can report the absolute line number itself.

vdice: I added some notes on my current thinking/solution for the scenario where there are duplicate offending keys that trigger the linter.  WDYT?  My first impulse was to go this approach as it seemed the lighter-weight option (as opposed to refactoring to pass in add'l (potentially structured) data into the mixin query)... but we can definitely go another route if this latest iteration still has issues.

vdice:
Closing this PR as we wish to pursue an alternate implementation as described in https://github.com/deislabs/porter/issues/1019

PR #1017, opened 2020-04-28
Skip params at runtime that don't apply

### LINKED ISSUE #1016: File Type Parameters don't respect applyTo at runtime
## Describe the bug
A file type parameter that is not applicable to the current action is still acted upon by the runtime.

## To Reproduce
1. Make a bundle with a file type parameter that only applies to install
2. Install the bundle
3. Upgrade the bundle and do not pass in the parameter.

```
Error: unable to acquire value for parameter FILEPARAM
```
See https://github.com/deislabs/porter/blob/5e79cd0062edabfca22dfb92072fbc59b95ad9b9/pkg/runtime/runtime-manifest.go#L380 for the code in question. It should check if the parameter applies to the current action. We need to do a sweep for related bugs.

## Expected behavior
The parameter should be ignored during upgrade.

## Porter Command and Output
```
$ porter upgrade
...
Digest: sha256:d094dc6a6e9445a34ad35be5e108aed75914fb83f88f673716be986693c7fcf1
Status: Downloaded newer image for squillace/functions-java-hello@sha256:d094dc6a6e9445a34ad35be5e108aed75914fb83f88f673716be986693c7fcf1
executing install action from azure-functions-java-hello (bundle instance: azure-functions-java-hello)
Error: unable to acquire value for parameter functionapp
Error: 1 error occurred:
        * failed to install the bundle: container exit code: 1, message: <nil>. fetching outputs failed: error copying outputs from container: Error: No such container:path: 3c3a6dac6bcfc5fb659e94724d0cb745192076b739eaf008fc651be3d85122fe:/cnab/app/outputs
```

## Version

porter v0.26.1-beta.1-30-g5e79cd00 (5e79cd00)

carolynvs:
# What does this change
* Do not inject parameters into the manifest template that do not apply to
  the current action.
* Do not process file type parameters that do not apply to the current
  action.

# What issue does it fix
Closes #1016 

# Notes for the reviewer
I looked around in the cnab/provider actions and the usages of the manifest parameters but couldn't think of where else we should be limiting by appliesTo.

I have a follow-up https://github.com/deislabs/porter/issues/1018 to have the linter catch people using `{{bundle.parameters.foo}}` in actions where it doesn't apply.

# Checklist
- [x] Unit Tests
- [x] Documentation - not affected
- [x] Schema (porter.yaml) - not affected

### DIFF
--- a/build/testdata/bundles/terraform/porter.yaml
+++ b/build/testdata/bundles/terraform/porter.yaml
@@ -35,8 +35,6 @@ status:
 uninstall:
   - terraform:
       description: "Uninstall Terraform assets"
-      vars:
-        file_contents: "{{bundle.parameters.file_contents}}"
 
 outputs:
   - name: file_contents
--- a/pkg/manifest/manifest.go
+++ b/pkg/manifest/manifest.go
@@ -174,6 +174,20 @@ func (pd *ParameterDefinition) DeepCopy() *ParameterDefinition {
 	return &p2
 }
 
+// AppliesTo returns a boolean value specifying whether or not
+// the Parameter applies to the provided action
+func (pd *ParameterDefinition) AppliesTo(action string) bool {
+	if len(pd.ApplyTo) == 0 {
+		return true
+	}
+	for _, act := range pd.ApplyTo {
+		if action == act {
+			return true
+		}
+	}
+	return false
+}
+
 // CredentialDefinition represents the structure or fields of a credential parameter
 type CredentialDefinition struct {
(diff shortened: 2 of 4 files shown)


build/testdata/bundles/terraform/porter.yaml
```diff
@@ -35,8 +35,6 @@ status:
 uninstall:
   - terraform:
       description: "Uninstall Terraform assets"
-      vars:
```
vdice, line 38: Nice; thank you for removing.

PR #1022, opened 2020-05-01
Cleaned up documentation for initial setup for developer tasks.

### LINKED ISSUE #1020: Update documentation to clear up contributing steps
## Describe the bug
The default steps to getting developer environment started don't work currently. Rather than tell folks to use go get, we can document folks getting started with git fork and clone instead. 

## To Reproduce

$ go get -u  get.porter.sh/porter
package get.porter.sh/porter: no Go files in /Users/sigje/go/src/get.porter.sh/porter

iennae:
# What does this change
In the initial setup, I was getting an error following

```
go get -u  get.porter.sh/porter
package get.porter.sh/porter: no Go files in /Users/sigje/go/src/get.porter.sh/porter
```

This commit cleans up the documentation around initial setup to match what I think
is going on more closely.

# What issue does it fix
Closes #1020 

# Notes for the reviewer

A couple of issues to think about:
* get.porter.sh url versus GitHub
* meta around what install does and whether more process documentation is needed around setup. 

I'm not sure whether there is a get.porter.sh url that would work for cloning directly. 

This

```Cloning into 'porter'...
fatal: unable to update url base from redirection:
  asked for: https://get.porter.sh/porter/info/refs?service=git-upload-pack
   redirect: https://github.com/deislabs/porter
```

didn't work and maybe there is a better way to do that or maybe it's just a "remember to update when the organization changes". 

Additionally, I did clean up the language in the what does `make install` do. I'm kind of wondering if there is a need to say something around creating a development environment separate from operating environment. 

Say you've installed porter following that process, decide you want to do some development work, if you follow the steps as written you just blew away your installation for the development version. That's probably a separate meta/docs issue, but I wanted to bring it up. 

# Checklist
- [ ] Unit Tests
- [X] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -184,8 +184,10 @@ maintainer.
 
 ## Initial setup
 
-1. Clone this repository with `go get -u get.porter.sh/porter`. Porter relies on being in the GOPATH.
-1. Run `make build install`.
+1. Clone this repository with `git clone https://github.com/deislabs/porter.git ~/go/src/get.porter.sh/porter`. Porter relies on being in the GOPATH.
+1. Run `make build install` from within the newly cloned repository.
+
+If you are planning on contributing back to the project, you'll need to [fork](https://guides.github.com/activities/forking/) and clone your fork. If you want to build porter from scratch, you can follow the process above and clone directly from the project.
 
 You now have canary builds of porter and all the mixins installed.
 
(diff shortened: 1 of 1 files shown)


iennae:
set PORTER_HOME and modify how they call from so folks don't have to install and override system installed porter. 
Example:

```$ export PORTER_HOME=~/.porter
$ ./bin/porter list
NAME                     CREATED      MODIFIED     LAST ACTION   LAST STATUS
porter-wordpress         2019-11-11   2019-11-11   install       success
porter-wordpress-mysql   2019-11-11   2019-11-11   install       success
```

PR #1028, opened 2020-05-08
Handle plugin command errors

dev-drprasad:
# What does this change
This change will catch std errors from `plugin run` command

# What issue does it fix
Since this change is small, I haven't created issue

# Notes for the reviewer
tried to write unit tests. but no luck as `pluginCommand.Path` got modified by `NewTestCommand` and execution failed

# Checklist
- [ ] Unit Tests

### DIFF
--- a/pkg/plugins/pluggable/loader.go
+++ b/pkg/plugins/pluggable/loader.go
@@ -81,11 +81,13 @@ func (l *PluginLoader) Load(pluginType PluginTypeConfig) (interface{}, func(), e
 		pluginType.Interface: pluginType.Plugin,
 	}
 
+	var errbuf bytes.Buffer
 	client := plugin.NewClient(&plugin.ClientConfig{
 		HandshakeConfig: plugins.HandshakeConfig,
 		Plugins:         pluginTypes,
 		Cmd:             pluginCommand,
 		Logger:          logger,
+		Stderr:          &errbuf,
 	})
 	cleanup := func() {
 		client.Kill()
@@ -95,6 +97,9 @@ func (l *PluginLoader) Load(pluginType PluginTypeConfig) (interface{}, func(), e
 	rpcClient, err := client.Client()
 	if err != nil {
 		cleanup()
+		if stderr := errbuf.String(); stderr != "" {
+			err = errors.Wrap(errors.New(stderr), err.Error())
+		}
 		return nil, nil, errors.Wrapf(err, "could not connect to the %s plugin", l.SelectedPluginKey)
 	}
 


dev-drprasad:
I found this comment in `porter-azure-plugins` after this PR raised
```
// if we write to stdout without the hclog, it will cause the plugin framework to blow up
```
I think this comment saying "we shouldn't read command errors as they are huge". But we can remove usage text in plugins using `SilenceUsage: true` while init-ing cobra so that we will have better understanding of plugin errors

`SilenceUsage: false`:

```
	* unable to resolve credential aks.TENANT_ID from value abcd: could not connect to the secrets.hashicorp.vault plugin: Unrecognized remote plugin message:

This usually means that the plugin is either invalid or simply
needs to be recompiled to support the latest protocol.: Error: invalid argument "secrets.hashicorp.vault" for "hashicorp run"
Usage:
  hashicorp run [implementation] [flags]

Flags:
  -h, --help   help for run
```

`SilenceUsage: true`

```
	* unable to resolve credential aks.SUBSCRIPTION_ID from value abdc: could not connect to the secrets.hashicorp.vault plugin: Unrecognized remote plugin message:

This usually means that the plugin is either invalid or simply
needs to be recompiled to support the latest protocol.: Error: invalid argument "secrets.hashicorp.vault" for "hashicorp run"
```

carolynvs:
>  // if we write to stdout without the hclog, it will cause the plugin framework to blow up

That comment was from me, sorry! It means that the plugin framework requires that plugins communicate using structured messages that it understands. Unstructured messages printed to stdout/stderr will cause the Porter (the initiator of the plugin connection) to abort with a message about being unable to understand the message from the plugin. So when plugins need to print a debug message or error they _must_ use hclog.

carolynvs:
Was there a scenario that cased you to look into implementing this? I haven't run into any cases where an error from the plugin wasn't reported back to porter. 

For any PR, having the reason behind why the PR is necessary explained either in an existing issue or in the pull request itself is required in order for us to properly review the change.

dev-drprasad:
@carolynvs 

Lets say user made typo while configuring `config.toml`, for example `plugin = "azure.vault"` instead of `plugin = "azure.keyvault"`.

now running `porter install` throws errors like
```
	* unable to resolve credential aks.SUBSCRIPTION_ID from secret secret1: could not connect to the secrets.azure.vault plugin: Unrecognized remote plugin message:

This usually means that the plugin is either invalid or simply
needs to be recompiled to support the latest protocol.
	* unable to resolve credential aks.TENANT_ID from secret secret2: could not connect to the secrets.azure.vault plugin: Unrecognized remote plugin message:

This usually means that the plugin is either invalid or simply
needs to be recompiled to support the latest protocol.
```
errors saying that porter could not connect to plugin but there is no why. I think this is bcoz porter not considering stderr thrown by plugin.

with current PR, error messages will look like:
```
	* unable to resolve credential aks.SUBSCRIPTION_ID from secret secret1: could not connect to the secrets.azure.vault plugin: Unrecognized remote plugin message:

This usually means that the plugin is either invalid or simply
needs to be recompiled to support the latest protocol.: 2020-05-14T14:26:41.735+0530 [ERROR] azure: invalid plugin key specified: "secrets.azure.vault"

	* unable to resolve credential aks.TENANT_ID from secret secret2: could not connect to the secrets.azure.vault plugin: Unrecognized remote plugin message:

This usually means that the plugin is either invalid or simply
needs to be recompiled to support the latest protocol.: 2020-05-14T14:26:41.747+0530 [ERROR] azure: invalid plugin key specified: "secrets.azure.vault"
```
these error messages giving more context (`2020-05-14T14:26:41.747+0530 [ERROR] azure: invalid plugin key specified: "secrets.azure.vault"`) on why `porter install` failed. Ofcourse, there is extra detail (`2020-05-14T14:26:41.747+0530 [ERROR] azure:`) bcoz `hclog`.

Since this PR captures stderr, from now on, we can redirect all errors to stderr except those happen between client and server while communicating. I agree that we need `hclog` between plugin-server and porter-client.

There is a reason why I am biased to read plugin errors from stderr in porter. In `porter-azure-plugins`,
https://github.com/deislabs/porter-azure-plugins/blob/f93ef852df2c4c7f0752cc3c20e3bece2f3543e6/pkg/azure/run.go#L21-L35,
we do validations which can be done by `cobra` with little config (ofcourse, we loose some customization) and usually `cobra`writes errors to stderr


(assuming plugins will be developed by "non porter authors"[1]) plugin authors will have flexibility to do validation by themself or leave it to `cobra`. Actually using `cobra` for validation reduces some boilerplate code.

[1] sorry, if it feels offensive. not so good at english

pkg/plugins/pluggable/loader.go
```diff
@@ -95,6 +97,9 @@ func (l *PluginLoader) Load(pluginType PluginTypeConfig) (interface{}, func(), e
 	rpcClient, err := client.Client()
 	if err != nil {
 		cleanup()
+		if stderr := errbuf.String(); stderr != "" {
```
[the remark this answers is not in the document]
dev-drprasad: I think, if plugin writes error to stderr, mostly probably plugin won't start it's rpcserver. So, porter receives error at `client.Client()` and returns error immediately without executing `rpcClient.Dispense`. So, we don't need to check for error after `rpcClient.Dispense`

I tried hours writing unit tests but I had some problems making them pass. Anyway I will add them again and lets see

dev-drprasad:
@vdice @carolynvs 
I added a test case to check errors thrown by plugins.
What I did to make test pass is completely not right way to do it. But, I dont know how to properly mock plugin execution.

pkg/context/helpers.go
```diff
@@ -51,8 +51,8 @@ func NewTestContext(t *testing.T) *TestContext {
 
 func NewTestCommand() CommandBuilder {
 	return func(command string, args ...string) *exec.Cmd {
-		testArgs := append([]string{command}, args...)
-		cmd := exec.Command(os.Args[0], testArgs...)
+		// testArgs := append([]string{command}, args...)
+		cmd := exec.Command(command, args...)
```
dev-drprasad, line 55: I don't understand purpose of `NewTestCommand`. I see, it changes command path to `$(pwd)/debug.test`. But for plugin execution to work, cmd path should be correct

dev-drprasad:
added fake plugin `badplugin` which is binary file to mock error. not able to come up with better idea to mock.

incase if you're curious, its source code is
```c
#include <stdio.h>

int main()
{
  fprintf(stderr, "invalid plugin key\n");
}
```

pkg/context/helpers.go
vdice: Hmmm, I'm not sure on the genesis of this code either...

carolynvs: NewTestCommand redirects commands executed by porter back to the test binary so that it can handle the command and assert that it 

1. was called
2. the proper arguments were passed

It works in conjunction with the TestMain functions declared in some of our test packages.

Otherwise executing a command from porter during a unit test, like calling a mixin, would try to call real binaries on the file system and you would be testing the binary's behavior. This wasn't what we wanted for unit tests.

We disable this behavior in some places in porter, for example in the integration tests:

https://github.com/deislabs/porter/blob/b22ec2bdbacaceb8a24521a47a145380cd94e4d1/pkg/porter/helpers.go#L68-L73

dev-drprasad: Got it. I don't see any mocked mixin execution to mock plugin execution. So, I decided to add it as integration test

dev-drprasad:
@carolynvs @vdice 
I added comment explaining why this change is necessary. I added test case. Running it without considering stderr will make it fail.  What do you say ? should we continue on this PR ?

pkg/plugins/pluggable/plugins_integration_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"path"
+	"testing"
+
+	"get.porter.sh/porter/pkg/config"
+	"get.porter.sh/porter/pkg/secrets"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPlugins_CatchStderr(t *testing.T) {
+	c := config.NewTestConfig(t)
```
carolynvs, line 16: This runs the test in a temp directory instead of in your real porter home directory.

```suggestion
        home, err := ioutil.TempDir("/tmp", "porter")
	require.NoError(t, err, "could not create porter home directory")

	c := config.NewTestConfig(t)
	c.SetHomeDir(home)
```

Makefile
```diff
@@ -79,8 +79,9 @@ test: clean-last-testrun test-unit test-integration test-cli
 test-unit:
 	$(GO) test ./...
 
-test-integration: build
-	$(GO) test -timeout 20m -tags=integration ./...
+test-integration: 
+	$(GO) build -o $(PORTER_HOME)/testplugin ./cmd/testplugin
+	PROJECT_ROOT=$(shell pwd) $(GO) test -timeout 20m -tags=integration ./...
```
dev-drprasad, line 84: I was not able to copy testplugin binary from `bin` directory, since relative path not working. Even if we run `make test-integration` from project root directory, `os.Getwd` will be parent directory of test file. So, I needed `PROJECT_ROOT` to access `bin` directory.

carolynvs:
The integration test is failing due to an unrelated problem with our CI build. The porter and mixin binaries are not being built suddenly during the integration job. I'm not sure why? 🤷‍♀️ Honestly, it's not clear why we were getting an implicit `make build` executed in the first place. So I have opened a PR #1061  to make the build step explicit like it was already for the CLI test job.

I am thinking that once that passes, and is merged, you can rebase and the tests will pass.

dev-drprasad:
@carolynvs lets run integration tests again since #1061 merged

dev-drprasad:
I love it when tests pass

PR #1033, opened 2020-05-13
fix(manifest.go): update default invocation image naming

vdice:
# What does this change
* Updates default invocation image naming to allow for registries of the domain:port pattern
# What issue does it fix
N/A, but discovered via investigation into https://github.com/deislabs/porter/issues/1030

# Notes for the reviewer
N/A
# Checklist
- [x] Unit Tests
- [x] Documentation - N/A

### DIFF
--- a/pkg/manifest/manifest.go
+++ b/pkg/manifest/manifest.go
@@ -532,7 +532,12 @@ func UnmarshalManifest(manifestData []byte) (*Manifest, error) {
 
 func (m *Manifest) SetDefaults() {
 	if m.Image == "" && m.BundleTag != "" {
-		registry := strings.Split(m.BundleTag, ":")[0]
+		tag := m.BundleTag
+		// Derive the last index of ":"
+		// This allows for a registry consisting of a domain:port to be included
+		li := strings.LastIndex(tag, ":")
+		// Split on this last index
+		registry := tag[0:li]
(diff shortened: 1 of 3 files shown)


pkg/manifest/manifest.go
```diff
@@ -532,7 +532,12 @@ func UnmarshalManifest(manifestData []byte) (*Manifest, error) {
 
 func (m *Manifest) SetDefaults() {
 	if m.Image == "" && m.BundleTag != "" {
-		registry := strings.Split(m.BundleTag, ":")[0]
+		tag := m.BundleTag
```
carolynvs, line 535: What do you think about using the docker library instead so that we aren't trying to replicate their logic? I think this should work, I ran it against your unit test

```go
ref, err := reference.ParseNormalizedNamed(m.BundleTag)
if err != nil {
	return errors.Wrapf(err, "invalid tag %s", m.BundleTag)
}
registry := reference.FamiliarName(ref)
```

Since ParseNormalizedName _could_ return an error when the tag is invalid, it would require `SetDefaults` changing to return an error though.

vdice: I like it; will update.  Thank you for the head start!

vdice: Updated!

PR #1042, opened 2020-05-15
Add airgap example

carolynvs:
# What does this change
Add an example of shoving a bundle over an airgap. 🐳 💨🚜 

TODO:
- [x] https://github.com/deislabs/porter/pull/1047#discussion_r431995706

# What issue does it fix
Needed it for a demo and people asked a lot about it in Slack.

# Notes for the reviewer
* As I wrote this I realized that our docs don't explain how to do this, so there will be follow-up PRs.
* Found a bug in how we rename images when we copy them to another registry, need to split that out into another PR. It's in here because it broke my demo...

# Checklist
- [ ] Unit Tests
- [x] Documentation
- [ ] Schema (porter.yaml)


carolynvs:
@vdice Okay this is finally ready for review

PR #1047, opened 2020-05-15
Selectively replace the org when moving images

### LINKED ISSUE #1040: publish string replacement is greedy
## Describe the bug
When we move a bundle from one repository to another, the logic to fixup the image names needs to be more strict, as it's replacing too much right now.

Src Name: `getporter/whalegap-installer`
Expected Target: `getporter.azurecr.io/whalegap-installer`
Actual Target: `whalegap.azurecr.io/whalegap-installer`

## To Reproduce
1. Archive a bundle from the getporter registry
2. Try to publish it to the getporter.azurecr.io registry

## Expected behavior
The bundle should be published to the new registry but instead it fails because the string replacement was too greedy.

## Porter Command and Output

```console
$ porter publish --archive whalegap.tgz --tag getporter.azurecr.io/whalegap:v0.1.0 --debug

Resolved porter binary from /usr/local/bin/porter to /Users/carolynvs/.porter/porter
Extracting bundle from archive /tmp/whalegap/whalegap.tgz...
Beginning bundle publish to getporter.azurecr.io/whalegap:v0.1.0. This may take some time.
Error: unable to push image whalegap.azurecr.io/whalegap/whalegap-installer:0.1.0: failed to write image sha256:dbe7633bd096ced6eb50bdac4c4e269383326803d083f3ff36129a949b73dbd2 
to whalegap.azurecr.io/whalegap/whalegap-installer:0.1.0: failed to write image 
whalegap.azurecr.io/whalegap/whalegap-installer:0.1.0: Get https://whalegap.azurecr.io/v2/: dial 
tcp: lookup whalegap.azurecr.io on 192.168.0.2:53: no such host
```

## Version
```
porter v0.26.2-beta.1-6-gfd0d726c (fd0d726c)
```

carolynvs:
# What does this change
Don't use a string replace because when the same text is used in the org
and the image name and the registry, it can end up changing more text
than it should.

Don't replace the org when there isn't one present. Some registries
embed the org in the domain, e.g. getporter.azurecr.io, in which case
the path of the tag is the name of the bundle, and we shouldn't be
appending that to the images.

# What issue does it fix
Closes #1040 

# Notes for the reviewer

# Checklist
- [x] Unit Tests
- [x] Documentation (working on that now)
- [x] Schema (porter.yaml) - Not Impacted

### DIFF
--- a/docs/content/distribute-bundles.md
+++ b/docs/content/distribute-bundles.md
@@ -96,7 +96,7 @@ porter publish -a mybunz1.1.tgz --tag getporter/megabundle:1.1.0
 ## Image References After Publishing
 
 When a bundle is published, all images [referenced][image-map] by the bundle are
-published **inside** bundle repository. Bundles should be written to use the
+published **inside** the bundle repository. Bundles should be written to use the
 relocated image locations.
 
 * REGISTRY/ORG/BUNDLE:TAG
@@ -150,7 +150,7 @@ Here we can see that the spring-music image was relocated inside the published b
 
 Publishing archived bundles will result in intermediate artifacts that are not
 used in the final bundle. New repositories are created for each image used by
-the bundle on the destination registry. These are a side effect of the publishing
+the bundle on the destination registry. These are side effects of the publishing
 process and they are not used by the final bundle.
 
 Using the example above, the following repositories are created and can be
--- a/docs/content/distribute-bundles.md
+++ b/docs/content/distribute-bundles.md
@@ -7,6 +7,11 @@ aliases:
 
 Once you have built a bundle with Porter, the next step is to share the bundle and invocation image so others can use it. Porter uses OCI (Docker) registries to share both CNAB bundle manifest and invocation images.
 
+* [Preparing For Bundle Publishing](#preparing-for-bundle-publishing)
+* [Bundle Publish](#bundle-publish)
+* [Publish Archived Bundles](#publish-archived-bundles)
+* [Image References After Publishing](#image-references-after-publishing)
+ 
 ## Preparing For Bundle Publishing
 
 Before you can publish your bundle, you must first run a `porter build` command. This will create the invocation image so it can be pushed to an OCI (Docker) registry along with your CNAB bundle manifest. It's a good idea to work with your bundle and test it locally before you publish it to a registry.
@@ -90,7 +95,18 @@ porter publish -a mybunz1.1.tgz --tag getporter/megabundle:1.1.0
 
 ## Image References After Publishing
 
-When a bundle is published, the images that it will use are copied into the location of the published bundle. This simplifies access control and management of artifacts in the repository. Consider the following `porter.yaml` snippet:
+When a bundle is published, all images [referenced][image-map] by the bundle are
+published **inside** bundle repository. Bundles should be written to use the
+relocated image locations.
+
+* REGISTRY/ORG/BUNDLE:TAG
+  * REGISTRY/ORG/BUNDLE@**INVOCATION_IMAGE_DIGEST**
+  * REGISTRY/ORG/BUNDLE@**REFERENCED_IMAGE_1_DIGEST**
+  * REGISTRY/ORG/BUNDLE@**REFERENCED_IMAGE_2_DIGEST**
+
+NOTE: Digest refers to the the repository digest (not the image id).
+
+Consider the following example:
 
 ```
 name: spring-music
@@ -107,7 +123,10 @@ images:
       digest: "sha256:8f1133d81f1b078c865cdb11d17d1ff15f55c449d3eecca50190eed0f5e5e26f"
 ```
 
-When this bundle is published, both the invocation image and the `spring-music` image will be copied and stored in the context of the bundle. To see this in action, you can use the `porter inspect` command to see what images will actually be used for a given bundle.
+When this bundle is published, both the invocation image and the spring-music
+image will be copied and stored in the context of the bundle. To see this in
+action, you can use the `porter inspect` command to see what images will
+actually be used for a given bundle.
 
 ```
 Name: spring-music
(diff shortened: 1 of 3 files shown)


carolynvs:
Should we have the same logic for renaming images between `porter publish --archive` and `porter copy`? This PR was fixing a bug in `porter publish --archive` and when I went to compare what `porter copy` does so that I could document them on the website (which is implemented in cnab-to-oci) I realized that they do different things... It seems confusing and honestly probably not what any user wants.

Here's what they do today:

**Archive**
* `porter publish --archive whalegap.tgz --tag example.com/org2/whalegap:v0.1.0`.
  * localhost:5000/org1/whalegap:v0.1.0 -> example.com/org2/whalegap:v0.1.0
  * localhost:5000/org1/whalegap-installer -> example.com/org2/whalegap-installer
  * nginx -> example.com/org2/nginx

**Copy**
* `porter copy --source localhost:5000/org1/whalegap:v0.1.0 --tag example.com/org2/whalegap:v0.1.0`.
  * localhost:5000/org1/whalegap-installer -> example.com/org2/whalegap/whalegap-installer
  * nginx -> example.com/org2/whalegap/nginx

The key difference is that **images are tagged under not only the org but the name of the bundle**.

I wish I could find the original issue where we discussed how the image relocation works in cnab-to-oci and why it puts things where it does. (can anyone find it?)

vdice:
Interesting.  `porter publish --archive` seems like the "right" approach to publishing/copying, from one org to another -- no extra "namespacing" on the resulting images.  Is it `porter copy` that delegates to the cnab-to-oci library?  Maybe we can engage discussion with the maintainers and/or at the next CNAB meeting to learn about the reasoning/design.

carolynvs:
> if a source bundle is org/bundle and a source installer is org/bundle/installer is that hierarchy preserved when porter publish --archive is called?

I just tested your scenario with a custom invocation image that uses `bundle/installer` instead of `bundle-installer`

`porter publish --archive --tag example.com/neworg/apache:v0.1.0`
* localhost:5000/myorg/apache/installer -> example.com/neworg/apache/installer

carolynvs:
>  Is it porter copy that delegates to the cnab-to-oci library? Maybe we can engage discussion with the maintainers and/or at the next CNAB meeting to learn about the reasoning/design.

Yes, porter copy uses cnab-to-oci. I can't seem to find where a decision was made to namespace under the bundle name. That logic happens from this code:

https://github.com/cnabio/cnab-to-oci/blob/c9599a06b976f4cd5f505092292776caa924b549/remotes/fixup.go#L202

@radu-matei Do you have a link to why cnab-to-oci behaves this way? Where images referenced by a bundle, when copied to another registry are put under `newregistry/ORG/BUNDLENAME/IMAGENAME`, instead of `newregistry/ORG/IMAGENAME`?

radu-matei:
`cnab-to-oci` [uses an OCI index])(https://github.com/cnabio/cnab-to-oci/#selection-of-oci-index) to represent a bundle in the registry - specifically, the bundle descriptor (`bundle.json`), the invocation image, and component images are part of the said index.

However, an index can only reference objects that are part of the same repository - this will eventually enable registries to treat all artifacts in the index as a whole (from vulnerability scanning to garbage collection).
(There is an ongoing discussion in the OCI community about allowing manifests to reference _foreign_ objects, and it could enable this scenario, but the discussion is not advanced enough to continue that discussion).

So technically, because `cnab-to-oci` publishes an index, it cannot publish the images referenced in the bundle under `newregistry/ORG/IMAGENAME`.
(There is an open question as to whether it should tag the individual images from `newregistry/ORG/BUNDLENAME/IMAGENAME` though.)

Ultimately, `cnab-to-oci` ensures your bundle and referenced artifacts are available in the same _repository_ - so if a user has access to the bundle, they also have access to pulling the images.

carolynvs:
In the cnab meeting yesterday we decided to fix just the bug here, and document what each command is doing under the hood, and what resulting repositories are created.

carolynvs:
I've updated https://deploy-preview-1047--porter.netlify.app/distribute-bundles/#image-references-after-publishing to clarify how images are renamed during publish normally and the intermediate artifacts generated during a `porter publish --archive`.

docs/content/distribute-bundles.md
```diff
@@ -90,7 +95,18 @@ porter publish -a mybunz1.1.tgz --tag getporter/megabundle:1.1.0
 
 ## Image References After Publishing
 
-When a bundle is published, the images that it will use are copied into the location of the published bundle. This simplifies access control and management of artifacts in the repository. Consider the following `porter.yaml` snippet:
+When a bundle is published, all images [referenced][image-map] by the bundle are
+published **inside** bundle repository. Bundles should be written to use the
```
radu-matei, line 99: I don't really follow what is the action a bundle author should take from this.
Is this part not covered by relocation maps?

docs/content/distribute-bundles.md
```diff
@@ -90,7 +95,18 @@ porter publish -a mybunz1.1.tgz --tag getporter/megabundle:1.1.0
...
+relocated image locations.
+
+* REGISTRY/ORG/BUNDLE:TAG
+  * REGISTRY/ORG/BUNDLE@**INVOCATION_IMAGE_DIGEST**
+  * REGISTRY/ORG/BUNDLE@**REFERENCED_IMAGE_1_DIGEST**
+  * REGISTRY/ORG/BUNDLE@**REFERENCED_IMAGE_2_DIGEST**
+
+NOTE: Digest refers to the the repository digest (not the image id).
```
radu-matei, line 107: Maybe link [this section of the OCI specification](https://github.com/opencontainers/image-spec/blob/master/descriptor.md#digests)?

docs/content/distribute-bundles.md
```diff
@@ -107,20 +123,40 @@ images:
...
+
+### Archived Bundle Artifacts
+
+Publishing archived bundles will result in intermediate artifacts that are not
+used in the final bundle. New repositories are created for each image used by
+the bundle on the destination registry. These are a side effect of the publishing
+process and they are not used by the final bundle.
+
+Using the example above, the following repositories are created and should be
+ignored:
```
radu-matei, line 157: While accurate for scenarios where users are _only_ interested in the bundle alone, as we saw in the airgapped environments meeting, having the images in their own repositories could be a valid use case for a subset of users.

So rather than a side effect, we _could_ potentially see this as a feature.

docs/content/distribute-bundles.md
carolynvs: There is another [PR](https://github.com/deislabs/porter/pull/1042) where I've been working on building up docs for how to actually use the relocation map properly in your bundle. For example just populating the `images` section isn't enough, you also then have to use helm charts that understand image digests (not tags which they all use) and then actually use the relocation map to get the final location of your images. This is something that porter helps with.

That PR is being held up by this bug at the moment. So I can explain more and then link to that once this is merged. I'll make a note of it on that PR?

docs/content/distribute-bundles.md
carolynvs: I'll change to "can be ignored" because for the purpose of the bundle, they are not used and we need to make that clear but yeah they could use it if they wanted to.

Issue #398, opened 2019-06-07
creds generate duplicates prompt when you arrow up/down


carolynvs:
When I arrow up/down in the `porter credentials generate` interactive prompt, it duplicates the "how would you like to set credential "X" [Use arrows to move...]" line every time I press the arrow key.

![Screen Shot 2019-06-07 at 10 22 27 AM](https://user-images.githubusercontent.com/1368985/59115094-2bdd9480-890e-11e9-980e-69f68004809f.png)

dev-drprasad:
@carolynvs this happen only when user terminal width is less content width. this is bug in [survey](https://github.com/AlecAivazis/survey) and there is already an issue https://github.com/AlecAivazis/survey/issues/101.

There is a hackfix mentioned in https://github.com/AlecAivazis/survey/issues/101#issuecomment-420923209. But it will cause overflown text to disappear like shown in below image
<img width="722" alt="Screenshot 2020-02-18 at 9 17 46 PM" src="https://user-images.githubusercontent.com/22687681/74752104-31102300-5294-11ea-96d8-0466784f22b9.png">
As you can see sentence `Use arrows to move, space to selec` not complete.

We can use above solution and change question to something like `Resolve "CLIENT_ID" from [Use arrows to move..........` to prevent overflow to some extent.

What do you say ?

dev-drprasad:
I realized hiding overflown text may be bad idea. Because it hides user entered values also if they are lengthy. I have another solution which doesn't solve problem completey but helps to some extent.

We can add newlines in questions to prevent overflow. For example, after adding newline after question, prompt will look like below
<img width="668" alt="Screenshot 2020-02-19 at 12 36 35 PM" src="https://user-images.githubusercontent.com/22687681/74810019-7d9b4300-5314-11ea-91e2-818c6b484de0.png">

What do you think ?

carolynvs:
Thanks for the great troubleshooting! I apologize for not seeing this until now. I love the solution. Let's go with what you came up with. 💯

carolynvs:
Fixed close enough in #1056. If we want to do more, let's open a separate issue. Since we want to move more towards generating files for people to edit, vs do interactive prompts, I am not sure we need to fix further.

PR #1056, opened 2020-05-27
🐛 Fix duplicate question while up/down-ing, while surveying

dev-drprasad:
# What does this change
Fixes duplicate question while up/down-ing, while surveying

# What issue does it fix
Partial prevents problem mentioned at #398 . But doesn't completely fix it

# Notes for the reviewer
   I couldn't resist adding extra space after question

# Checklist
- [ ] Unit Tests
- [x] Documentation - N/A

### DIFF
--- a/pkg/credentialsgenerator/generator.go
+++ b/pkg/credentialsgenerator/generator.go
@@ -93,13 +93,15 @@ func genEmptyCredentials(name string) (credentials.CredentialStrategy, error) {
 
 func genCredentialSurvey(name string) (credentials.CredentialStrategy, error) {
 
+	// extra space-suffix to align question and answer
 	sourceTypePrompt := &survey.Select{
-		Message: fmt.Sprintf("How would you like to set credential %q", name),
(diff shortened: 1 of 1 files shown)


carolynvs:
Just so we remember what the fix looks like, here's a screencap:

1. The "[Use arrows to move...]" prompt is put on a new line so that it is **less likely** to wrap and trigger the bug.
2. A space was added which indents that text a bit under the question.

Screen Shot 2020-06-08 at 3 49 54 PM

dev-drprasad:
@carolynvs 

> The "[Use arrows to move...]" prompt is put on a new line so that it is less likely to wrap and trigger the bug.

Yes

> A space was added which indents that text a bit under the question.

Space was added to align question with answer. Unfortunately it misaligns help text

Screenshot 2020-06-09 at 1 20 32 PM

One more thing, I removed the work `credential` from `Enter the %s that will be used to set credential %q` to make it short. Is that OK ?

I put it in draft mode cuz I want to write A test. But I couldn't find a way to simulate a terminal with specific width to trigger bug. So, Lets just merge it

Issue #937, opened 2020-03-04
Migrate claims to new structure


carolynvs:
The claim spec is changing and this breaks with the current structure used. We need to identify our migration plan (up-front or transparent like we did with credential sets) and then implement the migration from the old structure to the new. It will need to be supported for some defined period of releases (perhaps until 1.0?).

See 
* https://github.com/cnabio/cnab-spec/pull/339
* https://github.com/cnabio/cnab-spec/pull/344

carolynvs:
Closed by #1090

PR #1065, opened 2020-06-04
Use latest claims code in cnab-go

### LINKED ISSUE #770: Keep claim for uninstalled bundles
Instead of removing the claim when a bundle is uninstalled, and losing all record that it was installed, and when it was uninstalled. Let's investigate if we can record the uninstall in the claim instead.

Some of this is enshrined in cnab-go, perhaps because behavior from duffle was copied into that library and the spec is not clear on the uninstall behavior that must be implemented.

This is part of our general preference to preserve an audit trail and not just the last performed action.

### LINKED ISSUE #836: Outputs from previous actions are lost
## Describe the bug
Old outputs are overwritten by new actions.

## To Reproduce
1. Run an action that generates a bundle level output
2. Run another action that does not generate that same named output
3. Note that the output from the first action is no longer persisted in the claim

## Expected behavior
The original output should still be persisted and available.

## Version

`porter v0.22.0-beta.1 (8f3e429)`

### LINKED ISSUE #935: EPIC: Store claim revisions separately
We followed duffle's implementation for claims (and cnab-go), which overwrites the previous claim with the latest revision. This isn't how claims were intended to be used per the spec turns out.

We need to rework the storage of claims so that claim storage is like so:

claims/
* NAME/
  * revision A
  * revision B
  * revision C (latest)

When we retrieve a claim, it should retrieve latest by listing the revisions and returning the latest by revision (ULID).

After we have this in place, we can create follow-on issues to add more enhancements:
* History of actions performed on an installation TODO
* Fix losing outputs #836
* Do not delete the installation entirely when uninstall is executed #770

carolynvs:
# What does this change
This gets porter using the latest claims spec that is now supported in cnab-go. It doesn't take advantage yet of new features that we could do with the additional data, just ensures that we still build and work.

# What issue does it fix
Closes #935 Store claim revisions separately
Closes #836 Outputs from previous actions are lost
Closes #770 Keep Claim for uninstalled bundles

# Notes for the reviewer
* Can't merge until https://github.com/cnabio/cnab-go/pull/212 is merged
* Claims migration is necessary, so I'm skipping tests (#937)
* Parameter sources is necessary to replace outputs in claims, so those tests are skipped too (#1066)

# Checklist
- [ ] Unit Tests
- [ ] Documentation


pkg/cnab/provider/install.go
```diff
@@ -3,58 +3,79 @@ package cnabprovider
...
+		return errors.Wrap(err, "failed to install the bundle")
 	}
 
-	if d.Debug {
+	return a.SaveOperationResult(opResult, c, result)
+}
+
+// appendFailedResult creates a failed result from the operation error and accumulates
+// the error(s).
+func (r *Runtime) appendFailedResult(opErr error, c claim.Claim) error {
```
vdice, line 62: nit: should this function definition live in a more general-purpose file untied to an action?  (`runtime.go` perhaps?)

pkg/cnab/provider/uninstall.go
```diff
@@ -1,76 +1,71 @@
...
-	d.Extensions = exts
 
-	c.Parameters, err = d.loadParameters(&c, args.Params, string(manifest.ActionUninstall))
+	params, err := r.loadParameters(b, args.Params, claim.ActionUpgrade)
 	if err != nil {
 		return errors.Wrap(err, "invalid parameters")
 	}
 
-	driver, err := d.newDriver(args.Driver, c.Installation, args)
+	c, err := existingClaim.NewClaim(claim.ActionUpgrade, b, params)
```
vdice, line 39:
```suggestion
	c, err := existingClaim.NewClaim(claim.ActionUninstall, b, params)
```

pkg/cnab/provider/uninstall.go
```diff
@@ -1,76 +1,71 @@
...
+
+	// If the user didn't override the bundle definition, use the one
+	// from the existing claim
+	if args.BundlePath == "" {
+		b = existingClaim.Bundle
 	}
-	d.Extensions = exts
 
-	c.Parameters, err = d.loadParameters(&c, args.Params, string(manifest.ActionUninstall))
+	params, err := r.loadParameters(b, args.Params, claim.ActionUpgrade)
```
vdice, line 34:
```suggestion
	params, err := r.loadParameters(b, args.Params, claim.ActionUninstall)
```

pkg/porter/list.go
```diff
@@ -15,68 +17,101 @@ type ListOptions struct {
...
-			Name:     claim.Installation,
-			Created:  claim.Created,
-			Modified: claim.Modified,
-			Action:   claim.Result.Action,
-			Status:   claim.Result.Status,
+	var displayInstallations DisplayInstallations
+	for _, installation := range installations {
+		displayInstallation, err := NewDisplayInstallation(installation)
+		if err != nil {
+			// There isn't an installation for this result
```
vdice, line 87: Should we add a debug line here if p.Debug is set?

pkg/storage/pluginstore/store.go
```diff
@@ -61,11 +61,13 @@ func (s *Store) Connect() error {
 	}
 
 	s.BackingStore = crud.NewBackingStore(store)
+	s.BackingStore.AutoClose = false
 
 	return nil
 }
 
 func (s *Store) Close() error {
+	return nil
```
vdice, line 70: Was this line intentional?

carolynvs:
I moved this out of draft because otherwise I don't get warned about merge conflicts. I still need to address your comments though!

pkg/porter/list.go
carolynvs: Looking at that more closely the real problem is for situations where we can't determine when the bundle was installed, maybe because they have only run custom actions but not install yet. Let me handle that better.

pkg/storage/pluginstore/store.go
carolynvs: Nope! Thanks, that's weird not sure where that came from.

pkg/cnab/provider/uninstall.go
```diff
@@ -1,76 +1,70 @@
...
-	c, err := d.claims.Read(args.Claim)
-	if err != nil {
-		// Yay! It's already gone
-		if err == claim.ErrClaimNotFound {
-			return nil
-		}
-		return errors.Wrapf(err, "could not load bundle instance %s", args.Claim)
-	}
+func (r *Runtime) Uninstall(args ActionArguments) error {
+	// TODO: (carolynvs) consolidate logic across actions
```
vdice, line 11: Has this item been completed?  (It seems so.)

scripts/test/test-hello.sh
```diff
@@ -18,10 +18,11 @@ sed -i "s/getporter\/porter-hello/${REGISTRY}\/porter-hello/g" porter.yaml
 
 ${PORTER_HOME}/porter build
 ${PORTER_HOME}/porter install --debug
-cat ${PORTER_HOME}/claims/HELLO.json
+${PORTER_HOME}/porter instance show HELLO --debug
```
vdice, line 21: Nice catch/update!

pkg/cnab/provider/uninstall.go
carolynvs: I was thinking that I could consolidate the implementation of install/upgrade/invoke/uninstall into a single function, there's a lot of copy paste that is easy to mess up and not realize it. But there also aren't any unit tests either, just integration tests. Let me take a stab at it and see if I can change these safely.

carolynvs: OK! Just pushed a commit with some unit tests and all the install/upgrade/invoke/uninstall consolidated.

vdice: 👏  https://github.com/deislabs/porter/pull/1065/commits/487403ae297af20a1edc8436536ad5abafe137c5 looks great!

PR #1075, opened 2020-06-08
Change defaulting of bundle/image tags

### LINKED ISSUE #1062: Porter not respecting specified tag value
## Describe the bug
a porter.yaml file with the `tag: org/bundle:v0.1.0` ends up creating the installer image WITHOUT the `v` character.

## To Reproduce
1. `porter create`
2. modify the porter.yaml file to include `tag: foo/foobundle:v0.1.0`
3. `porter build`
4. `docker run -it --rm foo/foobundle:v0.1.0`

## Expected behavior
jump into the container and sit at the bash prompt

## Porter Command and Output
```
╭─ralph at surfacebook in ~/work/nogo/squillace/powerapp using
╰─○ porter publish
Pushing CNAB invocation image...
The push refers to repository [docker.io/squillace/powerapp-function-installer]
8408421d36e2: Preparing
ec8157663efe: Preparing
7d472ca9c113: Preparing
5597b553b85a: Preparing
fe9f7b8998ed: Preparing
2c995a2087c1: Preparing
2c995a2087c1: Waiting
fe9f7b8998ed: Mounted from squillace/hello-installer
7d472ca9c113: Pushed
8408421d36e2: Pushed
2c995a2087c1: Mounted from squillace/hello-installer
ec8157663efe: Pushed
5597b553b85a: Pushed
0.1.0: digest: sha256:7f3a2704f587846201671eafdddc7a21112d7e2db1c190fd7359742532028288 size: 1579
Rewriting CNAB bundle.json...
Starting to copy image squillace/powerapp-function-installer:0.1.0...
Completed image squillace/powerapp-function-installer:0.1.0 copy
Bundle tag docker.io/squillace/powerapp-function:v0.1.0 pushed successfully, with digest "sha256:de08c80d2fcd9b2913ec8176b27d3b2c94e4ccb1cc57f03e00b15554394c5b7b"
╭─ralph at surfacebook in ~/work/nogo/squillace/powerapp using
╰─○ docker run -it --rm squillace/powerapp-function-installer:v0.1.0 bash
Unable to find image 'squillace/powerapp-function-installer:v0.1.0' locally
docker: Error response from daemon: manifest for squillace/powerapp-function-installer:v0.1.0 not found: manifest unknown: manifest unknown.
See 'docker run --help'.
```

## Version

porter v0.26.2-beta.1 (e4aa6cd3)

carolynvs:
# What does this change
When bundle tag is omitted, default to the bundle version with a 'v' prefixed. For example:

```yaml
version: 1.2.3
tag: example.com/mybun
```

This bundle publishes to `example.com/mybun:v1.2.3`

When the invocation image is omitted, default to the same tag as the bundle. In the example above, the invocation image pushed would be `example.com/mybun-installer:v1.2.3`. This addresses the confusion where the invocation image would not have a 'v' prefix but the bundle would because the invocation image's tag was based on the bundle version.

# What issue does it fix
Closes #1062 

# Notes for the reviewer
I moved when we call `SetDefaults` to avoid how many tests were impacted by the change since many of our tests use partial porter.yaml files.

# Checklist
- [x] Unit Tests
- [x] Documentation


carolynvs:
cc @squillace This doesn't yet to the full defaulting where it could generate the entire bundle tag (we would still need to introduce `registry` to the manifest and `--registry` to publish but it gets us much closer. It should fix what you reported in #1062 where the invocation image and the bundle end up with different tags though.

PR #1076, opened 2020-06-09
ref(*): remove param file flag/logic

vdice:
# What does this change
* Removes `--param-file` flag/logic in favor of parameter set as intro'd in https://github.com/deislabs/porter/pull/1068

# What issue does it fix
Closes https://github.com/deislabs/porter/issues/1072
Contributes towards https://github.com/deislabs/porter/issues/1053

# Notes for the reviewer
N/A

# Checklist
- [x] Unit Tests
- [x] Documentation


pkg/porter/lifecycle_test.go
```diff
@@ -87,7 +87,7 @@ func TestInstallFromTagIgnoresCurrentBundle(t *testing.T) {
 func TestBundleLifecycleOpts_ToActionArgs(t *testing.T) {
 	cxt := context.NewTestContext(t)
 
-	cxt.AddTestFile("testdata/install/base-params.txt", "base-params.txt")
+	cxt.AddTestFile("testdata/install/base-params.json", "base-params.json")
```
carolynvs, line 90: This test file isn't used because reading the parameter set doesn't happen until the bundle is executed. Previously we would read the parameter file (with key value pairs) in this function, but now this isn't needed at all, we are just recording the name passed in.

pkg/porter/testdata/install/base-params.json
```diff
@@ -0,0 +1,11 @@
+{
```
carolynvs, line 1: I think we can delete this file (per comment above)?

PR #1088, opened 2020-06-15
fixes to install scripts for path set up

### LINKED ISSUE #1081: Fix installation script to include adding porter to your bash profile
## Describe the bug
When installing porter, I was not aware that I had to add the porter path to my bash profile. This was confusing for me when I tried to work with porter afterwards because the command was not recognized. 

## To Reproduce
1. Install porter
2. Open a new terminal window and type any porter command
3. See error

## Expected behavior
After installing porter, have clear directions that explain the path needs to be added to the bash profile and maybe show how. 

## Version
porter v0.26.3-beta.1 (c0dc3415)

gaurimadhok:
# What does this change
This changes the instructions given at the end of the install script to tell the user how to add porter to their path. 

# What issue does it fix
Closes #1081

### DIFF
--- a/docs/content/install.md
+++ b/docs/content/install.md
@@ -11,8 +11,7 @@ We have a few release types available for you to use:
 
 You can also install and manage [mixins](#mixins) using porter,
 and use the [Porter VS Code Extension][vscode-ext] for help
-authoring bundles. For Windows installation, you will need to
-create a [PowerShell Profile][ps-link] if you do not have one. 
+authoring bundles. 
 
 [vscode-ext]: https://marketplace.visualstudio.com/items?itemName=ms-kubernetes-tools.porter-vscode
 [ps-link]: https://www.howtogeek.com/126469/how-to-create-a-powershell-profile/
@@ -32,6 +31,8 @@ curl https://cdn.porter.sh/latest/install-linux.sh | bash
 ```
 
 ## Latest Windows
+You will need to create a [PowerShell Profile][ps-link] if you do not have one.
+
 ```
 iwr "https://cdn.porter.sh/latest/install-windows.ps1" -UseBasicParsing | iex
 ```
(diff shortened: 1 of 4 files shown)


docs/content/install.md
```diff
@@ -11,9 +11,11 @@ We have a few release types available for you to use:
 
 You can also install and manage [mixins](#mixins) using porter,
 and use the [Porter VS Code Extension][vscode-ext] for help
-authoring bundles.
+authoring bundles. For Windows installation, you will need to
```
carolynvs, line 14: Hmm, I see why you put this note here, because otherwise it would be copied 3 times on the page. However, people will link directly to the windows instructions like this:

https://deploy-preview-1088--porter.netlify.app/install/#latest-windows

and then they probably won't see the note. I recommend removing the message from this location and copy/pasting it into each windows installation section with the "For windows installation," removed since we are in the context of the windows section. 

I realize it seems silly to duplicate but hopefully it will make it more likely that they see the text.

scripts/install/install-mac.sh
```diff
@@ -27,5 +27,5 @@ $PORTER_HOME/porter mixin install gcloud --version $PKG_PERMALINK
 $PORTER_HOME/porter plugin install azure --version $PKG_PERMALINK
 
 echo "Installation complete."
-echo "Add porter to your path by running:"
+echo "Add porter to your path by adding the following line to your ~/.bash_profile:"
```
carolynvs, line 30: I forgot that the latest version of MacOS (Catalina) changed the default shell from Bash to ZSH. Their profile file is `~/.zprofile`. Let's change this a bit to account for the user having either:

```suggestion
echo "Add porter to your path by adding the following line to your ~/.bash_profile or ~/.zprofile:"
```

scripts/install/install-linux.sh
```diff
@@ -26,5 +26,5 @@ $PORTER_HOME/porter mixin install gcloud --version $PKG_PERMALINK
 $PORTER_HOME/porter plugin install azure --version $PKG_PERMALINK
 
 echo "Installation complete."
-echo "Add porter to your path by running:"
+echo "Add porter to your path by adding the following line to your ~/.bash_profile:"
```
carolynvs, line 29: I just double checked and on Debian, it comes with `~/.profile` by default instead of `~/.bash_profile`. Let's have them edit that instead.

carolynvs:
The preview of the changes to the install page look good! 👍 

https://deploy-preview-1088--porter.netlify.app/install/

PR #1101, opened 2020-06-29
Remove internal params in `explain` output

### LINKED ISSUE #1091: internal parameters should not be displayed
## Describe the bug
The porter-debug parameter is printed out during `porter explain`

## To Reproduce

Create a bundle with porter and then run porter explain on it.

## Expected behavior
Any internally generated parameters, porter-* should not be displayed during `porter explain` or `porter inspect`.

## Porter Command and Output
```
$ porter explain
Name: whalegap
Description: An example bundle that demonstrates how to sneak a whale-sized bundle through an airgap
Version: 0.1.0

Credentials:
Name         Description   Required
kubeconfig                 true

Parameters:
Name           Description                                                     Type      Default              Required   Applies To
msg            a message for the whales to speak                               string    whale hello there!   false      All Actions
porter-debug   Print debug information from Porter when executing the bundle   boolean   false                false      All Actions
release        helm release name                                               string    whalegap             false      All Actions
```

## Version
```
$ porter version
porter v0.26.3-beta.1-27-g637c4311 (637c4311)
```

dev-drprasad:
# What does this change
Don't show internal params in output of `explain`

# What issue does it fix
Closes #1091 

# Notes for the reviewer

# Checklist
- [x] Unit Tests
- [x] Documentation (N/A)
- [x] Schema (porter.yaml) (N/A)

### DIFF
--- a/pkg/porter/explain.go
+++ b/pkg/porter/explain.go
@@ -171,10 +171,10 @@ func (p *Porter) printBundleExplain(o ExplainOpts, pb *PrintableBundle) error {
 	}
 }
 
-func checkIsInternalParam(param string) bool {
+func checkIsInternalParam(inputParam string) bool {
 	isInternalParam := false
 	for _, internalParam := range porterInternalParams {
-		if param == internalParam {
+		if inputParam == internalParam {
 			isInternalParam = true
 			break
 		}
--- a/pkg/porter/explain.go
+++ b/pkg/porter/explain.go
@@ -114,6 +114,8 @@ func (s SortPrintableAction) Swap(i, j int) {
 	s[i], s[j] = s[j], s[i]
 }
 
+var porterInternalParams = []string{"porter-debug"}
+
 func (o *ExplainOpts) Validate(args []string, cxt *context.Context) error {
 	err := o.sharedOptions.Validate(args, cxt)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


dev-drprasad:
@carolynvs You mentioned, `porter inspect` shows parameter list in issue. But I see `porter inspect` showing only image and invocation image list. So, I added necessary changes to `explain` command only

pkg/porter/explain.go
```diff
@@ -114,6 +114,8 @@ func (s SortPrintableAction) Swap(i, j int) {
 	s[i], s[j] = s[j], s[i]
 }
 
+var porterInternalParams = []string{"porter-debug"}
```
carolynvs, line 117: If we make this a `map[string]struct{}` (optimized for memory) or `map[string]bool` (not optimized but can be easier to understand sometimes), and then checkIsInternalParam can be simplified, where it returns the second return value from querying the map

```golang
var porterInternalParams = map[string]struct{}{
    "porter-debug": struct{}{},
}

// ...

// ok holds if the specified key is present in the map
_, ok := porterInternalParams[inputParam]
return ok
```

pkg/porter/explain.go
```diff
@@ -169,6 +171,17 @@ func (p *Porter) printBundleExplain(o ExplainOpts, pb *PrintableBundle) error {
 	}
 }
 
+func checkIsInternalParam(inputParam string) bool {
```
carolynvs, line 174: suggestion: This would read better in English if we changed the name to `isInternalParam`, for example look at line 219 in generatePrintable, it would become `if isInternalParam(p) {`. 

It's a pattern that can be useful at times, naming a function "isNOUN", or "canVERB",  etc.

carolynvs:
Oh great! At one point inspect was going to do more, and we changed it's scope. I wasn't sure if it was affected or not so thanks for checking.

carolynvs:
I am looking into why the integration tests are failing, it does not appear to be related to your changes.

carolynvs:
Your change is fine, verified manually locally. I believe that this is a problem with how the CI action is merging your change with master. I'm going to merge and then see how it lands on main. If there is a problem, I'll resolve any conflicts.

PR #1102, opened 2020-06-29
Rename instances to installations

### LINKED ISSUE #1070: Rename instances to installations?
Here's some food for thought. We never liked using the term "claim" as  a user-facing noun in the CLI so we came up with "instance" to refer to an instance of a bundle, e.g. `porter instance list`.

After reworking the claim spec, it is clear that we could have used a different term from the spec, "installation". That was the speccy term referencing the same thing. Claim never meant what we thought it did, but we got cris-crossed due to how it was implemented in duffle/cnab-go (since claim was synonymous with an installation since the implementation didn't keep around multiple claims per installation).

PHEW! 😅 

Now that we have that figured out... Do we prefer the term "installation" to "instance"?

* It is more clear what we are referring to
* It aligns with the spec
* If we want to change it, we should do it before 1.0

Please vote with 👍 👎 or leave a comment.

gaurimadhok:
# What does this change
This modifies bundle instances to installations.

# What issue does it fix
Closes #1070 

# Notes for the reviewer

# Checklist
- [ ] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)


carolynvs:
I have changed the target branch to our next release (v0.28) so that we can group it together with my claims feature and they can all ship at the same time.

pkg/porter/show.go
```diff
@@ -31,9 +31,9 @@ func (so *ShowOptions) Validate(args []string, cxt *context.Context) error {
 	return so.ParseFormat()
 }
 
-// ShowInstances shows a bundle, or more properly a bundle claim, along with any
+// ShowInstallations shows a bundle, or more properly a bundle claim, along with any
```
vdice, line 34: Should we update to `or more properly a bundle installation`?  Not sure how much we're wanting to update legacy instances of the term `claim`, which certainly has meaning in the CNAB spec(s) but perhaps is mostly intended to be replaced by `installation` in Porter.  Granted, this is a function comment/description, so quite minor.

cmd/porter/bundle.go
```diff
@@ -92,11 +92,11 @@ The lint command is run automatically when you build a bundle. The command is av
 func buildBundleInstallCommand(p *porter.Porter) *cobra.Command {
 	opts := porter.InstallOptions{}
 	cmd := &cobra.Command{
-		Use:   "install [INSTANCE]",
-		Short: "Install a new instance of a bundle",
-		Long: `Install a new instance of a bundle.
+		Use:   "install [NAME]",
```
carolynvs, line 95: Would you change this one back to `[INSTALLATION]`? I made a mistake when I told you NAME on our chat today.

cmd/porter/installations.go
```diff
@@ -0,0 +1,80 @@
...
+import (
+	"get.porter.sh/porter/pkg/porter"
+	"github.com/spf13/cobra"
+)
+
+func buildInstallationCommands(p *porter.Porter) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "installations",
+		Aliases: []string{"inst", "installation"},
+		Short:   "Bundle Installation commands",
```
carolynvs, line 12: Let's go with `"Installation commands"` here.

docs/content/cli/_index.md
```diff
@@ -1,4 +0,0 @@
----
```
carolynvs, line 1: Let's undelete this file. My mistake! I now realize why the makefile target doesn't just delete that directory. This file helps us generate this file

https://porter.sh/cli/

which gives us a nice page with all of our commands.

docs/content/credentials.md
```diff
@@ -32,8 +32,8 @@ like `{{ bundle.credentials.github_token }}` with the actual credential value
 before executing the step.
 
 Once the bundle finishes executing, the credentials are NOT recorded in the
-bundle instance (claim). Parameters are recorded there so that you can view them
-later using `porter instances show NAME --output json`.
+installation (claim). Parameters are recorded there so that you can view them
```
carolynvs, line 35: This isn't strictly part of your change but it would be helpful to fix it here. 

It has to do with the larger features (rewriting claims), and the sentence will make more sense this way. Previously we stored a single file (called a claim) to represent an installation, but now we store many files to keep track of the history of what has happened to an installation. So we want to clarify that we won't store credentials in that history.

Would you please change this part?

```suggestion
installation history. Parameters are recorded there so that you can view them
```

pkg/cnab/provider/action.go
```diff
@@ -13,7 +13,7 @@ import (
 
 // Shared arguments for all CNAB actions
 type ActionArguments struct {
-	// Name of the instance.
+	// Name of the installation.
 	Claim string
```
carolynvs, line 17: While we are here, would you also change this field `Claim` to `Installation`? See my previous comment for why it was called claim, but it represented the same thing as "instance" and should be changed in this case.

pkg/porter/show.go
carolynvs: I think we should limit the scope of this PR to just updating the function names, fields and user facing text. Let's leave the explanatory comments to the maintainers to update because I think the changes will be nuanced and having someone else update them will be a game of telephone.

So don't worry too much about finding the word claim, I'll add a follow-up issue to do another pass as part of my claim milestone for the v0.28 release.

carolynvs: https://github.com/deislabs/porter/issues/1105

PR #1103, opened 2020-06-29
feat(*): support dependency params in Parameter Set

vdice:
# What does this change
* Adds support for specifying dependency parameters in a Parameter Set

# What issue does it fix
Closes https://github.com/deislabs/porter/issues/1096

# Notes for the reviewer

# Checklist
- [x] Unit Tests
- [x] Documentation
- [x] Schema (porter.yaml) (N/A)

### DIFF
--- a/cmd/porter/bundle.go
+++ b/cmd/porter/bundle.go
@@ -108,7 +108,7 @@ For example, the 'debug' driver may be specified, which simply logs the info giv
   porter bundle install MyAppFromTag --tag getporter/kubernetes:v0.1.0
 `,
 		PreRunE: func(cmd *cobra.Command, args []string) error {
-			return opts.Validate(args, p.CNAB, p.Context)
+			return opts.Validate(args, p)
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return p.InstallBundle(opts)
@@ -158,7 +158,7 @@ For example, the 'debug' driver may be specified, which simply logs the info giv
   porter bundle upgrade MyAppFromTag --tag getporter/kubernetes:v0.1.0
 `,
 		PreRunE: func(cmd *cobra.Command, args []string) error {
-			return opts.Validate(args, p.CNAB, p.Context)
+			return opts.Validate(args, p)
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return p.UpgradeBundle(opts)
@@ -209,7 +209,7 @@ For example, the 'debug' driver may be specified, which simply logs the info giv
   porter bundle invoke --action ACTION MyAppFromTag --tag getporter/kubernetes:v0.1.0
 `,
 		PreRunE: func(cmd *cobra.Command, args []string) error {
-			return opts.Validate(args, p.CNAB, p.Context)
+			return opts.Validate(args, p)
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return p.InvokeBundle(opts)
@@ -263,7 +263,7 @@ For example, the 'debug' driver may be specified, which simply logs the info giv
 
 `,
 		PreRunE: func(cmd *cobra.Command, args []string) error {
-			return opts.Validate(args, p.CNAB, p.Context)
+			return opts.Validate(args, p)
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return p.UninstallBundle(opts)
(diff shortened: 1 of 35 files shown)


jeffersonbenson:
LGTM

cmd/porter/bundle.go
```diff
@@ -108,7 +108,7 @@ For example, the 'debug' driver may be specified, which simply logs the info giv
   porter bundle install MyAppFromTag --tag getporter/kubernetes:v0.1.0
 `,
 		PreRunE: func(cmd *cobra.Command, args []string) error {
-			return opts.Validate(args, p.Context)
+			return opts.Validate(args, p.CNAB, p.Context)
```
carolynvs, line 111: If you want to simplify, in other places I have passed the Porter app `p` to `Validate` in cases where I need to pass multiple components that porter controls. You don't have to, just letting you know that we've done that before.

pkg/cnab/provider/install.go
```diff
@@ -35,7 +35,7 @@ func (d *Runtime) Install(args ActionArguments) error {
 	}
 	d.Extensions = exts
 
-	params, err := d.loadParameters(c, args.Params, args.ParameterSets, string(manifest.ActionInstall))
+	params, err := d.loadParameters(c, args.Params, string(manifest.ActionInstall))
```
carolynvs, line 38: Oh nice, this is much cleaner to have already resolved all parameters by the time we get to the runtime here.

pkg/cnab/provider/parameters.go
```diff
@@ -58,8 +46,8 @@ func (d *Runtime) loadParameters(claim *claim.Claim, rawOverrides map[string]str
 	return bundle.ValuesOrDefaults(overrides, bun, action)
 }
 
-// loadParameterSets loads parameter values per their parameter set strategies
-func (d *Runtime) loadParameterSets(b *bundle.Bundle, params []string) (valuesource.Set, error) {
+// LoadParameterSets loads parameter values per their parameter set strategies
+func (d *Runtime) LoadParameterSets(params []string) (valuesource.Set, error) {
```
carolynvs, line 50: I think that this function can be moved out of the pkg/cnab/provider and into pkg/porter. Nothing in pkg/cnab/provider package calls this, and none of the code here relies on being located in this package, right? So we could move it up closer to where it's used (the porter client) and not have to have this in the runtime.

pkg/porter/archive.go
```diff
@@ -22,12 +23,12 @@ type ArchiveOptions struct {
...
 
 // Validate performs validation on the publish options
-func (o *ArchiveOptions) Validate(args []string, ctx *context.Context) error {
+func (o *ArchiveOptions) Validate(args []string, runtime cnabprovider.CNABProvider, ctx *context.Context) error {
 	if len(args) < 1 || args[0] == "" {
 		return errors.New("Destination File is required")
 	}
 	o.ArchiveFile = args[0]
-	return o.BundleLifecycleOpts.Validate(args, ctx)
+	return o.BundleLifecycleOpts.Validate(args, runtime, ctx)
```
carolynvs, line 31: Yeah I think seeing passing the runtime to a client side function (Validate) makes it clear that LoadParameterSet really belongs on Porter, or somewhere else in a client package instead of on the runtime.

pkg/porter/cnab.go
```diff
@@ -200,12 +206,19 @@ func (o *bundleFileOptions) validateCNABFile(cxt *context.Context) error {
...
 }
 
-func (o *sharedOptions) validateParams(cxt *context.Context) error {
+func (o *sharedOptions) validateParams(runtime cnabprovider.CNABProvider, cxt *context.Context) error {
 	err := o.parseParams()
 	if err != nil {
 		return err
 	}
 
+	err = o.parseParamSets(runtime, cxt)
```
carolynvs, line 215: haha this code is unflushable! 🚽

pkg/porter/explain.go
```diff
@@ -115,10 +115,16 @@ func (s SortPrintableAction) Swap(i, j int) {
 }
 
 func (o *ExplainOpts) Validate(args []string, cxt *context.Context) error {
-	err := o.sharedOptions.Validate(args, cxt)
+	err := o.validateInstanceName(args)
```
carolynvs, line 118: What's odd here is that explain really shouldn't be using the shared options... That is meant for the actions, like install and upgrade. Thanks for fixing.

tests/rebuild_test.go
```diff
@@ -28,7 +28,7 @@ func TestRebuild_InstallNewBundle(t *testing.T) {
 
 	// Install a bundle without building first
 	installOpts := porter.InstallOptions{}
-	err = installOpts.Validate([]string{}, p.Context)
+	err = installOpts.Validate([]string{}, p.CNAB, p.Context)
```
carolynvs, line 31: Yeah seeing adding this extra arg everywhere, if we change it to passing `p` instead, I think that would help in the future to avoid cascading changes like this. What do you think?

vdice: Good call; simplified to just passing `p`.

pkg/cnab/provider/parameters.go
vdice: Indeed, thank you.  Now moved to `pkg/porter`.  PTAL when convenient.

PR #1113, opened 2020-07-02
feat(pkg/cnab): docker driver support using host network

vdice:
# What does this change
 * Adds option to use host network to docker extension configuration

# What issue does it fix
Related to https://github.com/deislabs/porter/issues/1079.  Lays the groundwork for using Porter with KinD k8s clusters.

# Notes for the reviewer

# Checklist
- [x] Unit Tests
- [x] Documentation

### DIFF
--- a/docs/content/author-bundles.md
+++ b/docs/content/author-bundles.md
@@ -357,6 +357,7 @@ When the bundle is executed, this elevated privilege must be explicitly granted 
 **Configuration:**
 
   * `privileged: BOOLEAN` - OPTIONAL. Whether or not the `--privileged` flag should be set when the bundle's invocation image runs. Defaults to false.
+  * `useHostNetwork: BOOLEAN` - OPTIONAL. Whether or not host networking should be used when the bundle's invocation image runs. Defaults to false.
 
 Example:
 
--- a/docs/content/configuration.md
+++ b/docs/content/configuration.md
@@ -38,9 +38,10 @@ combination of: `table`, `json` and `yaml`.
 `--allow-docker-host-access` controls whether or not the local Docker daemon
 should be made available to executing bundles. This flag is available for the
 following commands: [install], [upgrade], [invoke] and [uninstall]. When this
-value is set to true, bundles are executed in a privileged container with the
-docker socket mounted. This allows you to use Docker from within your bundle,
-such as `docker push`, `docker-compose`, or docker-in-docker.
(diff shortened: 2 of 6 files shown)


carolynvs:
The change looks fine.

However, I have a question about when an author would know that the bundle requires host networking. I'm wondering (but am not really sure) if it's the case that the person deploying would know if host networking is needed, because of the infrastructure on which it's being deployed, but the bundle could work either way.

I am feeling out if it is really a scenario of the bundle author wants to be able to handle it when host networking needs to be turned on by whoever is deploying based on where it's being deployed but the bundle itself isn't really requiring that it must be deployed that way. If that's the case, Porter's current mechanism for allowing bundles to get higher levels of access (driven by the bundle author, then allowed by the porter user) needs to be expanded to support this scenario with less indirection of intent.

I'm guessing this is related to KinD. Can you briefly explain when the bundle author knows that their bundle must have host networking?

vdice:
>  I'm wondering (but am not really sure) if it's the case that the person deploying would know if host networking is needed, because of the infrastructure on which it's being deployed, but the bundle could work either way.

That's exactly it.  This is a new case wherein the bundle installer may want to toggle a setting (say, enabling use of host networking) depending on underlying infrastructure -- it needn't/shouldn't be a hard-coded setting added to the bundle by the author.

How do we want to capture the work to enable install-time configuration?  For this feature specifically, we could update the [driver code](https://github.com/deislabs/porter/blob/main/pkg/cnab/provider/driver.go#L23) to enable host networking on the presence of a command (`install`, etc.) flag alone (and not require that the bundle have a `docker` extension required).  The flag could be the usual `--allow-docker-host-access` or a new one.

And/or, we might want a proposal around the general use case of allowing bundle installers to enable install-time configuration on a driver (mostly Docker at this time).  Besides enabling the host networking option on a Docker run command, I'd predict there might be use cases for mounting volumes, supplying env var values, etc. as well.

carolynvs:
At the maintainer call, we decided to timebox looking into if we can trick kind into making a usable kubeconfig. If not we will add a client side flag that gives host networking access (using the hierarchy that other flags have https://porter.sh/configuration/). Maybe something like `--use-host-network`?

vdice:
As @carolynvs mentioned, I believe we're going the route of configuring kind to enable Porter use (see https://github.com/deislabs/porter/pull/1138) rather than to add this host networking feature.  We can keep the branch around if we want to revisit this in the future.

PR #1116, opened 2020-07-06
Set PORTER_HOME in integration tests on plugins

carolynvs:
# What does this change
The integration tests don't use the PORTER_HOME env var, they programmatically set the porter home directory on the porter app. So when Porter creates a plugin, we need to explicitly set the home dir on the plugin when it's instantiated since it won't get PORTER_HOME automatically.

# What issue does it fix
Running integration tests and having it act against your real porter install. 😱 

# Notes for the reviewer
I have heartburn right now. 🚒 

# Checklist
- [ ] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/pkg/plugins/pluggable/loader.go
+++ b/pkg/plugins/pluggable/loader.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"os"
 	"os/exec"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


carolynvs:
I am seeing this a few times lately with our PR CI build. These specific tests will fail. But they pass locally and it's not clear at all why they are failing considering the sha's that were merged, it should contain the correct code to make the tests pass. 

In another PR, I merged the PR with the test failing on a hunch, and then it passed on main. I have this feeling that whatever code it is running in the CI isn't quite the right code, you know? But I bet I am missing something...

carolynvs:
That is so weird! I kicked the build and this time it passed? Maybe that test is a flake? 🤔 I'll keep an eye on it.

PR #1125, opened 2020-07-07
Add --version flag to porter

carolynvs:
# What does this change
People never know if a CLI supports --version or a version command. So let's just support both so whatever they try will work. MAGIC!

# What issue does it fix
Fighting muscle memory

# Notes for the reviewer
Inspired by https://twitter.com/michellenoorali/status/1280517451550199808

# Checklist
- [x] Unit Tests
- [x] Documentation
- [x] Schema (porter.yaml) - NA

### DIFF
--- a/cmd/porter/version_test.go
+++ b/cmd/porter/version_test.go
@@ -29,7 +29,7 @@ func TestVersion(t *testing.T) {
 		assert.Contains(t, out.String(), "porter v1.0.0 (abc123)")
 	})
 
-	t.Run("command", func(t *testing.T) {
+	t.Run("flag", func(t *testing.T) {
 		p := buildRootCommand()
 
 		// Capture output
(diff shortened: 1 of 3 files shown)


carolynvs:
OMG I looooove testing cobra in unit tests. I just realize that you can set `os.Args` and was able to make the unit test so much easier to work with. 💯

cmd/porter/version_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+		// Set the command to run
+		os.Args = []string{"porter", "version"}
+
+		err := p.Execute()
+		require.NoError(t, err)
+		assert.Contains(t, out.String(), "porter v1.0.0 (abc123)")
+	})
+
+	t.Run("command", func(t *testing.T) {
```
vdice, line 32: nit: should we update this test name to differ from the other above?

PR #1127, opened 2020-07-10
feat(publish.go): update invocation image per bundle tag on publish

vdice:
# What does this change

* Updates `porter publish` logic such that the invocation image name is updated when a bundle tag value (via `--tag`) is provided.  The updated image name inherits the registry and docker tag from the provided bundle tag value.
* A bit of refactoring was introduced to consolidate this image naming logic for use by multiple areas (in this case, publish as well as setting defaults on the manifest)

# What issue does it fix
Closes https://github.com/deislabs/porter/issues/1123

# Notes for the reviewer

# Checklist
- [x] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/pkg/manifest/manifest.go
+++ b/pkg/manifest/manifest.go
@@ -543,7 +543,8 @@ func (m *Manifest) SetDefaults() error {
 
 // SetInvocationImageFromBundleTag sets the invocation image name on the manifest
 // per the provided bundle tag, setting/overriding the original image name if
-// empty or if overrideImage is true
+// empty or if overrideImage is true and updating the manifest BundleTag value
+// if it initially lacks a Docker tag
 func (m *Manifest) SetInvocationImageFromBundleTag(bundleTag string, overrideImage bool) error {
 	bundleRef, err := reference.ParseNormalizedNamed(bundleTag)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/manifest/manifest.go
```diff
@@ -590,12 +581,41 @@ func (m *Manifest) SetDefaults() error {
 			}
 			m.Image = reference.FamiliarString(imageRef)
 		case reference.Digested:
-			return errors.New("invalid bundle tag format %q, must be an OCI image tag")
+			return errors.New("invalid bundle tag format, must be an OCI image tag")
```
carolynvs, line 584: oops! 😊

pkg/manifest/manifest.go
```diff
@@ -590,12 +581,41 @@ func (m *Manifest) SetDefaults() error {
...
 		}
 	}
+
 	return nil
 }
 
+// getDockerTagFromBundleRef returns the Docker tag portion of the bundle tag,
+// updating the manifest BundleTag value if it initially lacks a Docker tag
```
carolynvs, line 592: I feel like this function is doing too much now that it is called `get*`. I wouldn't expect a function called get to also do an update.

It would make be preferable from a separation of responsibility standpoint if it just get the tag, and then have `SetInvocationImageFromBundleTag` continue to manage setting `m.Tag` like it did before.

carolynvs:
Originally we let people set the invocation image, but now I don't think it makes sense anymore. What would you think about keeping all of our logic, but removing the field from the yaml for invocation image so that people can't override it anymore. Essentially it would only be defaulted going forward into v1.0?

pkg/manifest/manifest.go
vdice: Good call; will update such that setting the manifest BundleTag value occurs in `SetInvocationImageFromBundleTag`.

vdice:
> Originally we let people set the invocation image, but now I don't think it makes sense anymore.

I like this a lot.  We'd definitely be able to remove a good chunk of code if we remove the option to set the invocation image name from the manifest.  What do you think about filing an issue for a follow-up after this is merged?  Or would we rather knock it out in this PR?

carolynvs:
Definitely, let's do it in a follow-up issue+PR. I think what you have here is very close. Then we can go in to remove the edge cases around overrides of the invocation image separately.

vdice:
Great!  Issue here: https://github.com/deislabs/porter/issues/1132

Commit addressing other item of feedback here: https://github.com/deislabs/porter/pull/1127/commits/02b6b46c0a449ed7cc849868de2e60b54f57d18e

PR #1128, opened 2020-07-10
fix(publish.go): fix certain cases around image name updating

vdice:
# What does this change

When we go to update an invocation image based on a bundle tag, there were some cases when the resulting image had an incorrect or unexpected name.  This logic currently is used by `archive` but may also come in handy for `publish` if/when https://github.com/deislabs/porter/pull/1127 is merged.  As a bonus, this area of code has been simplified while implementing the fix.

One obvious case of brokenness was when the invocation image had a registry but no org:

```
	t.Run("has registry, no org", func(t *testing.T) {
		newInvImgName, err := getNewImageNameFromBundleTag("localhost:5000/apache-installer", "example.com/neworg/apache:v0.1.0")
		require.NoError(t, err, "getNewImageNameFromBundleTag failed")
		assert.Equal(t, "example.com/neworg/apache-installer", newInvImgName.String())
	})
```
Yielded:
```
    --- FAIL: TestPublish_getNewImageNameFromBundleTag/has_registry,_no_org (0.00s)
        publish_test.go:88:
            	Error Trace:	publish_test.go:88
            	Error:      	Not equal:
            	            	expected: "example.com/neworg/apache-installer"
            	            	actual  : "docker.io/neworg/example.com/apache-installer"
```

# What issue does it fix
(No corresponding GH issue.)

# Notes for the reviewer

Note the added unit tests around the corner cases where either/both of the invocation image and bundle tag have subdomains/subdirectories in their names.  This PR currently takes the position of inheriting the subdomain from the bundle tag, but I'm not sure if this is what is desired.

# Checklist
- [x] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/pkg/porter/publish.go
+++ b/pkg/porter/publish.go
@@ -276,40 +276,26 @@ func (p *Porter) updateBundleWithNewImage(bun bundle.Bundle, newImg image.Name, 
 // getNewImageNameFromBundleTag derives a new image.Name object from the provided original
 // image (string) using the provided bundleTag to glean registry/org/etc.
 func getNewImageNameFromBundleTag(origImg, bundleTag string) (image.Name, error) {
-	origImgName, err := image.NewName(origImg)
+	origName, err := image.NewName(origImg)
 	if err != nil {
 		return image.EmptyName, errors.Wrapf(err, "unable to parse image %q into domain/path components", origImg)
 	}
 
-	bundleTagName, err := image.NewName(bundleTag)
+	bundleName, err := image.NewName(bundleTag)
 	if err != nil {
 		return image.EmptyName, errors.Wrapf(err, "unable to parse bundle tag %q into domain/path components", bundleTag)
 	}
+	bundleHost := bundleName.Host() // e.g. docker.io
 
-	// Swap out Host
-	origHost := origImgName.Host()
-	newHost := bundleTagName.Host()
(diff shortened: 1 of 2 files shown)


pkg/porter/publish.go
```diff
@@ -276,40 +276,26 @@ func (p *Porter) updateBundleWithNewImage(bun bundle.Bundle, newImg image.Name,
...
-	} else {
-		newImgName, err := image.NewName(newImg)
-		if err != nil {
-			return image.EmptyName, errors.Wrapf(err, "unable to parse image %q into domain/path components", newImg)
-		}
-
-		newImg = path.Join(newImgName.Host(), strings.Replace(newImgName.Path(), origPathParts[0], newOrg, 1))
-	}
+	// Join for bundleHost/bundleOrg/bundleSubdir/origImgName
+	newImg := path.Join(bundleHost, bundleOrg, origImgName)
```
carolynvs, line 298: When I was looking at your simplification I saw that we could take this even further (replacing lines 288-298):

```suggestion
// Use the image name with the bundle location
newImg := path.Join(path.Dir(bundleName.Name()), path.Base(origName.Name()))
```

vdice: 🙏  So good!  Thanks for the even better simplification.  Just pushed a commit with this change (GH doesn't appear to support committing from suggestion if on deleted lines...)

PR #1129, opened 2020-07-10
Param source

### LINKED ISSUE #1066: Implement Parameter Sources extension
Now that outputs are no longer stored in claims, we need to implement the parameter sources extension from the spec to get outputs from a dependency into the bundle runtime.

https://github.com/cnabio/cnab-spec/blob/master/810-well-known-custom-extensions.md#parameter-sources

This should read the parameter sources from the bundle and load outputs from storage into the bundle runtime.

carolynvs:
# What does this change
Support the parameter source extension at runtime. See https://github.com/cnabio/cnab-spec/blob/master/810-well-known-custom-extensions.md#parameter-sources for the CNAB spec definition of how that extension works.

I decided that if we can't resolve the requested output, perhaps because the output wasn't generated, or the action it relies upon hasn't been run yet, we skip over it and let porter's general parameter resolution kick in. So if there's a default, that's applied, otherwise the user is told it's a required parameter that they need to set the value for.

# What issue does it fix
Closes #1066 

# Notes for the reviewer
* This doesn't support generating bundles with parameter-source extension data, just handles running bundles with it defined. That part comes next.
* I did some refactoring to prep for this change, so I recommend reviewing this commit by commit.

# Checklist
- [x] Unit Tests
- [x] Documentation - N/A

### DIFF
--- a/pkg/cnab/extensions/parameter_sources.go
+++ b/pkg/cnab/extensions/parameter_sources.go
@@ -8,11 +8,12 @@ import (
 )
 
 const (
-	// ParameterSourcesExtensionKey represents the full key for the Parameter Sources Extension
+	// ParameterSourcesExtensionKey represents the full key for the Parameter Sources Extension.
 	ParameterSourcesExtensionKey = "io.cnab.parameter-sources"
-	// DockerExtensionSchema represents the schema for the Docker Extension
-	ParameterSourcesExtensionSchema = "https://cnab.io/specs/v1/parameter-sources.schema.json"
-	ParameterSourceTypeOutput       = "output"
+	// ParameterSourcesExtensionSchema represents the schema for the Docker Extension.
+	ParameterSourcesExtensionSchema = "https://cnab.io/v1/parameter-sources.schema.json"
+	// ParameterSourceTypeOutput defines a type of parameter source that is provided by a bundle output.
+	ParameterSourceTypeOutput = "output"
 )
 
 // ParameterSourcesExtension represents a required extension that specifies how
@@ -44,7 +45,7 @@ func (ps *ParameterSources) SetParameterFromOutput(parameter string, output stri
 }
 
 type ParameterSource struct {
-	// Array of source types in the priority order that they should be used to
+	// Priority is an array of source types in the priority order that they should be used to
 	// populated the parameter.
 	Priority []string `json:"priority" mapstructure:"priority"`
 
@@ -139,7 +140,7 @@ func ParameterSourcesExtensionReader(bun bundle.Bundle) (interface{}, error) {
 	return ps, nil
 }
 
-// GetParameterSourcesExtension checks if the docker extension is present and returns its
+// GetParameterSourcesExtension checks if the parameter sources extension is present and returns its
 // extension configuration.
 func (e ProcessedExtensions) GetParameterSourcesExtension() (ParameterSources, bool, error) {
 	rawExt, required := e[ParameterSourcesExtensionKey]
--- a/pkg/cnab/extensions/parameter_sources.go
+++ b/pkg/cnab/extensions/parameter_sources.go
@@ -1,0 +1,153 @@
+package extensions
+
+import (
+	"encoding/json"
+
+	"github.com/cnabio/cnab-go/bundle"
+	"github.com/pkg/errors"
+)
+
+const (
+	// ParameterSourcesExtensionKey represents the full key for the Parameter Sources Extension
+	ParameterSourcesExtensionKey = "io.cnab.parameter-sources"
+	// DockerExtensionSchema represents the schema for the Docker Extension
+	ParameterSourcesExtensionSchema = "https://cnab.io/specs/v1/parameter-sources.schema.json"
+	ParameterSourceTypeOutput       = "output"
+)
+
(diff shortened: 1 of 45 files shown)


pkg/cnab/extensions/parameter_sources.go
```diff
@@ -0,0 +1,153 @@
...
+	err = json.Unmarshal(dataB, &ps)
+	if err != nil {
+		return nil, errors.Wrapf(err, "could not unmarshal the %q extension %q",
+			ParameterSourcesExtensionKey, string(dataB))
+	}
+
+	return ps, nil
+}
+
+// GetParameterSourcesExtension checks if the docker extension is present and returns its
```
vdice, line 142: typo: `docker extension`

pkg/cnab/extensions/paramter_sources_test.go
```diff
@@ -0,0 +1,43 @@
...
+		assert.Empty(t, ext, "parameter-sources should default to empty when not required")
+	})
+
+	t.Run("extension invalid", func(t *testing.T) {
+		processed := ProcessedExtensions{
+			ParameterSourcesExtensionKey: map[string]string{"ponies": "are great"},
+		}
+
+		ext, required, err := processed.GetParameterSourcesExtension()
+		require.Error(t, err, "GetParameterSourcesExtension failed")
```
vdice, line 39: nit: should the test comment be updated to something like: `GetParameterSourcesExtension succeeded`?

pkg/cnab/extensions/paramter_sources_test.go
```diff
@@ -0,0 +1,43 @@
...
+		assert.True(t, required, "parameter-sources should be a required extension")
```
vdice, line 40: These two assertions are interesting... even though the extension is invalid, it is still required, but empty?

pkg/cnab/extensions/required_test.go
```diff
@@ -34,45 +29,45 @@ func TestProcessRequiredExtensions(t *testing.T) {
...
 	})
 }
 
 func TestGetSupportedExtension(t *testing.T) {
-	t.Run("supported via shorthand", func(t *testing.T) {
-		ext, err := GetSupportedExtension("docker")
-		require.NoError(t, err)
-		require.Equal(t, DockerExtension.Key, ext.Key)
-	})
+	for _, supported := range SupportedExtensions {
```
vdice, line 58: Nice!

pkg/cnab/provider/parameters_test.go
```diff
@@ -179,16 +191,113 @@ func Test_loadParameters_fileParameter(t *testing.T) {
...
+		}
+		params, err := r.loadParameters(b, args)
+		require.NoError(t, err)
+		assert.Equal(t, "foo_override", params["foo"],
+			"expected param 'foo' to have parameter override value")
+	})
+
+	t.Run("merge parameter values", func(t *testing.T) {
+		// foo is set by a the user
+		// baz is set by a parameter source
```
vdice, line 273: nit: these appear to be swapped.  `baz` looks to be set by the default below, while `bar` is set by the parameter source.

pkg/cnab/extensions/parameter_sources.go
```diff
@@ -0,0 +1,153 @@
...
+
+	"github.com/cnabio/cnab-go/bundle"
+	"github.com/pkg/errors"
+)
+
+const (
+	// ParameterSourcesExtensionKey represents the full key for the Parameter Sources Extension
+	ParameterSourcesExtensionKey = "io.cnab.parameter-sources"
+	// DockerExtensionSchema represents the schema for the Docker Extension
+	ParameterSourcesExtensionSchema = "https://cnab.io/specs/v1/parameter-sources.schema.json"
```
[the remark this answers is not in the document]
carolynvs: Thanks, I'll fix the URL for dependencies too then because that's the form that it uses too.

pkg/cnab/extensions/parameter_sources.go
```diff
@@ -0,0 +1,153 @@
...
+	(*ps)[parameter] = ParameterSource{
+		Priority: []string{ParameterSourceTypeOutput},
+		Sources: ParameterSourceMap{
+			ParameterSourceTypeOutput: OutputParameterSource{OutputName: output},
+		},
+	}
+}
+
+type ParameterSource struct {
+	// Array of source types in the priority order that they should be used to
```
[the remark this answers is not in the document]
carolynvs: Ah yeah I was copying from the spec here. Will fix

pkg/cnab/extensions/paramter_sources_test.go
carolynvs: Will fix in the original test that I copied from too.

pkg/cnab/extensions/paramter_sources_test.go
carolynvs: The key was present but the associated data was malformed and could not be marshaled. So it's required because the key was there. But it's invalid data at the same time so Porter can't handle the bundle.

carolynvs:
I'll check if release/v0.28.0 needs to be merged with main again.

PR #1140, opened 2020-07-16
ref(*): remove invocation image from the Porter manifest

vdice:
# What does this change
* Removes ability to specify an invocation image name in the Porter manifest

# What issue does it fix
Closes https://github.com/deislabs/porter/issues/1132

# Notes for the reviewer

A couple questions:
 - The current approach explicitly errors out if the `invocationImage` is present in the manifest.  What do we think?
 - Did I visit all of the known places where logic can be simplified now that this value can't be user-specified?

# Checklist
- [x] Unit Tests
- [x] Documentation
- [x] Schema (porter.yaml)

### DIFF
--- a/pkg/manifest/manifest.go
+++ b/pkg/manifest/manifest.go
@@ -482,7 +482,7 @@ func (s *Step) GetMixinName() string {
 	return mixinName
 }
 
-func UnmarshalManifest(manifestData []byte) (*Manifest, error) {
+func UnmarshalManifest(cxt *context.Context, manifestData []byte) (*Manifest, error) {
 	// Unmarshal the manifest into the normal struct
 	manifest := &Manifest{}
 	err := yaml.Unmarshal(manifestData, &manifest)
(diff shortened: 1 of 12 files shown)


pkg/manifest/manifest.go
```diff
@@ -514,6 +515,10 @@ func UnmarshalManifest(manifestData []byte) (*Manifest, error) {
 		if _, found := knownFields[key]; found {
 			delete(unmappedData, key)
 		}
+		// Explicitly error out if deprecated invocationImage is supplied
+		if key == "invocationImage" {
```
carolynvs, line 519: I think this would be a rough transition for some people who have older manifests that set the invocationImage. What do you think about ignoring it and printing this message as a warning instead? We'd have to pass in the porter context to this function.

Otherwise someone updating to this release would suddenly have porter build completely stop working. Then perhaps in a later release we can make it a hard error, perhaps in 1.0 we can hard fail on unknown keys.

What do you think?

vdice: Agreed, I'll update to print a warning instead.

vdice: Updated in https://github.com/deislabs/porter/pull/1140/commits/abb969a4c9e0feb196e8b0682f9bc7f3e69cd6f4 - let me know how it looks when convenient.

vdice:
Manifest schema updated!  Thank you for the reminder 😅

PR #1144, opened 2020-07-16
Added suffix arguments to mixin executableStep

gaurimadhok:
# What does this change
This change adds an option for mixin commands to put arguments at the end of the command in addition to the current functionality of adding arguments at the beginning.

# What issue does it fix
This change was needed for the docker mixin in order to have some arguments come at the end of the command such as the path for docker build which comes after all the optional flags. 
- docker build -t tag -f file path ARGUMENTS --FLAGS

### DIFF
--- a/pkg/exec/builder/execute.go
+++ b/pkg/exec/builder/execute.go
@@ -28,8 +28,13 @@ type ExecutableAction interface {
 
 type ExecutableStep interface {
 	GetCommand() string
+	//GetArguments() puts the arguments at the beginning of the command
 	GetArguments() []string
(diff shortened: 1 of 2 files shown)


carolynvs:
It needs a follow-up issue to support suffix args in the exec mixin. This is the first step to support it in the builder, which Gauri is going to immediately use in her Docker mixin.

nunix:
Thank you so so much ❤️ 
Sorry for the delayed response, was in holidays but this week is gonna be Porter-crazy for the Corsair 😄

PR #1156, opened 2020-07-22
updates Makefile

simongdavies:
# What does this change

Updates makefile so make fails with an error if packr2 is not installed and  ${GOPATH}/bin is not in path


build/azure-pipelines.setup-go-workspace.sh
```diff
@@ -5,4 +5,4 @@ set -xeuo pipefail
 # Create GOPATH/bin and add it to our PATH so that
 # installed go binaries are available
 mkdir -p /home/vsts/go/bin/
-echo "##vso[task.prependpath]/home/vsts/go/bin/"
+echo "##vso[task.prependpath]/home/vsts/go/bin/"
```
carolynvs, line 8: Is this file change significant?

simongdavies: No seems like I removed the trailing new line, I can revert it?

carolynvs: No that's fine, I just wasn't sure if we needed to make a similar change for other builds. 😀

PR #1165, opened 2020-07-24
feat(.github): add proposal template

vdice:
# What does this change
* Adds an issue template for proposals

# What issue does it fix
Closes https://github.com/deislabs/porter/issues/1161

# Notes for the reviewer
Definitely appreciate feedback on the question prompts, wording, etc.... tried to find a balance between a suggested outline and something very specific/prescriptive (trying to avoid the latter).

# Checklist
- [ ] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)


carolynvs:
Do you think that the feature request and proposal should be merged? If not how do we clarify which to use when? Some of the questions in the feature request are useful, for example getting them to articulate what problem they are having that prompted the issue, alternatives attempted.

vdice:
> Do you think that the feature request and proposal should be merged? If not how do we clarify which to use when? Some of the questions in the feature request are useful, for example getting them to articulate what problem they are having that prompted the issue, alternatives attempted.

Mmm, good point.  Let me see what that would look like.

vdice:
After giving it some thought, I went the other way: I attempted to refocus the new proposal template to only cover design proposals.  It seemed to me that it was different enough in scope/intention from most feature requests to warrant a separate template.  If we end up feeling strongly about merging feature request with proposal into one template, I can revisit.

.github/ISSUE_TEMPLATE/design_proposal.md
```diff
@@ -0,0 +1,25 @@
...
+**Risks/Concerns**
+What are some potential concerns with the proposal? (For example: heavy development cost, extensive refactoring, high complexity.)
+
+**What other approaches were considered?**
+Provide other approaches considered prior to deciding on the submitted proposal.
+
+**Implementation details**
+If applicable, provide potential implementation details assuming the proposal is accepted.
+
+_Note that proposals are expected to remain open for at least one week after being created, to allow time for community feedback._
```
carolynvs, line 25: Should we mention here that proposals should be posted to the mailing list? I think once we have that this is good to go. 👍

vdice: Oh good call.  Switched this section up to a checklist w/ the addition of the mailing list note.  WDYT?

Issue #969, opened 2020-04-02
Random Execution of the dependencies at Runtime


MChorfa:
## Describe the Issue
Since the nested dependencies are not supported yet. So we decided to create multiple dependent bundles, and we want the execute bundle in a certain orders. But on each run we get a random elections at Runtime.

Here what we get:

##### Declared

MYDEP-core
MYDEP-s3
MYDEP-db
MYDEP-experiments

---

##### Executed

MYDEP-experiments
MYDEP-s3
MYDEP-core
MYDEP-db

## To Reproduce
1. Create bundle with multiple dependencies
2. install 
4. Observe the declaration and the execution plan

## Expected behavior
Get the dependent bundle to execute for now in the exact same way they were declared

## Version
porter v0.23.0-beta.1 (9c49056)

MChorfa:
After a quick scanning of the code, this issue will require changes all the up to CNAB dependencies spec. 
By changing from map `Dependencies map[string]Dependency  `yaml:"dependencies,omitempty"`` to to a list `Dependencies []Dependency  `yaml:"dependencies,omitempty"` will require a lot code changes. The other way would be to add an extra property like `order`  within the dependency object b but I wonder if this won't impact the spec?

MChorfa:
Hello @carolynvs, @vdice,
Potential solution: 
1. Add Prop: Dependency struct [dependencies.go] 
```go
// Dependency describes a dependency on another bundle
type Dependency struct {
	// Bundle is the location of the bundle in a registry, for example REGISTRY/NAME:TAG
	Bundle string `json:"bundle" mapstructure:"bundle"`

	// Version is a set of allowed versions
	Version *DependencyVersion `json:"version,omitempty" mapstructure:"version"`

	// Order is a way to indicate precedence within the dependencies
	Order int `json:"order,omitempty" mapstructure:"order"`
}
```
2. Add Prop: DependencyLock struct [solver.go] 
```go
type DependencyLock struct {
	Alias string
	Tag   string
	Order int
}
```
3. Sort the DependencyLocks within the ResolveDependencies func
```go
...
	q := make([]DependencyLock, 0, len(deps.Requires))
	for alias, dep := range deps.Requires {
		bundle := strings.Split(dep.Bundle, ":")[0]
		version, err := s.ResolveVersion(alias, dep)
		if err != nil {
			return nil, err
		}
		lock := DependencyLock{
			Alias: alias,
			Tag:   fmt.Sprintf("%s:%s", bundle, version),
[...]

PR #1170, opened 2020-07-26
User Defined Order#1147

### LINKED ISSUE #1147: Execute dependencies in a user-defined order
See https://github.com/cnabio/cnab-spec/issues/361

🚨 This is a breaking format change

MChorfa:
# What does this change
This addresses the issue of random execution dependencies at Runtime. This implementation reflects the proposition into the spec. see https://github.com/cnabio/cnab-spec/issues/361

# What issue does it fix
Closes #1147, #969

# Checklist
- [x] Unit Tests
- [x] Documentation
- [x] Schema (porter.yaml)


carolynvs:
@MChorfa Let me know if you need help with the test failure or if you are still working on that yourself.

MChorfa:
> @MChorfa Let me know if you need help with the test failure or if you are still working on that yourself.

@carolynvs yes, please. I am having an issue with the map Keys. The order is never guaranteed even after reordering with sequence definition. Also, I am worried about the `for loop range` behavior.  The order is also not guaranteed when going through a map.

carolynvs:
Oh, it's the ordering that's causing trouble? The way we fix that is often with a custom data structure that sorts according to our desired strategy. I will take a look later today and may push a commit with a data structure to help out.

carolynvs:
@MChorfa Your PR wasn't set up to allow commits from maintainers so I couldn't push a commit with a fix for your test. Here's the fix https://github.com/carolynvs/porter/commit/430738eb373cf7d78f4a98d7a9f7440816a06838

Wherever you need to read the dependencies in sorted order you would use the `ListBySequence` function, instead of reading by the map directly. In Go reading from a map is guaranteed to result in a different order each time to prevent you from relying upon it, and force you to do something like this instead.

pkg/cnab/extensions/dependencies_test.go
```diff
@@ -21,21 +20,27 @@ func TestReadDependencyProperties(t *testing.T) {
...
-	assert.NotNil(t, dep, "expected Dependencies.Requires to have an entry for 'mysql'")
-	assert.Equal(t, "somecloud/mysql", dep.Bundle, "Dependency.Bundle is incorrect")
-	assert.True(t, dep.Version.AllowPrereleases, "Dependency.Bundle.Version.AllowPrereleases should be true")
-	assert.Equal(t, []string{"5.7.x"}, dep.Version.Ranges, "Dependency.Bundle.Version.Ranges is incorrect")
+	assert.Len(t, deps, 2, "Dependencies.Requires is the wrong length")
+
+	for _, dep := range deps {
+		if dep.Name == "storage" {
+
+			assert.NotNil(t, dep, "expected Dependencies.Requires to have an entry for 'storage'")
```
carolynvs, line 28: Now that this has been changed to ranging over deps, checking for nil here doesn't apply. If it were nil we would have gotten an error on the previous line.

pkg/cnab/extensions/dependencies_test.go
```diff
@@ -21,21 +20,27 @@ func TestReadDependencyProperties(t *testing.T) {
...
+			assert.Equal(t, "somecloud/blob-storage", dep.Bundle, "Dependency.Bundle is incorrect")
+			assert.Nil(t, dep.Version, "Dependency.Version should be nil")
+
+		} else if dep.Name == "mysql" {
+
+			assert.NotNil(t, dep, "expected Dependencies.Requires to have an entry for 'mysql'")
```
carolynvs, line 34: Same here, no need to keep this nil check.

pkg/manifest/manifest.go
```diff
@@ -370,19 +370,19 @@ func (mi *MappedImage) Validate() error {
...
+	Name             string            `yaml:"name"`
 	Tag              string            `yaml:"tag"`
 	Versions         []string          `yaml:"versions"`
 	AllowPrereleases bool              `yaml:"prereleases"`
 	Parameters       map[string]string `yaml:"parameters,omitempty"`
 }
 
 func (d *Dependency) Validate() error {
+	if d.Name == "" {
+		return errors.New("dependency tag is required")
```
carolynvs, line 383: Let's update the error message to match what is being checked.

tests/dependencies_test.go
```diff
@@ -99,10 +99,19 @@ func installWordpressBundle(p *porter.TestPorter) (namespace string) {
 	return namespace
 }
 
-func cleanupWordpressBundle(p *porter.TestPorter) {
+func cleanupWordpressBundle(p *porter.TestPorter, targetDependencyName string) {
```
carolynvs, line 102: It's not clear why changing this test cleanup function was necessary? What problem were you running into that made you change it so I can help?

tests/dependencies_test.go
```diff
@@ -99,10 +99,19 @@ func installWordpressBundle(p *porter.TestPorter) (namespace string) {
...
+
+	// Question: Do need we handle the case of uninstalling all the Dependencies
```
carolynvs, line 104: This function explicitly cleans up the wordpress bundle in case the test failed, so that we don't leave mysql and/or wordpress installed on your dev machine. Porter handles uninstalling dependencies. The test does call uninstall on wordpress before this function. This is a failsafe in case something bad happens and will attempt to remove test data before the test completes.

So no changes should be necessary here.

MChorfa:
Thank you @carolynvs for the review. Still need to write more units tests and update the documentation

pkg/cnab/extensions/dependencies_test.go
```diff
@@ -20,16 +20,63 @@ func TestReadDependencyProperties(t *testing.T) {
...
-	assert.Nil(t, dep.Version, "Dependency.Version should be nil")
-
-	dep = deps.Requires["mysql"]
-	assert.NotNil(t, dep, "expected Dependencies.Requires to have an entry for 'mysql'")
-	assert.Equal(t, "somecloud/mysql", dep.Bundle, "Dependency.Bundle is incorrect")
-	assert.True(t, dep.Version.AllowPrereleases, "Dependency.Bundle.Version.AllowPrereleases should be true")
-	assert.Equal(t, []string{"5.7.x"}, dep.Version.Ranges, "Dependency.Bundle.Version.Ranges is incorrect")
+	assert.Len(t, deps, 2, "Dependencies.Requires is the wrong length")
+
+	for _, dep := range deps {
```
vdice, line 25: Can we not range via `Dependencies.Sequence` here?  (And/or, build in the expectation of a certain order/sequence into these assertions?)

pkg/cnab/extensions/dependencies.go
```diff
@@ -94,7 +101,27 @@ func HasDependencies(bun bundle.Bundle) bool {
 	return ok
 }
 
+
+// ListBySequence returns the dependencies by the defined sequence,
+// if none is specified, they are unsorted.
+func (d Dependencies) ListBySequence() []Dependency {
```
vdice, line 107: Why not return a `Dependencies` object as opposed to `[]Dependency`?

pkg/cnab/extensions/dependencies.go
```diff
@@ -51,18 +57,19 @@ type DependencyVersion struct {
 // ReadDependencies is a convenience method for returning a bonafide
 // Dependencies reference after reading from the applicable section from
 // the provided bundle
-func ReadDependencies(bun bundle.Bundle) (Dependencies, error) {
+func ReadDependencies(bun bundle.Bundle) ([]Dependency, error) {
```
vdice, line 60: Why not return a `Dependencies` object as opposed to `[]Dependency`?

carolynvs: It would be fine to return Dependencies from here, and then in solver.go ResolveDependencies immediately call `deps.ListBySequence()` and then iterate over the result.

Probably would make more sense from a separation of responsibilities to more faithfully return the full structure as represented in the bundle.json from this struct anyway.

pkg/cnab/extensions/dependencies.go
carolynvs: This function is responsible for understanding how to convert the data inside of a Dependencies (sequence and requires) and return a sequence of Dependency items for the consumer to iterate over. This way the logic from combining sequence and requires is in a single location.

It wouldn't really help to return a Dependencies struct from this method.

pkg/cnab/extensions/dependencies_test.go
carolynvs: If we stop calling ListBySequence inside of ReadDependencies (because it now returns a Dependencies struct), we would start calling it here immediately after ReadDependencies, then could assume the items are ordered without the for loop.

I wouldn't test the ordering much in this function because the very next function below tests ListBySequence in detail. Double testing that function doesn't make much sense IMO and I'd limit the test here to the scope of ReadDependencies.

pkg/porter/testdata/cache/README.md
```diff
@@ -1,5 +1,5 @@
 # Cache Cheat Sheet
 
 * deislabs/kubekahn:1.0 -> 887e7e65e39277f8744bd00278760b06
-* getporter/mysql:v0.1.2 -> 972978d3b715212b783a70fdc3449584
+* getporter/mysql:v0.1.3 -> 972978d3b715212b783a70fdc3449584
```
carolynvs, line 4: Oops, this should not have been changed. The cache keys in porter correspond to the bundle tag, so if the tag changed, the key would too. I assume this was a find/replace that should not have touched this file. Please revert this file.

pkg/manifest/manifest_test.go
```diff
@@ -78,6 +78,30 @@ func TestLoadManifestWithDependencies(t *testing.T) {
...
+	cxt := context.NewTestContext(t)
+
+	cxt.AddTestFile("testdata/porter-with-deps.yaml", config.Name)
+	cxt.AddTestDirectory("testdata/bundles", "bundles")
+
+	m, err := LoadManifestFrom(cxt.Context, config.Name)
+	require.NoError(t, err, "could not load manifest")
+
+	assert.NotNil(t, m)
+	assert.Equal(t, []MixinDeclaration{{Name: "exec"}}, m.Mixins)
```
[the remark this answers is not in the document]
MChorfa: Yes, you are right, these assertions are not relevant

pkg/porter/testdata/cache/README.md
MChorfa: Yes, it was a  find/replace 😅

pkg/cnab/extensions/dependencies.go
MChorfa: Changes applied  accordantly

MChorfa: Now the solver is responsible to call ListBySequence() before building the DependencyLock array

vdice:
@MChorfa also remembered we'll need to update the [manifest schema](https://github.com/deislabs/porter/blob/main/pkg/templates/templates/schema.json) 😅

vdice:
@MChorfa sorry that you're battling the schema unit test with subtle differences.  I think it can be resolved by copying the schema that is printed out when the test fails (from `make test-unit`) and copying that to `pkg/porter/testdata/schema.json`.

vdice:
Or, even simpler, copy this updated file: https://raw.githubusercontent.com/vdice/porter/mchorfa-user-defined-order/pkg/porter/testdata/schema.json  That should work 🙏

MChorfa:
> Or, even simpler, copy this updated file: https://raw.githubusercontent.com/vdice/porter/mchorfa-user-defined-order/pkg/porter/testdata/schema.json That should work 🙏

Ah! thank you :)

PR #1182, opened 2020-07-29
yaml for bundle to set up discourse

gaurimadhok:
# What does this change
This an example bundle that will set up and install Discourse. Currently, the bundle has a readme to explain the purpose of the bundle and a porter-discourse.yaml with a basic outline of what the bundle needs. There are still unimplemented parts, so this bundle cannot run in its current state. The porter.yaml is the hello example bundle and is added so that the bundle can run in the pipeline.

# What issue does it fix
The bundle will provide an example of how Porter can be used to build a complex application.


examples/porter-discourse/porter.yaml
```diff
@@ -0,0 +1,186 @@
...
+
+  - name: SMTP_user_name
+    type: string
+
+  - name: SMTP_password
+    type: string
+
+install:
+  - az:
+      description: "Azure ClI login"
```
iennae, line 86: ClI -> CLI

gaurimadhok: thanks for catching that!

examples/porter-discourse/README.md
```diff
@@ -0,0 +1,62 @@
...
...docker image and do a simple install. You will have to go through the complex process in order to set up the infrastructure and customize your Discourse. In order to install, first, you need to create a cloud server and access it. Then, you need to install docker and install discourse. To get email working, you will need to set up a mail server and get the credentials. You also need a domain name. Then, you will launch a discourse set up tool and answer questions about your hostname, email address, server address, etc. After Discourse is up and running, if an update needs to be made or maintenance is needed, another process is required. Additionally, if you want more features such as single-sign-on, plugins, or encryption, those need to be configured as well. Because of all these complex steps that are needed to install Discourse and upgrade it, putting Discourse in a bundle would greatly simplify the process. It would take away the need for individuals installing Discourse to read through all the installation instructions. Users would simply need to enter their credentials and any parameters and Porter would do the work for them. Discourse would be installed with just one command.
```
carolynvs, line 4: Make sure to preview your markdown files in your editor locally. That will help you proactively spot when there are formatting issues, such as a malformed markdown links like there are here.

Also if you use hard paragraph breaks every so often (I do every 80-100 characters) it is easier for reviewers to leave comments on the lines that have problems. Otherwise I can only comment on the entire paragraph and it's more difficult to suggest changes to help out. If you open your PR's files tab and try to leave a comment yourself on a single line in this paragraph, you will quickly see what I'm talking about. 😀

examples/porter-discourse/README.md
```diff
@@ -0,0 +1,62 @@
...
+    - need credentials
+* email (Mailgun, SendGrid, Mailjet)
+    - can configure reply via email
+    - need credentials and email as parameter
+* domain name
+    - hostname parameter
+* ssl certificate (Let’s Encrypt free certificate) 
+    - need certificate and key
+    - have to configure NGINX and a docker container
+* docker as a service (viable as a bundle) / Virtual Machine (current standard)
```
carolynvs, line 21: When we took these notes, we were talking about what _could be done_ but now this is documenting what we want the bundle to do for whoever will implement it. So let's point them in a firm direction and just say it will be a virtual machine.

examples/porter-discourse/README.md
```diff
@@ -0,0 +1,62 @@
...
+* configure SSO
+    - enable_sso parameter must be enabled
+    - sso_url: the offsite URL users will be sent to when attempting to log on
+    - sso_secret: a secret string used to hash SSO payloads
+* login via Google, Twitter, GitHub, Facebook
+* install plugins
+    - need plugin's git clone url 
+    - add it to app.yml
+    - rebuild the container
+* multisite configuration with docker
```
carolynvs, line 34: I am not sure what this bullet refers to?

examples/porter-discourse/README.md
```diff
@@ -0,0 +1,62 @@
+# Discourse Example Bundle
+
+## Why Discourse?
```
carolynvs, line 3: Let's put right at the top before we even say Why Discourse that this is a work in progress and is not a finished example bundle.

examples/porter-discourse/README.md
gaurimadhok: In the readme instructions for install for Discourse, it talked about how if you want to you could configure multiple Discourse sites on the same server and the link it provided explained how to do multisite configuration with docker. I added a bullet under to try to explain it but we can also remove this if it isn't needed.

carolynvs: I think let's remove that because it's pretty advanced and I predict that we are doing to encourage people to do a single site per bundle.

Issue #1169, opened 2020-07-26
Archive SEGV when destination folder does not exist


chris-crone:
## Describe the bug
When running `porter bundle archive folder-does-not-exist/bundle.tgz`, Porter segfaults.

## To Reproduce
1. Use any porter.yaml
1. Run `porter bundle archive folder-does-not-exist/bundle.tgz`
1. Segfault

## Expected behavior
Either Porter should: (1) error out saying the destination folder does not exist, or (2) create the destination folder.

I think that (1) is preferred.

## Porter Command and Output
```
$ porter bundle archive archive/kubecon-eu-2020.tgz
ERRO[0008] Can't add file /var/folders/cb/156z21pd32g06x7s9p3v7g9r0000gn/T/kubecon-eu-2020-0.1.0857665579/artifacts/layout/blobs/sha256/0b8eaf929c06a2baf68481a231287c519c58765d93125dfd53c7d0dd0641a336 to tar: io: read/write on closed pipe 
ERRO[0008] Can't close tar writer: io: read/write on closed pipe 
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x106cba2]

goroutine 1 [running]:
io.copyBuffer(0x0, 0x0, 0x2aafd40, 0xc00000e090, 0xc0009f2000, 0x8000, 0x8000, 0x19, 0x0, 0x0)
        /opt/hostedtoolcache/go/1.13.10/x64/src/io/io.go:404 +0x1d2
io.Copy(...)
        /opt/hostedtoolcache/go/1.13.10/x64/src/io/io.go:364
get.porter.sh/porter/pkg/porter.(*exporter).export(0xc0005e9b38, 0x0, 0x0)
        /home/vsts/work/1/s/pkg/porter/archive.go:130 +0x6a4
get.porter.sh/porter/pkg/porter.(*Porter).Archive(0xc000178160, 0x28141e0, 0xb, 0xc00027adc0, 0x11, 0x0, 0x0, 0x7ffeefbff98e, 0x1b, 0x0, ...)
        /home/vsts/work/1/s/pkg/porter/archive.go:70 +0x25c
main.buildBundleArchiveCommand.func2(0xc000256dc0, 0xc000254f60, 0x1, 0x1, 0x0, 0x0)
        /home/vsts/work/1/s/cmd/porter/bundle.go:342 +0x7e
[...]

PR #1186, opened 2020-07-30
fix check for GOPATH/bin

simongdavies:
# What does this change

Updates check in make file for GOPATH/bin in PATH to use =~ rather than extglob which when turned off results in the issue reported in #1169


chris-crone:
Seems to work on my machine now. Thanks @simongdavies and @vdice!

PR #1189, opened 2020-07-31
feat(*): add installation delete options/flags to relevant commands

vdice:
# What does this change
* Adds logic to enable deleting an installation, either via `porter uninstall` or `porter installation delete` (as designed in https://github.com/deislabs/porter/issues/1163)

# What issue does it fix
Closes https://github.com/deislabs/porter/issues/1163

# Notes for the reviewer
- Do we want any additional docs besides the auto-gen'd CLI docs?
- Suggestions for simplifying logic around uninstall +/- delete +/- force-delete welcome.
- Note how we are swallowing errors around the execution of the uninstall action if force-delete is true -- do we like this approach?  Or should we continue to return any/all execution errors after we force delete the installation?

# Checklist
- [x] Unit Tests
- [x] Documentation

### DIFF
--- a/pkg/cnab/provider/helpers.go
+++ b/pkg/cnab/provider/helpers.go
@@ -45,6 +45,8 @@ func (t *TestRuntime) LoadBundle(bundleFile string) (bundle.Bundle, error) {
 }
 
 func (t *TestRuntime) Execute(args ActionArguments) error {
-	args.Driver = debugDriver
+	if args.Driver == "" {
+		args.Driver = debugDriver
+	}
 	return t.Runtime.Execute(args)
 }
--- a/pkg/context/helpers.go
+++ b/pkg/context/helpers.go
@@ -135,6 +135,41 @@ func (c *TestContext) AddTestDirectory(srcDir, destDir string) {
 	}
 }
 
+func (c *TestContext) AddTestDriver(src, name string) string {
+	c.T.Helper()
+
+	data, err := ioutil.ReadFile(src)
+	if err != nil {
+		c.T.Fatal(err)
+	}
+
+	dirname, err := c.FileSystem.TempDir("", "porter")
+	if err != nil {
+		c.T.Fatal(err)
+	}
+
+	// filename in accordance with cnab-go's command driver expectations
+	filename := fmt.Sprintf("%s/cnab-%s", dirname, name)
+
+	newfile, err := c.FileSystem.Create(filename)
+	if err != nil {
+		c.T.Fatal(err)
+	}
+
+	if len(data) > 0 {
+		newfile.Write(data)
+	}
+
+	c.FileSystem.Chmod(newfile.Name(), os.ModePerm)
+	newfile.Close()
+	path := os.Getenv("PATH")
+	pathlist := []string{dirname, path}
+	newpath := strings.Join(pathlist, string(os.PathListSeparator))
+	os.Setenv("PATH", newpath)
+
+	return dirname
+}
+
 // GetOutput returns all text printed to stdout.
 func (c *TestContext) GetOutput() string {
 	return string(c.capturedOut.Bytes())
--- a/pkg/context/helpers.go
+++ b/pkg/context/helpers.go
@@ -157,11 +157,21 @@ func (c *TestContext) AddTestDriver(src, name string) string {
 	}
 
 	if len(data) > 0 {
-		newfile.Write(data)
+		_, err := newfile.Write(data)
+		if err != nil {
+			c.T.Fatal(err)
+		}
 	}
 
-	c.FileSystem.Chmod(newfile.Name(), os.ModePerm)
-	newfile.Close()
+	err = c.FileSystem.Chmod(newfile.Name(), os.ModePerm)
+	if err != nil {
+		c.T.Fatal(err)
+	}
+	err = newfile.Close()
+	if err != nil {
+		c.T.Fatal(err)
+	}
+
 	path := os.Getenv("PATH")
 	pathlist := []string{dirname, path}
 	newpath := strings.Join(pathlist, string(os.PathListSeparator))
--- a/pkg/porter/delete.go
+++ b/pkg/porter/delete.go
@@ -10,6 +10,9 @@ import (
 
 const installationDeleteTmpl = "deleting installation records for %s...\n"
 
+// ErrUnsafeInstallationDelete warns the user that deletion of an unsuccessfully uninstalled installation is unsafe
+var ErrUnsafeInstallationDelete = errors.New("it is unsafe to delete an installation when the last action wasn't a successful uninstall; if you are sure it should be deleted, retry the last command with the --force flag")
+
 // DeleteOptions represent options for Porter's installation delete command
 type DeleteOptions struct {
 	sharedOptions
@@ -45,7 +48,7 @@ func (p *Porter) DeleteInstallation(opts DeleteOptions) error {
 	result, err := p.Claims.ReadLastResult(claim.ID)
 
 	if (claim.Action != claims.ActionUninstall || result.Status != claims.StatusSucceeded) && !opts.Force {
-		return fmt.Errorf("not deleting installation as the last action was not a successful %s; use --force to override", claims.ActionUninstall)
+		return ErrUnsafeInstallationDelete
 	}
 
 	fmt.Fprintf(p.Out, installationDeleteTmpl, opts.Name)
--- a/pkg/porter/delete.go
+++ b/pkg/porter/delete.go
@@ -10,8 +10,13 @@ import (
 
 const installationDeleteTmpl = "deleting installation records for %s...\n"
 
-// ErrUnsafeInstallationDelete warns the user that deletion of an unsuccessfully uninstalled installation is unsafe
-var ErrUnsafeInstallationDelete = errors.New("it is unsafe to delete an installation when the last action wasn't a successful uninstall; if you are sure it should be deleted, retry the last command with the --force flag")
+var (
+	// ErrUnsafeInstallationDelete warns the user that deletion of an unsuccessfully uninstalled installation is unsafe
+	ErrUnsafeInstallationDelete = errors.New("it is unsafe to delete an installation when the last action wasn't a successful uninstall")
+
+	// ErrUnsafeInstallationDeleteRetryForce presents the ErrUnsafeInstallationDelete error and provides a retry option of --force
+	ErrUnsafeInstallationDeleteRetryForce = fmt.Errorf("%s; if you are sure it should be deleted, retry the last command with the --force flag", ErrUnsafeInstallationDelete)
+)
 
 // DeleteOptions represent options for Porter's installation delete command
 type DeleteOptions struct {
@@ -48,7 +53,7 @@ func (p *Porter) DeleteInstallation(opts DeleteOptions) error {
 	}
 
 	if (claim.Action != claims.ActionUninstall || installation.GetLastStatus() != claims.StatusSucceeded) && !opts.Force {
-		return ErrUnsafeInstallationDelete
+		return ErrUnsafeInstallationDeleteRetryForce
 	}
 
 	fmt.Fprintf(p.Out, installationDeleteTmpl, opts.Name)
--- a/pkg/porter/delete.go
+++ b/pkg/porter/delete.go
@@ -1,0 +1,53 @@
+package porter
+
+import (
+	"fmt"
+
+	"get.porter.sh/porter/pkg/context"
+	claims "github.com/cnabio/cnab-go/claim"
+	"github.com/pkg/errors"
+)
+
+const installationDeleteTmpl = "deleting installation records for %s...\n"
+
+// DeleteOptions represent options for Porter's installation delete command
+type DeleteOptions struct {
+	sharedOptions
+	Force bool
+}
+
+// Validate prepares for an installation delete action and validates the args/options.
+func (o *DeleteOptions) Validate(args []string, cxt *context.Context) error {
+	// Ensure only one argument exists (installation name) if args length non-zero
+	err := o.sharedOptions.validateInstallationName(args)
+	if err != nil {
+		return err
+	}
+
+	return o.sharedOptions.defaultBundleFiles(cxt)
+}
+
+// DeleteInstallation handles deletion of an installation
+func (p *Porter) DeleteInstallation(opts DeleteOptions) error {
+	err := p.applyDefaultOptions(&opts.sharedOptions)
+	if err != nil {
+		return err
+	}
+
+	// TODO: do we check to see if the installation claim has required deps
+	// declared and if so, attempt to delete them as well?
+
+	claim, err := p.Claims.ReadLastClaim(opts.Name)
+	if err != nil {
+		return errors.Wrapf(err, "unable to read last claim for installation %s", opts.Name)
+	}
+
+	result, err := p.Claims.ReadLastResult(claim.ID)
+
+	if (claim.Action != claims.ActionUninstall || result.Status != claims.StatusSucceeded) && !opts.Force {
+		return fmt.Errorf("not deleting installation as the last action was not a successful %s; use --force to override", claims.ActionUninstall)
+	}
+
+	fmt.Fprintf(p.Out, installationDeleteTmpl, opts.Name)
+	return p.Claims.DeleteInstallation(opts.Name)
+}
--- a/pkg/porter/delete.go
+++ b/pkg/porter/delete.go
@@ -37,17 +37,17 @@ func (p *Porter) DeleteInstallation(opts DeleteOptions) error {
 		return err
 	}
 
-	// TODO: do we check to see if the installation claim has required deps
-	// declared and if so, attempt to delete them as well?
-
-	claim, err := p.Claims.ReadLastClaim(opts.Name)
+	installation, err := p.Claims.ReadInstallationStatus(opts.Name)
 	if err != nil {
-		return errors.Wrapf(err, "unable to read last claim for installation %s", opts.Name)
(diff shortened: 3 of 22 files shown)


pkg/cnab/provider/helpers.go
```diff
@@ -45,6 +45,8 @@ func (t *TestRuntime) LoadBundle(bundleFile string) (bundle.Bundle, error) {
 }
 
 func (t *TestRuntime) Execute(args ActionArguments) error {
-	args.Driver = debugDriver
+	if args.Driver == "" {
```
carolynvs, line 48: 👍

pkg/context/helpers.go
```diff
@@ -135,6 +135,41 @@ func (c *TestContext) AddTestDirectory(srcDir, destDir string) {
...
+	// filename in accordance with cnab-go's command driver expectations
+	filename := fmt.Sprintf("%s/cnab-%s", dirname, name)
+
+	newfile, err := c.FileSystem.Create(filename)
+	if err != nil {
+		c.T.Fatal(err)
+	}
+
+	if len(data) > 0 {
+		newfile.Write(data)
```
carolynvs, line 160: Write can return an error and so can Chmod and Close. Let's use `require.NoError(c.T, err)` to check that they didn't return an error. 

The other helpers in porter use testify so it's okay to make use of that here as well.

pkg/porter/delete.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+// DeleteInstallation handles deletion of an installation
+func (p *Porter) DeleteInstallation(opts DeleteOptions) error {
+	err := p.applyDefaultOptions(&opts.sharedOptions)
+	if err != nil {
+		return err
+	}
+
+	// TODO: do we check to see if the installation claim has required deps
```
carolynvs, line 37: I believe all the logic for dependencies should be encapsulated in the actions, e.g. uninstall. So if they have made it to this point, either they are forcing the deletion or it was uninstalled and the deps were taken care of already.

pkg/porter/delete.go
```diff
@@ -0,0 +1,53 @@
...
+	// declared and if so, attempt to delete them as well?
+
+	claim, err := p.Claims.ReadLastClaim(opts.Name)
```
carolynvs, line 40: You can use `installation, err := p.Claims.ReadInstallationStatus(opts.Name)` to retrieve the last claim and result for the installation. Installation has `GetLastClaim()` to get the action, and `GetLastStatus()` to get succeeded or not. It will result in fewer connections this way.

pkg/porter/delete.go
```diff
@@ -0,0 +1,53 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "unable to read last claim for installation %s", opts.Name)
+	}
+
+	result, err := p.Claims.ReadLastResult(claim.ID)
+
+	if (claim.Action != claims.ActionUninstall || result.Status != claims.StatusSucceeded) && !opts.Force {
+		return fmt.Errorf("not deleting installation as the last action was not a successful %s; use --force to override", claims.ActionUninstall)
```
carolynvs, line 48: The error message is a bit confusing. How about  

```suggestion
		return fmt.Error("It is unsafe to delete an installation when the last action wasn't a successful uninstall. If you are sure it should be deleted, retry the last command with the --force flag.")
```

pkg/porter/delete_test.go
```diff
@@ -0,0 +1,64 @@
...
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
+			p := NewTestPorter(t)
+			p.TestConfig.SetupPorterHome()
+
+			var err error
+
+			// Create test claim
+			if tc.lastAction != "" {
+				c := p.TestClaims.CreateClaim("test", tc.lastAction, bundle.Bundle{}, nil)
```
carolynvs, line 36: You are using a convenience method "CreateClaim" that creates the claim with NewClaim, asserts that no error occured, persists the claim and returns just the claim. So you don't need to call SaveClaim or NoError again below.

pkg/porter/delete_test.go
```diff
@@ -0,0 +1,64 @@
...
+				err := p.Claims.SaveClaim(c)
+				require.NoError(t, err, "SaveClaim failed")
+				_ = p.TestClaims.CreateResult(c, tc.lastActionStatus)
+			}
+
+			opts := DeleteOptions{
```
carolynvs, line 42: Just a suggestion that when using embedded types, sometimes it's easier to not set fields when initializing the struct. depending on how nested it is, it can make your life a lot easier and save lines. For example:

```go
opts := DeleteOptions{}
opts.Name = "test"
opts.Force =  tc.force
```

pkg/porter/dependencies.go
```diff
@@ -248,20 +249,34 @@ func (e *dependencyExecutioner) executeDependency(dep *queuedDependency, parentA
...
+	var executeErrs error
 	fmt.Fprintf(e.Out, "Executing dependency %s...\n", dep.Alias)
 	err := e.CNAB.Execute(depArgs)
 	if err != nil {
-		return errors.Wrapf(err, "error executing dependency %s", dep.Alias)
+		executeErrs = multierror.Append(executeErrs, errors.Wrapf(err, "error executing dependency %s", dep.Alias))
+
+		if depArgs.Action == claim.ActionUninstall {
+			if e.parentOpts.Delete && !e.parentOpts.ForceDelete {
+				executeErrs = multierror.Append(executeErrs, fmt.Errorf("not deleting installation %s as uninstall was not successful; use --force-delete to override", depArgs.Installation))
```
carolynvs, line 261: Can you update this error message to match the other as well that you just updated?

pkg/porter/uninstall.go
```diff
@@ -32,16 +40,41 @@ func (p *Porter) UninstallBundle(opts UninstallOptions) error {
 		return err
 	}
 
+	var uninstallErrs error
 	fmt.Fprintf(p.Out, "uninstalling %s...\n", opts.Name)
 	err = p.CNAB.Execute(opts.ToActionArgs(deperator))
 	if err != nil {
+		uninstallErrs = multierror.Append(uninstallErrs, err)
+		if strings.Contains(err.Error(), claim.ErrInstallationNotFound.Error()) {
```
carolynvs, line 48: Can you add a comment on why this special handling for ErrInstallationNotFound is needed?

pkg/porter/uninstall.go
```diff
@@ -32,16 +40,41 @@ func (p *Porter) UninstallBundle(opts UninstallOptions) error {
...
 		if len(deperator.deps) > 0 {
-			return errors.Wrapf(err, "failed to uninstall the %s bundle, the remaining dependencies were not uninstalled", opts.Name)
-		} else {
-			return err
+			uninstallErrs = multierror.Append(uninstallErrs,
+				fmt.Errorf("failed to uninstall the %s bundle, the remaining dependencies were not uninstalled", opts.Name))
+		}
+
+		if opts.Delete && !opts.ForceDelete {
+			uninstallErrs = multierror.Append(uninstallErrs, fmt.Errorf("not deleting installation %s as uninstall was not successful; use --force-delete to override", opts.Name))
```
carolynvs, line 58: Sorry looks like that error message I had you update was used here as well.

pkg/porter/uninstall.go
```diff
@@ -32,16 +40,41 @@ func (p *Porter) UninstallBundle(opts UninstallOptions) error {
...
 		}
+		if !opts.ForceDelete {
+			return uninstallErrs
+		}
+		// else, we swallow uninstallErrs as opts.ForceDelete is true
```
carolynvs, line 63: I think it would be helpful to no matter what, print out the errors even on --force-delete. We just will have a 0 exit code for --force-delete even when there are errors. That way you know what happened.

pkg/porter/uninstall.go
```diff
@@ -32,16 +40,41 @@ func (p *Porter) UninstallBundle(opts UninstallOptions) error {
...
+		// and we wish to pass same option down to deps, if applicable
 	}
 
 	// TODO: See https://github.com/deislabs/porter/issues/465 for flag to allow keeping around the dependencies
-	return deperator.Execute()
+	err = deperator.Execute()
+	if err != nil {
```
carolynvs, line 69: Right now it's confusing how the ForceDelete/Delete is handled by both this function and the Deperator, and then they have all sorts of conditionals for handling the error, printing that warning about using --force, etc. I am concerned that it's going to be difficult to maintain without introducing bugs or fixing a bug accidentally in just one place (and forgetting to fix it in the dependencies.go for example).

🤔 Looking at it in a github code review it's hard to say "just change X" to simplify it. But I do wish there was a way to do so. Maybe we can take a second pass at it and see if there's a way to either extract the logic for dealing with it, or consolidate the logic for checking and printing the messages? I'm not sure what to suggest which is why I think trying again in a follow-up or whenever we next feel clever or inspired is fine.

One idea is to have a function takes in the flags (delete, force delete and the installation status) and it handles deciding if we should delete the installation and printing the warning. Then at least we could reuse that in 2+ places?

tests/dependencies_test.go
```diff
@@ -31,6 +31,8 @@ func TestDependenciesLifecycle(t *testing.T) {
 	invokeWordpressBundle(p, namespace)
 
 	uninstallWordpressBundle(p, namespace)
+
+	noReallyUninstallWordpressBundle(p, namespace)
```
carolynvs, line 35: 😂

tests/uninstall_test.go
```diff
@@ -0,0 +1,91 @@
...
+	testcases := []struct {
+		name               string
+		installed          bool
+		uninstallFails     bool
+		delete             bool
+		forceDelete        bool
+		installationExists bool
+		wantError          string
+	}{
+		{"not yet installed", false, false, false, false, false, "1 error occurred:\n\t* could not load installation HELLO: Installation does not exist\n\n"},
```
carolynvs, line 23: I suggest test cases like this, which are really long, from setting every field without using the field name (which is quite hard to read without the support of an editor) to using the field names and only setting the ones that are non-default values,

```go
{name: "not yet installed", wantError: "1 error occurred..."},
```

pkg/porter/uninstall.go
vdice: You're right, this area warrants further thought.  Thank you for the initial suggestions.  I'll spend some more time here and hopefully have an update w/ less fragility and duplication.

vdice: (Still working on this one...)

vdice: Ok, refactored for a bit of simplification... or at least some shared logic.  Not sure if better, but perhaps it might generate some further ideas: https://github.com/deislabs/porter/pull/1189/commits/dd5dd04dba74654b626a5ccf92eec5ebcc52b1b3

tests/dependencies_test.go
vdice: Oh, I just realized, should I just place this uninstall+delete logic into `cleanupWordpressBundle`?  Or keep it separate?

vdice: (I updated `cleanupWordpressBundle` w/ uninstall+delete logic in https://github.com/deislabs/porter/pull/1189/commits/6af718a9446dbd9c99821092c06f89822c553dde)

tests/dependencies_test.go
```diff
@@ -101,24 +99,29 @@ func installWordpressBundle(p *porter.TestPorter) (namespace string) {
...
-func cleanupWordpressBundle(p *porter.TestPorter) {
-	uninstallOpts := porter.UninstallOptions{}
-	uninstallOpts.CredentialIdentifiers = []string{"ci"}
-	uninstallOpts.Tag = p.Manifest.Dependencies["mysql"].Tag
-	err := uninstallOpts.Validate([]string{"wordpress-mysql"}, p.Porter)
-	assert.NoError(p.T(), err, "validation of uninstall opts failed for dependent bundle")
+func cleanupWordpressBundle(p *porter.TestPorter, namespace string) {
+	uninstallOptions := porter.UninstallOptions{}
+	uninstallOptions.CredentialIdentifiers = []string{"ci"}
+	uninstallOptions.Delete = true
```
carolynvs, line 105: 👍

pkg/porter/dependencies.go
```diff
@@ -248,22 +249,31 @@ func (e *dependencyExecutioner) executeDependency(dep *queuedDependency, parentA
...
-			return err
+		// Handle errors when/if the action is uninstall
+		executeErrs = e.parentOpts.handleUninstallErrs(e.Out, executeErrs)
+		if executeErrs != nil {
+			return executeErrs
 		}
+	}
 
-		dep.outputs = outputs
+	if e.parentOpts.shouldDelete() {
```
carolynvs, line 266: I think you put the `UninstallDeleteOptions` on the generic `BundleLifecycleOptions` so that you could make this check here, right?

Instead, since I'm asking further down to move that to the `UninstallOptions` type would be to do a type check

```go
if uninstOpts, ok := e.parentOpts.(UninstallOptions); ok {
...
}
```

Right now `parentOpts` is just a `BundleLifecycleOptions` but we could make it an interface (pseudo code, not tested, but something like below), so that we could cast and get at the original options as needed while still having access to the other data that we were using on the parentOpts the rest of the time.

What do you think?

```golang
type BundleAction interface {
    // Implemented for free because of the embedded BundleLifecycleOpts
    ToActionArgs()

    // Let us grab and get direct access to the bundle opts fields like we need in some funcs
    GetBundleLifecycleOptions()
}
```

pkg/porter/uninstall.go
```diff
@@ -2,17 +2,53 @@ package porter
...
+
+func (opts *UninstallDeleteOptions) shouldDelete() bool {
+	return opts.Delete || opts.ForceDelete
+}
+
+func (opts *UninstallDeleteOptions) unsafeDelete() bool {
+	return opts.Delete && !opts.ForceDelete
+}
+
+func (opts *UninstallDeleteOptions) handleUninstallErrs(out io.Writer, err error) error {
```
carolynvs, line 36: 👍

pkg/porter/dependencies.go
vdice: Great suggestion!  Thank you for providing the blueprint on this approach.  Here's what I've updated: https://github.com/deislabs/porter/pull/1189/commits/4b2b64746636e6997ee1f3dae8c87953c8b0cddc

How does it look?

pkg/porter/dependencies.go
```diff
@@ -250,20 +252,27 @@ func (e *dependencyExecutioner) executeDependency(dep *queuedDependency, parentA
...
 		// Handle errors when/if the action is uninstall
-		executeErrs = e.parentOpts.handleUninstallErrs(e.Out, executeErrs)
+		executeErrs = uninstallOpts.handleUninstallErrs(e.Out, executeErrs)
 		if executeErrs != nil {
 			return executeErrs
 		}
 	}
 
-	if e.parentOpts.shouldDelete() {
+	if uninstallOpts.shouldDelete() {
```
carolynvs, line 275: If uninstallOpts is an empty struct (because parentOpts was some other action), what does this call do?

pkg/porter/lifecycle.go
```diff
@@ -5,11 +5,22 @@ import (
...
 )
 
+var _ BundleAction = BundleLifecycleOpts{}
+
+// BundleAction is an interface that defines methods related to constructing
+// ActionArguments and supplying BundleLifecycleOptions.  The latter is useful
+// when the implementation contains further, action-specific options beyond
+// the stock BundleLifecycleOptions.
+type BundleAction interface {
+	ToActionArgs(*dependencyExecutioner) cnabprovider.ActionArguments
```
carolynvs, line 15: Looks like this isn't used anywhere? I thought it may be used depending on how the implementation worked out but guess not.

pkg/porter/lifecycle.go
```diff
@@ -28,8 +39,12 @@ func (o *BundleLifecycleOpts) Validate(args []string, porter *Porter) error {
 	return nil
 }
 
+func (o BundleLifecycleOpts) GetBundleLifecycleOptions() BundleLifecycleOpts {
```
carolynvs, line 42: Brilliant! I wouldn't have thought to implement this here, and would have put it on each of the actions but this is sneakier. 💯

pkg/porter/lifecycle.go
vdice: it is used here: https://github.com/deislabs/porter/blob/4b2b64746636e6997ee1f3dae8c87953c8b0cddc/pkg/porter/dependencies.go#L86

But, to your point, it could just as easily be removed from this interface and exist as it once was (a method on `BundleLifecycleOpts`), since at this time we don't see a need to have BundleAction implementations overriding its behavior.  Shall I change this back and remove the method from the interface?

carolynvs: Oh, thanks I obviously missed that. Either way is fine, up to you!

pkg/porter/dependencies.go
vdice: It *think* all methods on an empty struct (when action not uninstall) return the uninitialized/default values for the struct fields... so mostly boolean `false` values, which is what we'd want here.  The tests seem to agree.

However, I can see this approach/code being fragile... even if the assumptions are true.  What do we think?  Should we refactor the areas where we rely on uninstallOpts to run only if we are sure the struct is actually that type (and not empty)?

carolynvs: I think it would get pretty gnarly to conditionally call those, so the empty struct is fine. I was just double checking. Maybe let's just add a comment about if it's not uninstall, it's a no-op handler?

pkg/porter/lifecycle.go
vdice: Ok, removed from interface 👍 .

PR #1193, opened 2020-08-04
Support bundle dependencies with new claims spec

### LINKED ISSUE #1069: Generate parameter sources for dependency outputs
When we build a bundle, look for templating that uses `bundle.dependencies.DEPENDENCY_NAME.outputs.*` and create parameter source definition of type `dependencies.output`

```yaml
dependencies:
  mysql:
    tag: getporter/mysql:v0.1.2

install:
  - exec: "Install Wordpress"
      command: ./install.sh
      arguments:
        - "{{ bundle.dependencies.mysql.outputs.root-password }}"
```

should yield the following bundle.json

```json
{
    "custom": {
      "io.cnab.parameter-sources": {
        "porter-connstr": {
          "priority": ["dependencies.output"],
          "sources": {
            "dependencies.output": {
              "name": "connstr"
            }
          }
        }
      }
    }
```

Otherwise beyond the different source type of the parameter, it should work the same as #1067 .

This requires an addition to the cnab spec to support `dependencies.output` which we should prove out here.

carolynvs:
# What does this change
Generate bundle dependency parameter sources when `{{ bundle.dependencies.DEP.outputs.OUTPUT }}` is used in porter.yaml and then inject the output from the dependency into the bundle at runtime.

# What issue does it fix
Closes #1069  

# Notes for the reviewer

# Checklist
- [x] Unit Tests
- [ ] Documentation TODO
- [ ] Schema (porter.yaml)


carolynvs:
@vdice  this isn't ready, I'm just having trouble getting the build to run so I tried removing it from draft

pkg/cnab/provider/parameters.go
```diff
@@ -102,34 +107,46 @@ func (r *Runtime) resolveParameterSources(bun bundle.Bundle, args ActionArgument
...
-						strings.Contains(err.Error(), claim.ErrOutputNotFound.Error()) {
-						continue
-					}
-					// Otherwise, something else has happened, perhaps bad data or connectivity problems, we can't ignore it
-					return nil, errors.Wrapf(err, "could not set parameter %s from its parameter source, output %s", parameterName, source.OutputName)
-				}
+				installation = args.Installation
+				outputName = source.OutputName
+			case extensions.DependencyOutputParameterSource:
+				installation = fmt.Sprintf("%s-%s", args.Installation, source.Dependency)
```
vdice, line 117: Do we have a helper yet around this dep naming convention?  We do same/similar elsewhere, right?

carolynvs: No we don't have a helper and I totally messed up that naming pattern once already when developing this locally. Let me add a helper function for this.

PR #1198, opened 2020-08-05
Adds support for debugging plugins

simongdavies:
# What this change does

Adds the ability to have porter start delve and attach it to a plugin.

The following variables are required/used:

`PORTER_RUN_PLUGIN_IN_DEBUGGER` should be set to the name of the plugin to be debugged (e.g. secrets.azure.keyvault to debug the azure secrets plugin)
`PORTER_DEBUGGER_PORT` should be set to the port number where the delve API will listen, if not set it defaults to 2345
`PORTER_PLUGIN_WORKING_DIRECTORY` should be the path to the source code for the plugin being executed. (e.g. for the Azure plugin this value would be <PATH>/porter-azure-plugins/cmd/azure

The plugin to be debugged should be compiled and placed in porters plugin path (e.g. in the Azure plugin case the plugin should be copied to $PORTER_HOME/plugins/azure/.

A process will then be started using [delve attach](https://github.com/go-delve/delve/blob/master/Documentation/usage/dlv_attach.md) which exposes the delve [API ](https://github.com/go-delve/delve/tree/master/Documentation/api) so that any delve client can connect to the server and debug the plugin.

### DIFF
--- a/pkg/context/context.go
+++ b/pkg/context/context.go
@@ -21,14 +21,17 @@ const (
 type CommandBuilder func(name string, arg ...string) *exec.Cmd
 
 type Context struct {
-	Debug        bool
-	DebugPlugins bool
-	verbose      bool
-	FileSystem   *afero.Afero
-	In           io.Reader
-	Out          io.Writer
-	Err          io.Writer
-	NewCommand   CommandBuilder
+	Debug                  bool
+	DebugPlugins           bool
+	RunPlugInInDebugger    string
+	PlugInWorkingDirectory string
+	DebuggerPort           string
+	verbose                bool
+	FileSystem             *afero.Afero
+	In                     io.Reader
+	Out                    io.Writer
+	Err                    io.Writer
+	NewCommand             CommandBuilder
 }
 
 func (c *Context) SetVerbose(value bool) {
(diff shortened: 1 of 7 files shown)


pkg/context/context.go
```diff
@@ -21,14 +21,17 @@ const (
...
-	DebugPlugins bool
-	verbose      bool
-	FileSystem   *afero.Afero
-	In           io.Reader
-	Out          io.Writer
-	Err          io.Writer
-	NewCommand   CommandBuilder
+	Debug                  bool
+	DebugPlugins           bool
+	RunPlugInInDebugger    string
```
vdice, line 26: What do we think about creating a new struct to house plugin debugging config (e.g. all of the new values added here)?  Then we could add the one `DebugConfig` (or similar)  struct to `Context` here, keeping down the number of member fields.

vdice (commented):
Minor suggestion around encapsulating plugin debugging fields into a new struct.

Also, can a doc be added along the lines of what is mentioned in this PR description on how to hook everything up for debugging a plugin?  (What env vars are expected, etc.)  It could perhaps live somewhere under https://porter.sh/plugins/ ?

vdice:
(Tested in conjunction with https://github.com/deislabs/porter-azure-plugins/pull/24 ; functionality looks good!)

pkg/context/context.go
simongdavies: Done !!

pkg/plugins/pluggable/loader.go
```diff
@@ -133,6 +140,58 @@ func (l *PluginLoader) Load(pluginType PluginTypeConfig) (interface{}, func(), e
 	return raw, cleanup, nil
 }
 
+func (l *PluginLoader) setUpDebugger(client *plugin.Client, cleanup func()) (func(), error) {
+	debugContext := l.Context.PlugInDebugContext
+	if len(debugContext.RunPlugInInDebugger) > 0 && strings.ToLower(l.SelectedPluginKey.String()) == strings.TrimSpace(strings.ToLower(debugContext.RunPlugInInDebugger)) {
```
vdice, line 145: This line seems to be susceptible to a panic (seen from the [integration test](https://dev.azure.com/deislabs/porter/_build/results?buildId=6909&view=logs&j=0d1d2f05-4482-5088-4d55-986bcf02e524&t=bd076812-b150-58c7-9f90-9c2fa032ad9b)):

```
panic(0x2298ea0, 0x3738610)
	/opt/hostedtoolcache/go/1.13.10/x64/src/runtime/panic.go:679 +0x1b2
get.porter.sh/porter/pkg/plugins/pluggable.(*PluginLoader).setUpDebugger(0xc00036a8e0, 0xc00049a000, 0xc00046e610, 0x0, 0x0, 0x0)
	/home/vsts/work/1/s/pkg/plugins/pluggable/loader.go:145 +0x4b
```

vdice (commented):
Thanks for the consolidation into a new plugin debug context.

One section of the `setUpDebugger` method may need some attention; looks like the integration tests hit a panic in certain instances...

pkg/plugins/pluggable/loader.go
simongdavies: Arghhh will take a look

vdice (approved):

PR #1199, opened 2020-08-05
🐛 Skip bundle defaults if --tag passed

### LINKED ISSUE #1180: Porter parses local porter.yaml even when using tag flag
## Describe the bug
If there is a `porter.yaml` file in the current working directory, Porter systematically parses it even if one is trying to use a bundle directly from a registry.

## To Reproduce
```console
$ mkdir empty && cd empty
$ touch porter.yaml
$ porter install --tag deislabs/porter-hello-bundle:latest
Error: invalid tag : invalid reference format

$ rm porter.yaml
$ porter install --tag deislabs/porter-hello-bundle:latest
installing HELLO...
executing install action from HELLO (bundle instance: HELLO)
Install Hello World
Hello World
```

## Expected behavior
I would expect that when doing an operation with `--tag` that the local `porter.yaml` would be ignored.

## Porter Command and Output
See steps to reproduce

## Version
`porter v0.27.2 (aee93e98)`

dev-drprasad:
# What does this change
if `--tag` passed, porter still tries to validate local manifest file. This PR skips validating manifest file if `--tag` present

# What issue does it fix
Closes #1180

# Notes for the reviewer
N/A

# Checklist
- [x] Unit Tests
- [x] Documentation - N/A

### DIFF
--- a/pkg/porter/lifecycle_test.go
+++ b/pkg/porter/lifecycle_test.go
@@ -169,11 +169,11 @@ func TestManifestIgnoredWithTag(t *testing.T) {
 		opts.Tag = "deislabs/kubekahn:latest"
 
 		wd, _ := os.Getwd()
-		// Below line makes cnab.go#defaultBundleFiles#manifestExists true
+		// `path.Join(wd...` -> makes cnab.go#defaultBundleFiles#manifestExists `true`
+		// Only when `manifestExists` eq to `true`, default bundle logic will run
 		p.TestConfig.TestContext.AddTestFileContents([]byte(""), path.Join(wd, config.Name))
 		// When execution reach to `readFromFile`, manifest file path will be lost.
-		// So, had to empty default manifest content also
-		// Below line makes manifest.go#readFromFile#exists true
+		// So, had to use root manifest file also for error simuation purpose
 		p.TestConfig.TestContext.AddTestFileContents([]byte(""), config.Name)
 
 		err := opts.Validate(nil, p.Porter)
--- a/pkg/porter/lifecycle_test.go
+++ b/pkg/porter/lifecycle_test.go
@@ -160,17 +160,24 @@ func TestBundleLifecycleOpts_ToActionArgs(t *testing.T) {
 		assert.Equal(t, opts.Name, args.Installation, "Claim not populated correctly")
 		assert.Equal(t, opts.RelocationMapping, args.RelocationMapping, "RelocationMapping not populated correctly")
 	})
+}
 
+func TestManifestIgnoredWithTag(t *testing.T) {
+	p := NewTestPorter(t)
 	t.Run("ignore manifest in cwd if tag present", func(t *testing.T) {
 		opts := BundleLifecycleOpts{}
 		opts.Tag = "deislabs/kubekahn:latest"
 
-		// cnab.go#defaultBundleFiles uses `os.Getwd()`
 		wd, _ := os.Getwd()
-		p.TestConfig.TestContext.AddTestFileContents([]byte(""), path.Join(wd, "porter.yaml"))
+		// Below line makes cnab.go#defaultBundleFiles#manifestExists true
+		p.TestConfig.TestContext.AddTestFileContents([]byte(""), path.Join(wd, config.Name))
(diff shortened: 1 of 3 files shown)


dev-drprasad:
In draft, cuz thinking of adding unit tests

dev-drprasad:
@vdice I feel, your solution directly convey the problem we're trying to solve. Let incorporate those changes only

pkg/porter/lifecycle_test.go
```diff
@@ -158,6 +160,18 @@ func TestBundleLifecycleOpts_ToActionArgs(t *testing.T) {
 		assert.Equal(t, opts.Name, args.Installation, "Claim not populated correctly")
 		assert.Equal(t, opts.RelocationMapping, args.RelocationMapping, "RelocationMapping not populated correctly")
 	})
+
+	t.Run("ignore manifest in cwd if tag present", func(t *testing.T) {
```
vdice, line 164: Surprisingly, I couldn't get this test to fail when running against the current code on `main` (meaning, without the changes to cnab.go and lifecycle.go above).  Maybe double-check my work.

I succeeded in getting it to fail when explicitly specifying `opts.File = "porter.yaml"` (and it didn't seem like the working dir logic was strictly necessary):

```
+++ b/pkg/porter/lifecycle_test.go
@@ -1,8 +1,6 @@
 package porter

 import (
-       "os"
-       "path"
        "path/filepath"
        "testing"

@@ -164,13 +162,11 @@ func TestBundleLifecycleOpts_ToActionArgs(t *testing.T) {
        t.Run("ignore manifest in cwd if tag present", func(t *testing.T) {
                opts := BundleLifecycleOpts{}
                opts.Tag = "deislabs/kubekahn:latest"
+               opts.File = "porter.yaml"

-               // cnab.go#defaultBundleFiles uses `os.Getwd()`
-               wd, _ := os.Getwd()
-               p.TestConfig.TestContext.AddTestFileContents([]byte(""), path.Join(wd, "porter.yaml"))
+               p.TestConfig.TestContext.AddTestFileContents([]byte(""), "porter.yaml")
                err := opts.Validate(nil, p.Porter)
                require.NoError(t, err, "Validate failed")
-               p.TestConfig.TestContext.FileSystem.Remove(path.Join(wd, "porter.yaml"))
        })
 }
```

We don't necessarily need to go with the test mods above, but we should make sure that this test fails without the actual code changes elsewhere in this PR.

dev-drprasad: > Surprisingly, I couldn't get this test to fail when running against the current code on main

My bad! Thanks for catching this. Test case indeed failing in my local with `main` branch. But the reason was something else and I thought test case working as expected 🤦🏻‍♂️

`opts.File = ''porter.yaml"` simulates explicitly passing of `--file` arg. I wanted to test default behaviour without passing `--file` arg

pkg/porter/lifecycle_test.go
```diff
@@ -160,6 +162,25 @@ func TestBundleLifecycleOpts_ToActionArgs(t *testing.T) {
...
+		opts := BundleLifecycleOpts{}
+		opts.Tag = "deislabs/kubekahn:latest"
+
+		wd, _ := os.Getwd()
+		// Below line makes cnab.go#defaultBundleFiles#manifestExists true
+		p.TestConfig.TestContext.AddTestFileContents([]byte(""), path.Join(wd, config.Name))
+		// When execution reach to `readFromFile`, manifest file path will be lost.
+		// So, had to empty default manifest content also
+		// Below line makes manifest.go#readFromFile#exists true
+		p.TestConfig.TestContext.AddTestFileContents([]byte(""), config.Name)
```
dev-drprasad, line 177: I couldn't make this test fail with `main` branch with just one empty manifest file. 

`defaultBundleFiles#manifestExists` will be `true` only if I use `path.Join(wd....` and we need it to execute bundle default logic.
But `readFromFile` func gets only filename (not complete path) and reads  manifest file from root directory. So, I had to use root manifest file also

vdice: Ah I see.  The in-line comments help!  Thanks for updating.

PR #1205, opened 2020-08-11
fix(runtime-manifest.go): fix resolution for step- and bundle-level outputs

vdice:
# What does this change
* Fixes resolution for outputs that are both step- and bundle-level and referenced (templated) in the same step that they are produced

This one was subtle and took me a bit to understand.  When an output is both a step- and bundle-level output, if it is referenced (templated) in the same step, the previous mode of resolution (going straight to lookup the wiring parameter env var) always resulted in `null`, because that wiring parameter env var will only be populated with a value for _subsequent_ actions.

# What issue does it fix
N/A, found while testing.

# Notes for the reviewer

# Checklist
- [x] Unit Tests (integration)
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/examples/exec-outputs/cluster.sh
+++ b/examples/exec-outputs/cluster.sh
@@ -68,5 +68,12 @@ uninstall() {
   echo 'Uninstalling Cluster...'
 }
 
+assert-admin() {
+  if [ "$1" != "sally" ]; then
+    echo "$1 is not an admin."
+    exit 1
+  fi
+}
+
 # Call the requested function and pass the arguments as-is
 "$@"
--- a/examples/exec-outputs/cluster.sh
+++ b/examples/exec-outputs/cluster.sh
@@ -68,12 +68,5 @@ uninstall() {
   echo 'Uninstalling Cluster...'
 }
 
-assert-admin() {
-  if [ "$1" != "sally" ]; then
-    echo "$1 is not an admin."
-    exit 1
-  fi
-}
-
 # Call the requested function and pass the arguments as-is
 "$@"
--- a/pkg/runtime/runtime-manifest.go
+++ b/pkg/runtime/runtime-manifest.go
@@ -329,6 +329,12 @@ func (m *RuntimeManifest) buildSourceData() (map[string]interface{}, error) {
 					continue
 				}
 
+				// A bundle-level output may also be a step-level output
+				// If already set, do not override
+				if val, exists := m.outputs[ps.OutputName]; exists && val != "" {
+					continue
+				}
+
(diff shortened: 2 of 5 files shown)


pkg/runtime/runtime-manifest.go
```diff
@@ -100,7 +100,15 @@ func resolveCredential(cd manifest.CredentialDefinition) (string, error) {
 }
 
 func (m *RuntimeManifest) resolveBundleOutput(def manifest.OutputDefinition) (string, error) {
-	// Get the output's value from the injected parameter source
+	// The output value may be tracked on the runtime manifest for use during the same action,
```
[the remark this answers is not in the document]
vdice: Ah, I also think I prefer _a_ and can't currently think of an edge case... will work on this now.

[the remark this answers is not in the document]
vdice: Ok.  Actually, option b seemed the most straightforward approach when I revisited this.  We do currently set step outputs last (https://github.com/deislabs/porter/blob/main/pkg/runtime/runtime.go#L98-L112).  In this case, we'll have set the value of a step output (that also happens to be a bundle output) during a previous step and so when we go to build the source data for a subsequent step (during the same action), we can defer to this value instead of attempting to override with the bundle-level wiring value (which will be empty since we're still executing the same action).  WDYT?

pkg/runtime/runtime-manifest.go
```diff
@@ -323,6 +323,12 @@ func (m *RuntimeManifest) buildSourceData() (map[string]interface{}, error) {
 					continue
 				}
 
+				// A bundle-level output may also be a step-level output
```
carolynvs, line 326: Oh nice! This is a very small change. 👏

examples/exec-outputs/cluster.sh
```diff
@@ -68,5 +68,12 @@ uninstall() {
   echo 'Uninstalling Cluster...'
 }
 
+assert-admin() {
```
carolynvs, line 71: Instead of putting the porter integration test in the bundle, which isn't how we normally do example bundles, how about updating this test and put in any asserts that you want?

https://github.com/deislabs/porter/blob/1cff9e4713002cec674ee5aad516ad97832825aa/tests/outputs_test.go#L98

That way the example continues to demonstrate bundle techniques to people and the tests reside all in the same test file.

vdice: Good point.  As we discussed offline, we still can't currently inspect docker driver output in the int tests, so moving the modifications to the bundle in `tests/testdata` (as opposed to an example bundle in `examples`) seemed to be a good compromise.

PR #1206, opened 2020-08-11
fix(runtime-manifest): defer to bundle-level output def sensitivity

vdice:
# What does this change
* Updates logic around masking output values.  We still regard step outputs as sensitive by default, but this updates step outputs that are also bundle-level outputs to inherit the bundle-level output setting.  (Note that bundle-level outputs are _not_ sensitive by default.)

# What issue does it fix
N/A
# Notes for the reviewer

# Checklist
- [x] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)


vdice:
Thanks @carolynvs ; indeed, your recent updates have made this change set much simpler!  Rebased and ready for another look when convenient.

PR #1211, opened 2020-08-14
fix(kubernetes): trim quotes from output before returning

vdice:
# What does this change
* Adds logic to remove quotes from a Kubernetes resource output, if applicable, before returning

Proposed fix for https://github.com/deislabs/porter/issues/1210, assuming it's agreed it is an issue we wish to fix.

# What issue does it fix
Closes https://github.com/deislabs/porter/issues/1210

# Notes for the reviewer
Do we want to trim both `"` and `'` as I have it currently?  Or limit to only `'`?

Also, please let me know if I'm missing a way to do this either via jsonpath or some other, simpler approach.

# Checklist
- [x] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/pkg/kubernetes/helpers.go
+++ b/pkg/kubernetes/helpers.go
@@ -20,3 +20,13 @@ func NewTestMixin(t *testing.T) *TestMixin {
 		TestContext: c,
 	}
 }
+
+// trimQuotes receives data and returns it with double or single wrapping quotes removed
+func trimQuotes(data []byte) []byte {
+	if len(data) >= 2 {
+		if c := data[len(data)-1]; data[0] == c && (c == '"' || c == '\'') {
+			return data[1 : len(data)-1]
+		}
+	}
+	return data
+}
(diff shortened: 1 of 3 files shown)


carolynvs:
I just dug into the mixin and I believe the problem is caused by how we are calling kubectl and passing the jsonpath. 

When I edited the mixin and removed the single quotes around the jsonpath, it fixed the problem. 

https://github.com/deislabs/porter/blob/05bffc73397ba85abc9883cf27f78d85b6f1dea4/pkg/kubernetes/kubernetes.go#L97

```golang
args = append(args, fmt.Sprintf("-o=jsonpath=%s", jsonPath))
```

I don't think the single quotes are necessary, and they are being wrapped around the value extracted by the jsonpath?

vdice:
👏 Nice find!  Agreed, I think we can just remove the single quotes from that linked line.

vdice:
Closing in favor of simpler approach suggested by @carolynvs : https://github.com/deislabs/porter/pull/1227

Issue #1203, opened 2020-08-10
Make migrations explicit


carolynvs:
_Whelp I wrote this up once already and it got lost in the ether. This one is going to be shorter..._ 🤦‍♀️ 

Migrating Porter data needs to be an explicit choice that can be delayed until the user is prepared to backup data and take on the additional risk.

```console
$ porter list
Useful message that says that we can't run the command because porter is new and their data is old. Link to release notes or page that explains it in more detail. Tell the user to backup their data then run 

    porter storage migrate
```

```console
$ porter storage migrate
Credentials schema is up-to-date
Storage schema is out-of-date (want: WANT got: GOT)
Migrating claims storage...
Saved migration logs to ~/.porter/migrate-TIMESTAMP.log
```

PR #1218, opened 2020-08-18
Migrate credential and parameter sets to have schemaVersion

carolynvs:
# What does this change
Migrate credentials during porter storage migrate:

* Track credential-set schema in schema.json under the "credentials",
e.g.   "credentials": "cnab-credentialsets-1.0.0-DRAFT-b6c701f"
* A migration will upgrade everything that is out of date, if claims
fail, it will keep migrating credentials then report all failures.
* Removed unused credential set migration from yaml to json. We haven't
used it for months and it isn't possible anymore with the filesystem
store / crud.Store interface.
* Renamed mybuns testdata to mybun for consistency and decency.
* Removed extra exclamation points in the migration logs. Now that we
are logging to a file, and using a dedicated command, grabbing people's
attention isn't as critical.
* Tweak when we log Migration Complete/Failed and the log file path so
that it is captured appropriately in the console and in the log file
without duplication.
* Aggregate errors from migration so that we continue on error but
report that an error occurred during migration.
* Implement interface methods in storage.Manager so that our custom
HandleConnect is called exactly when we want it, instead of the
HandleConnect of our datastore.

# What issue does it fix
Follow-up to #1203 

# Notes for the reviewer
* ~I'm going to add parameter migration to this PR. Working on that now.~
* ~Need to write schemaVersion in our credentialSet and ParameterSet when we write them.~
* 🚨 I don't think this can merge until I have a spec version for parameter sets to reference. Also I'm running into circular references because parameter sets aren't in cnab-go. Should I wait for that to get moved? Or refactor to remove the problem?

# Checklist
- [x] Unit Tests
- [ ] Documentation - NA
- [ ] Schema (porter.yaml) - NA


carolynvs:
@vdice I was able to get it working using the untyped crud.Store interface. I'll come back and update it after the parameter store moves to cnab-go and use that when it's ready.

Issue #1136, opened 2020-07-14
Download the kubectl client binary dynamically at runtime


carolynvs:
**Is your feature request related to a problem? Please describe.**
I don't always know the version of the kubernetes cluster that I am connecting to at build time.

**Describe the solution you'd like**
I want the kubernetes mixin to dynamically download the kubectl client binary at runtime if the versions don't match (controlled by a config flag). The helm mixin does this already minus the configuration to bypass this functionality.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

MChorfa:
@carolynvs I can help with this one

PR #1221, opened 2020-08-21
Added Client Version for Kubernetes Mixin

### LINKED ISSUE #1135: Configure kubernetes mixin version
K8s bumps their version every 3 months, we don't. 😀 

I want to be able to specify the version of kubectl to build into the bundle just like we can with the helm mixin.

```yaml
mixins:
- kubernetes:
    clientVersion: v1.17.0
```

See #1136 for having the client version selected dynamically at runtime.

**Additional context**
Here is the [build code](https://github.com/deislabs/porter-helm/blob/main/pkg/helm/build.go) that helm uses to configure and download the specified version of helm.

donmstewart:
Changes to the Kubernetes Mixin in order the user can pass in the version of kubectl to install which then downloads kubectl in the build action.

# What issue does it fix
Closes #1135


vdice:
Added tests look great @donmstewart - thank you.

One quibble: I don't think https://github.com/deislabs/porter/issues/1136 should be listed as being closed by this PR, as it appears to relate to runtime version matching as opposed to build-time version fetching, which is what this PR accomplishes.

donmstewart:
I don't think thats a quibble I think thats a perfectly valid point. Can we mark this just for #1135 for now and I'll do #1136 as a new PR after this is in?

PR #1227, opened 2020-08-24
fix(kubernetes.go): remove jsonpath quotes

### LINKED ISSUE #1210: Kubernetes resource outputs are single-quoted
## Describe the bug
Bundle outputs can come from Kubernetes resources, as provided by the `kubernetes` mixin and now the `porter-helm` mixin (https://github.com/deislabs/porter-helm/pull/70).  See an example in the [kubernetes example bundle](https://github.com/deislabs/porter/blob/main/examples/kubernetes/porter.yaml#L20-L24).

However, output values always appear to be wrapped in single quotes.  This can present issues when the output values are intended to be used in subsequent steps and actions.

## To Reproduce
1. From the root of this Porter repo, `cd examples/kubernetes`
2. Run `porter generate credentials` to generate a credential set (requires a kubeconfig)
3. Run `porter install -c kubernetes`
4. Check out the output via `porter installation outputs show IP_ADDRESS`:

```
 $ porter installation outputs show IP_ADDRESS
'10.104.11.4'
```

Whereas, if I grab the value via `kubectl` directly, there are no single quotes:

```
 $ kubectl get service nginx-deployment -o=jsonpath='{.spec.clusterIP}' && echo
10.104.11.4
```

## Expected behavior
I would expect the outputs to not be wrapped in single quotes.

The fix, in my opinion, would therefore be to ensure output values are not wrapped in single quotes, both in the Kubernetes and Helm mixins.

## Version
The latest versions of the kubernetes and porter-helm mixins built from source (`main` branch).

```
 $ porter version
porter v0.27.2-35-gb5e3e34a (b5e3e34a)
```

vdice:
# What does this change
* Removes wrapping single quotes from the `-o jsonpath` flag when retrieving outputs from Kubernetes

All credit for the fix goes to @carolynvs 🙇 (see https://github.com/deislabs/porter/pull/1211#issuecomment-679299614)

# What issue does it fix
Closes #1210 

# Notes for the reviewer

* I opted for an integration test since we have the pre-existing kubernetes example bundle... which brings the added benefit of testing the example... but I can understand if we don't wish to introduce this overhead.

* Some auto-formatting changes came in to `kubernetes.go`, I guess from my VS Code IDE/golang extension(s)... if we feel strongly, I can see about keeping those out.

# Checklist
- [x] ~~Unit Tests~~ Integration test
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/tests/outputs_test.go
+++ b/tests/outputs_test.go
@@ -9,7 +9,6 @@ import (
 
 	"get.porter.sh/porter/pkg/porter"
 	"get.porter.sh/porter/pkg/printer"
-	"github.com/cnabio/cnab-go/claim"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
@@ -163,9 +162,4 @@ func TestKubernetesMixinOutputs(t *testing.T) {
 
 	err = p.UninstallBundle(uninstallOptions)
 	require.NoError(p.T(), err, "uninstall of bundle failed")
-
-	// Verify that the installation is deleted
-	i, err := p.Claims.ReadInstallation("kubernetes")
-	require.EqualError(p.T(), err, "Installation does not exist")
-	require.Equal(p.T(), claim.Installation{}, i)
 }
--- a/tests/outputs_test.go
+++ b/tests/outputs_test.go
@@ -9,6 +9,7 @@ import (
 
 	"get.porter.sh/porter/pkg/porter"
 	"get.porter.sh/porter/pkg/printer"
+	"github.com/cnabio/cnab-go/claim"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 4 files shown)


tests/outputs_test.go
```diff
@@ -127,3 +128,44 @@ func TestStepLevelAndBundleLevelOutputs(t *testing.T) {
...
+	// Verify the output value has no wrapping single quotes
+	output, err := p.ReadBundleOutput("IP_ADDRESS", "kubernetes")
+	require.NoError(p.T(), err, "could not read bundle output")
+	require.NotEmpty(p.T(), output, "expected the output value to not be empty")
+	require.NotContains(p.T(), "'", output, "did not expect the output value to contain any single quote characters")
+
+	// Uninstall and delete installation
+	uninstallOptions := porter.UninstallOptions{}
+	uninstallOptions.CredentialIdentifiers = []string{"ci"}
+	uninstallOptions.Delete = true
```
carolynvs, line 159: Best feature ever 😀

tests/outputs_test.go
```diff
@@ -127,3 +128,44 @@ func TestStepLevelAndBundleLevelOutputs(t *testing.T) {
...
+
+	err = uninstallOptions.Validate([]string{}, p.Porter)
+	require.NoError(p.T(), err, "validation of uninstall opts for bundle failed")
+
+	err = p.UninstallBundle(uninstallOptions)
+	require.NoError(p.T(), err, "uninstall of bundle failed")
+
+	// Verify that the installation is deleted
```
carolynvs, line 167: Since we aren't testing that feature here, I don't think we need to assert that uninstall worked as intended. We can just use it and expect it to work. Let the tests that are for that feature do the deeper asserts that check that it is working properly and scope our test to only assert on relevant edge cases for this test (that there aren't the single quotes in the k8s output).

vdice:
Shoot, I broke the 20m timeout w/ the added test 😭 .  Do we bump the timeout or nix the test?

carolynvs:
I would like to make an effort in the next 2 months to reduce how many integration tests we have in Porter's integration test run and improve our running time. Don't know about you but it makes it really hard to stay focused when running the tests take so long.

So how about we bump the timeout for now and then this will be one of the tests we move out of Porter's standard integration test run. I'm thinking the k8s mixin should be a standalone mixin now and only exec should live in porter.

vdice:
>  Don't know about you but it makes it really hard to stay focused when running the tests take so long.

Indeed, same.  Also in support of k8s migrating out of core (and with it any/all int tests).

I'll try bumping to 30m for now.

PR #1238, opened 2020-08-31
docs(dependencies.md): mention ordering

vdice:
# What does this change
* Updates the dependency doc to mention required fields and ordering (now that https://github.com/deislabs/porter/pull/1170 is in 🎉 )

# What issue does it fix
n/a

# Notes for the reviewer
n/a
# Checklist
- [ ] Unit Tests
- [x] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/docs/content/dependencies.md
+++ b/docs/content/dependencies.md
@@ -15,14 +15,28 @@ Here is a [full example][example] of a Porter manifest that uses dependencies.
 
 ## Define a dependency
 
-In the manifest, add entries for each dependency of your bundle. The key used is a short name for the dependent bundle that
+In the manifest, add entries for each dependency of your bundle. The `name` field takes a short name for the dependent bundle that
 you will use to reference the dependent bundle elsewhere in the bundle. For example you can reference the dependent bundle's
-outputs via `{{ bundle.dependencies.KEY.outputs }}`.
+outputs via `{{ bundle.dependencies.KEY.outputs }}`.  The `tag` field takes the bundle tag of the dependency.  Both `name` and `tag`
+are required fields.
 
 ```yaml
 dependencies:
   - name: mysql
     tag: getporter/mysql:v0.1.3
+```
+
+## Ordering of dependencies
+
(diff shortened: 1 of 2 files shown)


docs/content/dependencies.md
```diff
@@ -15,16 +15,30 @@ Here is a [full example][example] of a Porter manifest that uses dependencies.
 
 ## Define a dependency
 
-In the manifest, add entries for each dependency of your bundle. The key used is a short name for the dependent bundle that
+In the manifest, add entries for each dependency of your bundle. The `name` field takes a short name for the dependent bundle that
 you will use to reference the dependent bundle elsewhere in the bundle. For example you can reference the dependent bundle's
-outputs via `{{ bundle.dependencies.KEY.outputs }}`.
+outputs via `{{ bundle.dependencies.KEY.outputs }}`.  The `tag` field takes the bundle tag of the dependency.  Both `name` and `tag`
```
carolynvs, line 20: Since the dependency now has a name, and we don't need to reference it by a map key, I think it would be helpful to to change how we mention it in our templating suggestion (below). I am only changing it here but I'm sure we use that form (and variations of it, so it would be good to search for `bundle.dependencies.` to check for them, like bundle.dependencies.DEP or ALIAS.

```suggestion
outputs via `{{ bundle.dependencies.NAME.outputs }}`.  The `tag` field takes the bundle tag of the dependency.  Both `name` and `tag`
```

carolynvs (commented):
Looks good, just one suggestion to improve our templating prompt.

docs/content/dependencies.md
vdice: Oh good call, thank you.

vdice:
PTAL when convenient @carolynvs

vdice:
Added an update to the `author-bundles` doc as well.

vdice:
...aaand revisited the dependencies doc to update the last vestiges of the older `key` references.

carolynvs (approved):
🚢 Thanks!

PR #1242, opened 2020-09-03
fix(dependencies.go): verify dep match before checking params

vdice:
# What does this change
* Adds a check to be sure the manifest dep matches the current dep being processed before validating param def and setting param value

# What issue does it fix
Previously, we'd erroneously validate param definitions on the processed dep _against every dep listed in the manifest_ via the for loop, so we'd naturally hit errors.

# Notes for the reviewer

# Checklist
- [ ] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)


MChorfa (approved):

MChorfa:
LGTM!

PR #1246, opened 2020-09-07
WIP: Make docs preview run into docker

MChorfa:
This will allow to run docs preview into docker 

# Checklist
- [ ] Unit Tests
- [x] Documentation
- [ ] Schema (porter.yaml)


Makefile
```diff
@@ -104,7 +104,13 @@ docs-gen:
...
 
 docs-preview:
-	hugo serve --source docs/
+	make docs-stop-preview
+	@export DOCKER_BUILDKIT=1
+	@docker build -f "Dockerfile.Docs" -t porter-docs:latest "."
+	@docker run -d  -p 1313:80/tcp --name porter-docs porter-docs:latest
+	@python -m webbrowser -t "http://127.0.0.1:1313/docs/"
+docs-stop-preview:
+	@docker rm porter-docs > /dev/null 2>&1 &
```
carolynvs, line 113: I have a couple comments on the makefile targets here that I'm combining. I ended up trying it out locally and getting it to work.

* Instead of calling `make docs-stop-preview` inline, let's have it be a prerequisite
* Tweaked docs-stop-preview so that it didn't run the background and is always successful (in case someone removed the container themselves).
* The export won't work because each line in make is in its own shell session. The next line in the target won't have it defined.
* I added a dockerignore file so that we only send the docs directory as the docker context
* Use open instead of python so that we don't introduce another dependency.

What do you think?

```makefile
docs-preview: docs-stop-preview
	@DOCKER_BUILDKIT=1 docker build -f "Dockerfile.Docs" -t porter-docs:latest "."
	@docker run -d -p 1313:80/tcp --name porter-docs porter-docs:latest
	@open http://localhost:1313/docs/

docs-stop-preview:
	@docker rm -f porter-docs &> /dev/null || true
```

**Dockerfile.Docs.dockerignore**

```dockerignore
**
!/docs/**
```

carolynvs (changes requested):
Thanks for improving this! It will make it much easier for people to contribute to our documentation.

Makefile
MChorfa: Thank you @carolynvs for your comments and suggestions :) 
We have one annoying thing is that when we try to access a page that point directly to the base url. somehow the port number gets strip out.
For example: 
I need to access the mixins page. the generated URL is  http://localhost:1313/mixins, but somehow the port number disappears. Will investigate

carolynvs: I see now. The problem is that you have changed how the files are served, with nginx instead of hugo's webserver (or in production Netlify which also handles / redirects properly). Some of the links do not have a trailing `/` and it is not being redirected the same and the redirect is losing the porter number. The links that do have a trailing `/` are working fine.

This is a problem, but another problem I realize is that we aren't getting the live updates anymore from hugo. Normally the hugo server watches for changes to the files and will live refresh the web browser so that you can edit the site's files and not have to re-run `make docs-preview`.

Have you already tried running `hugo serve` in the container instead of using nginx? If not, that would be the preferable solution. What is here currently initially looked right but upon deeper testing, I realize is too different from how we normally use hugo to work on the site, and render in production.

MChorfa: Yes you are right @carolynvs, thank you for the pointers 👍 

A couple of changes: 
- Now we have a Hugo base image and we can pass `Hugo` commands when running the docker container 
- The files are now watched so upon changes the hot reload will reflect the changes. 
- Moved the `Dockerfile` to the docs folder. I think it makes more sense to have it there.
- Fixed the trailing that could cause issues with URL redirection.

⚠️  One small caveat with safari when the page open we have to refresh the page to display the content, but with the other browsers [Chrome, Firefox, Brave]  it works just fine

docs/config.toml
```diff
@@ -22,20 +22,20 @@ defaultContentLanguage = "en"
   weight = 2
   [[menu.main]]
     name = "Install Porter"
-    url = "/install"
+    url = "/install/"
```
carolynvs, line 25: Since we are using hugo to preview locally, this change is not needed. I had mentioned the trailing slash as the reason why the redirects were working differently with nginx vs. hugo but it's not something that we should change. Both hugo and netlify append a slash automatically for us. 

If you do want to clean up the links to make things consistent, let's do that in another PR as it doesn't impact this one.

Makefile
```diff
@@ -103,14 +103,14 @@ docs:
...
-docs-preview:
-	make docs-stop-preview
-	@export DOCKER_BUILDKIT=1
-	@docker build -f "Dockerfile.Docs" -t porter-docs:latest "."
-	@docker run -d  -p 1313:80/tcp --name porter-docs porter-docs:latest
-	@python -m webbrowser -t "http://127.0.0.1:1313/docs/"
+docs-preview: docs-stop-preview
+	@DOCKER_BUILDKIT=1 docker build -f "$$PWD/docs/Dockerfile" -t porter-docs:latest "."
+	@docker run -d -v $$PWD/docs:/docs -p 1313:1313 -u hugo --name porter-docs \
+	porter-docs:latest hugo server --noHTTPCache --watch --bind=0.0.0.0
```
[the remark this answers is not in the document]
MChorfa: Yes, this is a very well done docker image 🚀 . I tested it and works! Thank you!

docs/config.toml
MChorfa: You are right. File reversed. Will revisit the trailing issue within another ticket

docs/content/plugins/_index.md
```diff
@@ -16,6 +16,6 @@ plugin binary may contain multiple implementations.
 See the [Search Guide][search-guide] on how to search for available plugins and/or
 add your own to the list.
 
-[vs]: /mixins-vs-plugins/
-[types]: /plugins/types/
-[search-guide]: /package-search/
+[vs]: (/mixins-vs-plugins/)
```
carolynvs, line 19: Would you please revert the changes to this file?

carolynvs (commented):
Just one more thing, then this is ready to go. 

Let's update the [CONTRIBUTING.md](https://github.com/deislabs/porter/blob/main/CONTRIBUTING.md#preview-documentation) so that people don't think that they need to install hugo or manually open the page anymore. We should probably say that they should refresh the page if it doesn't show up the first time. Sometimes it loads too fast, before hugo is ready and I need to refresh it the first time. Or maybe it just needs a sleep, not sure which is better, the doc or a sleep or both? 🤔

carolynvs (approved):
Looks great, so much more simple for people to contribute. Thank you! 💖

Issue #1162, opened 2020-07-24
Rename, move, and hide Porter's generated Dockerfile


gaurimadhok:
**Is your feature request related to a problem? Please describe.**
The dockerfile that Porter generates and uses to build the invocation image is named Dockerfile and is in the directory with the rest of the bundle. If someone is trying to use Docker within their bundle and wants to make their own dockerfile, they will have to name it something else. 

**Describe the solution you'd like**
A potential solution suggested by @carolynvs is to rename the generated Dockerfile to Dockerfile.porter or .Dockerfile.porter and put it in the .cnab directory. The .cnab directory is already hidden to prevent confusion,  so adding the Dockerfile within that directory to hide it will also help limit confusion. When building the invocation image, the docker build command could be adapted by simply adding .cnab/ before the name of the dockerfile and changing the name of the file.

carolynvs:
Yeah I think if we can make this command work essentially (but automated through porter, we don't really run a raw docker command), then it would be less confusing for people:

```
docker build -t INVOCATIONIMAGE -f .cnab/Dockerfile .
```

If we can get it to work with the Dockerfile moved under .cnab, then we wouldn't even need to rename it. 👍

carolynvs:
Heads up when this is implemented, we will need to tweak the exclude fix with the updated Dockerfile/dockerignore names per https://github.com/getporter/porter/pull/1263

PR #1263, opened 2020-09-14
fix(docker.go): don't exclude Dockerfile and .dockerignore from build context

vdice:
# What does this change
* Uses the docker cli [TrimBuildFilesFromExcludes](https://github.com/docker/cli/blob/master/cli/command/image/build/dockerignore.go#L28-L39) helper to be sure _not_ to exclude the Dockerfile and .dockerignore files when building the installer image.  Though, as noted in the helper docs, both will be removed from the final context (if both exist in the user-supplied .dockerignore).

# What issue does it fix
@squillace was hitting the following when `Dockerfile` was added to a bundle's `.dockerignore` file:

```
$ porter build
...
Error: unable to build CNAB invocation image: Error response from daemon: Cannot locate specified Dockerfile: Dockerfile
```

# Notes for the reviewer

# Checklist
- [x] ~Unit~ Integration Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)


carolynvs:
Are these files (Dockerfile, .dockerignore) excluded by default? Or were they added by the user? I think I'm missing some context into the troubleshooting and what the root cause was here.

If the files were called Dockerfile.foo and Dockerfile.foo.dockerignore would they have been excluded?

squillace:
it's very common -- and a practice I use -- to reject all files in the .dockerignore by default and then add IN the app artifacts you really need. BUT... it's easy to NOT know what files the container MUST have added in, so in this case, if you have a .dockerignore file but have forgotten to enable Dockerfile you get this error, but you have no idea why it happened. 

It's a basic ux thing that an expert like you would not do (because you'd always know that docker needs the Dockerfile inside the context to build it). :-) Does that help?

vdice:
This fix just ensures that Porter can build the installer/invocation image even if the .dockerignore file in the bundle directory has an entry matching the name of the Dockerfile that Porter uses/needs to build from.  So, the .dockerignore might look like:

```
stuff-i-don't-want-in-the-installer-image
Dockerfile # which I also don't want in the installer image
.dockerignore # ditto
```

And, indeed, all of the files above _won't_ end up in the final image, but with this fix we ensure that `Dockerfile` and `.dockerignore` are accessible to the daemon performing the build.

> If the files were called Dockerfile.foo and Dockerfile.foo.dockerignore would they have been excluded?

Yes.  The helper matches ".dockerignore" [specifically](https://github.com/docker/cli/blob/master/cli/command/image/build/dockerignore.go#L32) and the Dockerfile string config'd on Porter's [build options](https://github.com/deislabs/porter/blob/main/pkg/build/provider/docker.go#L43) (currently "Dockerfile") -- all other files/filenames present in the user-supplied .dockerignore _will_ be excluded in all contexts.

carolynvs:
Thanks for clarifying. So when we implement https://github.com/getporter/porter/issues/1162, we will need to update this fix if we rename Dockerfile/.dockerfileignore. Gotcha. 👍

carolynvs (approved):
LGTM

PR #1290, opened 2020-09-21
Template installation name and custom metadata

### LINKED ISSUE #1035: Add the instance name to the avilable template variables
It would be useful when I'm installing a bundle if I could access the instance name of the bundle, not just the published name of the bundle. For example, I could use it to default the release name of my helm releases to something more unique.

Here is what I would like to do, note that this relies on being able to use templates throughout the bundle, not just in the action.

```yaml
parameters:
  - name: release
    description: helm release name
    type: string
    default: "{{ installation.name }}"
```

I am also assuming that we have already changed the term `instance` to `installation` at this point. I'd like to be consistent with whatever we decide to do with that.

carolynvs:
# What does this change
* Template installation name
* Template custom metadata

# What issue does it fix
Closes #1035

# Notes for the reviewer
We need to redo the wiring page. It really is a page about how to do porter templating but I don't think we have a page that documents that? I think the page should be renamed and tweaked, removing how to define parameters, etc moving that into authoring bundles leaving just how to do templating.

# Checklist
- [x] Unit Tests
- [x] Documentation
- [ ] Schema (porter.yaml) - Not impacted


vdice (approved):
🎉

PR #1295, opened 2020-09-22
fix custom metadata serialisation

### LINKED ISSUE #1292: custom entries with object or array values in porter.yaml cause build failures
## Describe the bug
Adding custom data to a porter manifest will cause porter build to fail unless the value for the custom key is a scalar value. This is because yaml unmarshall will de-serialise the value into map[interface]interface{} type which subsequently cannot be serialised to json based bundle.

## To Reproduce
1. Use porter.yaml from https://gist.github.com/simongdavies/f48db979bbfa99aac57a30e920fe51db
1. Run `porter build`
1. Result: error: error writing to .cnab/bundle.json: json: unsupported type: map[interface {}]interface {}

## Porter Command and Output
```
$porter build
Copying porter runtime ===>
Copying mixins ===>
Copying mixin exec ===>

Writing Dockerfile =======>

Error: error writing to .cnab/bundle.json: json: unsupported type: map[interface {}]interface {}
```

## Version

porter v0.27.2-58-g48e94401 (48e94401)

simongdavies:
# What does this change
 Updates marshalling of customdata YAML in order that the data can be serialised to JSON. See https://github.com/go-yaml/yaml/issues/139

# What issue does it fix
Closes #1292

# Checklist
- [X] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/pkg/cnab/config-adapter/adapter_test.go
+++ b/pkg/cnab/config-adapter/adapter_test.go
@@ -675,6 +675,39 @@ func TestManifestConverter_generateCustomMetadata(t *testing.T) {
 	require.NoError(t, err, "ToBundle failed")
 	assert.Len(t, bun.Custom, 2)
 
-	fooCustomData := bun.Custom["foo"]
-	assert.Equal(t, "bar", fooCustomData)
+	val, ok := bun.Custom["foo"].(map[string]interface{})
+	require.True(t, ok, "Cannot cast foo value to map[string]interface{}")
+
+	val1, ok := val["test1"].(bool)
+	require.True(t, ok, "Cannot cast test1 value to bool")
+	require.True(t, val1, "test1 value is unexpected")
+
+	val2, ok := val["test2"].(int)
+	require.True(t, ok, "Cannot cast test2 value to int")
+	require.Equal(t, 1, val2, "test2 value is unexpected")
+
+	val3, ok := val["test3"].(string)
+	require.True(t, ok, "Cannot cast test3 value to string")
+	require.Equal(t, "value", val3, "test3 value is unexpected")
+
+	val4, ok := val["test4"].([]interface{})
+	require.True(t, ok, "Cannot cast test4 value to interface{} array")
+	val5, ok := val4[0].(string)
+	require.True(t, ok, "Cannot cast test4[0] value to string")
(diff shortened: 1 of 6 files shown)


carolynvs:
Hey Simon, the original bug that your created doesn't reproduce for me on porter `v0.29.0-31-g6ce58563`. Please take a look and if you can get it to error out, update the PR with a failing regression test for the creating the bundle.json.

A good spot for the test would be in `TestManifestConverter_generateCustomMetadata` https://github.com/getporter/porter/blob/main/pkg/cnab/config-adapter/adapter_test.go and then use code like below to write the file out.

```go
        f, err := p.Config.FileSystem.OpenFile(build.LOCAL_BUNDLE, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644)
	defer f.Close()
	if err != nil {
		...
	}
	_, err = b.WriteTo(f)
        ...
```

carolynvs (changes requested):
These aren't the right places to make the changes because we don't serialize the bundle.json from the Manifest directly. I've left a comment on the PR with further direction.

simongdavies:
@carolynvs I've made the changes, I left the tests in the manifest test as I modified a test that was already present

pkg/cnab/config-adapter/adapter_test.go
```diff
@@ -675,6 +675,39 @@ func TestManifestConverter_generateCustomMetadata(t *testing.T) {
 	require.NoError(t, err, "ToBundle failed")
 	assert.Len(t, bun.Custom, 2)
 
-	fooCustomData := bun.Custom["foo"]
-	assert.Equal(t, "bar", fooCustomData)
+	val, ok := bun.Custom["foo"].(map[string]interface{})
```
carolynvs, line 678: This test should use the original code I pointed you to for Marshaling a bundle to json, not repeat the tests from the manifest test suite. What we need to ensure is that after your change that we have a regression test that checks that we can create bundle.json files that contain nested custom metadata structures.

So let's write the json, compare it to a known string in a testdata file and this should be all set! 👍

carolynvs (changes requested):

carolynvs (approved):
Looks great! Thanks for identifying and fixing this. 👍

PR #1298, opened 2020-09-23
Print the captured output for jsonPath outputs

carolynvs:
# What does this change
In debug mode, print the full captured document and the query we are using to match output so that the user can troubleshoot when the output's format or contents do not contain the expected output value.

# What issue does it fix
Makes @squillace happy. I will update the az mixin to use this after it is merged.

# Notes for the reviewer
N/A

# Checklist
- [x] Unit Tests
- [x] Documentation
- [ ] Schema (porter.yaml)


vdice (approved):
Nice!

PR #1300, opened 2020-09-23
Add blog framework

### LINKED ISSUE #1126: Add a blog
Right now we blog on deislabs.io, but at some point we should make a project blog on porter.sh and copy our existing blog posts over there. Then define a way for anyone to submit blog posts, like [SMI does](https://github.com/servicemeshinterface/smi-spec.io#how-to-add-a-blog-post).

carolynvs:
# What does this change
* ~There is a bug in Hugo so the frontmatter for the date is 'postDate' instead of 'date'.~
* The CNCF post is draft that won't publish on the live site. I will write it in a follow-up.
* Moved old posts over to the site

https://deploy-preview-1300--porter.netlify.app/blog/

# What issue does it fix
Closes #1126 

# Notes for the reviewer
I am not a designer but we can iterate later. This just needs to be functional and have the content up.

# Checklist
- [ ] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)


carolynvs:
Posts can specify their own twitter card, otherwise it falls back to the default

Screen Shot 2020-09-24 at 12 07 24 PM


Screen Shot 2020-09-24 at 12 06 39 PM

iennae (approved):
LGTM! 

I do see some weirdness like on https://deploy-preview-1300--porter.netlify.app/blog/using-docker-in-bundles/ why is docker cat off in that one side but I think it's something to do with the content just pushing over as on other pages the content goes past the date. As you mentioned, we can iterate on the design/flow though.

carolynvs:
On that one post there is an image inside the top of the post and the text is wrapping around it with `align="right"`.

carolynvs:
I have updated the blog post layout so that it uses the same logo in the header as the other pages.

PR #1306, opened 2020-09-30
Ensure content digest exists when pulling bundle via tag

### LINKED ISSUE #1305: Ensure content digest is populated for all published bundles
* When you run a bundle locally, we should allow the content digest to not be set
* When it's a bundle run from --tag, it MUST be set, we will error out otherwise
* Ensure that the digest is always set when porter publishes a bundle

vdice:
# What does this change
* adds logic to fail bundle pull (for main and any dependency bundle(s)) if a given bundle's invocation image is missing a content digest value

# What issue does it fix
Closes #1305 

Note that the other items mentioned in #1305 are already present in the code base, namely:

- When you run a bundle locally, we should allow the content digest to not be set
- Ensure that the digest is always set when porter publishes a bundle (Porter does so [here](https://github.com/getporter/porter/blob/main/pkg/porter/publish.go#L305) and [here](https://github.com/getporter/porter/blob/main/pkg/porter/publish.go#L252))

# Notes for the reviewer
n/a
# Checklist
- [x] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)


pkg/porter/dependencies.go
```diff
@@ -181,6 +181,11 @@ func (e *dependencyExecutioner) prepareDependency(dep *queuedDependency) error {
 	dep.CNABFile = cachedDep.BundlePath
 	dep.RelocationMapping = cachedDep.RelocationFilePath
 
+	invocationImage := cachedDep.Bundle.InvocationImages[0]
```
carolynvs, line 184: Do you think we could check the digest when we pull the bundle, instead of later when we try to run it? That way the logic would be in a single spot instead of split across dependencies and the other code paths?

pkg/porter/dependencies.go
```diff
@@ -181,6 +181,11 @@ func (e *dependencyExecutioner) prepareDependency(dep *queuedDependency) error {
...
+	if invocationImage.Digest == "" {
+		return fmt.Errorf("no content digest is present for dependency image %s", invocationImage.Image)
```
carolynvs, line 186: So if I saw this error, I would have no idea what to do about it or what the problem is. Let's try to explain the security issue and how the invocation image for the bundle doesn't have a digest so we can't verify that it is is the image referenced by the bundle and hasn't been tampered with. Or something like that.

Basically we need to throw the bundle under the bus, say it's invalid and could be compromised.

pkg/porter/testdata/cache/887e7e65e39277f8744bd00278760b06/cnab/bundle.json
```diff
@@ -7,7 +7,7 @@
     "description": "",
     "images": null,
     "invocationImages": [{
-        "digest": "sha256:f858bc025ad34099fe67ebe6152e03b4c91b34cc7a77d1aa10aaf1dc1389c2c2",
+        "contentDigest": "sha256:f858bc025ad34099fe67ebe6152e03b4c91b34cc7a77d1aa10aaf1dc1389c2c2",
```
carolynvs, line 10: Did you see this same mistake anywhere else in our repo? Would you mind taking a quick look?

pkg/porter/testdata/cache/README.md
```diff
@@ -1,5 +1,5 @@
 # Cache Cheat Sheet
 
 * deislabs/kubekahn:1.0 -> 887e7e65e39277f8744bd00278760b06
-* getporter/mysql:v0.1.2 -> 972978d3b715212b783a70fdc3449584
-* getporter/wordpress:v0.1.2 -> bfa9f2b528338f9ac1bc9b8b984c3997
+* getporter/mysql:v0.1.3 -> 4f17468b1ad86cb50cd76d9b8148d249
```
carolynvs, line 4: Thanks for updating my beloved cheat sheet!

pkg/porter/testdata/cache/887e7e65e39277f8744bd00278760b06/cnab/bundle.json
vdice: Ooh, yea, there are def other spots.  In fact, for the `images`/ImageMap part of a manifest, the yaml tag is still `digest`, so there will be quite a few changes.  Do we want to add these as an additional commit to this PR or as a follow-up?

pkg/porter/dependencies.go
vdice: 🚌 💨

pkg/porter/testdata/cache/887e7e65e39277f8744bd00278760b06/cnab/bundle.json
carolynvs: Follow-up please! 🙇‍♀️

pkg/porter/dependencies.go
vdice: Good call, thank you.  Updated.

pkg/porter/dependencies.go
vdice: Error text has been updated to better convey the issue around this scenario.  Thank you!

pkg/porter/testdata/cache/README.md
vdice: I backed out the changes in this directory now that we've placed the logic in PullBundle (and, hence, cached bundles wouldn't hit this case).  For another day!

carolynvs (approved):
Looks great! 💯

Issue #1307, opened 2020-09-30
Update digest field in images map to use contentDigest


vdice:
As discovered in https://github.com/getporter/porter/pull/1306, the digest field for a Mapped Image is still `Digest` with a yaml tag of `digest`.  These need to be updated to `ContentDigest`/`contentDigest` in accordance w/ the spec (https://github.com/cnabio/cnab-spec/blob/master/101-bundle-json.md).

PR #1308, opened 2020-10-01
fix(images): update digest field for a mapped image

vdice:
# What does this change
* Updates the (former) `digest` field for a mapped image to `contentDigest`, per the CNAB core spec
* Also updates the similar field for an invocation image in some test data files

# What issue does it fix
Closes https://github.com/getporter/porter/issues/1307

# Notes for the reviewer

# Checklist
- [x] Unit Tests
- [ ] Documentation
- [x] Schema (porter.yaml)


pkg/runtime/runtime-manifest.go
```diff
@@ -376,7 +376,7 @@ func (m *RuntimeManifest) buildSourceData() (map[string]interface{}, error) {
 		typeOfT := val.Type()
 		for i := 0; i < val.NumField(); i++ {
 			f := val.Field(i)
-			name := strings.ToLower(typeOfT.Field(i).Name)
+			name := toLowerCase(typeOfT.Field(i).Name)
```
carolynvs, line 379: Oh so this "jsonify's" the name. Previously we were making the entire name lower case which would mess up something like "MyThing" turning it into "mything" when it should be "myThing". Do I have that right?

pkg/runtime/runtime-manifest.go
```diff
@@ -518,3 +518,13 @@ func resolveImage(image *manifest.MappedImage, refString string) error {
 	}
 	return nil
 }
+
+// toLowerCase returns a string with the first character lowercased
+func toLowerCase(str string) string {
```
carolynvs, line 523: The name of this function is misleading. Perhaps it should be called `toCamelCase`?

https://medium.com/better-programming/string-case-styles-camel-pascal-snake-and-kebab-case-981407998841

carolynvs (changes requested):
I'm glad you noticed that digest wasn't conformant anymore. Thanks for fixing this!

I just had one question about how the struct name was being turned into a string, then this is good to go.

pkg/runtime/runtime-manifest.go
vdice: Yup, that's correct!

pkg/runtime/runtime-manifest.go
vdice: 👍 `toCamelCase` it shall be.

carolynvs (approved):
LGTM 👍

PR #1322, opened 2020-10-09
Update the landing page for Porter Docs removing the porter CLI commands

iennae:
# What does this change
Cleans up the Porter Docs landing page. 

# What issue does it fix
Documentation update. 

# Checklist

- [x] Documentation

### DIFF
--- a/docs/content/docs.md
+++ b/docs/content/docs.md
@@ -3,59 +3,14 @@ title: Porter Docs
 description: All the magic of Porter explained
 ---
 
-Porter takes the work out of creating CNAB bundles. It provides a declarative authoring 
-experience that lets you to reuse existing bundles, and understands how to translate 
-CNAB actions to Helm, Terraform, Azure, etc.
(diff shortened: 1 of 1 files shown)


docs/content/docs.md
```diff
@@ -3,59 +3,14 @@ title: Porter Docs
...
 
-$ porter build
-created Dockerfile
-created bundle.json
-created cnab/app/run
+Porter has a lot of documentation and we're in the process of reorganizing and updating it. We welcome [contributions](/contribute/) from the community. 
 
-$ porter install
-```
+## Quicklinks
```
carolynvs, line 12: Do you mind if we continue calling this section "Next Steps"? My goal was to have a section like this at the bottom of pages that lead readers to good things to read next. Quicklinks doesn't quite imply the same thing.

iennae: What about having both? As the landing page I was thinking this is like Step 0 so here are a set of areas you could go to versus "Next"

carolynvs: Ah yeah I see how the main docs page is difference. Makes sense. 👍

carolynvs (approved):
LGTM

PR #1337, opened 2020-10-22
Refactor when parameters flags are parsed

### LINKED ISSUE #1320: Using file type parameters from parameterset on published bundle fails with base64 encoding issue
## Describe the bug
When I have a parameter of type 'file', it is correctly available when invoking actions for the bundle in the current working directory, but if invoking the action on a published bundle (porter install --tag ...) it fails with "invalid base64 value"

## To Reproduce
1. Define a parameter of type 'file' and set a destination path.
2. Create a parameterset and specify a value for the parameter
3. Publish the bundle to a registry
4. Run "porter install -p <name of parameterset>"
5. See error

## Expected behavior
I would expect that the file is mounted as is, just like it is if you run it from the local bundle or if you set the parameter using "--param <param_name>=<path_to_file>
It seems to only happen when the parameter is in a parameterset

## Porter Command and Output
```
$ porter publish
$ porter parameters generate temp
$ porter install --tag localhost:5000/temp:v0.1.0 -p temp
installing HELLO...
Error: invalid parameters: cannot use value: dummy: value as parameter dummy_file: invalid base64 value: dummy: value
```
If I base64 encode the content of the file, then it works just like when the bundle is local or if using a --param option instead of parameterset.

## Version
porter v0.29.1 (2e0f3648)

carolynvs:
# What does this change

Parse parameters after bundle is loaded
* ExecuteAction performs any action but uninstall since that operation is performed backwards.
* Move resolution of parameters later, when we convert the options into cnab arguments, so that we have the bundle definition available
* Only validate that --param values are in the proper format, PARAM=VALUE

# What issue does it fix
Fixes #1320

# Notes for the reviewer
I'd appreciate feedback on if this is an improvement or not. I was trying to have us parse parameters once, and move towards reuse with our new BundleAction interface.

# Checklist
- [x] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/cmd/porter/bundle.go
+++ b/cmd/porter/bundle.go
@@ -90,7 +90,7 @@ The lint command is run automatically when you build a bundle. The command is av
 }
 
 func buildBundleInstallCommand(p *porter.Porter) *cobra.Command {
-	opts := porter.InstallOptions{}
+	opts := porter.NewInstallOptions()
 	cmd := &cobra.Command{
 		Use:   "install [INSTALLATION]",
 		Short: "Create a new installation of a bundle",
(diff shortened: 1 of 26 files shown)


vdice:
This looks like a great improvement to me!  Definitely in favor of this refactor.

carolynvs:
Okay, the build is working again so this is ready for review.

vdice (approved):
LGTM!

Issue #1333, opened 2020-10-22
Manifest: Use registry and repository instead of tag


vdice:
Implementation for the corresponding section of https://github.com/getporter/porter/issues/1151:

Currently, the naming and versioning fields in the Porter manifest are: `name`, `version` and `tag`.  The last one of these has heretofore been used to capture the full name of the bundle reference, e.g. `myregistry/myorg/mybuns:v0.1.0`.

We intend to preserve `name` (as the common name of the bundle) and `version` (as the semantic version of the bundle) but replace `tag` with `registry` (and, optionally, `reference`).

* `registry`: this is the OCI registry and org/subdomain for the bundle, e.g. `myregistry.com/myorg`.  Porter will then construct the full bundle reference using this and the values for `name` and `version`, e.g. `myregistry.com/myorg/mybuns:v0.1.0` from the example below.

* `reference`: (OPTIONAL) this can be used to specify the full bundle reference when needed.  When the bundle is published this value will be used rather than the value constructed from the `name`, `version` and `registry` values.

```
name: mybuns
version: 0.1.0
registry: myregistry.com/myorg
# Optional: can use reference to specify the exact bundle reference (correlate to deprecated `tag` field)
# reference: myregistry.com/myorg/mybuns:v0.1.0
```

## AC

- the `tag` field is removed from the Porter manifest
- the `registry` and `reference` fields are added to the Porter manifest
- legacy manifests using the `tag` field continue to be supported
- docs updated/added

vdice:
@carolynvs ah, had forgotten about the dependency declarations in a manifest.  Previously, we still used `tag` to specify the full bundle reference: https://github.com/getporter/porter/blob/main/pkg/manifest/manifest.go#L504

Should we also update this field?  It would be updated to `reference`.

carolynvs:
Yup! I'm so glad you caught that. 💯

PR #1343, opened 2020-10-23
feat(*): manifest versioning updates

vdice:
# What does this change
* Updates the following fields on the Porter manifest:
  - the `tag` field is deprecated (if present, a warning is printed, though its value will still currently be used)
  - the `registry` field is added
  - the `reference` field is added
  - for any/all dependencies:
    - the `tag` field is deprecated (if present, a warning is printed, though its value will still currently be used)
    - the `reference` field takes its place

See https://github.com/getporter/porter/issues/1333 for further details/context.

# What issue does it fix
Closes https://github.com/getporter/porter/issues/1333

# Notes for the reviewer

# Checklist
- [x] Unit Tests
- [x] Documentation
- [x] Schema (porter.yaml)


pkg/manifest/manifest.go
```diff
@@ -756,21 +792,43 @@ func UnmarshalManifest(cxt *context.Context, manifestData []byte) (*Manifest, er
...
-	bundleRef, err := reference.ParseNormalizedNamed(bundleTag)
+// SetInvocationImageAndBundleTag sets the invocation image name and the
+// bundle tag on the manifest per the provided reference or via the
+// registry or repository values on the manifest.
+func (m *Manifest) SetInvocationImageAndBundleTag(ref string) error {
+	// Set bundle tag to repo value (combination of registry and name) by default
+	// Note: the bundle version will be appended as the Docker tag below
+	if ref != "" {
+		m.BundleTag = ref
+	} else if m.Registry != "" && m.BundleTag == "" {
```
carolynvs, line 806: This if/else tree here has two concerns mixed up that makes it a bit hard to follow.

1. Composing reference from registry and bundle name when it's not set.
1. Figuring out if it should use tag or reference.
1. Validating the manifest (when everything is empty).

I'm a bit unsure what would happen if both tag and reference are set which is another validation. Maybe move all the validation into `Manifest.Validate`, default reference to tag when reference is empty, and then have a bit of code here that just deals with default the full reference using registry and name if needed.

I think that would make this easier to maintain and deprecate tag when the time comes.

pkg/manifest/manifest_test.go
```diff
@@ -50,11 +51,13 @@ func TestLoadManifest(t *testing.T) {
...
 	cxt := context.NewTestContext(t)
 
-	cxt.AddTestFile("testdata/porter-with-image.yaml", config.Name)
+	cxt.AddTestFile("testdata/porter-with-deprecated-fields.yaml", config.Name)
 
 	m, err := LoadManifestFrom(cxt.Context, config.Name)
 	require.NoError(t, err, "expected no error")
 	require.NotNil(t, m, "manifest was nil")
+
+	require.Equal(t, "getporter/porter-hello:v0.1.0", m.BundleTag, "manifest has incorrect bundle tag")
```
carolynvs, line 60: We should be asserting off of the new field Reference, not BundleTag now. After a manifest is loaded, `Reference` should contain the final full reference, and bundletag shouldn't have been modified.

pkg/manifest/manifest_test.go
```diff
@@ -195,38 +207,91 @@ func TestReadManifest_File(t *testing.T) {
...
+	t.Run("registry and reference provided", func(t *testing.T) {
+		m := Manifest{
+			Name:      "mybun",
+			Version:   "1.2.3-beta.1",
+			Registry:  "myregistry/myorg",
+			Reference: "getporter/org/mybun:v1.2.3",
+		}
+		err := m.SetDefaults()
+		require.NoError(t, err)
+		assert.Equal(t, "getporter/org/mybun:v1.2.3", m.BundleTag)
```
carolynvs, line 279: Again let's update Reference and have it hold the final value, and not modify BundleTag (deprecated).

pkg/manifest/manifest_test.go
```diff
@@ -195,38 +207,91 @@ func TestReadManifest_File(t *testing.T) {
...
+			Version:  "1.2.3-beta.1",
+			Registry: "getporter/myorg",
+		}
+		err := m.SetDefaults()
+		require.NoError(t, err)
+		assert.Equal(t, "getporter/myorg/mybun:v1.2.3-beta.1", m.BundleTag)
+		assert.Equal(t, "getporter/myorg/mybun-installer:v1.2.3-beta.1", m.Image)
+	})
+
+	t.Run("registry and reference provided", func(t *testing.T) {
```
carolynvs, line 270: Based on this test case, seems like when you set both registry and reference that we ignore registry. Let's warn about that during `Manifest.Validate` so people aren't surprised.

pkg/manifest/manifest_test.go
```diff
@@ -195,38 +207,91 @@ func TestReadManifest_File(t *testing.T) {
...
+			Registry:  "myregistry/myorg",
+			Reference: "getporter/org/mybun:v1.2.3",
+		}
+		err := m.SetDefaults()
+		require.NoError(t, err)
+		assert.Equal(t, "getporter/org/mybun:v1.2.3", m.BundleTag)
+		assert.Equal(t, "getporter/org/mybun-installer:v1.2.3", m.Image)
+	})
+
+	t.Run("tag (deprecated) and registry provided", func(t *testing.T) {
```
carolynvs, line 283: Can you add one more test case for when tag and reference is specified?

pkg/manifest/testdata/simple.porter.yaml
```diff
@@ -4,7 +4,8 @@ mixins:
 name: hello
 description: "An example Porter configuration"
 version: 0.1.0
-tag: getporter/porter-hello
+registry: getporter
+reference: getporter/hello:v0.1.0
```
carolynvs, line 8: Why are we setting both in this test file?

pkg/runtime/runtime-manifest_test.go
```diff
@@ -657,10 +657,11 @@ func TestDependency_Validate(t *testing.T) {
...
 		wantError string
 	}{
-		{"version in tag", manifest.Dependency{Name: "mysql", Tag: "deislabs/azure-mysql:5.7"}, ""},
-		{"version ranges", manifest.Dependency{Name: "mysql", Tag: "deislabs/azure-mysql", Versions: []string{"5.7.x-6"}}, ""},
-		{"missing tag", manifest.Dependency{Name: "mysql", Tag: ""}, "dependency tag is required"},
-		{"version double specified", manifest.Dependency{Name: "mysql", Tag: "deislabs/azure-mysql:5.7", Versions: []string{"5.7.x-6"}}, "dependency tag can only specify REGISTRY/NAME when version ranges are specified"},
+		{"tag (deprecated) supplied", manifest.Dependency{Name: "mysql", Tag: "deislabs/azure-mysql:5.7"}, ""},
+		{"version in reference", manifest.Dependency{Name: "mysql", Reference: "deislabs/azure-mysql:5.7"}, ""},
+		{"version ranges", manifest.Dependency{Name: "mysql", Reference: "deislabs/azure-mysql", Versions: []string{"5.7.x-6"}}, ""},
+		{"missing tag", manifest.Dependency{Name: "mysql", Reference: ""}, "dependency reference is required"},
```
carolynvs, line 663:
```suggestion
		{"missing reference", manifest.Dependency{Name: "mysql", Reference: ""}, "dependency reference is required"},
```

pkg/runtime/runtime-manifest_test.go
```diff
@@ -657,10 +657,11 @@ func TestDependency_Validate(t *testing.T) {
 		dep       manifest.Dependency
 		wantError string
 	}{
-		{"version in tag", manifest.Dependency{Name: "mysql", Tag: "deislabs/azure-mysql:5.7"}, ""},
-		{"version ranges", manifest.Dependency{Name: "mysql", Tag: "deislabs/azure-mysql", Versions: []string{"5.7.x-6"}}, ""},
-		{"missing tag", manifest.Dependency{Name: "mysql", Tag: ""}, "dependency tag is required"},
-		{"version double specified", manifest.Dependency{Name: "mysql", Tag: "deislabs/azure-mysql:5.7", Versions: []string{"5.7.x-6"}}, "dependency tag can only specify REGISTRY/NAME when version ranges are specified"},
+		{"tag (deprecated) supplied", manifest.Dependency{Name: "mysql", Tag: "deislabs/azure-mysql:5.7"}, ""},
```
carolynvs, line 660: I think we need one more test case for when tag and reference is specified.

workshop/gcloud-compute/porter.yaml
```diff
@@ -1,7 +1,7 @@
 name: porter-gcloud-compute
```
carolynvs, line 1: maybe call this `name: gcloud-compute` just to match the change you have been making to remove the extra porter- prefix on our bundles?

workshop/wordpress/porter.yaml
```diff
@@ -1,6 +1,6 @@
-name: wordpress
+name: porter-workshop-wordpress
```
carolynvs, line 1: Sam here maybe just call it `workshop-wordpress`. I think the only special bundle that gets to have the porter- prefix is `porter-hello`.

carolynvs (changes requested):
This is gonna be soooo much better!

docs/content/author-bundles.md
```diff
@@ -31,17 +31,21 @@ although Porter does have extra fields that are specific to making Porter bundle
...
 * `version`: The version of the bundle, uses [semver](https://semver.org). Should not have a 'v' prefix.
-* `tag`: The tag to use when the bundle is published to a registry. The format is `REGISTRY/IMAGE`, or optionally `REGISTRY/IMAGE:TAG` 
-   if you do not want the default behavior of defaulting the docker image TAG portion to the version of the bundle. The invocation
-   image name will be based on this value.  For example, if the bundle `tag` is `getporter/porter-hello` and the `version` is `0.1.0`,
-   then the invocation image name will be `getporter/porter-hello-installer:v0.1.0`
+* `registry`: The registry to use for publishing the bundle. The format is `REGISTRY/ORG`.
+    Both the final bundle reference and invocation image name will be based on this value.
+    For example, if the bundle name is `porter-hello`, registry is `getporter` and the version is `0.1.0`,
+    the bundle reference will be `getporter/porter-hello:v0.1.0` and the invocation image name will be `getporter/porter-hello-installer:v0.1.0`
+* `reference`: OPTIONAL. The full bundle reference which takes precedence over any values set for the `registry`, `name` and `version` fields.
```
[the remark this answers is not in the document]
vdice: Defaulting the Docker tag portion of the `reference`, when not provided (e.g. `reference: getporter/wordpress`) will work just like it did for `tag`, i.e. the `version` value will be used.  If `version: 0.1.0`, the final bundle ref will be `getporter/wordpress:v0.1.0`.

Should I add a clarification here around this scenario?

pkg/manifest/testdata/simple.porter.yaml
vdice: Hmmm good Q.  Must be a mistake.  Will remove.

pkg/manifest/manifest.go
```diff
@@ -756,21 +792,43 @@ func UnmarshalManifest(cxt *context.Context, manifestData []byte) (*Manifest, er
...
 
-// SetInvocationImageFromBundleTag sets the invocation image name on the manifest
-// per the provided bundle tag, also updating the manifest BundleTag value
-// if it initially lacks a Docker tag
-func (m *Manifest) SetInvocationImageFromBundleTag(bundleTag string) error {
-	bundleRef, err := reference.ParseNormalizedNamed(bundleTag)
+// SetInvocationImageAndBundleTag sets the invocation image name and the
+// bundle tag on the manifest per the provided reference or via the
+// registry or repository values on the manifest.
+func (m *Manifest) SetInvocationImageAndBundleTag(ref string) error {
```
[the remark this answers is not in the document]
vdice: Good call, indeed.  Will update.

docs/content/author-bundles.md
```diff
@@ -31,17 +31,21 @@ although Porter does have extra fields that are specific to making Porter bundle
 name: azure-wordpress
 description: Install Wordpress on Azure
 version: 0.1.0
-tag: getporter/azure-wordpress
+registry: getporter
+reference: getporter/azure-wordpress:v0.1.0
```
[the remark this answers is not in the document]
carolynvs: Excellent! How about we continue to leave the TAG off in that case, preserving the simplified example from before?

I liked how it demonstrated your defaulting magic where the version of the bundle is appended to the tag/reference. Encouraging people to not double specify the version should be a best practice that we nudge people towards.

docs/content/author-bundles.md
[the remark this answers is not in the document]
carolynvs: Yeah I think mentioning that it will automatically append the version as the TAG, maybe say that we recommend not specifying the TAG part of `REGISTRY_HOST/ORG/BUNDLE:TAG` and let porter fill it in?

pkg/cache/cache_test.go
```diff
@@ -174,7 +174,7 @@ func TestStoreManifest(t *testing.T) {
...
+						"manifest": "bmFtZTogbXlidW4KdmVyc2lvbjogMC4xLjAKcmVnaXN0cnk6IGV4YW1wbGUuY29tCgptaXhpbnM6CiAgLSBleGVjCgpkZXBlbmRlbmNpZXM6CiAgLSBuYW1lOiBteXNxbAogICAgcmVmZXJlbmNlOiAiZ2V0cG9ydGVyL2F6dXJlLW15c3FsOjUuNyIKICAgIHBhcmFtZXRlcnM6CiAgICAgIGRhdGFiYXNlLW5hbWU6IHdvcmRwcmVzcwoKaW5zdGFsbDoKICAtIGV4ZWM6CiAgICAgIGRlc2NyaXB0aW9uOiAiSW5zdGFsbCBIZWxsbyBXb3JsZCIKICAgICAgY29tbWFuZDogYmFzaAogICAgICBmbGFnczoKICAgICAgICBjOiBlY2hvIEhlbGxvIFdvcmxkCiAgICAgICAgCnVuaW5zdGFsbDoKICAtIGV4ZWM6CiAgICAgIGRlc2NyaXB0aW9uOiAiVW5pbnN0YWxsIEhlbGxvIFdvcmxkIgogICAgICBjb21tYW5kOiBiYXNoCiAgICAgIGZsYWdzOgogICAgICAgIGM6IGVjaG8gR29vZGJ5ZSBXb3JsZAoKY3VzdG9tOgogIGZvbzogYmFyCgpyZXF1aXJlZDoKICAtIHJlcXVpcmVkRXh0ZW5zaW9uMQogIC0gcmVxdWlyZWRFeHRlbnNpb24yOgogICAgICBjb25maWc6IHRydWUKCg==",
```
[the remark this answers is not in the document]
vdice: Oh, thank you for catching this.  I don't know what I used here 😂 -- will remove my changes as they don't apply to this PR.

pkg/manifest/manifest.go
```diff
@@ -512,12 +534,19 @@ func (d *Dependency) Validate() error {
 		return errors.New("dependency name is required")
 	}
 
-	if d.Tag == "" {
-		return errors.New("dependency tag is required")
+	depRef := d.Reference
```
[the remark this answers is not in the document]
vdice: > Perhaps use reference instead and just warn about tag

I like this.  I've gone this route and added testing around the expected warning(s).

pkg/manifest/manifest.go
vdice: Ok, moved some logic into Validate (or, rather, a helper `validateMetadata` func).  PTAL when convenient.  Still open to suggestions for re-org/simplification.

carolynvs (approved):
LGTM, thanks! 💯

PR #1356, opened 2020-11-05
Fix defaulting of reference on windows

### LINKED ISSUE #1354: reference in metadata is documented as OPTIONAL, but build fails if it is absent
## Describe the bug
The 'reference' field is documented as OPTIONAL and will take precedence if present, but if it is omitted, 'porter build' fails while trying to parse the bundle reference.

## To Reproduce
1. Run `porter create`
2. Run `porter build`
3. See error

## Expected behavior
I would expect that it would build the bundle using a calculated reference based on 'registry' and 'name' as described in the docs: https://porter.sh/author-bundles/#bundle-metadata 

## Porter Command and Output
```
$ porter create
creating porter configuration in the current directory

$ porter build
Error: invalid bundle reference %!s(<nil>): invalid reference format
```

## Version
porter v0.30.0 (7a18716b)

carolynvs:
# What does this change
We were accidentally using filepath when joining components of a docker reference, which creates an invalid value registry\bun (wrong slash).

The exiting unit tests would have caught this but we don't (yet) run CI on windows. I just hit this bug when adding CI on Windows so this will have coverage soon.

# What issue does it fix
Fixes #1354 

# Notes for the reviewer
I have manually tested on Windows.

# Checklist
- [ ] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/pkg/manifest/manifest.go
+++ b/pkg/manifest/manifest.go
@@ -4,7 +4,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
-	"path/filepath"
+	"path"
 	"reflect"
 	"regexp"
 	"sort"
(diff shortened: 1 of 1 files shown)


vdice (approved):
LGTM!

PR #1367, opened 2020-11-16
docs(kind.md): clarify host networking option

vdice:
# What does this change
* Updates a section in the KinD doc to clarify current supported options

# What issue does it fix
In response to a question posted in https://github.com/getporter/porter/discussions/1366

# Notes for the reviewer
N/A
# Checklist
- [ ] Unit Tests
- [x] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/docs/content/best-practices/kind.md
+++ b/docs/content/best-practices/kind.md
@@ -11,10 +11,8 @@ By default, KinD sets up the Kubernetes API server IP to be the local loopback a
 
 There are two main options:
 
-1. Configure the Docker container that Porter runs to use [host networking](https://docs.docker.com/network/host/)
-1. Configure the KinD cluster to use an IP address that is resolvable from within Porter's Docker container
-
-Both routes lead to considerable security implications and neither are advised for clusters hosting actual workloads or sensitive information.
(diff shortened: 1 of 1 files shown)


docs/content/best-practices/kind.md
```diff
@@ -11,8 +11,8 @@ By default, KinD sets up the Kubernetes API server IP to be the local loopback a
 
 There are two main options:
 
-1. Configure the Docker container that Porter runs to use [host networking](https://docs.docker.com/network/host/)
-1. Configure the KinD cluster to use an IP address that is resolvable from within Porter's Docker container
+1. Configure the Docker container to use [host networking](https://docs.docker.com/network/host/).  There is an [open issue](https://github.com/getporter/porter/issues/1347) for supporting this in Porter.
```
carolynvs, line 14: Sounds like we don't support host networking yet, so perhaps instead we change the first bullet with the workaround from the issue and link to the issue about a longer term fix? I'm not sure if that's the best way to document what's not working though? I saw the comment about microk8s not working with porter either. Do you think that workaround ... works? 😀

vdice: Good point.  I've revised the first option to mention the public DNS route instead.

carolynvs (approved):
LGTM thanks!

Issue #1352, opened 2020-11-03
Generate immutable Porter manifest in the .cnab directory


vdice:
We'd like to start rendering an immutable Porter manifest and placing it in the `.cnab/app` directory  (alongside the `bundle.json`, etc.) when building a bundle.  This manifest would be a direct translation of the user's Porter manifest in their bundle working directory, but with the potential that certain fields may be updated dynamically, for example if the user builds the same bundle with a differing version or name (see https://github.com/getporter/porter/issues/1334).

Porter will then place this manifest into the installer image (as opposed to the user-provided manifest).  Similarly, on publish, this manifest will be the one referenced.

To be noted, this approach will also open up the ability for users to define an intermediate representation of the Porter manifest in the future (see https://github.com/getporter/porter/issues/207).

Issue #1334, opened 2020-10-22
Build: Support name/version overrides


vdice:
Implementation of the corresponding section of https://github.com/getporter/porter/issues/1151:

There's long been a wish to override the values above at bundle build-time.  See https://github.com/deislabs/porter/issues/694 and https://github.com/deislabs/porter/issues/814.  Being able to dynamically override these values opens up CI/automation possibilities for building and testing different bundle variants.

Assuming we proceed with the aforementioned manifest updates, the following added functionality is proposed to be added to the `porter build` command.

### Added flags

* `--name`: override the bundle name value
* `--version`: override the bundle version value
* ~`--registry`: override the bundle registry value~ *edit: we're currently leaning towards only allowing an override of this field on publish (see https://github.com/getporter/porter/issues/1335)*
* ~`--reference`: override the full bundle reference/~ *edit: we're currently leaning towards only allowing an override of this field on publish (see https://github.com/getporter/porter/issues/1335)*

For all of the above, Porter would update the manifest being referenced with the supplied overrides and save these changes.

## AC

- the flags mentioned above are added to the `porter build` command
- ~use of the `reference` flag should take precedence over any supplied `name`/`version`/`registry` flags (or be mutually exclusive)~

Depends on https://github.com/getporter/porter/issues/1333

PR #1368, opened 2020-11-17
Generate/use canonical manifest for build and publish

vdice:
# What does this change

Initially, I intended to restrict the change set to the creation of the canonical manifest.  However, in order to test the various interactions between the dynamic overrides (for name and version) on build that we wish to support (https://github.com/getporter/porter/issues/1334), I included these changes here as well.

Canonical Manifest changes:
- Store a canonical version of the Porter manifest in .cnab/app/porter.yaml (also included in the invocation image)
- Use yqlib to decode/encode manifest (and potentially apply updates)
  - We went this route as there are a handful of hidden fields (yaml tag of `"-"`) preventing a straight up Marshal
  - Also suits our need of single yaml field updates (to be used in https://github.com/getporter/porter/issues/1334)
- Delay invocation image build until after both the manifest and bundle.json is ready for inclusion

Build flag additions:
- Add name and version flags to the build command
- Use these when generating the canonical Porter manifest

Misc:
- Pin the manifest stamp generation to the user's original manifest (config.Name)

# What issue does it fix
Closes https://github.com/getporter/porter/issues/1352
Closes https://github.com/getporter/porter/issues/1334

# Notes for the reviewer

# Checklist
- [x] Unit Tests
- [x] Documentation
- [ ] Schema (porter.yaml)


pkg/porter/build.go
```diff
@@ -20,24 +21,61 @@ type BuildProvider interface {
...
+		if err := p.preLint(); err != nil {
 			return err
 		}
 	}
 
+	// Build bundle so that resulting bundle.json is available for inclusion
+	// into the invocation image.
+	// Note: the content digest field on the invocation image section of the
+	// bundle.json will *not* be correct until the image is actually pushed
+	// to a registry.  The bundle.json will need to be updated after publishing
```
carolynvs, line 73: 👍 Thanks for the detailed comment. I know we will need it!

pkg/porter/build_test.go
```diff
@@ -121,3 +121,34 @@ func TestPorter_paramRequired(t *testing.T) {
...
+	}{{
+		name:      "no opts",
+		opts:      BuildOptions{},
+		wantError: "",
+	}, {
+		name:      "invalid version set",
+		opts:      BuildOptions{metadataOpts: metadataOpts{Version: "latest"}},
+		wantError: `invalid bundle version "latest".  Cannot be parsed as semver: Invalid Semantic Version`,
+	}, {
+		name:      "valid name and value set",
```
carolynvs, line 139: Let's add one more test case for "v1.0.0", which should fail.

pkg/porter/generate_manifest.go
```diff
@@ -0,0 +1,90 @@
...
+func createUpdateCommand(path, value string) yqlib.UpdateCommand {
+	var valueParser = yqlib.NewValueParser()
+	var parsedValue = valueParser.Parse(value, "", "", "", false)
+	return yqlib.UpdateCommand{Command: "update", Path: path, Value: parsedValue, Overwrite: true}
+}
+
+func (p *Porter) initYqLogger() {
+	// The yq lib that we use makes frequent calls to a logger
+	// Here we set up to log to Porter's Err stream at log level ERROR
+	var backend = logging.AddModuleLevel(logging.NewLogBackend(p.Err, "", 0))
```
carolynvs, line 87: Cool! I didn't know about this before.

pkg/porter/publish.go
```diff
@@ -88,17 +87,33 @@ func (p *Porter) publishFromFile(opts PublishOptions) error {
...
+		// If the manifest file is the default/user-supplied manifest,
+		// hot-swap in Porter's canonical translation (if exists) from
+		// the .cnab/app directory, as there may be dynamic overrides for
+		// the name and version fields to inform invocation image naming.
+		canonicalExists, err := p.FileSystem.Exists(build.LOCAL_MANIFEST)
+		if err != nil {
+			return err
+		}
+
+		if opts.File == config.Name && canonicalExists {
```
carolynvs, line 99: Why the check for if opts.File == "porter.yaml"?

pkg/porter/build.go
```diff
@@ -20,24 +21,61 @@ type BuildProvider interface {
...
 		return err
 	}
 
+	if err := p.generateInternalManifest(opts.metadataOpts); err != nil {
+		return errors.Wrap(err, "unable to generate manifest")
+	}
+
+	// Publish may invoke this method and the manifest will already be
+	// populated.  Only load if still empty.
+	if p.Manifest == nil {
```
carolynvs, line 57: I am concerned that this if check here could cause line 51 to generate a different manifest than what is already loaded. It may be safer long term to unconditionally load the manifest that we just generated.

Especially since we came into this method from Publish deciding the manifest it has is out of date?

pkg/porter/generate_manifest.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	output, err := p.FileSystem.OpenFile(build.LOCAL_MANIFEST, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644)
+	if err != nil {
+		return errors.Wrapf(err, "error creating %s", build.LOCAL_MANIFEST)
+	}
+	defer output.Close()
+
+	// Encode the updated manifest to the proper location
+	var encoder = yqlib.NewYamlEncoder(output, 2, false)
```
carolynvs, line 72: In these calls to yqlib in this function, just reading this PR I can't tell what `2, false` are doing. It would be helpful to use constants for these so that people don't need intellisense or lookup the docs to follow-along.

pkg/porter/generate_manifest.go
```diff
@@ -0,0 +1,90 @@
...
+	return errors.Wrapf(encoder.Encode(&node), "unable to encode the manifest at %s", build.LOCAL_MANIFEST)
+}
+
+// createUpdateCommand creates a yqlib.UpdateCommand using the supplied YAML
+// path and replacement value
+func createUpdateCommand(path, value string) yqlib.UpdateCommand {
+	var valueParser = yqlib.NewValueParser()
+	var parsedValue = valueParser.Parse(value, "", "", "", false)
```
carolynvs, line 80: Yeah same here I have no idea what `"", "", "", false` are doing here. Even just a single comment if you don't want to do constants would really help out.

carolynvs (changes requested):
Sorry for all the questions. This was a really complex change, thanks for taking it on!

pkg/porter/build.go
```diff
@@ -20,24 +21,61 @@ type BuildProvider interface {
...
 
+func (o BuildOptions) Validate() error {
+	if o.Name != "" {
+		// TODO: What sort of validation, if any, do we want on the bundle name?
+		// Originally I had used the *installation* validation from cnab-go,
+		// but that is most likely more restrictive than what we want (no spaces allowed, etc.)
+	}
+
+	if o.Version != "" {
+		if _, err := semver.NewVersion(o.Version); err != nil {
```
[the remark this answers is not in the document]
vdice: Ah good point; it does allow the leading "v".

A few options come to mind:

1. Use our own, stricter regex and do not allow the leading "v", or
2. Allow users to supply `--version v0.1.0` (technically not valid semver) but remove the `v` prior to updating the manifest (manifest would then get `version: 0.1.0`) 

WDYT?

pkg/porter/build.go
```diff
@@ -20,24 +21,61 @@ type BuildProvider interface {
 
 type BuildOptions struct {
 	contextOptions
+	metadataOpts
 	NoLint bool
 }
 
+func (o BuildOptions) Validate() error {
+	if o.Name != "" {
```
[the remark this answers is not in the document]
vdice: Nice, ok, will update to not have any sort of name validation in this PR.

pkg/build/dockerfile-generater.go
```diff
@@ -72,8 +72,8 @@ func (g *DockerfileGenerator) buildDockerfile() ([]string, error) {
 		lines = append(pretoken, append(mixinLines, posttoken...)...)
 	}
 
-	lines = append(lines, g.buildCNABSection()...)
 	lines = append(lines, g.buildPorterSection()...)
+	lines = append(lines, g.buildCNABSection()...)
```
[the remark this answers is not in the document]
vdice: No significance, I'm pretty sure.  I believe during dev I had deleted the `buildPorterSection` but then re-added in this flipped spot.  As there isn't any reasoning beyond this, I'll move it back to its original spot.

pkg/cnab/config-adapter/stamp.go
```diff
@@ -64,7 +64,7 @@ func (c *ManifestConverter) GenerateStamp() (Stamp, error) {
 	stamp := Stamp{}
 
 	// Remember the original porter.yaml, base64 encoded to avoid canonical json shenanigans
-	rawManifest, err := manifest.ReadManifestData(c.Context, c.Manifest.ManifestPath)
+	rawManifest, err := manifest.ReadManifestData(c.Context, config.Name)
```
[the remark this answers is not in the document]
vdice: Ah yes.  Now that we frequently load Porter's canonical manifest (`build.LOCAL_MANIFEST`/`.cnab/app/porter.yaml`), which sets the corresponding `ManifestPath` field on the manifest in memory, the stamp logic was using this instead of the user's manifest.  I went the approach here of hard-coding to the user's manifest (`config.Name`), since this is the version we'd like to monitor for changes (to spark auto-rebuilds, etc.) but we could also go the route of updating `p.Manifest.ManifestPath` whenever we load the canonical version?  Or, there might be another approach? WDYT?

pkg/porter/build_test.go
vdice: Can do, pending resolution on https://github.com/getporter/porter/pull/1368#discussion_r528882681

pkg/build/dockerfile-generater.go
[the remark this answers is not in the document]
vdice: Oops, I take that back.  We indeed want to remove the user's porter.yaml first, as the canonical version will migrate via its location in .cnab (in the following buildCNABSection).  I have a comment to this effect in the buildPorterSection function -- but would it make sense to move and/or add a similar comment here?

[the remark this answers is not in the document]
carolynvs: No that's fine! I just wasn't sure what this was doing. Going forward it's fine and we will see something break in the tests if we mess with it.

pkg/porter/build.go
vdice: Yes, this is a tricky area.  This check was added in response to the following flow in particular:

1. `porter publish --tag diffreg/diffbundle:v0.1.0` -> Currently, we update the bundle reference and invocation image on the manfiest from this supplied tag, which triggers a rebuild owing to the updated manifest fields.
2. `porter build` is therefore invoked, however, currently, we don't pass along this tag value.  The updated manifest fields in memory are then lost when we call `p.LoadManifestFrom(build.LOCAL_MANIFEST)` here.
3. `porter publish --tag diffreg/diffbundle:v0.1.0` continues, albeit with the reloaded manifest and values for image and reference from `build.LOCAL_MANIFEST` and not the correct values derived from the supplied `tag`...

A few options we might entertain (also eager to consider any others!):

- Start passing the `tag` value along whenever publish triggers a build, so that build can use this value to make sure the name, version and 
- Update the latter (#3) logic in publish to use bundle ref and image values parsed from the supplied `tag` value when publishing the bundle and image to the registry, as opposed to using what is on the manifest itself.

pkg/porter/publish.go
vdice: My thinking was that if `opts.File` was different than the default/`config.Name` manifest, we'd want to skip this section and quickly drop down to the check to see if the bundle is up-to-date and trigger a re-build, which will happen assuming the manifest file provided by `opts.File` is different from the default.  However, further updates would be needed to fully support publishing from a manifest file other than the default, including updating the `generateInternalManifest` function added in this PR to accept a filepath variable indicating the original manifest (right now, we are hard-coding it to `config.Name`).  Shall I start focusing on this flow and make updates as needed?

pkg/cnab/config-adapter/stamp.go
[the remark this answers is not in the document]
carolynvs: Nope, that's fine. I just wasn't immediately getting the difference between the two.

* ManifestPath = the filepath that we loaded, which could be our rendered .cnab/porter.yaml
* config.Name = always the user's ./porter.yaml

pkg/porter/build.go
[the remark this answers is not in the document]
carolynvs: Ooh, I like the second option but let's do that in a follow-up.

pkg/porter/publish.go
carolynvs: Since this PR isn't about supporting loading the manifest from anything other than ./porter.yaml (and we both agree that loading from another file _does not work right now) I say don't add anything for that use case. We can follow-up on it later when/if it's necessary.

Let's not add anything else to this PR that isn't related to the main issue cuz this is a chonky PR. 😀

pkg/porter/build.go
carolynvs: Let me restate to make sure I understand. Is this what your change does? How close is your PR to this?

1. The bundle is originally built with `name: mybun`, `version: 1.0.0`, `registry: someplace/`. The invocation image name stored in the bundle.json is `someplace/mybun:v1.0.0`.
2. The bundle is published to an alternate location that doesn't match the existing name + version + registry. For example, `someplace/mybun:latest` or `anotherplace/mybun:v1.0.0`.
3. On publish we are parsing that tag and *updating the in-memory manifest* to match the values from tag (name, version and registry). That normally triggers a rebuild. We shouldn't rewrite the rendered manifest to disk.

I'm going to stop at this point. Does this match what you are thinking? We don't want to rebuild in this scenario.

Instead we want to push a bundle to a different location without modifying the invocation image through the use of the relocation mapping file. We push the original invocation image to the new location, map the original invocation image to the new one, and give the mapping file to cnab-to-oci.

Essentially specifying `porter publish --reference/tag` should not cause the metadata of the bundle to change. It only says "push it to someplace other than the default generated from the metadata".

\* I am ignoring https://github.com/getporter/porter/discussions/1360 for now.

pkg/porter/generate_manifest.go
vdice: Good call; added.

pkg/porter/generate_manifest.go
vdice: Good call; added.

pkg/porter/publish.go
vdice: Ok, check removed.

pkg/porter/build.go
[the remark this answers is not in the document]
vdice: Ok, updated so that this iteration disallows version values w/ a leading "v"

pkg/porter/build_test.go
vdice: Test added!

pkg/porter/build.go
vdice: Yup, you've got it.  As mentioned, currently, 3. _does_ trigger a rebuild, but we'd like to soon move to utilizing the relocation mapping file such that a rebuild is not needed.  Ideally we can prioritize this as a follow-up.

In this PR, I'm attempting to retain the status quo where publishing w/ a different tag/reference correctly triggeres a rebuild... (otherwise, if we don't rebuild -- assuming we don't yet use the relo map -- publish will fail as it can't find the invocation image locally).  Perhaps I can revisit to see if there's a better way to trigger this without sparking a non-trivial refactor...

carolynvs: No that's fine! You cleared it up for me.

carolynvs (approved):
LGTM, thanks for your patience!

PR #1370, opened 2020-11-18
Support running tests in parallel by iosolating process state 

carolynvs:
I will rebase before merging, I know there are a lot of commits... 😅 

# What does this change
* Updated integration tests to run in parallel because I hate waiting. This brought the CI time from 31 -> 24 minutes. We run our CI on a small linux machine, it would go much faster on our Windows agent which is a better spec (more cores = GOMAXPROCS is higher and more tests can run at the same time). I am working on that right now,  a lot of our tests assume they are not on Windows but it will come in a follow-up.
* The jsonschema library used by cnab isn't threadsafe, so I am using a fork that helps us.
* Store process level values such as environment values and current working directory in our context struct.
* Resolve env vars and other process dependent values against our custom context.
* I fixed cnab-go so that our command driver doesn't rely on global process state.

Note that to run a test matrix in parallel you need to avoid accidentally using go closures wrong:

```
t.Run(tc.name, func(t *testing.T) {
  t.Parallel()
  tc := tc // Ensure that the testcase is in the closure
```

# What issue does it fix
This allow us to run our tests in parallel.

# Notes for the reviewer
* Let me know if this seems maintainable, or a mess
* I am going to try to put additional changes in a follow-up, e.g. marking more tests with t.Parallel, running tests on a different agent, etc because this is getting big.
* An alternative to this would be change all the integration tests to e2e tests, which run in their own process space because they always start from the porter command line. We wouldn't be able to run unit tests in parallel, and would need to change all of those tests to work from the command-line. I am not sure if this would be better or not (it may be harder to test from the outside, we would have to always use real implementations, ever mocked ones that can simulate failures) so your thoughts are appreciated.

# Checklist
- [x] Unit Tests
- [x] Documentation
- [ ] Schema (porter.yaml)


vdice (approved):
I think the benefits of speed and thread-safe-ness outweigh added complexity.  LGTM.

carolynvs:
I've updated the PR with the one fix and squashed a bunch of commits. Will merge when the build says go 🚦

Issue #1389, opened 2020-12-16
Windows E2E failure


vdice:
Seeing the following CI failure around the Windows E2E tests: https://dev.azure.com/getporter/525ee0af-eff4-4f16-9127-da829779137b/_apis/build/builds/593/logs/70

It seems to time out on invoking `exec.exe --version ...`

Not sure if this is CI/Win agent related or a bug in our Windows binary build(s) -- or other!  But it seems to be cropping up lately in CI and thus failing builds for unrelated PRs, etc.

carolynvs:
There were three problems that caused this:

1. Docker Desktop doesn't start the docker service when it detects that an update is available. I have updated our windows agent image to disable checking for updates. We will just need to refresh periodically. ✅
2. My logic in our magefile that waits for docker service to come up, and it using debug logging instead of info logs so we didn't see that we were waiting on docker to start.
3. That same logic is waiting waay too long. I meant for it to wait a minute, not an hour! 😂

carolynvs:
This is fixed, builds are passing again.

PR #1391, opened 2020-12-21
Improve starting docker and local registry before running tests

carolynvs:
# What does this change
* You only see messages from log.Println when you call mage with -v. I have gone through our magefile and switched any output lines that should always be printed, like the messages about waiting for the docker service to start on the windows agent, to use fmt.Println.
* Start docker first, then start the local docker registry
* Handle the local docker registry container being stopped when attempting to start the docker registry for tests
* Execute the DockerDesktop.exe command differently so that we don't try to wait for that process to exit. (Fixes hang)

# What issue does it fix
Related to #1389

# Notes for the reviewer
~I am not sure why the build got stuck for an hour on this. I just tested my polling logic and it should only be waiting a minute. I'm going to just get this logging fixed and then we'll have more info if it happens again.~ The hang was caused by using sh.Run which waits for the command to exit. I have switched to using os/exec.Command.Start instead, which fixes the hang.

Here's what I saw after changing the timeout from 60 to 5 (note the 5 periods). It did quickly fail so it seems to work ... on my machine. 🔥 

```console
$ mage StartDocker
Waiting for the docker service to be ready.....
Error: a timeout was reached waiting for the docker service to become unavailable
```

# Checklist
- [ ] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,7 +92,7 @@ test-integration: clean-last-testrun start-local-docker-registry
 	$(GO) build -o $(PORTER_HOME)/testplugin ./cmd/testplugin
 	PROJECT_ROOT=$(shell pwd) $(GO) test -timeout 30m -tags=integration ./...
 
-teste2e: clean-last-testrun start-local-docker-registry
+teste2e:
 	go run mage.go teste2e
 
 .PHONY: docs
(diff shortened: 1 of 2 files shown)


vdice (approved):

carolynvs:
I fixed some flaws in the code that starts local docker registry. There was a race between starting docker and the registry and also it didn't gracefully handle an existing stopped registry container.

I still don't know what causes the hang but this should help narrow it down.

carolynvs:
Fixed the hang!

PR #1393, opened 2020-12-22
Set GOPATH/bin on PATH properly

carolynvs:
# What does this change
We were assuming a linux agent with /home/vsts/go/bin. We should have been using the discovered location using GOAPTH/bin instead.

# What issue does it fix
We are getting `'mage' is not recognized as an internal or external command,` on our windows build when running `mage TestE2E` because GOPATH/bin is not in PATH.

# Notes for the reviewer
I think this was passing before because the windows agent was "dirty" and had local changes I had made when debugging. When I reimaged and we had a fresh agent, this popped up.

# Checklist
- [ ] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/magefile.go
+++ b/magefile.go
@@ -48,7 +48,7 @@ func ConfigureAgent() error {
 	if err != nil {
 		return errors.Wrapf(err, "could not mkdir -p %s", gobin)
 	}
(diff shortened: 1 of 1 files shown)


carolynvs:
self merging cuz this fixes the build and it broke 😀

PR #1397, opened 2020-12-23
Bump cnab-go to v0.15.0

carolynvs:
# What does this change
This contains the fix so that docker writes to op.Out instead of stdout directly. This will let us better capture and only display bundle output (instead of all the logging too) like we want to for #1353. It also fixes our test output when in verbose mode or when tests fail, so the lines print in order.

# What issue does it fix

# Notes for the reviewer
NA

# Checklist
- [ ] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)


vdice (approved):
🎉

PR #1399, opened 2020-12-28
Stop build if we can't install a mixin

carolynvs:
# What does this change
The build right now doesn't stop when we can't install a mixin into the build. This allows the build to get all the way to the end before something fails because it needs that mixin. Fail fast.

* Move the get-mixins target to mage
* Fail the target if any mixins fail to install
* Only install the mixin if it's not present in the bin

# What issue does it fix
Makes broken build fail faster

# Notes for the reviewer
We really need to fix how we use our CDN so that we don't overwrite existing files. I'm not sure why it's so flaky but that is probably part of it.

# Checklist
- [ ] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,9 +32,6 @@ FILE_EXT=
 endif
 
 INT_MIXINS = exec
-EXT_MIXINS = helm arm terraform kubernetes
-MIXIN_TAG ?= canary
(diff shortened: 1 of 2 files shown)


carolynvs:
Fixing build with my admin powers... doot doot

Issue #1335, opened 2020-10-22
Publish: Update/support tag and repository flags


vdice:
Implementation of the corresponding section of https://github.com/getporter/porter/issues/1151:

Today, Porter's publish command offers the `-t/--tag` flag to override the bundle reference (and update the invocation image name accordingly) prior to pushing to an OCI registry.  The flag naming was directly inspired by the correlate from the `docker build` command, but we've since realized that we'd like to restrict it to just representing a Docker tag override, rather than an override for the bundle reference.  Therefore, we propose to redefine this flag's purpose and add a new flag, `--reference`, to serve the purpose of this flag's previous responsibility.

* `-t/--tag`: override the Docker tag of the bundle reference (and invocation image), e.g. `latest`, `v0.1.1`, etc.
* `--reference`: override the bundle reference entirely (and rename the invocation image accordingly), e.g. `otherregistry.com/otherorg/mybuns:v0.1.0`
* `--registry`: override the bundle registry

## AC

- the `tag` flag is updated to represent the Docker tag portion of the bundle for `porter publish`
- the `reference` flag is added for `porter publish`
- the `registry` flag is added

carolynvs:
I just realized that the proposal is to have `--repository` be the entire destination, e.g. `docker.io/myorg/mybundle:v1`. That will be confusing, just like tag is today, because repository refers to this part only `docker.io/myorg/mybundle`. Here's the Docker [documentation](https://docs.docker.com/docker-hub/repos) to help show the relationship: one repository has many tags.

> A single Docker Hub repository can hold many Docker images (stored as tags).

Since we are trying to match the terms better, let's not redefine repository. I have been searching all over for an accepted term for the entire reference and haven't found one. Hah, pretty sure this is why we used tag to begin with (besides following the convention set by the docker cli).

Breaking down the parts of a reference we have `REGISTRYHOST/ORG/BUNDLE_NAME:TAG` where:

* Really the term registry refers to just `docker.io` or `quay.io`. However since swapping just the host of the registry is usually not what people do, they also change the org and other subpaths as well, we are calling this entire part `REGISTRYHOST/ORG` the registry. So someone can change where they are publishing a bundle using `--registry quay.io/differentuser/subproject` **Do you think this is confusing or helpful?**

* The repository is composed of `REGISTRYHOST/ORG/BUNDLE_NAME`

* The tag refers to the final component of the reference `TAG` e.g. `latest` or `v1`

Some examples using the new flags

```
[...]

Issue #1151, opened 2020-07-20
Proposal: Naming/Versioning Manifest updates


vdice:
# Proposal: Naming/Versioning Manifest updates

This issue proposes to update the naming and versioning fields in the Porter manifest, as well as prescribe some immediate follow-ups around Porter build and publish functionality to utilize the simplification these updates bring.

## Manifest updates

- [x] https://github.com/getporter/porter/issues/1333

Currently, the naming and versioning fields in the Porter manifest are: `name`, `version` and `tag`.  The last one of these has heretofore been used to capture the full name of the bundle reference, e.g. `myregistry/myorg/mybuns:v0.1.0`.

We intend to preserve `name` (as the common name of the bundle) and `version` (as the semantic version of the bundle) but replace `tag` with `registry` (and, optionally, `reference`).

* `registry`: this is the OCI registry and org/subdomain for the bundle, e.g. `myregistry.com/myorg`.  Porter will then construct the full bundle reference using this and the values for `name` and `version`, e.g. `myregistry.com/myorg/mybuns:v0.1.0` from the example below.

* `reference`: (OPTIONAL) this can be used to specify the full bundle reference when needed.  When the bundle is published this value will be used rather than the value constructed from the `name`, `version` and `registry` values.

```
name: mybuns
version: 0.1.0
registry: myregistry.com/myorg
# Optional: can use reference to specify the exact bundle reference (correlate to deprecated `tag` field)
# reference: myregistry.com/myorg/mybuns:v0.1.0
```

For a few issues and added context that led to the genesis of this proposed update, see https://github.com/deislabs/porter/issues/459 and https://github.com/deislabs/porter/issues/1146.

## Build functionality

- [x] https://github.com/getporter/porter/issues/1334

There's long been a wish to override the values above at bundle build-time.  See https://github.com/deislabs/porter/issues/694 and https://github.com/deislabs/porter/issues/814.  Being able to dynamically override these values opens up CI/automation possibilities for building and testing different bundle variants.

Assuming we proceed with the aforementioned manifest updates, the following added functionality is proposed to be added to the `porter build` command.

### Added flags

* `--name`: override the bundle name value
* `--version`: override the bundle version value
* `--registry`: override the bundle registry value
* `--reference`: override the full bundle reference

For all of the above, Porter would update the manifest being referenced with the supplied overrides and save these changes.

## Publish functionality

- [ ] https://github.com/getporter/porter/issues/1335

[...]

PR #1405, opened 2021-01-08
Updates to publishing: flags, invocation image tagging

vdice:
# What does this change

Updates `porter publish` per the manifest update epic in https://github.com/getporter/porter/issues/1151:

- Effectively renames the former `tag` flag to `reference` for specifying a full bundle reference for publishing
  - Note: as this flag is part of the BundlePullOptions, the name changes for other commands as well, e.g. `porter install --reference mybuns:v0.1.0`
- Updates the `tag` flag to just designate the Docker tag portion of the published bundle
- Adds the `registry` flag to designate the Docker registry (and org/subdomain) portion of the published bundle
- Updates logic in publish to tag (as in `docker tag`) an updated invocation image (say, if a new registry value is provided on publish), so that the image is found for publishing.  Previously, a new invocation image name would trigger a full new re-build.

Ref https://github.com/getporter/porter/issues/1335

# What issue does it fix
Closes https://github.com/getporter/porter/issues/1335
Closes https://github.com/getporter/porter/issues/1151

# Notes for the reviewer

# Checklist
- [ ] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/cmd/porter/bundle.go
+++ b/cmd/porter/bundle.go
@@ -292,7 +292,9 @@ func buildBundlePublishCommand(p *porter.Porter) *cobra.Command {
 	cmd := cobra.Command{
 		Use:   "publish",
 		Short: "Publish a bundle",
-		Long:  "Publishes a bundle by pushing the invocation image and bundle to a registry.",
+		Long: `Publishes a bundle by pushing the invocation image and bundle to a registry.
+
+Note: if overrides for registry/tag/reference are provided, this command only re-tags the invocation image and bundle; it does not re-build the bundle.`,
 		Example: `  porter bundle publish
   porter bundle publish --file myapp/porter.yaml
   porter bundle publish --archive /tmp/mybuns.tgz --reference myrepo/my-buns:0.1.0
--- a/cmd/porter/bundle.go
+++ b/cmd/porter/bundle.go
@@ -103,8 +103,8 @@ The first argument is the name of the installation to create. This defaults to t
 Porter uses the Docker driver as the default runtime for executing a bundle's invocation image, but an alternate driver may be supplied via '--driver/-d'.
 For example, the 'debug' driver may be specified, which simply logs the info given to it and then exits.`,
 		Example: `  porter bundle install
-  porter bundle install MyAppFromTag --tag getporter/kubernetes:v0.1.0
-  porter bundle install --tag localhost:5000/getporter/kubernetes:v0.1.0 --insecure-registry --force
+  porter bundle install MyAppFromReference --reference getporter/kubernetes:v0.1.0
+  porter bundle install --reference localhost:5000/getporter/kubernetes:v0.1.0 --insecure-registry --force
   porter bundle install MyAppInDev --file myapp/bundle.json
   porter bundle install --parameter-set azure --param test-mode=true --param header-color=blue
   porter bundle install --cred azure --cred kubernetes
@@ -149,8 +149,8 @@ The first argument is the installation name to upgrade. This defaults to the nam
 Porter uses the Docker driver as the default runtime for executing a bundle's invocation image, but an alternate driver may be supplied via '--driver/-d'.
 For example, the 'debug' driver may be specified, which simply logs the info given to it and then exits.`,
 		Example: `  porter bundle upgrade
-  porter bundle upgrade --tag getporter/kubernetes:v0.1.0
-  porter bundle upgrade --tag localhost:5000/getporter/kubernetes:v0.1.0 --insecure-registry --force
+  porter bundle upgrade --reference getporter/kubernetes:v0.1.0
+  porter bundle upgrade --reference localhost:5000/getporter/kubernetes:v0.1.0 --insecure-registry --force
   porter bundle upgrade MyAppInDev --file myapp/bundle.json
   porter bundle upgrade --parameter-set azure --param test-mode=true --param header-color=blue
   porter bundle upgrade --cred azure --cred kubernetes
@@ -196,8 +196,8 @@ The first argument is the installation name upon which to invoke the action. Thi
 Porter uses the Docker driver as the default runtime for executing a bundle's invocation image, but an alternate driver may be supplied via '--driver/-d'.
 For example, the 'debug' driver may be specified, which simply logs the info given to it and then exits.`,
 		Example: `  porter bundle invoke --action ACTION
-  porter bundle invoke --tag getporter/kubernetes:v0.1.0
-  porter bundle invoke --tag localhost:5000/getporter/kubernetes:v0.1.0 --insecure-registry --force
+  porter bundle invoke --reference getporter/kubernetes:v0.1.0
+  porter bundle invoke --reference localhost:5000/getporter/kubernetes:v0.1.0 --insecure-registry --force
   porter bundle invoke --action ACTION MyAppInDev --file myapp/bundle.json
   porter bundle invoke --action ACTION  --parameter-set azure --param test-mode=true --param header-color=blue
   porter bundle invoke --action ACTION --cred azure --cred kubernetes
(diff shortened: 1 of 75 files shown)


carolynvs:
Looks like the stamp constant changed for one of the tests, once that's updated the tests should pass.

pkg/cnab/config-adapter/stamp.go
```diff
@@ -102,9 +102,6 @@ func (c *ManifestConverter) DigestManifest() (string, error) {
 	v := pkg.Version
 	data = append(data, v...)
 
-	// (for instance, during publish), so add this to the data
-	data = append(data, c.Manifest.Image...)
```
carolynvs, line 106: Sorry I know we talked about this but I've gotten fuzzy over the holidays. Can you remind me why we can remove this? I am just wondering if the info is now in the generate manifest, or if we can't put it in anymore because the flag may have changed the value?

pkg/manifest/manifest.go
```diff
@@ -47,6 +47,10 @@ type Manifest struct {
 	// and isn't meant to be user-specified
 	BundleTag string `yaml:"-"`
```
carolynvs, line 48: Do we use this field, `BundleTag`, to hold the final computed bundle reference? Or can we remove it?

pkg/porter/build.go
```diff
@@ -54,16 +57,8 @@ func (p *Porter) Build(opts BuildOptions) error {
...
-	// populated.  Only load if still empty.
-	// For instance, Publish may be called with a full, new bundle reference
-	// via --tag, which will update the invocation image name and spark a new
-	// build here.  If we re-load from the local manifest, we will lose these
-	// values.
-	if p.Manifest == nil {
-		if err := p.LoadManifestFrom(build.LOCAL_MANIFEST); err != nil {
-			return err
-		}
+	if err := p.LoadManifestFrom(build.LOCAL_MANIFEST); err != nil {
```
carolynvs, line 60: What is the flow when someone does `publish --reference`. Do we do a full rebuild, or do we only retag? I'm wondering why it's safe now to unconditionally load the generated manifest.

I'm not you asking to change anything, just making sure I understand the new workflow with these tags.

cmd/porter/bundle.go
```diff
@@ -295,7 +295,9 @@ func buildBundlePublishCommand(p *porter.Porter) *cobra.Command {
 		Long:  "Publishes a bundle by pushing the invocation image and bundle to a registry.",
 		Example: `  porter bundle publish
   porter bundle publish --file myapp/porter.yaml
-  porter bundle publish --archive /tmp/mybuns.tgz --tag myrepo/my-buns:0.1.0
+  porter bundle publish --archive /tmp/mybuns.tgz --reference myrepo/my-buns:0.1.0
+  porter bundle publish --tag latest
```
carolynvs, line 299: It's going to be a rough transition because we changed the meaning of this flag. How about in a follow-up we add validation for the publish options that checks if they are using `--tag` to specify the entire reference and error out explaining that the flags have changed?

We can identify  that `--tag` contains a reference by checking for `:`/`@`, neither of which are allowed in a docker tag.

carolynvs (commented):
Looks great! Thank you for the massive rename and logic switcharoo. Just had some questions to make sure I fully understand the change and then we are good to go.

pkg/manifest/manifest.go
carolynvs: I see it used in code but it's super confusing and I think a relic from a previous config setting we removed. Now may be a real good time to get rid of it completely (in a follow-up).

vdice: > Do we use this field, BundleTag, to hold the final computed bundle reference? Or can we remove it?

Not anymore - now we use `Reference` to track the bundle reference.  Indeed, `BundleTag` is ripe for removal.  I can create an issue.

pkg/porter/build.go
vdice: Good question.  Now, when a `--reference` override is supplied, we only retag (invocation image and final bundle reference).  This does mean the bundle.json and porter.yaml files in the invocation image will be from the unchanged internal manifest.

pkg/cnab/config-adapter/stamp.go
vdice: I/we had added this previously to automatically trigger a rebuild whenever a bundle's image was updated.  (Say, when a new registry or tag value is supplied on publish.)  However, in this PR, instead of rebuilding the bundle, we re-tag images/bundle references and use these for publishing.

As noted elsewhere, it does mean that the actual contents (bundle.json/porter.yaml) included in the installer/invocation image may have different registry/reference values from the final published versions.  For instance, the built bundle might have `registry: myregistry.com` but if I publish the bundle via `porter publish --registry localhost:5000`, the published invocation image (`localhost:5000/mybuns:v0.1.0`) will have a `porter.yaml` file with the original `registry: myregistry.com` field.

cmd/porter/bundle.go
vdice: Good point.  I almost think it's warranted to delay this PR until another commit including this logic is pushed.  WDYT?  I can work on this.

pkg/cnab/config-adapter/stamp.go
carolynvs: Thanks for clearing that up. I know some users want to be able to publish to alternate locations, either because they moved it across an airgap or something else, and have the bundle _not_ contain the original information. One workaround in that case would be to use non-sensitive information when building the bundle and then always set a --registry, etc during publish.

cmd/porter/bundle.go
carolynvs: If you are up for it, that's great.

pkg/porter/build.go
carolynvs: Can you  document this new "re-tag not rebuild" behavior in the `LongDescription` of the `porter publish`  command?

vdice: 👍 Added.

cmd/porter/bundle.go
vdice: Ok, I added logic to error accordingly for the tag option on publish.

Do we want to do something similar for the other commands that previously also had a `tag` opt (courtesy BundlePullOptions)?  As it stands, users who try to invoke `porter install --tag mybuns:v0.1.0` will just see a standard `Error: unknown flag: --tag` after this is merged (and need to check the docs to see `--reference` is the new opt).  Naturally, in the release notes, we can note this breaking cli opt change... not sure if we should also bake in a deprecation warning or error.

carolynvs: Yeah, we could add back a Tag field and flag definitions and then check if it was populated and give a better error message. It would just be temporary until 1.0 when hopefully we stop making breaking changes for a bit.

pkg/porter/publish.go
```diff
@@ -57,6 +57,21 @@ func (o *PublishOptions) Validate(cxt *portercontext.Context) error {
...
+	}
+
+	return nil
+}
+
+// validateTag checks to make sure the supplied tag is of the expected form.
+// A previous iteration of this flag was used to designate an entire bundle
+// reference.  If we detect this attempted use, we return an error and
+// explanation
+func (o *PublishOptions) validateTag() error {
```
carolynvs, line 71: Thanks for adding this! It will really help avoid a bunch of very unhappy users.

cmd/porter/bundle.go
vdice: Ok, added it back as a deprecated flag (in Cobra terms this means it'll hide the flag by default and produce a deprecation warning when a value is provided).  When provided, we just assign reference to the value instead of explicitly erroring out -- WDYT?

cmd/porter/main.go
```diff
@@ -133,11 +133,17 @@ func ShouldShowUngroupedCommand(cmd *cobra.Command) bool {
...
 func addBundlePullFlags(f *pflag.FlagSet, opts *porter.BundlePullOptions) {
+	addDeprecatedTagFlag(f, opts)
 	addReferenceFlag(f, opts)
 	addInsecureRegistryFlag(f, opts)
 	addForcePullFlag(f, opts)
 }
 
+func addDeprecatedTagFlag(f *pflag.FlagSet, opts *porter.BundlePullOptions) {
+	f.StringVar(&opts.Tag, "tag", "", "")
+	f.MarkDeprecated("tag", "use --reference to declare a full bundle reference")
```
carolynvs, line 144: Will this print a warning so that they know it's been deprecated and that they should stop using it? If not, let's an the warning to where you do the reference = tag assignment.

vdice: It will.  This is what it currently looks like:
```
 $ p install --tag localhost:5000/cert-manager:v0.1.0 -c mk -p cert-manager
Flag --tag has been deprecated, use --reference to declare a full bundle reference
installing cert-manager...
...
```

Do we want stronger verbiage (e.g. `WARNING ...`)?

carolynvs: Nope! I just hadn't used the cobra deprecation flag and wasn't sure how it worked.   👍

carolynvs (approved):
One step closer to 1.0!

PR #1421, opened 2021-01-15
fix(*): continue to support dependency tag use for time being

vdice:
# What does this change

Alas, another area where we need to support the deprecated tag field for the time being:

 - Adds the `tag` yaml tag back in (along with corresponding logic) so that older manifests still function if they use this field, until 1.0
 
# What issue does it fix

# Notes for the reviewer

# Checklist
- [x] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/pkg/cnab/config-adapter/adapter.go
+++ b/pkg/cnab/config-adapter/adapter.go
@@ -307,12 +307,6 @@ func (c *ManifestConverter) generateDependencies() *extensions.Dependencies {
 	}
 
 	for _, dep := range c.Manifest.Dependencies {
-		// Use the deprecated Tag value if Reference not set
-		// We should remove this field completely before 1.0
-		if dep.Tag != "" && dep.Reference == "" {
-			dep.Reference = dep.Tag
-		}
-
 		dependencyRef := extensions.Dependency{
 			Name:   dep.Name,
 			Bundle: dep.Reference,
--- a/pkg/cnab/config-adapter/adapter.go
+++ b/pkg/cnab/config-adapter/adapter.go
@@ -307,6 +307,12 @@ func (c *ManifestConverter) generateDependencies() *extensions.Dependencies {
 	}
 
 	for _, dep := range c.Manifest.Dependencies {
+		// Use the deprecated Tag value if Reference not set
+		// We should remove this field completely before 1.0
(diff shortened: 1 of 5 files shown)


pkg/cnab/config-adapter/adapter.go
```diff
@@ -307,6 +307,12 @@ func (c *ManifestConverter) generateDependencies() *extensions.Dependencies {
 	}
 
 	for _, dep := range c.Manifest.Dependencies {
+		// Use the deprecated Tag value if Reference not set
+		// We should remove this field completely before 1.0
+		if dep.Tag != "" && dep.Reference == "" {
```
carolynvs, line 312: I would have expected that when we load the porter.yaml with tag set for a dependency, that dep.Reference would be populated with the tag value (the backwards compatibility fix). Are you sure this is needed?

I don't think it's hurting anything but I'm wondering if I am missing a code path when thinking about this fix.

pkg/manifest/manifest.go
```diff
@@ -557,36 +557,37 @@ type Dependency struct {
...
-	depRef := d.Reference
 	if d.Tag != "" {
-		fmt.Println("WARNING: the tag field has been deprecated in favor of reference; " +
-			"please update the Porter manifest accordingly")
-		if depRef == "" {
-			depRef = d.Tag
+		fmt.Fprintf(cxt.Out, "WARNING: the tag field for dependency %q has been deprecated "+
+			"in favor of reference; please update the Porter manifest accordingly\n",
+			d.Name)
+		if d.Reference == "" {
```
carolynvs, line 577: I think the defaulting right here makes the logic in the adapter unnecessary unless there's another code path that targets?

pkg/cnab/config-adapter/adapter.go
vdice: Great point -- indeed, we needed to update the Dependencies field to be a slice of pointers so that the `Validate` logic in manifest.go would persist.  Updated.

pkg/manifest/manifest.go
vdice: ditto https://github.com/getporter/porter/pull/1421#discussion_r558503530

carolynvs (approved):
Glad you spotted the bug!

PR #1422, opened 2021-01-16
Include version of the Porter when using `porter explain` command

### LINKED ISSUE #1407: Explain should include the Porter version inside the bundle
When I run `porter explain` I want to see which version of Porter was used to build the bundle, and therefore what version is embedded _inside_ the bundle.

```console
$ porter explain  --tag getporter/porter-hello:v0.1.0
Name: porter-hello
Description: An example Porter configuration
Version: 0.1.0
Porter Version: v0.30.1
```

In order to accomplish this we must add an entry into the bundle.json's custom section with information about Porter during `porter build`:

```json
{
  "custom": {
    "sh.porter": {
      "version": "v0.30.0",
      "commit": "3b7c85ba"
    }
//... rest of bundle definition
}
```

This change can be made in the following function:

Save the porter version (get.porter.sh/pkg -> pkg.Version + pkg.Commit) during `porter build`:
https://github.com/getporter/porter/blob/5efddb87b292184baceaf64a3301639e3949e4d1/pkg/cnab/config-adapter/stamp.go#L63

https://github.com/getporter/porter/blob/5efddb87b292184baceaf64a3301639e3949e4d1/pkg/cnab/config-adapter/stamp_test.go#L17

Print the porter version during `porter explain`:
1. Add PorterVersion field to https://github.com/getporter/porter/blob/5efddb87b292184baceaf64a3301639e3949e4d1/pkg/porter/explain.go#L22
2. Load the stamp from the bundle using pkg/cnab/config-adapter.LoadStamp(bundle) and set PorterVersion https://github.com/getporter/porter/blob/5efddb87b292184baceaf64a3301639e3949e4d1/pkg/porter/explain.go#L186-L191
3. Update the test here: https://github.com/getporter/porter/blob/5efddb87b292184baceaf64a3301639e3949e4d1/pkg/porter/explain_test.go#L123

joshuabezaleel:
# What does this change
Includes version of the Porter used when a bundle was built in the output of `porter explain` command when explaining that particular bundle. 

# What issue does it fix
Fixes #1407 

# Notes for the reviewer

# Checklist
- [x] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)


joshuabezaleel:
Thank you so much @carolynvs for explaining crystal clear the steps that I need to implement on the issue 🙂 
Looking forward to the review! 😁

pkg/cnab/config-adapter/stamp_test.go
```diff
@@ -29,6 +29,8 @@ func TestConfig_GenerateStamp(t *testing.T) {
 	require.NoError(t, err, "DigestManifest failed")
 	assert.Equal(t, simpleManifestDigest, stamp.ManifestDigest)
 	assert.Equal(t, map[string]MixinRecord{"exec": {}}, stamp.Mixins, "Stamp.Mixins was not populated properly")
+	assert.Equal(t, pkg.Version, stamp.Version)
```
carolynvs, line 32: During unit tests, pkg.Version and Commit are empty strings by default. So even if you weren't setting these values on the stamp the test would pass. Here's what I recommend you do so that we can test that it's actually being set:

At the top of the test set pkg.Version/Commit to values and then use a defer to cleanup the change (those are package level variables and the change would persist for the next test).

```go
pkg.Version = "v1.2.3"
pkg.Commit = "abc123"
defer func() {
  pkg.Version = ""
  pkg.Commit = ""
}
```

pkg/porter/explain.go
```diff
@@ -20,14 +21,15 @@ type ExplainOpts struct {
...
-	Version      string                `json:"version" yaml:"version"`
-	Parameters   []PrintableParameter  `json:"parameters,omitempty" yaml:"parameters,omitempty"`
-	Credentials  []PrintableCredential `json:"credentials,omitempty" yaml:"credentials,omitempty"`
-	Outputs      []PrintableOutput     `json:"outputs,omitempty" yaml:"outputs,omitempty"`
-	Actions      []PrintableAction     `json:"customActions,omitempty" yaml:"customActions,omitempty"`
-	Dependencies []PrintableDependency `json:"dependencies,omitempty" yaml:"dependencies,omitempty"`
+	Name          string                `json:"name" yaml:"name"`
+	Description   string                `json:"description,omitempty" yaml:"description,omitempty"`
+	Version       string                `json:"version" yaml:"version"`
+	PorterVersion string                `json:"porter_version" yaml:"porter_version"`
```
carolynvs, line 27: Can you change this to use camel case (`porterVersion`) instead of snake case (`porter_version`)? That's what we are using elsewhere when writing to json.

pkg/porter/explain_test.go
```diff
@@ -198,6 +198,13 @@ func TestExplain_generatePrintableBundleParams(t *testing.T) {
 				Required:   true,
 			},
 		},
+		Custom: map[string]interface{}{
```
carolynvs, line 201: I don't think this change is used by the test since it's verifying bundle parameters. So this can be removed.

pkg/porter/explain_test.go
```diff
@@ -224,6 +231,13 @@ func TestExplain_generatePrintableBundleOutputs(t *testing.T) {
 				Definition: "string",
 			},
 		},
+		Custom: map[string]interface{}{
```
carolynvs, line 234: Same, we only need this extra testdata in the test where we are asserting on the porter version in the explain output.

pkg/porter/explain_test.go
```diff
@@ -245,6 +259,13 @@ func TestExplain_generatePrintableBundleCreds(t *testing.T) {
 				Description: "a cred",
 			},
 		},
+		Custom: map[string]interface{}{
```
carolynvs, line 262: This one can be removed too

pkg/porter/explain_test.go
```diff
@@ -259,6 +280,33 @@ func TestExplain_generatePrintableBundleCreds(t *testing.T) {
...
+				"commit":   "3b7c85ba",
+			},
+		},
+	}
+
+	pb, err := generatePrintable(bun)
+	assert.NoError(t, err)
+
+	assert.Equal(t, "v0.30.0", pb.PorterVersion)
+	assert.Equal(t, 0, len(pb.Parameters))
```
carolynvs, line 304: For porter tests, it's okay to not assert for things unrelated to the test, so it's okay to just test `pb.PorterVersion` and not all the other fields in this test.

pkg/porter/explain_test.go
```diff
@@ -281,6 +329,11 @@ func TestExplain_generatePrintableBundleDependencies(t *testing.T) {
 					},
 				},
 			},
+			"sh.porter": map[string]interface{}{
```
carolynvs, line 332: This one can be removed too.

carolynvs (changes requested):
This looks great! 💯 I just have some very small cleanup items and then we are good to go!

carolynvs:
@joshuabezaleel Let me know if there's anything you are stuck on. I'm happy to help get this PR (and the other) finished and merged!

joshuabezaleel:
Hi @carolynvs , I am truly sorry for a really late reply. Currently in a process of moving so things got pretty busy. Will submit a patch by the end of this week or max the start of next week. Thank you so much in advance for the really kind offer and reminder! 🙂

carolynvs:
I was just checking in to make sure you didn't have any questions. There is never a rush to get things merged fast. I hope your move goes well!

pkg/cnab/config-adapter/stamp_test.go
joshuabezaleel: I'm not sure the value of the `pkg.Version` and `pkg.Commit` that I should assign at the beginning of the test. 

Assigning any kind of value in advance to the `pkg.Version` at the beginning of the test will make the result of `stamp.ManifestDigest` vary thus making the assertion of `assert.Equal(t, simpleManifestDigest, stamp.ManifestDigest)` always fails because it differs to the `simpleManifestDigest` that has been declared at the top of the test file, while the value of `pkg.Commit` does not change at all. 

image

pkg/porter/explain_test.go
joshuabezaleel: The reason why I added the `Custom` property to the bundle is because it yields panic if it's not present when we want to load the stamp when calling `generatePrintable`. 

image

joshuabezaleel:
Hi @carolynvs sorry for just replying. I've made some replies to the review since I still got several questions. Sorry if I'm still not really familiar to the project and thank you in advance for the really kind help, truly appreciate it! Hope you have a great week ahead 🙂

pkg/cnab/config-adapter/stamp_test.go
carolynvs: Sorry for the late reply!

I see what you mean, so why don't we add a new test based on that existing one that only checks if we set the version and commit on the stamp properly? That way we can avoid the problem you discovered, where the digest changes.

```go
func TestConfig_GenerateStamp_IncludeVersion(t *testing.T) {
	// Do not run this test in parallel
	// Still need to figure out what is introducing flakey-ness

	pkg.Version = "v1.2.3"
	pkg.Commit = "abc123"
	defer func() {
		pkg.Version = ""
		pkg.Commit = ""
	}()

	c := config.NewTestConfig(t)
	c.TestContext.AddTestFile("../../manifest/testdata/simple.porter.yaml", config.Name)

	m, err := manifest.LoadManifestFrom(c.Context, config.Name)
	require.NoError(t, err, "could not load manifest")

	a := NewManifestConverter(c.Context, m, nil, nil)
	stamp, err := a.GenerateStamp()
	require.NoError(t, err, "DigestManifest failed")
	assert.Equal(t, "v1.2.3", stamp.Version)
	assert.Equal(t, "abc123", stamp.Commit)
}
```

pkg/porter/explain_test.go
carolynvs: Oops, sorry about that you are right. 👍

joshuabezaleel:
@carolynvs ah I see, I thought that the `stamp.Version` should be compared to the `pkg.Version` on the test's assertion. I've made changes according to the review, thank you so much for the really kind and helpful one! 😄

pkg/porter/explain.go
```diff
@@ -22,14 +23,15 @@ type ExplainOpts struct {
...
-	Version      string                `json:"version" yaml:"version"`
-	Parameters   []PrintableParameter  `json:"parameters,omitempty" yaml:"parameters,omitempty"`
-	Credentials  []PrintableCredential `json:"credentials,omitempty" yaml:"credentials,omitempty"`
-	Outputs      []PrintableOutput     `json:"outputs,omitempty" yaml:"outputs,omitempty"`
-	Actions      []PrintableAction     `json:"customActions,omitempty" yaml:"customActions,omitempty"`
-	Dependencies []PrintableDependency `json:"dependencies,omitempty" yaml:"dependencies,omitempty"`
+	Name          string                `json:"name" yaml:"name"`
+	Description   string                `json:"description,omitempty" yaml:"description,omitempty"`
+	Version       string                `json:"version" yaml:"version"`
+	PorterVersion string                `json:"porterVersion" yaml:"porterVersion"`
```
carolynvs, line 29: Let's make it so that if this isn't a Porter bundle, then `porterVersion` is not included when we marshal it to json or yaml. We can do that by using omitempty

```
PorterVersion string                `json:"porterVersion,omitempty" yaml:"porterVersion,omitempty"`
```

pkg/porter/explain.go
```diff
@@ -187,11 +189,17 @@ func (p *Porter) printBundleExplain(o ExplainOpts, pb *PrintableBundle) error {
 	}
 }
 
-func generatePrintable(bun bundle.Bundle, action string) (*PrintableBundle, error) {
+func generatePrintable(bun bundle.Bundle) (*PrintableBundle, error) {
+	stamp, err := configadapter.LoadStamp(bun)
```
carolynvs, line 193: Instead of returning an error, one way to deal with non-porter created bundles is to set Stamp to an empty struct when it can't be loaded, e.g. `Stamp{}`. Then only print "Porter Version" in the output if it's not empty.

You can make a duplicate of your new test where the stamp isn't set to validate that it's not included in the output.

carolynvs (changes requested):
Sorry I should have mentioned this up-front in the issue but even I forgot about it. 😊 We should be able to use porter explain even with bundles that were not created by Porter.

Everything else looks good, we just need to make sure that after this is merged, explain works for both Porter created bundles and vanilla cnab bundles.

joshuabezaleel:
Hi @carolynvs , truly sorry for a late reply and update... and also I think I messed up with how I merged the main branch to mine, I'm really sorry in advance .. I've updated according to the review for handling non-Porter bundle. Looking forward to the review :)

pkg/porter/explain.go
joshuabezaleel: This is for handling the non-Porter bundle by assigning Stamp's empty struct when it returns error.

pkg/porter/explain.go
```diff
@@ -298,6 +308,9 @@ func (p *Porter) printBundleExplainTable(bun *PrintableBundle) error {
 	fmt.Fprintf(p.Out, "Name: %s\n", bun.Name)
 	fmt.Fprintf(p.Out, "Description: %s\n", bun.Description)
 	fmt.Fprintf(p.Out, "Version: %s\n", bun.Version)
+	if bun.PorterVersion != "" {
+		fmt.Fprintf(p.Out, "Porter Version: %s\n", bun.PorterVersion)
```
joshuabezaleel, line 312: This is to omit printing `porterVersion` when its empty.

pkg/porter/explain_test.go
```diff
@@ -324,6 +352,45 @@ func TestExplain_generatePrintableBundleCreds(t *testing.T) {
...
+		},
+	}
+
+	pb, err := generatePrintable(bun, "")
+	assert.NoError(t, err)
+
+	assert.Equal(t, "v0.30.0", pb.PorterVersion)
+}
+
+func TestExplain_generatePrintableBundlePorterVersionNonPorterBundle(t *testing.T) {
```
joshuabezaleel, line 378: @carolynvs and this is the new test for the non-Porter bundle with no `Custom` property.

carolynvs:
@joshuabezaleel I did my best to fixup this PR so that it only has your changes in it. Want to take a look and let me know if this looks like all of your work?

https://github.com/getporter/porter/compare/main...carolynvs:pr1422

If it looks okay, I can update your PR to match. You will still be the committer, it just shows my name on it too since I touched it up.

joshuabezaleel:
> @joshuabezaleel I did my best to fixup this PR so that it only has your changes in it. Want to take a look and let me know if this looks like all of your work?
> 
> [main...carolynvs:pr1422](https://github.com/getporter/porter/compare/main...carolynvs:pr1422)
> 
> If it looks okay, I can update your PR to match. You will still be the committer, it just shows my name on it too since I touched it up.

Yes, the files mentioned on the branch was my work except for the first two markdown files under the `/docs/content/cli/` folder. So sorry for the trouble and thank you very much for the help, @carolynvs ..

carolynvs:
Thanks for confirming! I'll get your PR updated to match. Those extra content files I think are good to check-in as well, I'll double check but I think I forgot to update the generated docs for 2 pages on the website and they need to be checked-in anyway.

carolynvs:
I just confirmed that your changes survived my cleanup of the PR.

```
$ porter explain
Name: porter-hello
Description: An example Porter configuration
Version: 0.1.0
Porter Version: v0.35.0-12-g2b226484

No credentials defined

Parameters:
Name     Description   Type     Default            Required   Applies To
config                 object   {"color":"blue"}   false      All Actions

No outputs defined

No custom actions defined

No dependencies defined


$ porter explain -r getporter/porter-hello:v0.1.0
Name: HELLO
Description: An example Porter configuration
Version: 0.1.0

No credentials defined

No parameters defined

No outputs defined

No custom actions defined

No dependencies defined
```

carolynvs (approved):
This is great! 💯 Thank you for adding in this functionality. I will merge as soon as the builds pass and then cut a new release with your PR.

carolynvs:
@joshuabezaleel Let me know if you are looking for another issue to tackle!

joshuabezaleel:
Thank you very much for the really kind guide and help, @carolynvs ! 🙂 Truly grateful for it, thank you for making open source a learning safe-space for beginners like me. 
Yep, I have several issues in mind to have a shot on. Will comment on the issue and also tinker with Porter on my own too! 🙂

PR #1425, opened 2021-01-26
Add action option to explain command

### LINKED ISSUE #1416: Add --action to porter explain
When I am using `porter explain` I often am just looking for what I _have to do_ to run a certain action, usually install. It would be great if I could type `porter explain --action install` and have porter filter out unused parameters (and credentials once we have applyTo for creds too).

joshuabezaleel:
Signed-off-by: joshuabezaleel <joshua.bezaleel@gmail.com>

# What does this change
Enable `--action <action>` flag on `explain` command and only print corresponding credentials and parameters according to the action.

# What issue does it fix
Closes #1416 

# Notes for the reviewer
Screenshot is shown below:
image

# Checklist
- [x] Unit Tests
- [x] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/pkg/porter/explain.go
+++ b/pkg/porter/explain.go
@@ -236,7 +236,7 @@ func generatePrintable(bun bundle.Bundle, actionOption string) (*PrintableBundle
 		pp.Required = v.Required
 		pp.Description = v.Description
 
-		if pp.ApplyTo == "All Actions" || pp.ApplyTo == actionOption {
+		if pp.ApplyTo == "All Actions" || pp.ApplyTo == actionOption || actionOption == "" {
 			params = append(params, pp)
 		}
 	}
@@ -257,7 +257,7 @@ func generatePrintable(bun bundle.Bundle, actionOption string) (*PrintableBundle
 		po.ApplyTo = generateApplyToString(v.ApplyTo)
 		po.Description = v.Description
 
-		if po.ApplyTo == "All Actions" || po.ApplyTo == actionOption {
+		if po.ApplyTo == "All Actions" || po.ApplyTo == actionOption || actionOption == "" {
 			outputs = append(outputs, po)
 		}
 	}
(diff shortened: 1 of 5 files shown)


carolynvs:
I am looking into why the build is failing on an unrelated area of the test suite.

pkg/porter/explain.go
```diff
@@ -253,7 +257,9 @@ func generatePrintable(bun bundle.Bundle) (*PrintableBundle, error) {
 		po.ApplyTo = generateApplyToString(v.ApplyTo)
 		po.Description = v.Description
 
-		outputs = append(outputs, po)
+		if po.ApplyTo == "All Actions" || po.ApplyTo == actionOption || actionOption == "" {
```
carolynvs, line 260: Thanks for catching that outputs also should be filtered! 👍 

The `Output` type also has an AppliesTo function that we can use

```suggestion
		if v.AppliesTo(action){
```

pkg/porter/explain_test.go
```diff
@@ -210,7 +210,7 @@ func TestExplain_generatePrintableBundleParams(t *testing.T) {
 		},
 	}
 
-	pb, err := generatePrintable(bun)
+	pb, err := generatePrintable(bun, "")
```
carolynvs, line 213: Right now the existing tests aren't enough to validate your change. Let's copy this test `TestExplain_generatePrintableBundleParams` and change it so that one of the parameters has `ApplyTo` set to "install". Then we can check that when we provided "install" it shows up and when we provide "upgrade" as the action, it is not present in the generated parameters.

carolynvs (changes requested):
This is on the right track! I have made some suggestions to simplify the code and once we have a test in place this should be all set. 👍

joshuabezaleel:
Hi @carolynvs ! I've made changes according to the comments, thank you very much for the review! 🙂 
image

carolynvs (approved):
Perfect! Thank you so much for implementing this, I know that @squillace is going to use it and so lots of other users. 🎉

squillace:
oh YES!!!!

carolynvs:
@joshuabezaleel If you like, please open a PR and include your name on our contributors page! 🎉 

https://github.com/getporter/porter/blob/main/CONTRIBUTORS.md

PR #1428, opened 2021-01-27
Fix the service fabric example bundle

carolynvs:
# What does this change
Bump the base image and update the commands to match the updated instructions.

# What issue does it fix
Broken builds

# Notes for the reviewer
We should review our examples and consider moving them into a separate repo and build, triggered when new releases of porter are created. Some of the examples when they fail have nothing to do with if porter is working or not, and it's are slow build build them all.

# Checklist
- [ ] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)


carolynvs:
Merge like the build is broken! 💨

Issue #1449, opened 2021-02-16
Cannot build with a different version and then publish


carolynvs:
## Describe the bug
When I override the version of a bundle during porter build, porter publish always fails.

## To Reproduce
1. Create a bundle with `porter create` and change the name of the bundle to something that isn't already in your docker registry, e.g. `name: unique-bug-report`
2. Run porter build and set a version that isn't the same as what is in the porter.yaml, e.g. `porter build --version 0.0.0`
3. Run porter publish (either with or without specifying --registry).

## Expected behavior
The bundle is published using the version specified during build. Instead it fails to publish.

## Porter Command and Output
```
$ porter publish
Error: could not tag image getporter/fix-bug-installer:v0.1.0 with value getporter/fix-bug-installer:v0.0.0: Error response from daemon: No such image: getporter/fix-bug-installer:v0.1.0
```

## Version
v0.33.0 (991a5a23)

PR #1450, opened 2021-02-16
fix(publish.go): get original image value after loading internal manifest

vdice:
# What does this change

* Re-orders logic in publish to load the internal manifest prior to getting the original invocation image name (which may be updated and compared later on).

# What issue does it fix
Closes https://github.com/getporter/porter/issues/1449

# Notes for the reviewer

# Checklist
- [x] ~~Unit~~ Integration Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/pkg/porter/publish.go
+++ b/pkg/porter/publish.go
@@ -96,9 +96,6 @@ func (p *Porter) publishFromFile(opts PublishOptions) error {
 		return err
 	}
 
-	// Capture original invocation image name as it may be updated below
-	origInvImg := p.Manifest.Image
-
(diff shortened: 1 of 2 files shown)


carolynvs (commented):
The fix looks great. Do you see any existing integration tests that we can modify to test this out? Otherwise it would be great to have a test that uses a small bundle with a unique name, does a build with --version and then tries to publish to the local docker registry.

It seems like it would be really easy to get the ordering of this lost during a refactor so I'd feel better with a regression test.

carolynvs (approved):
🚢 I'll cut a release after this merges. Thank you!

PR #1452, opened 2021-02-16
Enable testing cdn from alt domain

carolynvs:
Allows us to test from the cdn2.porter.sh URL as well.


carolynvs:
Merge like everything is on fire 🔥

PR #1455, opened 2021-02-17
Publish mixins feed to /mixins/atom.xml

carolynvs:
# What does this change
Originally we had only a single feed, just for mixins, and it was located at /atom.xml. When we introduced plugins, we changed the location of the feed to /mixins/atom.xml but for backwards compatibility reasons we continued to publish the file to /atom.xml and used a redirect for /mixins/atom.xml.

That was over a year ago and at this point, no one should be using clients that old. So this feels safe to move.

By publishing the file where it Porter looks for it when installing mixins, we don't need a rule on our CDN to do that rewrite which makes things less complicated for us.

# What issue does it fix
None, but it helps with our current CDN issues.

# Notes for the reviewer
N/A

# Checklist
- [ ] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -123,9 +123,9 @@ publish-mixins:
 	$(MAKE) $(MAKE_OPTS) publish MIXIN=exec -f mixin.mk
 
 	# Generate the mixin feed
-	az storage blob download -c porter -n atom.xml -f bin/atom.xml
-	bin/porter mixins feed generate -d bin/mixins -f bin/atom.xml -t build/atom-template.xml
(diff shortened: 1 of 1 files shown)


carolynvs:
People with v0.22.2-beta.1 or later are unaffected. If someone is using an older version (from before 2020-01-22) then they need to upgrade to see new mixin versions published to our mixin feed.

vdice (approved):

PR #1460, opened 2021-02-19
Move common publish build scripts into mage

### LINKED ISSUE #750: Mixin feed updates should use a blob lease
When we publish mixins, we should use blob leases, so that we don't accidentally clobber the atom.xml file.

carolynvs:
# What does this change
Move the publish logic for mixins and the mixin feed into a shared library so that other repositories can reuse it. As long as I was in here changing the publish logic, I implemented blob leases to remove the publish race condition around atom.xml.

Mixins should begin moving to using mage as the backing implementation behind their make targets, starting with `make publish`.

The following import in the magefile imports the shared targets declared in Porter.
```go
// +mage import
import "get.porter.sh/porter/mage/releases"
```

```console
$ mage -l
Targets:
  publishMixin        Publish a mixin's binaries.
  publishMixinFeed    Generate an updated mixin feed and releases it.
```

See https://github.com/getporter/kubernetes-mixin/pull/28 for what the full migration looks like.

# What issue does it fix
As we change the publish logic (like this week we moved atom.xml) this will allow mixins to easily keep up-to-date by bumping their go.mod. Some mixins may choose to always use latest of porter as well which is something we should think about.

Closes #750 

# Notes for the reviewer
None

# Checklist
- [ ] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)


mage/releases/publish.go
```diff
@@ -0,0 +1,73 @@
...
+	mgx.Must(err)
+	defer shx.RunE("az", "storage", "blob", "lease", "release", "-c", ContainerName, "-b", mixinFeedBlob, "--lease-id", leaseId)
+
+	must.RunE("az", "storage", "blob", "download", "-c", ContainerName, "-n", mixinFeedBlob, "-f", mixinFeedFile, "--lease-id", leaseId)
+	must.RunV("bin/porter", "mixins", "feed", "generate", "-d", filepath.Dir(mixinFeedFile), "-f", mixinFeedFile, "-t", "build/atom-template.xml")
+	must.RunV("az", "storage", "blob", "upload", "-c", ContainerName, "-n", mixinFeedBlob, "-f", mixinFeedFile, "--content-cache-control", VolatileCache, "--lease-id", leaseId)
+}
+
+// Tries to get a lock on the mixin feed in blob storage, returning the lease id
+func lockMixinFeed(ctx context.Context) (string, error) {
```
vdice, line 48: It could be handy for this method to return the necessary cleanup (lease release) function... then `PublishMixinFeed` could be updated to `defer leaseReleaseFunc` (or some such).

vdice (approved):
Such a good idea!  It hadn't clicked until this work that Mage, in addition to other gains, could serve to unify build/CI logic spread out among Porter's various repos (in this case, mixins).  This is much-needed.  LGTM!

PR #1461, opened 2021-02-19
Enable dependabot for cnab-go

carolynvs:
Let's try out using dependabot on our project and start with just cnab-go.
I don't want to be overwhelmed with a bunch of updated deps and would like
to see first what it does with our biggest dependency.

Oops! I am testing this on my fork before we merge this, ignore please!

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    allow:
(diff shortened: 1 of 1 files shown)


carolynvs:
I'm not sure what I wanted it to do but it doesn't seem to pay attention to the replace directive. So it's bumping the base dependency which has no effect.

https://github.com/carolynvs/porter/pull/2

carolynvs:
Okay yeah found it in their docs/repo: They don't update replace directives because they are considered "pinned" https://github.com/dependabot/dependabot-core/pull/1221

PR #1466, opened 2021-02-23
Do not publish untagged or latest to mixin feed

carolynvs:
# What does this change
The file structure in the bin is used when generating the mixin feed. So
if there is a folder named "latest" then a version "latest" is put in
the atom.xml. A previous commit caused an untagged release to be in the
bin, instead of being renamed to canary and uploaded to our feed which
broke installing porter.

I have moved the directory prep into mage, calling it in make xbuild-all
so that it is guaranteed to happen. I have also added a check in the
atom feed generation to ignore those releases if found.

Now after you cross build, your bin looks like this
```
bin/
  mixins/
    exec/
      canary/... OR
      v1.2.3/...
```

# What issue does it fix
The recent change in #1460 caused a regression in generating the mixin feed. I have already manually fixed the atom.xml (because it is appended to during publish, so we needed to get the bad entries out).

# Notes for the reviewer
Oops, sorry!

Here is the draft PR for fixing the k8s mixin which has already adopted the previous change.
https://github.com/getporter/kubernetes-mixin/pull/30

# Checklist
- [x] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/pkg/pkgmgmt/feed/generate.go
+++ b/pkg/pkgmgmt/feed/generate.go
@@ -70,11 +70,14 @@ func (feed *MixinFeed) Generate(opts GenerateOptions) error {
 		if len(matches) > 0 {
 			version := matches[2]
 
-			// As a safety measure, skip versions that shouldn't be put in the feed, "latest" and non tagged releases.
-			// Normally the makefile prepares the bin ahead of time, just bail out just in case.
-			untaggedRegex := regexp.MustCompile(`v\d+\.\d+\.\d+-\d+-g[a-z0-9]{8}`)
-			if version == "latest" || untaggedRegex.MatchString(version) {
-				return nil
+			// As a safety measure, skip versions that shouldn't be put in the feed, we only want canary and tagged releases.
+			if version != "canary" {
+				versionRegex := regexp.MustCompile(`v\d+\.\d+\.\d+(-\d+-g[a-z0-9]{8})?`)
+				matches := versionRegex.FindStringSubmatch(version)
+				if len(matches) == 0 || // e.g. latest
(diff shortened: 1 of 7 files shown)


pkg/pkgmgmt/feed/generate.go
```diff
@@ -69,6 +69,14 @@ func (feed *MixinFeed) Generate(opts GenerateOptions) error {
 		matches := mixinRegex.FindStringSubmatch(path)
 		if len(matches) > 0 {
 			version := matches[2]
+
+			// As a safety measure, skip versions that shouldn't be put in the feed, "lates" and non tagged releases.
```
vdice, line 73:
```suggestion
			// As a safety measure, skip versions that shouldn't be put in the feed, "latest" and non tagged releases.
```

pkg/pkgmgmt/feed/generate_test.go
```diff
@@ -54,6 +54,15 @@ func TestGenerate(t *testing.T) {
 	tc.FileSystem.Chtimes("bin/canary/exec-linux-amd64", up10, up10)
 	tc.FileSystem.Chtimes("bin/canary/exec-windows-amd64.exe", up10, up10)
 
+	// Create extraneous release directories that should be ignores
```
vdice, line 57:
```suggestion
	// Create extraneous release directories that should be ignored
```

vdice (approved):

carolynvs:
@vdice Can you take another look? I made more changes that kind of invalidate your first approval. 😀

vdice (approved):

PR #1474, opened 2021-02-25
Simplify publish

carolynvs:
Create directories for latest/canary and version. It results in more files during publish but avoids us looking for binaries in the wrong spot. The mixin feed publish is smart now and will ignore extra directories.


carolynvs:
The build is failing because the latest version out there is constantly failing with HTTP_1_1_REQUIRED errors. My fix for http2 isn't live yet because of the bug I'm trying to fix in this PR. ☠️

vdice (approved):

PR #1475, opened 2021-02-26
Add debugging to porter downloads

carolynvs:
When we interact with cdn.porter.sh pass in extra flags so that if the command fails we can get a debug reference. When we download a file using porter, set the azure debug header and our own unique per request user agent string so that we can correlate errors on the client and server logs.

Also always set HTTP/1 when building the porter docker images, since it just failed with one of the errors we are tracking down:

Step 4/5 : RUN sh -c 'curl https://cdn.porter.sh/${PERMALINK}/install-linux.sh | sh' &&     ln -s /root/.porter/porter /usr/local/bin/porter
 ---> Running in b81a52ac1ca9
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  1237  100  1237    0     0   2181      0 --:--:-- --:--:-- --:--:--  2177
Installing porter to /root/.porter
curl: (56) OpenSSL SSL_read: Connection reset by peer, errno 104

### DIFF
--- a/build/images/client/Dockerfile
+++ b/build/images/client/Dockerfile
@@ -3,7 +3,7 @@ FROM alpine:3
 ARG PERMALINK
 
 RUN apk add curl --no-cache
-RUN sh -c 'curl https://cdn.porter.sh/${PERMALINK}/install-linux.sh | sh' && \
+RUN sh -c 'curl --http1.1 -v -H "X-Azure-DebugInfo: 1" -A "curl build-porter-client" https://cdn.porter.sh/${PERMALINK}/install-linux.sh | sh' && \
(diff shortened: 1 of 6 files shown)


vdice (approved):
🕵️‍♀️

carolynvs:
I need our builds to be like 2 minutes max so that I don't have to wait.

i hate waiting from the princess bride

PR #1476, opened 2021-02-26
Support applyTo for credentials

### LINKED ISSUE #877: Don't require credentials when applyTo makes them unnecessary

carolynvs:
# What does this change
Add the applyTo field to credentials. Works just like parameters.

# What issue does it fix
Closes #877 

# Notes for the reviewer
* This uses a pending PR in cnab-go, so once it is merged I will bump our cnab-go reference.

# Checklist
- [x] Unit Tests
- [x] Documentation
- [x] Schema (porter.yaml)


carolynvs:
@vdice Okay, the change I was relying upon in cnab-go has been merged. PTAL! 👀

vdice (approved):
LGTM!

PR #1477, opened 2021-02-26
Set a PORTER_TRACE agent value in install scripts

carolynvs:
I'm having trouble correlating failures when building the porter docker client or during check install so I'm adding a unique PORTER_TRACE agent value there as well.

### DIFF
--- a/build/images/client/Dockerfile
+++ b/build/images/client/Dockerfile
@@ -3,7 +3,7 @@ FROM alpine:3
 ARG PERMALINK
 
 RUN apk add curl --no-cache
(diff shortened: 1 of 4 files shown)


vdice (approved):

carolynvs:
I did not rewrite the powershell script because I don't believe there is a way to change just the http protocol version with the method we are using currently. We'd have to switch something else, like, but I'm concerned about rewriting it completely just to add the troubleshooting header. If we need to, we can sync up with @squillace and come up with the appropriate replacement and test it.

```
$r = [system.net.httpwebrequest]::create(URL)
$r.ProtocolVersion = [system.net.httpversion]::version10
```

PR #1482, opened 2021-03-04
Fix yaml <-> json roundtrip

### LINKED ISSUE #1483: null values in custom definitions become the string "<nil>" once published
## Describe the bug
When a field of an object defined in the custom section contains a null value, this becomes the literal string "&lt;nil>" instead of an actual null in the json of the manifest published to the repository.

## To Reproduce
1. Create a porter.yaml containing a 'custom' section like this:
```
custom:
  someObject:
    someValue: null 
``` 
2. Run `porter publish` to push the CNAB to a repository
3. Examine the manifest of the pushed CNAB; the JSON for the 'custom' property will contain this:
```
"someObject":{"someValue":"<nil>"}
```

## Expected behavior
The value of the field in the manifest should be null, not some string.

## Version
porter v0.33.1 (4359277d)

carolynvs:
# What does this change
I had a bug report for problems parsing a --yes flag as a bool instead of a string, causing the flag to be passed as --true.

```
install:
- exec:
    description: "use a yes flag"
    command: az
    arguments:
    - something
    flags:
      yes: ""
```
This is fixed in v3 of the yaml package, becauase it uses a newer version of yaml that doesn't interpret yes/no as boolean keys.

This version bump does change the indentation strategy of serializing to yaml but other than impacting our test strings, it's still valid yaml, just uses more spaces.

# What issue does it fix
Fixes #1483 where null wasn't being handled properly and other reports of roundtrip problem between yaml to json.

# Notes for the reviewer
I'm not sure if using v3 gives us any other capabilities. We were actually using it already in _one_ of our packages because of it's node manipulation. This just has us use the same version _everywhere_.

# Checklist
- [x] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)


pkg/exec/builder/action.go
```diff
@@ -6,8 +6,10 @@ import (
 	"io/ioutil"
 
 	"get.porter.sh/porter/pkg/context"
+	//"get.porter.sh/porter/pkg/yaml"
```
vdice, line 9: Looks like this line can be removed?

vdice (approved):

PR #1484, opened 2021-03-04
Bandaid on publish

carolynvs:
# What does this change
* Split publish into two steps. The build is failing on building the docker images because of our CDN problems. This makes it easier to rerun just the docker image part, without retriggering the Azure bug.
* Build the docker images using local binaries. Copy the local binaries from the bin instead of running porter's install script. It will reduce build errors and is more accurate in case of the CDN lagging.

# What issue does it fix
N/A

# Notes for the reviewer
I am testing this on a special test release pipeline for porter so that I will know that it works before merging. I'll link to the passed build when ... it passes. https://dev.azure.com/getporter/porter/_build/results?buildId=973&view=results

I think using two pipelines based on the same yaml, that we can possibly have a single script instead of maintaining two versions, one for PRs and another for the release. The pipelines can be configured with skips based on if the secret is set, instead of if it's a PR. The docker login step would need to change to use secrets instead of a configured service (dockerRegistryEndpoint).

*  automatic PR pipeline doesn't have access to secrets.
*  release pipeline has access to secrets for the prod env and cannot be triggered by a PR. 
*  test-release pipeline has access to secrets for the test env and can only be triggered by a maintainer with `/azp run test-porter-release`.

# Checklist
- [ ] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/build/azure-pipelines.release.yml
+++ b/build/azure-pipelines.release.yml
@@ -128,17 +128,34 @@ stages:
               path: bin
           - script: go run mage.go UseXBuildBinaries
             displayName: "Setup Bin"
+          - script: |
+              export AZURE_STORAGE_CONNECTION_STRING=$(AZURE_STORAGE_CONNECTION_STRING)
+              make publish-bin publish-mixins
+            condition: and(succeeded(), ne(variables['Build.Reason'], 'PullRequest'))
+            displayName: "Publish Porter Binaries"
+
+      - job: publish_docker
+        steps:
+          - task: GoTool@0
+            inputs:
+              version: "$(GOVERSION)"
+          - task: DownloadPipelineArtifact@2
+            displayName: "Download Cross-Compiled Porter Binaries"
+            inputs:
+              source: current
+              artifact: xbuild-bin
+              path: bin
+          - script: go run mage.go UseXBuildBinaries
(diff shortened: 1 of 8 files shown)


build/azure-pipelines.release.yml
```diff
@@ -118,52 +120,81 @@ stages:
       - job: publish_binaries
         steps:
           - task: GoTool@0
+            condition: and(succeeded(), eq(variables.HAVE_SECRETS, 'true'))
```
carolynvs, line 123: I tried a bunch of different ways to put the condition on the stage or job with zero success, so it's on every task...

vdice: Hah, I just returned from looking up the Azure Pipelines docs to see if we could do!  Of course, you would've tried :)

vdice (approved):

build/azure-pipelines.release.yml
carolynvs: If you can figure it out, we can fix in a followup. I spent an entire day trying different variants (on a fake pipeline so it would fail fast) and was unable to get it to work when the condition relied transiently upon a secret.

PR #1489, opened 2021-03-10
Add examples for using Porter as a library

carolynvs:
# What does this change
Porter can be used as a library in Go. The API is not stable. This demonstrates how to use Porter programmatically.

I've realized that we need to start making our library a bit easier to use programmatically. For example defaults set in the cmd/porter package aren't used, so they need to start being defaulted in the Validate command.

Ideally the command itself should do the validation and we should stop doing PreValidate with cobra but that would need to be a larger refactoring.

https://deploy-preview-1489--porter.netlify.app/library/ (Note this won't look right until we publish since has URLs to files that are introduced in this PR, see image below for screenshot)

# What issue does it fix
People have asked for examples of how to use Porter with Go instead of the CLI

# Notes for the reviewer
N/A

# Checklist
- [x] Unit Tests
- [x] Documentation
- [ ] Schema (porter.yaml)


carolynvs:
This is what the doc looks like after the files are in the final location. I'm using this so that our examples can get compile/test support without embedding code in the docs.

Screen Shot 2021-03-10 at 11 10 23 AM

pkg/porter/show.go
```diff
@@ -28,44 +28,54 @@ func (so *ShowOptions) Validate(args []string, cxt *context.Context) error {
...
+func (p *Porter) ShowInstallation(opts ShowOptions) error {
+	displayInstallation, err := p.GetInstallation(opts)
+	if err != nil {
+		return err
+	}
+
 	switch opts.Format {
 	case printer.FormatJson:
-		return printer.PrintJson(p.Out, installation)
+		return printer.PrintJson(p.Out, displayInstallation)
```
carolynvs, line 76: We always should have been outputing the dispaly installation for yaml|json, so this is a bug fix too.

vdice (approved):

PR #1494, opened 2021-03-11
Build the porter-agent image for the operator

carolynvs:
# What does this change
* The porter-agent image is customized for running Porter in kubernetes as
part of the Porter Operator.
* Dual publish to ghcr.io/getporter and Docker Hub. Eventually we will migrate to just GHCR (since they won't expire our images).

https://deploy-preview-1494--porter.netlify.app/docker-images/agent/

# What issue does it fix
No more building the operator agent locally on Carolyn's machine!

# Notes for the reviewer
N/A

# Checklist
- [ ] Unit Tests
- [x] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/build/azure-pipelines.release-template.yml
+++ b/build/azure-pipelines.release-template.yml
@@ -15,6 +15,9 @@ parameters:
   - name: shouldPublish
     type: boolean
     default: false
+  - name: dualPublish # Publish to the registry parameter and to ghcr.io while we migrate to GHCR
+    type: boolean
+    default: true
 
 stages:
   - stage: Validate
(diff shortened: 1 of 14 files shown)


carolynvs:
@vdice Sorry about that I'm still testing and accidentally clicked the "review for review button". It's still not working the way I want.

build/azure-pipelines.release-template.yml
```diff
@@ -158,12 +161,18 @@ stages:
           - script: go run mage.go UseXBuildBinaries
             displayName: "Setup Bin"
           - task: Docker@1
-            displayName: Docker Login
+            displayName: Docker Hub Login
             inputs:
               containerRegistryType: Container Registry
-              dockerRegistryEndpoint: getporter-registry
+              dockerRegistryEndpoint: ${{parameters.registry}}
```
carolynvs, line 167: In Azure DevOps I have created a couple docker registry authentication sets, so that depending on where we are publishing to, it uses the proper credentials.

carolynvs:
@vdice I have finally found the right combination of ✨ and 🔥 to get this working on the test-porter-release pipeline. PTAL!

build/azure-pipelines.release-template.yml
```diff
@@ -164,13 +167,23 @@ stages:
...
-            displayName: "Publish Docker Images"
+            displayName: "Publish Docker Images to ${{parameters.registry}}"
+          - task: Docker@1 # Support publishing to our old registry on Docker Hub and to our new location on GHCR while we migrate
+            displayName: GHCR Login
+            condition: ${{parameters.dualPublish}}
+            inputs:
+              containerRegistryType: Container Registry
+              dockerRegistryEndpoint: ghcr.io/getporter
+              command: login
+          - script: REGISTRY=ghcr.io/getporter make publish-images
```
vdice, line 184: Question here.  This line should either omit this `REGISTRY` value _or_ be set to `REGISTRY=${{parameters.registry}}`, right?  Otherwise, if I'm reading the make target correctly, it will publish to `ghcr.io/getporter` twice.

vdice (commented):
Looks great; just one area of uncertainty.

build/azure-pipelines.release-template.yml
carolynvs: What I am trying to do is always publish to parameters.registry (which is currently set to docker.io/getporter). Then when dual publish is set, publish to ghcr.io/getporter as well.

So the pairs of tasks are 

* login to parameters.registry (which is either getporter or getporterci)
* push to parameters.registry
* login to ghcr.io/getporter
* push to ghcr.io/getporter

Looks like azure devops supports for loops now, so an alternate way we could try it is to define parameters.registry as an array, https://stackoverflow.com/a/62871342

vdice (approved):

PR #1496, opened 2021-03-12
Use pipeline parameters

carolynvs:
# What does this change
1. Skip publishing with the new pipeline parameters feature in Azure Devops. This removes all those extra condition checks for publishing.
2. Have a "test" version of the release pipeline that is safe to run on a PR as needed to check changes to the build scripts. It publishes to getporterci in docker hub and a test storage account in azure.  Eventually we could use this same pipeline for PRs too, just vary up if it should publish or not.

The way we do separate pipelines using the same file is make different instances of the pipeline in Azure DevOps. 

* porter-release: has access to production env secrets, and publishes to getporter on docker hub. Triggers _only_ on main and tags, never PRs.
* test-porter-release: has access to test env secrets, publishes to getporterci on docker hub. Can be triggered manually with a comment from a maintainer.
* (potentially) porter: has no secrets, skips publish.

Example Run:
https://dev.azure.com/getporter/porter/_build/results?buildId=1025&view=results

# What issue does it fix
Removes a bunch of condition checks.

# Notes for the reviewer
N/A

# Checklist
- [ ] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)


carolynvs:
@vdice This is ready to review finally!

vdice (approved):
Oooh super clean!  LGTM.

Issue #1498, opened 2021-03-16
Porter publish does not reach into a specific file


jeffersonbenson:
## Describe the bug

In a repository I'm working with, if I am in the root of the directory and the porter.yaml file exists in a porter folder, I am unable to run `porter publish --file porter/porter.yaml` in order to publish the bundle. Additionally specifying an absolute path to the porter.yaml file seems to produce the same issue. 

It would appear that porter is failing to take into account a folder named porter, and removes it from the path

## To Reproduce
1. Place porter.yaml file in a folder called porter
2. From the higher folder run `porter publish --file porter/porter.yaml`
3. See error

## Expected behavior

Porter should be able to publish the bundle with the specified file

## Porter Command and Output
```
$ porter publish --file porter/porter.yaml
Building bundle ===>
Error: unable to generate manifest: could not read the manifest at "porter.yaml": open /<redacted>/porter.yaml: no such file or directory

```

## Version
Copy the output of `porter version` below

[e.g. porter v0.14.1-beta.2 (72dd5df)]

porter v0.34.4 (ccebf071)

carolynvs:
Gotcha, thanks for the issue. This is both a known problem with porter (always using the current directory for the build context) and looks like a newly introduced bug around locating the porter.yaml.

I think we need to introduce a second flag, similar to how `docker build -f /path/to/Dockerfile /path/to/build/context` works so that we aren't tied to the current directory. Then have the paths in the porter.yaml be in relation to the build context directory.

Let me know if that aligns with what you were hoping Porter would do, or if I'm misunderstanding the use case!

jeffersonbenson:
So for us the build context can totally be in the same directory as the porter.yaml, and probably should imho. 

For a larger context we're using a CI/CD pipeline to build and test bundles on PR from our devs, so right now the quick and dirty solution is just to have our build script cd into the directory with the porter bits and go from there, but it was a bit confusing to see the option for `--file` without the expected functionality.

carolynvs:
Agreed! That flag doesn't work the way it should. Thanks for the feedback on how you plan on using the flag.

A workaround until we fix the bug with --file is to ensure that the porter.yaml is in in the current directory where you run porter build. Depending on people's setup that may involve copying the porter.yaml some other place than where it is stored in the repo. 

1. named porter.yaml
2. run porter build without the --file flag from the same directory as porter.yaml

This really should be fixed, I just can't guarantee how soon, which is why I'm making sure people understand how it's broken and the workaround.

[...]

PR #1515, opened 2021-03-22
feat(build/publish): add/use --dir and --file flags

vdice:
# What does this change

* Allows users to build and publish with an optional path to the build/bundle context directory (`--dir`) and/or an optional path to the Porter manifest (`--file`, previously only exposed to `porter publish` and [not fully implemented](https://github.com/getporter/porter/issues/1498)).

# What issue does it fix
Closes https://github.com/getporter/porter/issues/1498

# Notes for the reviewer

Some "icky" bits of logic have been added to continue support for auto-building (if changes to the user-supplied manifest are detected) as well as conditionally adding the Dockerfile line that attempts to remove the user-supplied manifest from `BUNDLE_DIR` inside the installer image (now that this manifest -- optionally denoted by `--file` -- may exist separate from the build context dir -- optionally denoted by `--dir` -- we don't want to attempt to remove it when/if it hasn't been copied into `BUNDLE_DIR` in the first place).  See the latter two commits on this PR to take a closer look into these areas.

Also, as mentioned in some code comments, this PR currently uses the following assumptions:

1. If `--file path/to/porter.yaml` is provided only, ~~we assume all build/bundle assets should exist/be created in the directory path that the manifest exists in,  e.g. `path/to` in this example~~, we still assume build context is in the current working dir (".")
2. If `--file path/to/porter.yaml` and `--dir otherpath/to/bundle`, we assume all build/bundle assets should exist/be created in the directory path designated by `--dir`, e.g. `otherpath/to/bundle` in this example
3. If `--dir otherpath/to/bundle` is provided only, we assume that the Porter manifest exists in the current working directory (named `porter.yaml`) and *not* in the provided build context directory.   All build/bundle assets are assumed to exist (and will be created/updated) in `otherpath/to/bundle`.

Thoughts/suggestions/concerns?  For instance, for 3., I could potentially envision that we would want to update our assumption and expect that the Porter manifest indeed be located in the provided build context dir (designated by `--dir`).

# Checklist
- [x] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/cmd/porter/bundle.go
+++ b/cmd/porter/bundle.go
@@ -54,8 +54,11 @@ func buildBundleBuildCommand(p *porter.Porter) *cobra.Command {
   porter build --name newbuns
   porter build --version 0.1.0
   porter build --file path/to/porter.yaml
-  porter build --dir path/to/bundle
+  porter build --dir path/to/build/context
 `,
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			return opts.Validate(p.Context)
+		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return p.Build(opts)
 		},
--- a/cmd/porter/bundle.go
+++ b/cmd/porter/bundle.go
@@ -50,6 +50,12 @@ func buildBundleBuildCommand(p *porter.Porter) *cobra.Command {
 		Use:   "build",
 		Short: "Build a bundle",
 		Long:  "Builds the bundle in the current directory by generating a Dockerfile and a CNAB bundle.json, and then building the invocation image.",
+		Example: `  porter build
+  porter build --name newbuns
+  porter build --version 0.1.0
+  porter build --file path/to/porter.yaml
+  porter build --dir path/to/bundle
+`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return p.Build(opts)
 		},
@@ -60,6 +66,10 @@ func buildBundleBuildCommand(p *porter.Porter) *cobra.Command {
 	f.BoolVarP(&opts.Verbose, "verbose", "v", false, "Enable verbose logging")
 	f.StringVar(&opts.Name, "name", "", "Override the bundle name")
 	f.StringVar(&opts.Version, "version", "", "Override the bundle version")
+	f.StringVarP(&opts.File, "file", "f", "",
+		"Path to the Porter manifest. Defaults to `porter.yaml` in the current directory.")
+	f.StringVarP(&opts.Dir, "dir", "d", "",
+		"Path to the build context directory where all bundle assets are located.")
 
 	return cmd
 }
@@ -72,6 +82,10 @@ func buildBundleLintCommand(p *porter.Porter) *cobra.Command {
 		Long: `Check the bundle for problems and adherence to best practices by running linters for porter and the mixins used in the bundle.
 
 The lint command is run automatically when you build a bundle. The command is available separately so that you can just lint your bundle without also building it.`,
+		Example: `  porter lint
+  porter lint --file path/to/porter.yaml
+  porter lint --output plaintext
+`,
 		PreRunE: func(cmd *cobra.Command, args []string) error {
 			return opts.Validate(p.Context)
 		},
@@ -297,6 +311,7 @@ func buildBundlePublishCommand(p *porter.Porter) *cobra.Command {
 Note: if overrides for registry/tag/reference are provided, this command only re-tags the invocation image and bundle; it does not re-build the bundle.`,
 		Example: `  porter bundle publish
   porter bundle publish --file myapp/porter.yaml
+  porter bundle publish --dir myapp
   porter bundle publish --archive /tmp/mybuns.tgz --reference myrepo/my-buns:0.1.0
   porter bundle publish --tag latest
   porter bundle publish --registry myregistry.com/myorg
@@ -311,6 +326,8 @@ Note: if overrides for registry/tag/reference are provided, this command only re
 
 	f := cmd.Flags()
 	f.StringVarP(&opts.File, "file", "f", "", "Path to the Porter manifest. Defaults to `porter.yaml` in the current directory.")
+	f.StringVarP(&opts.Dir, "dir", "d", "",
+		"Path to the build context directory where all bundle assets are located.")
 	f.StringVarP(&opts.ArchiveFile, "archive", "a", "", "Path to the bundle archive in .tgz format")
 	f.StringVar(&opts.Tag, "tag", "", "Override the Docker tag portion of the bundle reference, e.g. latest, v0.1.1")
 	f.StringVar(&opts.Registry, "registry", "", "Override the registry portion of the bundle reference, e.g. docker.io, myregistry.com/myorg")
--- a/pkg/build/dockerfile-generator.go
+++ b/pkg/build/dockerfile-generator.go
@@ -45,7 +45,7 @@ func (g *DockerfileGenerator) GenerateDockerFile() error {
 		fmt.Fprintln(g.Out, contents)
 	}
 
-	err = g.FileSystem.WriteFile("Dockerfile", []byte(contents), 0644)
+	err = g.FileSystem.WriteFile(DOCKER_FILE, []byte(contents), 0644)
 	return errors.Wrap(err, "couldn't write the Dockerfile")
 }
 
@@ -111,6 +111,7 @@ func (g *DockerfileGenerator) readAndValidateDockerfile(s *bufio.Scanner) ([]str
 
 func (g *DockerfileGenerator) getBaseDockerfile() ([]string, error) {
 	var reader io.Reader
+
 	if g.Manifest.Dockerfile != "" {
 		exists, err := g.FileSystem.Exists(g.Manifest.Dockerfile)
 		if err != nil {
@@ -185,7 +186,7 @@ func (g *DockerfileGenerator) buildMixinsSection() ([]string, error) {
 
 func (g *DockerfileGenerator) PrepareFilesystem() error {
 	// clean up previously generated files
-	g.FileSystem.Remove("Dockerfile")
+	g.FileSystem.Remove(DOCKER_FILE)
 
 	fmt.Fprintf(g.Out, "Copying porter runtime ===> \n")
 
--- a/pkg/build/dockerfile-generator.go
+++ b/pkg/build/dockerfile-generator.go
@@ -19,16 +19,24 @@ import (
 type DockerfileGenerator struct {
 	*config.Config
 	*manifest.Manifest
+	ManifestPath string
 	*templates.Templates
 	Mixins pkgmgmt.PackageManager
 }
 
-func NewDockerfileGenerator(config *config.Config, m *manifest.Manifest, tmpl *templates.Templates, mp pkgmgmt.PackageManager) *DockerfileGenerator {
+func NewDockerfileGenerator(
+	config *config.Config,
+	m *manifest.Manifest,
+	manifestPath string,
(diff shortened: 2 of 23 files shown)


vdice:
(Example build failure intended to be fixed in https://github.com/getporter/porter/pull/1517)

carolynvs:
I just merged the fix and kicked off the build again.

carolynvs:
Oh it just failed again with the same error. I guess you will need to rebase.

vdice:
😭 It looks like I hit a _different_ error: https://dev.azure.com/getporter/porter/_build/results?buildId=1098&view=logs&j=80fc20b5-bf9c-52e8-5f6f-1fdb6790c904&s=4304c483-acc4-5f4c-d091-add65b819655&t=b6215415-61fa-5101-eccb-e1abfdceed27&l=134

Trying a rerun before attempting to hunt it down...

carolynvs:
Success!

pkg/build/dockerfile-generator.go
```diff
@@ -143,11 +151,24 @@ func (g *DockerfileGenerator) getBaseDockerfile() ([]string, error) {
...
 func (g *DockerfileGenerator) buildPorterSection() []string {
-	return []string{
-		// Remove the user-provided Porter manifest as the canonical version
-		// will migrate via its location in .cnab
-		`RUN rm $BUNDLE_DIR/porter.yaml`,
+	// The user-provided manifest may be located separate from the build context directory.
+	// Therefore, only add lines to remove it from BUNDLE_DIR if its filepath
+	// contains the current working directory.
+	if strings.Contains(g.ManifestPath, g.Getwd()) {
+		manifestSubpaths := strings.SplitAfterN(g.ManifestPath, g.Getwd(), 2)
```
carolynvs, line 158: I'm having trouble following what's going on here inside the if clause. Are you trying to go from

ManifestPath: ~/stuff/porter.yaml
pwd: ~/stuff

to

[pwd, relative path to porter.yaml] ?

pkg/build/dockerfile-generator.go
```diff
@@ -143,11 +151,24 @@ func (g *DockerfileGenerator) getBaseDockerfile() ([]string, error) {
...
+		if len(manifestSubpaths) != 2 {
+			return []string{}
+		}
+		manifestSubpath := manifestSubpaths[1]
+		if exists, _ := g.FileSystem.Exists(manifestSubpath); exists {
```
carolynvs, line 163: Is manifest subpath supposed to be the relative path to the manifest in the current directory? I think if we use filepath.Rel for both the if statement and the logic here we can improve the readability of this code. What do you think?

pkg/build/dockerfile-generator_test.go
```diff
@@ -50,6 +50,90 @@ func TestPorter_buildDockerfile(t *testing.T) {
...
+
+	manifestPath := "separate/porter.yaml"
+	c.TestContext.AddTestFileContents(configTpl, manifestPath)
+
+	m, err := manifest.LoadManifestFrom(c.Context, manifestPath)
+	require.NoError(t, err, "could not load manifest")
+
+	// Now remove the local manifest to simulate it existing separate
+	// from the build context directory
+	c.TestContext.FileSystem.RemoveAll(manifestPath)
```
carolynvs, line 109: I think we can change this test to more closely mimic what we are testing for: the manifest is not in the build context directory.

```go
manifestPath := "/tmp/porter.yaml"
// check but I think pwd for the test is in a different directory than /tmp, so this should be enough to trigger the condition?
```

pkg/build/dockerfile-generator_test.go
```diff
@@ -214,7 +298,7 @@ func TestPorter_prepareDockerFilesystem(t *testing.T) {
 	require.NoError(t, err, "could not load manifest")
 
 	mp := mixin.NewTestMixinProvider()
-	g := NewDockerfileGenerator(c.Config, m, tmpl, mp)
+	g := NewDockerfileGenerator(c.Config, m, filepath.Join(c.Context.Getwd(), config.Name), tmpl, mp)
```
carolynvs, line 301: Does NewDockerfileGenerator assume that the manifest path is absolute? Does it call filepath.Abs to ensure that it is? If we did that, one it would catch errors if people call it wrong and two would let us be lazy in our tests.

pkg/cnab/config-adapter/adapter.go
```diff
@@ -20,14 +20,22 @@ const SchemaVersion = "v1.0.0"
...
+	cxt *context.Context,
+	manifest *manifest.Manifest,
+	manifestPath string,
+	imageDigests map[string]string,
+	mixins []mixin.Metadata,
+) *ManifestConverter {
 	return &ManifestConverter{
 		Context:      cxt,
 		Manifest:     manifest,
+		ManifestPath: manifestPath,
```
carolynvs, line 38: I'm curious if you looked into using `manifest.ManifestPath` instead of passing it around as an extra argument? Was the value not reliable?

If we could rely on that value, then we could revert a lot of the changes in this PR.

pkg/porter/build.go
```diff
@@ -123,6 +125,9 @@ func (p *Porter) getUsedMixins() ([]mixin.Metadata, error) {
 
 	var usedMixins []mixin.Metadata
 	for _, installedMixin := range installedMixins {
+		if p.Manifest == nil {
```
carolynvs, line 128: Did you notice which code path resulted in us trying to do a build when the manifest is nil? Seems like that shouldn't ever happen?

pkg/porter/cnab.go
```diff
@@ -131,8 +154,12 @@ func (o *sharedOptions) validateInstallationName(args []string) error {
 // defaultBundleFiles defaults the porter manifest and the bundle.json files.
 func (o *bundleFileOptions) defaultBundleFiles(cxt *context.Context) error {
 	if o.File != "" { // --file
-		bundleDir := filepath.Dir(o.File)
-		o.CNABFile = filepath.Join(bundleDir, build.LOCAL_BUNDLE)
+		// If o.Dir not set, assume the build context dir is in the same
```
carolynvs, line 157: Oh wait, this doesn't line up with what we discussed. I thought we were going to use the same semantics as docker build, which defaults the build context to the current working directory when unspecified.

So if you run `porter build -f porter/porter.yaml` and the build context is set to `.`.

If we are consistent with how docker does it, then we can set the default value of o.Dir to `"."` and remove a bit of logic in this file, no?

pkg/porter/stamp.go
```diff
@@ -22,7 +22,10 @@ func (p *Porter) ensureLocalBundleIsUpToDate(opts bundleFileOptions) error {
 
 	if !upToDate {
 		fmt.Fprintln(p.Out, "Building bundle ===>")
-		return p.Build(BuildOptions{})
+		// opts.File is non-empty, which overrides opts.CNABFile if set
+		// (which may be if a cached bundle is fetched e.g. when running an action)
+		opts.CNABFile = ""
```
carolynvs, line 27: I'm unclear on why clearing opts.CNABFile is necessary. Was this an existing bug? Something that came up after this PRs changes?

pkg/porter/build.go
vdice: Yes, I believe this was was hit in lint-related test(s) in build_test; updated the tests and removed this check.

pkg/cnab/config-adapter/adapter.go
vdice: Good thinking!  Forgot all about that field.  Removed the extra arg.

pkg/build/dockerfile-generator.go
vdice: Yup, love the suggestion to use filepath.Rel -- way cleaner.

pkg/build/dockerfile-generator.go
vdice: (Yes; but this bit removed now that filepath.Rel is being used)

pkg/build/dockerfile-generator.go
```diff
@@ -143,11 +151,24 @@ func (g *DockerfileGenerator) getBaseDockerfile() ([]string, error) {
...
 
 func (g *DockerfileGenerator) buildPorterSection() []string {
-	return []string{
-		// Remove the user-provided Porter manifest as the canonical version
-		// will migrate via its location in .cnab
-		`RUN rm $BUNDLE_DIR/porter.yaml`,
+	// The user-provided manifest may be located separate from the build context directory.
+	// Therefore, only add lines to remove it from BUNDLE_DIR if its filepath
+	// contains the current working directory.
+	if strings.Contains(g.ManifestPath, g.Getwd()) {
```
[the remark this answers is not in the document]
vdice: Ah, thank you.  Much prefer filepath.Rel!

pkg/porter/cnab.go
vdice: Oops, indeed.  I was fooled/forgot that discussion when I looked at the previous behavior.  It should all be updated now to reflect what you mentioned above.  Thank you.

pkg/porter/stamp.go
vdice: Yes, I believe it was introduced now that we are passing in bundleFileOptions on the following line (previously, we called `p.Build()` without any).  We pass these in now so that when File is supplied on publish and an auto-rebuild is detected, that same File value migrates to the triggered build.

cmd/porter/bundle.go
```diff
@@ -60,6 +69,10 @@ func buildBundleBuildCommand(p *porter.Porter) *cobra.Command {
 	f.BoolVarP(&opts.Verbose, "verbose", "v", false, "Enable verbose logging")
 	f.StringVar(&opts.Name, "name", "", "Override the bundle name")
 	f.StringVar(&opts.Version, "version", "", "Override the bundle version")
+	f.StringVarP(&opts.File, "file", "f", "",
+		"Path to the Porter manifest. Defaults to `porter.yaml` in the current directory.")
+	f.StringVarP(&opts.Dir, "dir", "d", "",
+		"Path to the build context directory where all bundle assets are located.")
```
carolynvs, line 75: :excellent:

pkg/manifest/manifest.go
```diff
@@ -24,7 +24,7 @@ import (
 const invalidStepErrorFormat = "validation of action \"%s\" failed"
 
 type Manifest struct {
-	// ManifestPath is the location from which the manifest was loaded, such as the path on the filesystem or a url.
+	// ManifestPath is location to the original, user-supplied manifest, such as the path on the filesystem or a url
```
carolynvs, line 27: Oh nice, looks like you didn't end up needing to track the generated manifest path separately. 👍

pkg/porter/cnab_test.go
```diff
@@ -314,24 +329,8 @@ func Test_bundleFileOptions(t *testing.T) {
...
-				} else if tc.opts.File != "" && tc.opts.Dir != "" {
-					// if opts.File is set and opts.Dir is set, opts.CNABFile should use the dir from opts.Dir
-					require.Equal(t, tc.opts.File, tc.opts.File)
-					require.Equal(t, filepath.Join(tc.opts.Dir, build.LOCAL_BUNDLE), tc.opts.CNABFile)
-				} else {
-					// if opts.File and opts.Dir are unset, expect local defaults
-					require.Equal(t, config.Name, tc.opts.File)
-					require.Equal(t, build.LOCAL_BUNDLE, tc.opts.CNABFile)
-				}
+				require.Equal(t, tc.wantFile, tc.opts.File)
```
carolynvs, line 332: Dang, this test cleaned up real well. 👏

carolynvs (approved):
I really like how this ended up working out. Thanks for adding this into Porter!

PR #1534, opened 2021-04-08
Support mixins without descriptions

### LINKED ISSUE #1480: Mixin descriptions should be optional
Allow a mixin step to not specify a description, because I am lazy. The mixin is allowed to still require one but porter should be able to handle it being empty. When it is empty, if porter needs to print the description of the step, use some form of  "NAME mixin [index]"

ThorstenHans:
# What does this change
This PR adds support for mixins without a description.

# What issue does it fix
Closes #1480

# Notes for the reviewer
If changed `testdata/porter.yaml` and removed a description for `install>exec`

# Checklist
- [x] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/pkg/manifest/manifest.go
+++ b/pkg/manifest/manifest.go
@@ -770,7 +770,7 @@ func (s *Step) GetDescription() (string, error) {
 	children := s.Data[mixinName]
 	d, ok := children.(map[string]interface{})["description"]
 	if !ok {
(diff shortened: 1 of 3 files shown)


carolynvs (approved):
Looks great, thank you! 💥

Issue #1526, opened 2021-03-31
Porter archive --reference returns different filesizes


scottbuckel:
## Describe the bug
Porter archive returns different filesizes, so when running `signy verify`, the sha256 is different

## To Reproduce
```
`porter archive --reference [dockerhub]/intoto-demo-bundle6:v0.1.7 porter-bundle1.tgz`
`porter archive --reference [dockerhub]/intoto-demo-bundle6:v0.1.7 porter-bundle2.tgz`
`porter archive --reference [dockerhub]/intoto-demo-bundle6:v0.1.7 porter-bundle3.tgz`
`porter archive --reference [dockerhub]/intoto-demo-bundle6:v0.1.7 porter-bundle4.tgz`
```

Notice the filesizes are all within a few hundred bytes of each other

```
ls -al

drwxr-xr-x   8 [username] staff        256 Mar 31 15:10 .
drwxr-xr-x  13 [username] staff        416 Mar 31 14:13 ..
-rw-r--r--@  1 [username] staff       6148 Mar 31 14:32 .DS_Store
-rw-r--r--   1 [username] staff  234303385 Mar 31 14:51 porter-bundle.tgz
-rw-r--r--   1 [username] staff  234303448 Mar 31 15:08 porter-bundle2.tgz
-rw-r--r--   1 [username] staff  234303413 Mar 31 15:09 porter-bundle3.tgz
-rw-r--r--   1 [username] staff  234303444 Mar 31 15:10 porter-bundle4.tgz
-rw-r--r--   1 [username] staff  234303463 Mar 31 15:10 porter-bundle5.tgz
```

```➜  bin git:(master) ✗ shasum -a 256 porter-bundle.tgz
6b96f524d410d8f3f25346e6ab2ed987c67f310203197782d197772eb241bbcf  porter-bundle.tgz
➜  bin git:(master) ✗ shasum -a 256 porter-bundle2.tgz
7cc44f835bf13337d92abe78d238a1910be256bfcf167016fff873aeb2195bf4  porter-bundle2.tgz
➜  bin git:(master) ✗ shasum -a 256 porter-bundle3.tgz
108b4482a016583c9aac4519b4478f2eb657c5a2c71f6cd5e96638e25882c27c  porter-bundle3.tgz
➜  bin git:(master) ✗ shasum -a 256 porter-bundle4.tgz
2421bae898c2564b798eb6205476db003fd101f399734eb9f9819e72f3dedfed  porter-bundle4.tgz
```

[...]

PR #1537, opened 2021-04-08
fix(archive.go): repeatable shasum on archive

vdice:
# What does this change
* Adds logic to ensure identical archive shasums are produced, provided the same bundle is used

Uses the approach of zero-ing out the timestamps for all archive files to achieve this.  As seen in https://unix.stackexchange.com/questions/346789/compressing-two-identical-folders-give-different-result and suggested by @sebumd in the original issue https://github.com/getporter/porter/issues/1526

# What issue does it fix
Closes https://github.com/getporter/porter/issues/1526

# Notes for the reviewer

* The original timestamps for all files in an expanded archive will no longer be preserved
* For testing, I updated the archive integration test to _not_ use a randomized bundle name and reference, so that we get a deterministic archive hash for matching on

# Checklist
- [x] ~Unit~ Integration Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/pkg/porter/archive.go
+++ b/pkg/porter/archive.go
@@ -6,7 +6,9 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"time"
 
+	"github.com/carolynvs/aferox"
 	"github.com/cnabio/cnab-go/bundle"
 	"github.com/cnabio/cnab-go/imagestore"
 	"github.com/cnabio/cnab-go/imagestore/construction"
@@ -72,6 +74,7 @@ func (p *Porter) Archive(opts ArchiveOptions) error {
 	dest, err := p.Config.FileSystem.OpenFile(opts.ArchiveFile, os.O_RDWR|os.O_CREATE, 0644)
 
 	exp := &exporter{
+		fs:                    p.Config.FileSystem,
 		out:                   p.Config.Out,
 		logs:                  p.Config.Out,
 		bundle:                bun,
@@ -86,6 +89,7 @@ func (p *Porter) Archive(opts ArchiveOptions) error {
 }
 
 type exporter struct {
+	fs                    aferox.Aferox
 	out                   io.Writer
 	logs                  io.Writer
 	bundle                bundle.Bundle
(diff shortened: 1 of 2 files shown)


vdice:
Still trying to figure out why the archive shasum is still non-deterministic in the integration test on CI... thoughts @carolynvs ? Repeated runs locally pass... 🤔

carolynvs:
Sorry been focused 100% on the infra migration from this week and haven't had time yet to consider this problem. I won't get a chance to look at this until Monday.

carolynvs:
The error reproduces on my machine! TO THE DEBUGGER! 💨

carolynvs:
I am still looking into this. After your match the diff between the files is MUCH closer but drifts towards the end of the file.

carolynvs:
I just pushed an extra commit that uses github.com/mholt/archiver to handle creating the file. Locally that fixed the problem for me. Let's see what happens on CI.

A couple thoughts after I have worked on this:
1. I am reading that tgz will still have differences depending on the os it was run on. We may want to move to a different compression format, such as zip or maybe? bzip2 which doesn't have as much volatile stuff in the headers.
2. I switched to archiver because I wanted to eliminate "we just screwed up writing the archive code". We can either keep it (because it's easy to use) or figure out the difference and implement it ourselves again. Note that the current logic in my fix can write using _any_ compression format it goes off of whatever file extension they specified with `porter archive mybuns.EXT`. So if we want to keep this fix, we should decide if we care about the format, which format is reproducible and possibly add validation up-front on the extension.

carolynvs:
Also I want to update the archive test to run archive twice and compare the hash. So that we don't get false positives on a dev machine.

scottbuckel:
> 1. I am reading that tgz will still have differences depending on the os it was run on. We may want to move to a different compression format, such as zip or maybe? bzip2 which doesn't have as much volatile stuff in the headers.

That probably explains why `docker save`  defaults to .tar files I'm guessing? I bet .tar doesn't have the same behavior.

carolynvs:
I'll give just plain .tar files a go too!

carolynvs:
@vdice I checked docker save and also get different hashes on different computers. But on the same computer the hash is stable. I've updated the test to only check if the hash is the same between calls to porter archive, but doesn't assert that it matches some canonical hash (because there doesn't seem to be one).

carolynvs (approved):
LGTM. If we get more clever in the future, we can always revisit! 😀

PR #1542, opened 2021-04-12
Allow overriding home dir and download url during install

carolynvs:
# What does this change
When installing, let the user override with PORTER_HOME and PORTER_MIRROR the location where porter should be installed, or the location from which it should be downloaded.

I have updated the porter CLI to understand the new configuration value PORTER_MIRROR, which and also be specified as a flag or config file setting.

This lets us test out alternate download infra from end-to-end, and also allows a user to set up a mirror of Porter's assets.

I am also getting rid of the troubleshooting I had added for Azure's storage and CDN now that we aren't using it anymore (e.g. dynamically generated PORTER_TRACE user agents).

# What issue does it fix
Fixes https://groups.io/g/porter/message/30

# Notes for the reviewer
N/A

# Checklist
- [x] Unit Tests
- [x] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/docs/content/install.md
+++ b/docs/content/install.md
@@ -13,7 +13,7 @@ We have a few release types available for you to use:
 * [Older Version](#older-version)
 
 You can also install and manage [mixins](#mixins) and [plugins](#plugins) using
-porter, and use the [Porter VS Code Extension][vscode-ext] for help author
+porter, and use the [Porter VS Code Extension][vscode-ext] to help author
 bundles.
 
 All the scripts for Porter v0.37.3+ support [customizing the installation through parameters](#install-script-parameters).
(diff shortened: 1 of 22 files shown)


docs/content/install.md
```diff
@@ -13,9 +13,11 @@ We have a few release types available for you to use:
 * [Older Version](#older-version)
 
 You can also install and manage [mixins](#mixins) and [plugins](#plugins) using
-porter, and use the [Porter VS Code Extension][vscode-ext] for help authoring
+porter, and use the [Porter VS Code Extension][vscode-ext] for help author
```
vdice, line 16: Perhaps change back to `authoring` -- or '.. to help author bundles' ?

docs/content/install.md
```diff
@@ -129,3 +131,64 @@ All of the Porter-authored plugins are published to `https://cdn.porter.sh/plugi
 
 [releases]: https://github.com/getporter/porter/releases
 
+
+
+# Install Script Parameters
+
+The installation scripts provide the following parameters. Paramters can be specified with environment variables for the macOS and Linux scripts, and on Windows they are named parameters in the script.
```
vdice, line 138:
```suggestion
The installation scripts provide the following parameters. Parameters can be specified with environment variables for the macOS and Linux scripts, and on Windows they are named parameters in the script.
```

vdice (approved):

docs/content/install.md
carolynvs: I fixed a couple places with that same typo. I really have trouble with typing that word! 😊

PR #1545, opened 2021-04-26
Generate Dockerfile into .cnab/ directory

### LINKED ISSUE #1162: Rename, move, and hide Porter's generated Dockerfile

carolynvs:
# What does this change
* Generate a Dockerfile at .cnab/Dockerfile instead of in the root of the bundle directory. This avoids us overwriting a file that isn't owned by the bundle (e.g. they have the bundle defined at the root of their repo and also have their own Dockerfile). It also keeps all the rando files generated by Porter in a single tidy location we can all ignore.
* Cleanup the entire .cnab/ directory between builds instead of just the Dockerfile. Otherwise we copy in old mixins that may not be used anymore and generally open ourselves up to non-repeatable builds.

# What issue does it fix
Closes #1162 

# Notes for the reviewer

# Checklist
- [x] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -6,7 +6,7 @@ import (
 
 var (
 	// DOCKER_FILE is the file generated before running a docker build.
-	DOCKER_FILE = "Dockerfile"
+	DOCKER_FILE = filepath.Join(LOCAL_CNAB, "Dockerfile")
 
 	// LOCAL_CNAB is the generated directory where porter stages the /cnab directory.
 	LOCAL_CNAB = ".cnab"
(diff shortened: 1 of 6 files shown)


carolynvs:
I will squash when this is merged.

vdice (approved):
LGTM!

PR #1551, opened 2021-04-29
add support for setting working directory per step

### LINKED ISSUE #1403: Add working directory to mixins
I want to run a command from a particular directory in my bundle

I want a new field added to mixins where I can set the working directory. For example:

```yaml
exec:
  description: "Run ls in /tmp"
  command: bash
  dir: /tmp
  flags:
    c: ls
```

* At first we thought we could just use cd in an exec stanza and then run our command in the next stanza, but the changes made by cd do not persist between mixin executions.
* A workaround is ... a helper script which isn't porter-ish.

**Additional context**
I am typing this up for @squillace ❤️

ThorstenHans:
# What does this change
This PR adds support to change the working directory per step

```yaml
# ...
install:
  - exec:
      description: "uname"
      command: uname
      dir: "/home"
```

# What issue does it fix
Closes #1403

# Notes for the reviewer
I wasn't sure how to verify modification of the working directory when running tests. That's why I create a draft PR before diving into docs and schema

# Checklist
- [ ] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/pkg/exec/action.go
+++ b/pkg/exec/action.go
@@ -99,6 +99,7 @@ type Step struct {
 type Instruction struct {
 	Description     string        `yaml:"description"`
 	Command         string        `yaml:"command"`
+	WorkingDir      string        `yaml:"dir,omitempty"`
 	Arguments       []string      `yaml:"arguments,omitempty"`
 	SuffixArguments []string      `yaml:"suffix-arguments,omitempty"`
 	Flags           builder.Flags `yaml:"flags,omitempty"`
@@ -132,6 +133,10 @@ func (s Step) GetOutputs() []builder.Output {
 		outputs[i] = s.Outputs[i]
 	}
 	return outputs
+}
+
+func (s Step) GetWorkingDir() string {
+	return s.WorkingDir
 }
 
 var _ builder.OutputRegex = Output{}
--- a/pkg/exec/builder/execute_test.go
+++ b/pkg/exec/builder/execute_test.go
@@ -1,6 +1,9 @@
 package builder
 
 import (
+	"fmt"
+	"os"
+	"path/filepath"
 	"testing"
 
 	"get.porter.sh/porter/pkg/context"
@@ -133,4 +136,23 @@ func TestExecuteStep_HasOrderedArguments(t *testing.T) {
 
 	_, err := ExecuteStep(c.Context, step)
 	require.NoError(t, err, "ExecuteStep failed")
+}
+
+func TestExecuteStep_SpecifiesCustomWorkingDirectory(t *testing.T) {
+	c := context.NewTestContext(t)
+	c.UseFilesystem()
+	wd, _ := filepath.EvalSymlinks(os.TempDir())
+
+	step := TestOrderedStep{
+		TestStep: TestStep{
+			Command:          "pwd",
+			Arguments:        []string{},
+			WorkingDirectory: wd,
+		},
+		SuffixArguments: []string{},
+	}
+
+	_, err := ExecuteStep(c.Context, step)
+	assert.Equal(t, fmt.Sprintln(wd), c.GetOutput())
+	require.NoError(t, err, "Execute Step failed")
 }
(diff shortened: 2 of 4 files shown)


pkg/exec/builder/execute_test.go
```diff
@@ -134,3 +135,22 @@ func TestExecuteStep_HasOrderedArguments(t *testing.T) {
 	_, err := ExecuteStep(c.Context, step)
 	require.NoError(t, err, "ExecuteStep failed")
 }
+
+func TestExecuteStep_SpecifiesCustomWorkingDirectory(t *testing.T) {
+	c := context.NewTestContext(t)
+	wd := os.TempDir()
```
carolynvs, line 141:
```suggestion
      // Do not use a unit test filesystem. When a command is run, it should actually run the command on the computer.
      c.UseFilesystem()
      // Get the absolute path of the wording directory, resolving symbolic links
      wd := os.TempDir()
      wd, _ = filepath.EvalSymlinks(wd)
```

pkg/exec/builder/execute_test.go
```diff
@@ -134,3 +135,22 @@ func TestExecuteStep_HasOrderedArguments(t *testing.T) {
...
+
+	step := TestOrderedStep{
+		TestStep: TestStep{
+			Command:          "uname",
```
carolynvs, line 145: Let's have the command return us the current working directory so that we know that it was set properly
```suggestion
			Command:          "pwd",
```

pkg/exec/builder/execute_test.go
```diff
@@ -134,3 +135,22 @@ func TestExecuteStep_HasOrderedArguments(t *testing.T) {
...
+			Arguments:        []string{},
+			WorkingDirectory: wd,
+		},
+		SuffixArguments: []string{},
+	}
+
+	// 🤔 how to get the actually used cmd.Dir
+	c.Setenv(test.ExpectedCommandEnv, "uname")
```
carolynvs, line 153:
```suggestion
// We can remove these lines, we are going to have the command actually execute so we don't need to mock anything
```

pkg/exec/builder/execute_test.go
```diff
@@ -134,3 +135,22 @@ func TestExecuteStep_HasOrderedArguments(t *testing.T) {
...
+	_, err := ExecuteStep(c.Context, step)
+	require.NoError(t, err, "Execute Step failed")
```
carolynvs, line 155:
```suggestion
	require.NoError(t, err, "Execute Step failed")
	
	// Read the output from running the command
	wantOutput := wd
	gotOutput := c.GetOutput()
	assert.Contains(t, gotOutput, wantOutput, "pwd did not return the expected directory")
```

pkg/exec/action.go
```diff
@@ -99,6 +99,7 @@ type Step struct {
 type Instruction struct {
 	Description     string        `yaml:"description"`
 	Command         string        `yaml:"command"`
+	WorkingDir      string        `yaml:"dir"`
```
carolynvs, line 102: Since dir is optional, let's use `omitempty` here. This means that when dir is an empty string, it is not included in the yaml when marshaled to a string.

```suggestion
	WorkingDir      string        `yaml:"dir,omitempty"`
```

carolynvs (commented):
I've added suggestions to your test that should get it working. Please give it a try and let me know if the test passes when you apply the change.

pkg/exec/action.go
ThorstenHans: yeah. Makes sense, obviously.

carolynvs (approved):
Looks great! Thank you for implementing this.

The exec mixin will be able to immediately use this change once I make a new release. The other mixins can use it if they update the version of porter in their go.mod file and then add a function for GetWorkingDir().

If you are interested in updating any of the mixins, let me know. Otherwise I will get to them next week.

carolynvs:
@ThorstenHans There is one more thing that you need to do before I can merge this pull request. Can you please sign your commits, so that the DCO (developer certificate of origin) check passes? That's what lets us use your code contributions in our open source project.

https://porter.sh/contribute/guide/#signing-your-commits

If you aren't sure how to sign all of the commits (there are 3), it's okay to just use the command in that link and sign the last one. I will squash all of your commits into a single commit when I merge your pull request. Feel free to DM me for help if you run into trouble with git!

ThorstenHans:
Sure. No prob have squashed and force pushed with signing activated.

PR #1560, opened 2021-05-05
Blog: Migrate from Docker App to Porter

carolynvs:
This is a blog post that explains how to migrate from Docker App to Porter.

https://deploy-preview-1560--porter.netlify.app/blog/migrate-from-docker-app/


carolynvs:
This is failing because the new example uses docker-compose, which isn't installed by default by our makefile. Once #1558 is merged, this should be fixed I think. 🤞

vdice (approved):

PR #1567, opened 2021-05-08
Add new build driver for buildkit

### LINKED ISSUE #1268: Toggle buildkit features in porter.yaml
Porter `build` fails when using the `terraform` mixin with git-based modules.
```sh
Error: unable to build CNAB invocation image: failed to stream docker build output: The command '/bin/sh -c cd /cnab/app/terraform && terraform init -backend=false' returned a non-zero code: 1
```
Using the `--debug` flag clearly shows:
```sh
Error: Failed to download module         
                                                           
Could not download module "app" (main.tf:248) source code from
"git::ssh://git@my.gitlab.com/terraform-modules/app?ref=v1.0.0":
error downloading
'ssh://git@my.gitlab.com/terraform-modules/app?ref=v1.0.0':
/usr/bin/git exited with 128: Cloning into
'.terraform/modules/app'...
Host key verification failed.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

#### Quick and dirt solution
Execute `terraform init` before `porter build`.

#### More elaborated solution
Add a toggle to allow Porter (and mixins) to enable buildkit features on demand, _i.e_
```yaml
name: my-cnab
version: 0.1.0

mixins:
  - terraform:
      clientVersion: 0.12.24
      sshAgentForwarding: true
```

The mixin would then be instructed to use `--mount=type=ssh` when generating its Dockerfile instructions, _i.e_
```dockerfile
RUN --mount=type=ssh /usr/bin/terraform init -backend=false
```

Circumvented the build issue by directly create my own Dockerfile. **(not tested a 100%)**

Here is a Dockerfile sample:
```dockerfile
# syntax=docker/dockerfile:experimental

FROM alpine:3.12

ARG BUNDLE_DIR
ENV TERRAFORM_VERSION=0.12.24

RUN apk add --update-cache --no-cache ca-certificates openssh-client git
RUN mkdir -p -m 0600 /root/.ssh && ssh-keyscan my.gitlab.com >> /root/.ssh/known_hosts

RUN wget https://releases.hashicorp.com/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_linux_amd64.zip && \
    unzip terraform_${TERRAFORM_VERSION}_linux_amd64.zip -d /usr/bin

COPY . $BUNDLE_DIR

RUN --mount=type=ssh  cd /cnab/app/terraform && terraform init -backend=false

COPY . $BUNDLE_DIR
RUN rm -fr $BUNDLE_DIR/.cnab
COPY .cnab /cnab
COPY porter.yaml $BUNDLE_DIR/porter.yaml
WORKDIR $BUNDLE_DIR
CMD ["/cnab/app/run"]
```

**Additional context**
```
$ porter --version
porter v0.29.0 (5e7240cf)

$ porter mixins list
Name         Version         Author
arm          v0.8.1-beta.1   Porter Authors
aws          v0.3.1-beta.1   Porter Authors
az           v0.5.1-beta.1   Porter Authors
exec         v0.29.0         Porter Authors
gcloud       v0.4.1-beta.1   Porter Authors
helm         v0.13.2         Porter Authors
kubernetes   v0.28.1         Porter Authors
terraform    v0.6.0          Porter Authors
```

carolynvs:
# What does this change
- [x] Add --experimental flag to root porter command
- [x] Add experimental flags to config file. These features are disabled by default.
- [x] Add experimental flag for "build-drivers". This allows specifying the build driver, e.g. docker, buildkit, in the config file.
- [x] Add driver for buildkit.
- [x] Document drivers and how to configure them.
- [x] Document experimental flags
- [x] `porter create` generates a Dockerfile that uses buildkit when that feature is enabled

❓ As a follow-on should a bundle be able to say "I use buildkit!" and then fail if it's not enabled, or automatically pick that driver

example configuration to enable the feature:

**~/.porter/config.toml**
```toml
experimental = ["build-drivers"]
build-driver = "buildkit"
```

You can also set it with environment variables
```
export PORTER_EXPERIMENTAL=build-drivers (comma seperated)
export PORTER_BUILD_DRIVER=buildkit
```

or with flags
```
porter build --experimental build-drivers --driver buildkit
```

# What issue does it fix
Closes #1268 

# Notes for the reviewer
As I was working on this I realized that we have some problems around when the configuration is loaded, and accessing it without accidentally getting a nil pointer exception. So I did a bit of refactoring to address since it was blocking the --experimental flag. It was prohibitive to refactor in a separate PR so hopefully having this smooshed together is reviewable.

* Refactor porter's config and data structures so that we can't get nil pointer exceptions or try to access the configuration data before it is loaded.
* Move helper functions to Config so that they are easier to find.
* Remove helper functions for config values that aren't needed anymore because its safe to directly access the bound field going forward.
* Use a fork of viper with a patch for a bug in how we load configuration when a config file is not found. We should still bind to environment variables.
  * https://github.com/spf13/viper/pull/1135
  * other pr tbd

# Checklist
- [x] Unit Tests
- [x] Documentation
- [ ] Schema (porter.yaml)


cmd/porter/bundle.go
```diff
@@ -73,6 +76,13 @@ func buildBundleBuildCommand(p *porter.Porter) *cobra.Command {
 		"Path to the Porter manifest. Defaults to `porter.yaml` in the current directory.")
 	f.StringVarP(&opts.Dir, "dir", "d", "",
 		"Path to the build context directory where all bundle assets are located.")
+	f.StringVar(&opts.Driver, "driver", porter.BuildDriverDefault,
```
carolynvs, line 79: I added this flag to build but not publish (for auto-build). I have really mixed feels about the auto-build feature right now. It's super helpful but also makes configuration a bit more of a mess. I was wondering if it should just be a message that you need to rebuild, instead of automatic, just so that we aren't adding flags onto publish that would be confusing and result in making scripts repeat themselves. 🤔

pkg/config/config_test.go
```diff
@@ -36,3 +38,32 @@ func TestConfig_GetHomeDirFromSymlink(t *testing.T) {
 	// and on Windows, that means flipping the afero `/` to `\`.
 	assert.Equal(t, filepath.Join("/root", ".porter"), home)
 }
+
+func TestConfig_GetFeatureFlags(t *testing.T) {
+	t.Parallel()
+
+	t.Run("build builders defaulted to disabled", func(t *testing.T) {
```
vdice, line 45:
```suggestion
	t.Run("build drivers defaulted to disabled", func(t *testing.T) {
```

pkg/config/config_test.go
```diff
@@ -36,3 +38,32 @@ func TestConfig_GetHomeDirFromSymlink(t *testing.T) {
...
+		c := Config{}
+		assert.False(t, c.IsFeatureEnabled(experimental.FlagBuildDrivers))
+	})
+
+	t.Run("build builders enabled", func(t *testing.T) {
```
vdice, line 50:
```suggestion
	t.Run("build drivers enabled", func(t *testing.T) {
```

pkg/config/datastore.go
```diff
@@ -10,12 +11,20 @@ type Data struct {
 	// Only define fields here that you need to access from code
 	// Values are dynamically applied to flags and don't need to be defined
 
+	// BuildDriver is the driver to use when building bundles.
+	// Available values are: docker, buildx.
```
vdice, line 15:
```suggestion
	// Available values are: docker, buildkit.
```

pkg/context/helpers.go
```diff
@@ -133,6 +145,14 @@ func (c *TestContext) AddTestFileContents(file []byte, dest string) error {
...
 }
 
+// AddTestDirectoryFromRoot adds a test directory where the filepath is relative to the root of the repository.
+// Use this when the directory you are referencing is in a different directory than the test.
+func (c *TestContext) AddTestDirectoryFromRoot(srcDir, destDir string) {
+	pathFromRoot := filepath.Join(c.FindRepoRoot(), srcDir)
+	c.AddTestDirectory(pathFromRoot, destDir)
+}
+
+// AddTestDirectory adds a test directory where the filepath is relative to the test directory.
```
vdice, line 155: Do we want a similar check if srcDir contains `..` and call `T.Fatal(..)`?

pkg/experimental/experimental.go
```diff
@@ -0,0 +1,26 @@
...
+// FeatureFlags is an enum of possible feature flags
+type FeatureFlags int
+
+const (
+	// FlagBuildDrivers indicates if configurable build drivers are enabled.
+	FlagBuildDrivers FeatureFlags = iota + 1
+)
+
+// ParseFlags converts a list of feature flag names into a bit map for faster lookups.
+func ParseFlags(flags []string) FeatureFlags {
```
vdice, line 17: Neat, I'll have to keep this technique in mind.

pkg/porter/build.go
```diff
@@ -39,12 +39,38 @@ func (o *BuildOptions) Validate(cxt *context.Context) error {
...
+		// Syncing value back to the config and we will always use the config
+		// to determine the driver
+		// This would be less awkward if we didn't do an automatic build during publish
+		p.Data.BuildDriver = o.Driver
+	}
+
+	return o.bundleFileOptions.Validate(p.Context)
+}
+
+func stringSliceContains(allowedValues []string, value string) bool {
```
vdice, line 58: Should we move this to a more generic helpers file/location, or nah?

cmd/porter/bundle.go
vdice: I'd be down to create an issue from those thoughts.  Definitely relate to the sentiment that the code complexity around auto-build may be more trouble than it is worth...

vdice (commented):
Awesome work!  Only had minor suggestions.

One other minor issue I was curious about.  When I supply an invalid/unsupported build driver, the error isn't as expected:
```
 $ p build --experimental build-drivers --driver buildx
Copying porter runtime ===>
Copying mixins ===>
Copying mixin exec ===>

Error: unable to generate Dockerfile: error generating the Dockerfile: error loading default Dockerfile template: open templates/build/buildx.Dockerfile: file does not exist
```
In other words, I'm wondering if the cli can instead bubble up the specific error that the spec'd build driver isn't supported?

docs/content/custom-dockerfile.md
```diff
@@ -49,6 +49,15 @@ has root ssl certificates installed. *You must use a base image that is
 debian-based, such as `debian` or `ubuntu` with apt installed.* Mixins assume
 that apt is available to install packages.
 
+# Buildkit
+
+You can use [Buildkit] in your Dockerfile by enabling the [experimental] feature
+for [build-drivers], and then specifying buildkit as the driver.
+
+
```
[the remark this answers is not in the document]
carolynvs: Thanks there was supposed to be a corresponding example of the template for buildkit enabled Dockerfiles

pkg/context/helpers.go
carolynvs: Oh yeah! 👍 The goal of these is to let us use a path to the test file that doesn't get messed up as much when the test is moved. Relative paths are still vulnerable but usually the testdata dir moves with a test.

pkg/experimental/experimental.go
carolynvs: I could have converted to a map and then used key lookups, but what is nice about this is that it's faster than a map lookup and uses less space too.

pkg/porter/build.go
carolynvs: I think we could use this in a couple places. I just was dithering about naming a package `util`, and imagined what people would think. 😂 

Screen Shot 2021-05-14 at 8 05 57 AM

carolynvs:
I have created https://github.com/getporter/porter/discussions/1582 to get feedback on auto-build for publish.

I did have code that tried to validate the driver, looks like bad values can still slip through. I'll fix that.

vdice (commented):
One last thing to check.  When I supply a bogus driver in Porter's config, I'm not seeing the cli error out on `porter build` (no flags or env vars set).  Do you see the same?  I'm using:
```
experimental = ["build-drivers"]
build-driver = "foo"
```

vdice (approved):
LGTM!

carolynvs:
Sorry this got a bit weird. I realized that there were some gaps in our configuration scheme exposed by what I wanted to do in this PR. Hopefully we've got enough tests now to cover all the weird cases.

Issue #1571, opened 2021-05-10
Move dependpencies template variable from bundle to installation


carolynvs:
Change dependencies templating from bundle.dependencies.* to installation.dependencies.* and make more information available from the runtime (i.e. the installation name of a dependency).

bundle.* should be used for static information about the bundle, such as its name. The dependencies are things resolved at runtime.

carolynvs:
I feel like this is repainting the bike shed color. Closing.

PR #1573, opened 2021-05-10
Rearrange dependencies syntax to be nested in 'requires' object

### LINKED ISSUE #1570: Rearrange dependencies syntax in porter.yaml
We know that we want to change the dependencies structure to allow for a "provides" section. As long as we are having a 1.0, let's make the change to "requires" now so that when we roll out dependencies it's not another breaking change.

```yaml
dependencies:
  - name: mysql
     reference: mysql:v1
```

to 

```yaml
dependencies:
  requires:
    - name: mysql
       reference: mysql:v1
```

ThorstenHans:
# What does this change

This PR embeds dependencies in an `requires` object (`porter.yaml`).

# What issue does it fix
Closes #1570

# Notes for the reviewer

`pkg/porter/install_test.go` relies on test data which is pre-populated in the repo. At this point, I was not sure how test data has been generated and if I can replace it. Would be great to chat about that

# Checklist
- [ ] Unit Tests
- [x] Documentation
- [x] Schema (porter.yaml)


docs/content/dependencies.md
```diff
@@ -17,13 +17,14 @@ Here is a [full example][example] of a Porter manifest that uses dependencies.
 
 In the manifest, add entries for each dependency of your bundle. The `name` field takes a short name for the dependent bundle that
 you will use to reference the dependent bundle elsewhere in the bundle. For example you can reference the dependent bundle's
-outputs via `{{ bundle.dependencies.NAME.outputs }}`.  The `reference` field takes the bundle reference of the dependency.
+outputs via `{{ bundle.dependencies.requires.NAME.outputs }}`.  The `reference` field takes the bundle reference of the dependency.
```
carolynvs, line 20: I don't think that we want to change the template variables. Another issue for v1 (#1571) is to move the variable for dependencies out of bundle, under "installation", to make it more clear that it's a runtime variable.

carolynvs (changes requested):
Sweet! This is nearly there. You should do a sweep for any other porter.yaml files that have `dependencies` in it. That is why the one test is failing but we want to make sure that we find all of them.

Here is the one causing the test to fail: https://github.com/getporter/porter/blob/main/build/testdata/bundles/wordpress/porter.yaml

carolynvs (approved):
Thank you for tackling this larger change!

PR #1575, opened 2021-05-10
Fix using kind in integration tests

carolynvs:
# What does this change
This moves the cluster logic logic into a reusable mage package. People can override some values, like the name of the kind cluster, with KIND_NAME, etc.

# What issue does it fix
One of the tests is failing because I was calling the integration tests directly from mage instead of with the bash script. This removes that confusion.

# Notes for the reviewer
N/A

# Checklist
- [ ] Unit Tests
- [x] Documentation
- [ ] Schema (porter.yaml)


build/azure-pipelines.integration.yml
```diff
@@ -7,15 +7,9 @@ trigger:
     include:
       - refs/heads/main
 
-# Only test a PR when a smoke test or the runtime has changed
+# Only test a PR when an integration test or the runtime has changed
```
vdice, line 10: Should this comment be removed or updated now that this currently only runs manually on PRs?

vdice (approved):

PR #1594, opened 2021-05-18
Bunch of small updates to docs and blog

ThorstenHans:
# What does this change 

I went through the docs and applied following modifications:

- code blocks use `bash` not `console`
- consistent use of code-fences instead of code-fence / italic font mix
- fix smaller typos
- ensure language is specified for several code blocks 

# Checklist
- [ ] Unit Tests
- [x] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -99,7 +99,7 @@ before you've spent a lot of time implementing it.
 
 Since Porter is a CLI, the "solution" will usually look like this:
 
-```console
+```bash
 $ porter newcommand [OPTIONAL] [--someflag VALUE]
 example output
 ```
--- a/docs/content/author-bundles.md
+++ b/docs/content/author-bundles.md
@@ -157,7 +157,7 @@ install:
 
 The syntax to pass a parameter to porter is the same for both regular and file parameters:
 
-```console
+```bash
 $ porter install --param mytar=./my.tar.gz
 ```
 
@@ -397,7 +397,7 @@ Here is a breakdown of all the supported fields on an image in this section of t
 
 A last note on `digest`.  Taking the example of the library `nginx` Docker image, we can get the repository digest like so:
 
-```console
+```bash
  $ docker inspect nginx | jq -r '.[].RepoDigests'
 [
   "nginx@sha256:a93c8a0b0974c967aebe868a186e5c205f4d3bcb5423a56559f2f9599074bbcd"
--- a/docs/content/author-bundles.md
+++ b/docs/content/author-bundles.md
@@ -28,8 +28,7 @@ We have full [examples](https://porter.sh/src/examples) of Porter manifests in t
 
 ## Bundle Metadata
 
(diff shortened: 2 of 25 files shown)


CONTRIBUTING.md
```diff
@@ -99,7 +99,7 @@ before you've spent a lot of time implementing it.
 
 Since Porter is a CLI, the "solution" will usually look like this:
 
-```console
+```bash
```
carolynvs, line 102: There are a few reasons why we use console instead of bash

1. Very often we are demonstrating a command and it's output in the same block. We don't want the output colored as it's incorrect and distracting (like the random green sha below).

Screen Shot 2021-05-25 at 10 36 17 AM

2. Most code blocks are not bash specific. Porter works on all operating systems, including Windows in which case people are often using powershell.

3. Not much is added by highlighting different words of an example command differently. With console, the command is black text and the output is blue text. I have found that is a more meaningful highlighting schema for what we are doing (command+example).

docs/content/author-bundles.md
```diff
@@ -64,8 +62,8 @@ they need to run, such as a CLI or config files, and how to execute their steps
 Anyone can [create a mixin](/mixin-dev-guide/), here's a list of the mixins that are installed with Porter by default:
 
 * exec - run shell scripts and commands
-* helm - use the helm cli
-* azure - provision services on the Azure cloud
+* helm - use the Helm CLI
+* az - Interact with Azure using the Azure CLI
```
carolynvs, line 66: 👍

docs/content/author-bundles.md
```diff
@@ -142,7 +140,7 @@ parameters:
 
 Porter supports passing a file as a parameter to a bundle.
 
-For instance, a bundle might declare a parameter mytar of type file, located at /root/mytar when the bundle is run:
+For instance, a bundle might declare a parameter `mytar` of type `file`, located at `/root/mytar` when the bundle is run:
```
carolynvs, line 143: You wouldn't know this since it's very inconsistent in the docs, but we are trying to move away from using inline code blocks. It renders funny and distracts the reader.

For example, this is hard to read and the blocks throw off the flow of the text:

Screen Shot 2021-05-25 at 10 41 08 AM

If you'd like to do a sweep for that, and have us switch to some other formatting, which could mean none at all for most things, that would be helpful but we should sync up first on what we want it to look like.

carolynvs (changes requested):
In general, submitting fixes for typos and other clear errors without an issue are just fine. As you have found out here, for larger changes, it's always best to open an issue first and coordinate on what the solution should be first.

I'll hold off on reviewing the rest until you've had a chance to back out some changes, so we can just look at the typos and other stuff. Let's undo others as I have commented below. Thanks!

carolynvs:
@ThorstenHans Just checking in to see if you have questions or a different viewpoint on the review feedback. Or if you want to get some of this merged quickly, you could submit a PR with just the typos.

ThorstenHans:
Hey, had quite some load during the last weeks. Will close this PR and come up with smaller PRs instead

PR #1598, opened 2021-05-18
fix(manifest.go): fix faulty param/output validation

vdice:
# What does this change

This PR fixes a few items in the parameter and output definition validation logic:

1. The `Validate` calls should be used with the default param/output value, if exists -- not the entire definition
2. If a validation error is returned, we don't want to use `errors.Wrap(err, ...)` as this will always return nil b/c `err` would be nil from the previous check.

Notes:
I'd been digging into these areas on the cnab-go library and wanted to revisit/double-check this logic.  As a follow-up, we can add validation to the param/output definition schema itself. 

The cnab-go library will soon supply all of these validation checks whenever we call `bundle.Validate()` on a generated bundle, so we are technically duplicating logic here -- but right now we need to as Porter has its custom `"file"` type for params/outputs.  Perhaps we might entertain proposing adding support for this type in cnab-go?  Then Porter would probably be in a spot to remove the duplicated param/output definition and default value validation.

# What issue does it fix

# Notes for the reviewer

# Checklist
- [x] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/pkg/manifest/manifest.go
+++ b/pkg/manifest/manifest.go
@@ -342,12 +342,14 @@ func (pd *ParameterDefinition) Validate() error {
 		pdCopy.ContentEncoding = "base64"
 	}
 
-	schemaValidationErrs, err := pdCopy.Schema.Validate(pdCopy)
-	if err != nil {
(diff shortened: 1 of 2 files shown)


carolynvs (approved):
LGTM!

PR #1599, opened 2021-05-19
Slugify installation name when set via bundle name

### LINKED ISSUE #1260: Porter builds bundle with invalid name, then throws exception on install
## Describe the bug
Porter will build a porter.yaml file that contains an invalid name. On installation, the error will appear and the "invalidly named" bundle will fail to install.

## To Reproduce
1. Build a bundle with spaces in the name: `name: Github Accelerator Demo`
2. Install the bundle and receive this error

image

## Expected behavior
IF it's an invalid name, it should not build. IF it's a valid name, it should install. Pick one. :-)

## Version

porter v0.29.0 (5e7240cf)

ThorstenHans:
Signed-off-by: Thorsten Hans <thorsten.hans@gmail.com>

# What does this change

This PR will ensure installation name is valid when set via bundle name. 

# What issue does it fix

Closes #1260

# Checklist
- [x] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/pkg/porter/lifecycle.go
+++ b/pkg/porter/lifecycle.go
@@ -1,6 +1,8 @@
 package porter
 
 import (
+	"unicode"
+
 	cnabprovider "get.porter.sh/porter/pkg/cnab/provider"
 	"github.com/pkg/errors"
 )
@@ -93,7 +95,7 @@ func (p *Porter) prepullBundleByReference(opts *BundleActionOptions) error {
 	opts.RelocationMapping = cachedBundle.RelocationFilePath
 
 	if opts.Name == "" {
-		opts.Name = cachedBundle.Bundle.Name
+		opts.Name = slugify(cachedBundle.Bundle.Name)
 	}
 
 	if cachedBundle.Manifest != nil {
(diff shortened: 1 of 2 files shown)


pkg/porter/lifecycle.go
```diff
@@ -102,3 +104,19 @@ func (p *Porter) prepullBundleByReference(opts *BundleActionOptions) error {
...
+	for _, v := range value {
+		if unicode.IsLetter(v) {
+			slug += string(unicode.ToLower(v))
+		}
+		if unicode.IsDigit(v) {
+			slug += string(v)
+		}
+		if unicode.IsSpace(v) {
+			slug += "-"
+		}
```
vdice, line 119: Perhaps we should add a case to allow `-` chars in as well?

carolynvs: I think there are even more characters allowed in the name. The biggest limiting factor in what's okay in the installation name being able to reference and store it properly. I recommend that we use the allowed characters that we are adopting for the installation state spec.

> The name segment is required and MUST be 63 characters or less, beginning and ending with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between.

This helps us avoid having names that cannot be stored in certain formats, like kubernetes or as an index value in a database.

PR #1601, opened 2021-05-21
QuickStart: Parameters

carolynvs:
# What does this change
Second part of the quickstart that walks through how to set parameters with a flag or parameter set.

# What issue does it fix
N/A

# Notes for the reviewer
Ideally there was be sub menus under quickstart but I'd rather get the content out first then work with @flynnduism to get our nav bar to do what we want.

# Checklist
- [ ] Unit Tests
- [x] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/docs/content/quickstart/parameters.md
+++ b/docs/content/quickstart/parameters.md
@@ -1,0 +1,147 @@
+---
+title: "QuickStart: Parameters"
+descriptions: Learn how to use a bundle with parameters
+layout: single
+---
+
+Now that you know how to install a bundle, let's look at how to specify parameters to customize how that bundle is installed.
+Bundles can define parameters to allow the end-user to tweak how the bundle is configured.
+A parameter can be a string, integer, boolean or even a json object.
+Some examples of how parameters can be used in a bundle are:
+
+* Log Level: Default the log level for an application to info. At any time a user can upgrade the bundle to change that parameter to a different value.
+* Deployment Region: Let the user specify which region, such as eastus-1, where the application should be deployed.
+* Helm Release Name: A bundle that uses Helm often will define a parameter that allows the user to set the release name for the Helm release.
+
+For optional parameters, bundles set a default value that is used when the user does not specify a value for the parameter.
+
+Let's look at a bundle with parameters:
+
+```console
+$ porter explain --reference getporter/hello-llama:v0.1.1
+Name: hello-llama
+Description: An example Porter bundle with parameters
+Version: 0.1.0
+Porter Version: v0.38.1-32-gb76f5c1c
+
+No credentials defined
+
+Parameters:
+Name   Description                           Type     Default   Required   Applies To
+name   Name of to whom we should say hello   string   llama     false      All Actions
+
+No outputs defined
+
+No custom actions defined
+
+No dependencies defined
+```
+
+The output tells us that the bundle has one optional string parameter, name, which defaults to "llama".
+The name parameter applies to "All Actions" meaning that this parameter can be specified with every action that the bundle defines.
+Since it says that no custom actions are defined, the bundle only supports the built-in actions of install, upgrade, and uninstall.
+
+## Use the default parameter values
+
+
+First install the bundle and do not specify any parameters so that you can observe how the bundle installs without any customization.
+
+```console
+$ porter install hello-llama --reference getporter/hello-llama:v0.1.1
+installing hello-llama...
+executing install action from hello-llama (installation: hello-llama)
+Hello, llama
+execution completed successfully!
+```
+
+The bundle printed "Hello, llama" using the default value for the name parameter.
+
+## Specify a parameter with a flag
+Next upgrade the installation and change the name parameter to another value.
+Parameters are specified with the \--param flag: 
+
+```console
+$ porter upgrade hello-llama --param name=Porter
+upgrading hello-llama...
+executing upgrade action from hello-llama (installation: hello-llama)
+Porter 2.0
+execution completed successfully!
+```
+
+## Create a Parameter Set
+Setting a parameter value individually on the command line with the \--param flag works well when trying out a bundle.
+When working with a bundle often though, remembering and typing out every parameter values every time is error prone.
+Parameter Sets store a set of parameter values to use with a bundle.
+Even more powerful, the value of the parameters can come from the following sources:
+
+* hard-coded value
+* environment variable
+* file
+* command output
+* secret
+
+Some parameters may be sensitive, for example a database connection string or oauth token.
+For improved security, and to limit exposure of sensitive values, it is recommended that you source those parameter values from a secret store such as Hashicorp Vault or Azure Key Vault.
+
+Create a parameter set for the hello-llama with the `porter parameters generate` command. It is an interactive command that walks through setting values for every parameter in the specified bundle.
+
+```console
+$ porter parameters generate hello-llama --reference getporter/hello-llama:v0.1.1
+Generating new parameter set hello-llama from bundle hello-llama
+==> 2 parameters declared for bundle hello-llama
+
+? How would you like to set parameter "name"
+   [Use arrows to move, space to select, type to filter]
+  secret
+> specific value
+  environment variable
+  file path
+  shell command
+
+? Enter the value that will be used to set parameter "name"
+  Porter
+```
+
+This creates a parameter set named hello-llama.
+View the parameter set with the `porter parameters show` command:
+
+```console
+$ porter parameters show hello-llama
+Name: hello-llama
+Created: 4 minutes ago
+Modified: 4 minutes ago
+
+-----------------------------------
+  Name  Local Source  Source Type
+-----------------------------------
+  name  Porter       value
+```
+
+The output shows that the parameter set has one parameter stored in it, the name parameter, and it will use the hard-coded value "Porter".
+
+## Specify a parameter with a Parameter Set
+
+Now re-run the upgrade command, this time specifying the name of the parameter set instead of individually specifying each parameter value with flags.
+Parameter sets are specified with the \--parameter-set or -p flag.
+
+```console
+$ porter upgrade hello-llama --parameter-set hello-llama
+upgrading hello-llama...
+executing upgrade action from hello-llama (installation: hello-llama)
+Porter 2.0
+execution completed successfully!
+```
+
+## Cleanup
+
+To clean up the resources installed from this QuickStart, use the `porter uninstall` command. 
+
+```
+porter uninstall hello-llama
+```
+
+## Next Steps 
+
+In this QuickStart, you learned how to see the parameters defined on a bundle, their default values, and customize the installation of a bundle by specifying alternate values.
+
+* [Understanding how parameters are resolved](/parameters)
--- a/docs/content/quickstart/parameters.md
+++ b/docs/content/quickstart/parameters.md
@@ -4,8 +4,8 @@ descriptions: Learn how to use a bundle with parameters
 layout: single
 ---
 
-Now that you know how to install a bundle, let's look at how to specify parameters to customize how that bundle is installed.
-Bundles can define parameters to allow the end-user to tweak how the bundle is configured.
+Now that you know how to install a bundle, let's look at how to customize the bundle installation with parameters.
+Bundle authors define parameters in bundles so that end-users can tweak how the bundle is configured at installation.
 A parameter can be a string, integer, boolean or even a json object.
 Some examples of how parameters can be used in a bundle are:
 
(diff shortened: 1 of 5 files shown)


carolynvs:
screenshot

Just trying out the new netlify preview functionality

<details><summary>Browser metadata</summary>





```
Path:      /quickstart/parameters/
Browser:   Firefox 88.0 on Mac OS 10.15
Viewport:  1440 x 692 @2x
Language:  en-US
Cookies:   Enabled
```

[View on BrowserStack](https://live.browserstack.com/dashboard#url=https%3A%2F%2Fdeploy-preview-1601--porter.netlify.app%2Fquickstart%2Fparameters%2F&os=OS+X&os_version=Catalina&browser=Firefox&browser_version=88.0&resolution=1440x692&speed=1&start=true)



</details>

[Open Deploy Preview](https://deploy-preview-1601--porter.netlify.app/quickstart/parameters/) · [Mark as Resolved](https://app.netlify.com/cdp/resolve?deployID=60a839984e0df3000781080f&commentID=60a83af7d275f100bde80389&resolution=resolved)

docs/content/quickstart/parameters.md
```diff
@@ -0,0 +1,147 @@
+---
+title: "QuickStart: Parameters"
+descriptions: Learn how to use a bundle with parameters
+layout: single
+---
+
+Now that you know how to install a bundle, let's look at how to specify parameters to customize how that bundle is installed.
```
iennae, line 7: customize bundle installation.

iennae, line 7: Now that you know how to install a bundle, let's look at how to customize the bundle installation with parameters.

docs/content/quickstart/parameters.md
```diff
@@ -0,0 +1,147 @@
...
+Bundles can define parameters to allow the end-user to tweak how the bundle is configured.
```
iennae, line 8: Needs to be active here. 

Bundle authors add parameters to bundles so that end-users can tweak how the bundle is configured at installation. An example of where this would be useful is supplying a username and password value for a database.

docs/content/quickstart/parameters.md
```diff
@@ -0,0 +1,147 @@
...
+A parameter can be a string, integer, boolean or even a json object.
+Some examples of how parameters can be used in a bundle are:
+
+* Log Level: Default the log level for an application to info. At any time a user can upgrade the bundle to change that parameter to a different value.
+* Deployment Region: Let the user specify which region, such as eastus-1, where the application should be deployed.
+* Helm Release Name: A bundle that uses Helm often will define a parameter that allows the user to set the release name for the Helm release.
+
+For optional parameters, bundles set a default value that is used when the user does not specify a value for the parameter.
```
iennae, line 16: To make the parameter optional, you must set a default value for the parameter within the bundle. 

Q: is the parameter optional here? Or really just optional at being specified. Just think this doesn't feel exactly precise.

docs/content/quickstart/parameters.md
```diff
@@ -0,0 +1,147 @@
...
+
+Let's look at a bundle with parameters:
+
```
iennae, line 19: I'm wondering if starting this out looking at the same bundle that was used in the quickstart (getporter/wordpress:v0.1..3) would be useful here.

docs/content/quickstart/parameters.md
iennae: Just realized in the quickstart that the parameters are 

namespace
wordpress-name
wordpress-password 

and both namespace and wordpress-name are optional (required is set to false) and only wordpress-name has a default setting.

docs/content/quickstart/parameters.md
```diff
@@ -0,0 +1,147 @@
...
+name   Name of to whom we should say hello   string   llama     false      All Actions
+
+No outputs defined
+
+No custom actions defined
+
+No dependencies defined
+```
+
+The output tells us that the bundle has one optional string parameter, name, which defaults to "llama".
```
iennae, line 40: I keep reading this example as llama as in llamar "to call" you call me ..

iennae: In the Parameters section of the output returned by explain, there is a single optional string parameter, name, with a default of "llama" that applies to "All Actions".

docs/content/quickstart/parameters.md
```diff
@@ -0,0 +1,147 @@
...
+The name parameter applies to "All Actions" meaning that this parameter can be specified with every action that the bundle defines.
```
iennae, line 41: This means that the name parameter can be specified with each action that the bundle defines.

docs/content/quickstart/parameters.md
```diff
@@ -0,0 +1,147 @@
...
+Since it says that no custom actions are defined, the bundle only supports the built-in actions of install, upgrade, and uninstall.
```
iennae, line 42: In the custom actions section of the output returned by explain, there are no custom actions defined. The hello-llama bundle only supports the built-in actions of install, upgrade, and uninstall.

docs/content/quickstart/parameters.md
```diff
@@ -0,0 +1,147 @@
...
+
+## Use the default parameter values
+
+
+First install the bundle and do not specify any parameters so that you can observe how the bundle installs without any customization.
```
iennae, line 47: Install the bundle without specifying any parameters so that you can see the default behavior of the bundle install.

docs/content/quickstart/parameters.md
```diff
@@ -0,0 +1,147 @@
...
+installing hello-llama...
+executing install action from hello-llama (installation: hello-llama)
+Hello, llama
+execution completed successfully!
+```
+
+The bundle printed "Hello, llama" using the default value for the name parameter.
+
+## Specify a parameter with a flag
+Next upgrade the installation and change the name parameter to another value.
```
iennae, line 60: Let's see the behavior when customizing the bundle with parameters. Upgrade the bundle and specify the name parameter.

docs/content/quickstart/parameters.md
```diff
@@ -0,0 +1,147 @@
...
+Parameters are specified with the \--param flag:
```
iennae, line 61: This feels like it should be explained earlier before we start experimenting.

docs/content/quickstart/parameters.md
```diff
@@ -0,0 +1,147 @@
...
+```
+
+The bundle printed "Hello, llama" using the default value for the name parameter.
+
+## Specify a parameter with a flag
+Next upgrade the installation and change the name parameter to another value.
+Parameters are specified with the \--param flag: 
+
+```console
+$ porter upgrade hello-llama --param name=Porter
```
iennae, line 64: Maybe use a different name than Porter so it isn't confusing to new folks. (Michele?)

docs/content/quickstart/parameters.md
```diff
@@ -0,0 +1,147 @@
...
+upgrading hello-llama...
+executing upgrade action from hello-llama (installation: hello-llama)
+Porter 2.0
+execution completed successfully!
+```
+
+## Create a Parameter Set
+Setting a parameter value individually on the command line with the \--param flag works well when trying out a bundle.
```
iennae, line 72: While trying out a bundle, it might work well when you set individual parameter values on the command line with the \--param flag.

docs/content/quickstart/parameters.md
```diff
@@ -0,0 +1,147 @@
...
+When working with a bundle often though, remembering and typing out every parameter values every time is error prone.
+Parameter Sets store a set of parameter values to use with a bundle.
```
iennae, line 74: To alleviate errors and the requirement of remembering and manually configuring parameters at the command line, you can store a set of parameter values to use with a bundle with a parameter set.

docs/content/quickstart/parameters.md
```diff
@@ -0,0 +1,147 @@
...
+Even more powerful, the value of the parameters can come from the following sources:
```
iennae, line 75: more powerful than what?

docs/content/quickstart/parameters.md
```diff
@@ -0,0 +1,147 @@
...
+
+* hard-coded value
+* environment variable
+* file
+* command output
+* secret
+
+Some parameters may be sensitive, for example a database connection string or oauth token.
+For improved security, and to limit exposure of sensitive values, it is recommended that you source those parameter values from a secret store such as Hashicorp Vault or Azure Key Vault.
```
iennae, line 84: For improved security, and to limit exposure of sensitive values, it is recommended that you source sensitive parameter values from a secret store such as HashiCorp Vault, Azure Key Vault or Google Secret Manager.

docs/content/quickstart/parameters.md
```diff
@@ -0,0 +1,147 @@
...
+  secret
+> specific value
+  environment variable
+  file path
+  shell command
+
+? Enter the value that will be used to set parameter "name"
+  Porter
+```
+
```
iennae, line 104: What if I want to accept a default value for a particular parameter?

iennae: Can there only be one parameter set per bundle? What happens if I re-run generate?

docs/content/quickstart/parameters.md
carolynvs: I am going to leave off the second sentence because a) those are most likely credentials and I don't want to confuse the issue of creds vs params just yet, and b) there are 3 examples of parameters in the next paragraph.

docs/content/quickstart/parameters.md
carolynvs: I'm a bit lost here and I think some of the confusion may being caused by the old/unused parameter field `required`. That isn't supported anymore to specifying if a parameter must be specified or not. If a default is set, then the parameter is optional and does not need to be specified by the user. If a default is not set, then the parameter is required. In a separate PR I will clean up any example porter.yaml files that are still using `required: false` or anything like that.

Does that help clear up this question? Or does the text still need to change?

docs/content/quickstart/parameters.md
carolynvs: I decided to not use that bundle because it requires credentials, which we have not introduced yet, and a kubernetes cluster (more work for the reader that isn't relevant).

I used wordpress as an example earlier in the first quickstart because it is a bundle that has credentials, parameters, outputs and custom actions defined. That was useful because I was highlighting the output of porter explain, and if all the output said was "no credentials/parameters/outputs defined" then that wouldn't demonstrate the command very well. This is why later in the same quickstart I switched to the hello world bundle, so that we weren't using features that have not been explained yet.

docs/content/quickstart/parameters.md
carolynvs: Yup! that is the hidden joke. 😀

docs/content/quickstart/parameters.md
carolynvs: I can't mention google because no one has made a plugin for that yet.

docs/content/quickstart/parameters.md
carolynvs: > What if I want to accept a default value for a particular parameter?

Let me make an issue for that. Right now the answer is manually edit the parameter set to remove the parameter mapping from the file.

carolynvs: > Can there only be one parameter set per bundle? What happens if I re-run generate?

I have answers but I'm not sure if that level of detail belongs in the quickstart, which is a task focused document and isn't intended to be a complete reference of the topic. A better place to answer those questions is porter.sh/parameters I think. We already say in the command-line documentation for the --parameter-set flag that it can be used more than once.

1. Yes you can define multiple parameter sets and use them when installing a bundle.
2. Re-running generate, runs you through the same process as the first time. If you choose the same parameter set name the previous parameter set is overwritten.

carolynvs:
@iennae Thanks for the thoughtful suggestions! I've updated it with your feedback.

docs/content/quickstart/parameters.md
iennae: Ahhhhhh! The output is confusing which is why I'm lost. so I think there is more nuance here and I'll need to chat to you more about it because optional feels like the wrong word

docs/content/quickstart/parameters.md
iennae: AHHH. just cut the Google Secret Manager out and keep the "sensitive parameter values" (instead of those)

docs/content/quickstart/parameters.md
iennae: Cool, wasn't saying that these things need to defined here but wondering if folks have more questions if they were answered somewhere.

iennae (approved):
LGTM

Issue #1446, opened 2021-02-10
Have installations aware of the bundle reference and version


carolynvs:
**What design is being proposed?**
I want to be able to upgrade a bundle to a newer version without having to respecify the entire reference again. It should reuse the reference from previous actions such as install.

First when we display an installation we should show what bundle it came from, the reference you got it from (if available, it may not be if it was installed from a local file) and the version of the bundle that we used last.

```console
# Current output
$ porter installation show porterops
Name: porterops
Created: 2021-02-05
Modified: 2 minutes ago

# Desired Output
$ porter installation show porterops
Name: porterops
Bundle: porter-operator
Reference: ghcr.io/getporter/porter-operator
Version: v0.1.0
Created: 2021-02-05
Modified: 2 minutes ago
```

Now when I want to upgrade my bundle to a new version I should be able to just specify the version. Porter should use the known reference to figure out the full reference:
```console
$ porter upgrade porterops --version v0.2.0
# really runs porter upgrade porterops --reference ghcr.io/getporter/porter-operator:v0.2.0
```

**Additional Context**
N/A

**Risks/Concerns**
An open question is when you run `porter upgrade --reference NEWREF` is that the new base reference for the bundle? Or do you need to explicitly say you want to use a new reference now? 

I am hoping we can look at the last claim and figure everything out from there but we'd need to vet that is safe.

**What other approaches were considered?**
> This is the way - The Mandalorian

**Implementation details**
I hope we can accomplish this without changing the data stored and only tweak the CLI to better display available fields, and use existing claim data before executing the action.

# Checklist
- [ ] An announcement of this proposal has been sent to the Porter mailing list: https://porter.sh/mailing-list
- [ ] This proposal has remained open for at least one week, to allow time for community feedback.

carolynvs:
I am implementing this as part of the verification of the Installation State Specification.

PR #1605, opened 2021-05-24
Display more installation information with show

### LINKED ISSUE #1585: Hide porter managed bundle outputs
When I added logs as a generated output for bundles, it started showing up in the bundle commands like show and porter installation output list.

I think it's confusing for people to see there, and we want to encourage people to look at the logs using the dedicated porter logs command instead. Just like we hide porter generated parameters, like debug, I think we should also hide porter generated outputs as well.

In the example below we shouldn't show `io.cnab.outputs.invocationImageLogs`:

```console
$ porter show porter-hello
Name: porter-hello
Created: 2021-03-27
Modified: 2021-05-03

Outputs:
-------------------------------------------------------------------------------
  Name                                 Type    Value
-------------------------------------------------------------------------------
  io.cnab.outputs.invocationImageLogs  string  DEBUG: defaulting action
                                               to CNAB_ACTION (install)
                                               executi...
```

carolynvs:
# What does this change
Add the following fields to the porter show output:

* Bundle Repository, e.g. getporter/porter-hello
* Bundle Version
* Parameters used last

All of this data comes from the last claim associated with the installation.

There was a CNAB Spec bug that tripped up my testing. Porter was not setting the bundle reference on the claim. It's not required by the spec, but since we _have_ the data and now we want to show it, we should set it whenever it's available.

I have also consolidated the print logic for parameters and outputs so that they both truncate and mask sensitive values the same way. If someone wants the full unmasked values, right now they need to use --ouput json|yaml instead.

```console
$ porter show mysql
Name: mysql
Bundle: getporter/mysql
Version: 0.1.3
Created: 7 minutes ago
Modified: 39 seconds ago

Parameters:
------------------------------------------
  Name           Type    Value
------------------------------------------
  mysql-name     string  porter-ci-mysql
  mysql-user     string  mysql-admin
  namespace      string
  database-name  string  mydb

Outputs:
---------------------------------------
  Name                 Type    Value
---------------------------------------
  mysql-password       string  ******
  mysql-root-password  string  ******

History:
-------------------------------------------------------------------------
  Run ID                      Action   Timestamp       Status  Has Logs
-------------------------------------------------------------------------
  01F6FZD5Z0NMSM4TZ9SGDF6Z6G  install  7 minutes ago   failed  true
```

# What issue does it fix
Fixes #1585
Part of #1446 (just changes show, doesn't add the --version flag to the other command)

# Notes for the reviewer

# Checklist
- [x] Unit Tests
- [ ] Documentation TODO
- [ ] Schema (porter.yaml) N/A


carolynvs:
@vdice Right now this shows the parameters on the last claim. Based on what we talked about today, when I implement the installation document, we would show the parameters associated with the installation when that's implemented. I think showing this data right now though is still useful?

docs/content/operators/logs.md
```diff
@@ -49,15 +49,17 @@ ID** uniquely identifies a run, or execution of a bundle.
 ```console
 $ porter show whalegap
 Name: whalegap
+Bundle:
```
vdice, line 52: Is the bundle name missing here?  Or is this how it looks when we don't know it for some reason?

pkg/porter/testdata/show/expected-output.json
```diff
@@ -0,0 +1,65 @@
...
+      "parameters": {
+        "logLevel": 5,
+        "token": "top-secret"
+      },
+      "timestamp": "2020-04-18T01:02:03.000000004Z",
+      "status": "succeeded",
+      "hasLogs": "false"
+    },
+    {
+      "claimID": "3",
```
vdice, line 52: Just found this curious: a jump from 1 to 3 for claimID...  A quirk of the `CreateClaim` helper?

vdice (approved):

pkg/porter/testdata/show/expected-output.json
carolynvs: it's a quirk of how the test generates repeated id's for claims and results. I was lazy and the number didn't really matter, it just had to be consistent between tests. 😀

docs/content/operators/logs.md
carolynvs: Let me change the output here to use a bundle that was installed using `--reference`. When that flag isn't used, i.e. we installed it from source, then we won't have this bundle value.

We could display `(installed from source)` instead of an empty string?

vdice: > We could display (installed from source) instead of an empty string?

Oh maybe so, that could be a good option.  In any case, sounds like a candidate for a follow-up.

vdice (approved):
LGTM

PR #1608, opened 2021-05-29
Count only the parameters in the parameter set

### LINKED ISSUE #1600: Incorrect number of parameters when generating parameter sets
## Describe the bug
Porter is showing the TOTAL number of parameters in the bundle, including porter specific parameters which are not included in the parameter set.

## To Reproduce
Run porter parameters generate for any bundle. Note that it shows a number one higher than the number of parameters defined in porter.yaml because it is counting PORTER_DEBUG.

## Expected behavior
It should show the number of parameters defined in porter.yaml.

## Porter Command and Output
```
$ porter parameters generate hello-llama -r getporter/hello-llama:v0.1.0
Generating new parameter set hello-llama from bundle hello-llama
==> 2 parameters declared for bundle hello-llama
```

## Version
porter v0.38.1-32-gb76f5c1c (b76f5c1c)

### How to Fix
* Create your fix based on the **v1** branch. When you submit your PR, set the destination to the **v1** branch.
* This is the incorrect line of code. We should loop over the parameters in the bundle and count only parameters that are not internal. You can use this function [IsInternal](https://github.com/getporter/porter/blob/784d61cf283bbd5b9c05a42ea718d3d042803305/pkg/parameters/parameters.go#L55) to know if it should be counted or not.

    https://github.com/getporter/porter/blob/784d61cf283bbd5b9c05a42ea718d3d042803305/pkg/porter/parameters.go#L132

The output from this command is hard to unit test, since it is interactive, so testing manually is fine. You can run the command in the bug report to verify the fix is working.

divbhasin:
Signed-off-by: div_bhasin <divbest99@gmail.com>

# What does this change
When using `porter parameters generate...`, we see only the number of parameters in the parameter set being generated now. Look at the "Porter Command and Output" screenshot in #1600 to see what was wrong before. For the same command, we now get:

Screen Shot 2021-05-29 at 9 44 49 AM

# What issue does it fix
Closes #1600 

# Notes for the reviewer
For the "1 parameters declared for bundle..." in the above screenshot, should I change the "parameters" to "parameter(s)" to account for the above case where we might only have 1 parameter in the parameter set? Alternatively, we could also do a switch/if statement to be more precise.

# Checklist
- [ ] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/pkg/cnab/config-adapter/testdata/porter-with-maintainers.yaml
+++ b/pkg/cnab/config-adapter/testdata/porter-with-maintainers.yaml
@@ -25,8 +25,9 @@ credentials:
       - uninstall
 
 dependencies:
-  - name: mysql
-    reference: "getporter/azure-mysql:5.7"
+  requires:
+    - name: mysql
+      reference: "getporter/azure-mysql:5.7"
     
 mixins:
 - exec
--- a/pkg/porter/parameters.go
+++ b/pkg/porter/parameters.go
@@ -129,7 +129,15 @@ func (p *Porter) GenerateParameters(opts ParameterOptions) error {
 		Bundle: bundle,
 	}
 	fmt.Fprintf(p.Out, "Generating new parameter set %s from bundle %s\n", genOpts.Name, bundle.Name)
-	fmt.Fprintf(p.Out, "==> %d parameters declared for bundle %s\n", len(bundle.Parameters), bundle.Name)
+	numInternalParams := 0
+
+	for name := range bundle.Parameters {
+		if parameters.IsInternal(name, bundle) {
+			numInternalParams += 1
+		}
+	}
+
+	fmt.Fprintf(p.Out, "==> %d parameters declared for bundle %s\n", numInternalParams, bundle.Name)
 
 	pset, err := genOpts.GenerateParameters()
 	if err != nil {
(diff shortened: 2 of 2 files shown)


pkg/cnab/config-adapter/testdata/porter-with-maintainers.yaml
```diff
@@ -25,8 +25,9 @@ credentials:
       - uninstall
 
 dependencies:
-  - name: mysql
-    reference: "getporter/azure-mysql:5.7"
+  requires:
```
divbhasin, line 28: This change is unrelated to the PR. Without it, the "TestManifestConverter_generatedMaintainers" was failing.

carolynvs (approved):
Thank you for fixing this, and also the bonus patch for our main build being broken! 💯

carolynvs:
Sorry I forgot to respond to your question about "parameter(s)". Yes, if you would like to submit a PR to change it to "parameter(s)" that would be great.

Also congrats on your first pull request to Porter! 🎉  We like to thank our contributors by having them add their name to our [contributors.md](https://github.com/getporter/porter/blob/main/CONTRIBUTORS.md) file at the root of the repository. You are welcome to add your name to the file anytime.

PR #1611, opened 2021-06-02
Bump cnab-go to v0.20.0

carolynvs:
Big bump energy

* jsonschema-lock is finally merged and we can get rid of our fork
* Allow numbers in bundle.json again
* Add support for params of type array
* Add ValidateSchema function
* Add output validation
* Validate parameter defaults
* Validate param regardless of if override or default
* Fix retrieval of outputs when characters in the result id are in the
output name
* Add support for params of object type in ConvertValue (#251)
* Use alternate canonical json library
* Bump docker libraries to newer version

### DIFF
--- a/pkg/cnab/provider/testdata/bundle-with-param-sources.json
+++ b/pkg/cnab/provider/testdata/bundle-with-param-sources.json
@@ -99,7 +99,8 @@
   },
   "definitions": {
     "connstr": {
-      "type": "string"
+      "type": "string",
+      "default": ""
     },
     "foo": {
       "type": "string",
--- a/go.mod
+++ b/go.mod
@@ -3,11 +3,6 @@ module get.porter.sh/porter
 go 1.16
 
 replace (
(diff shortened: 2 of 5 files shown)


vdice:
🎉 So excited!

We need to update or remove the dep override [here](https://github.com/getporter/porter/blob/main/go.mod#L9) right?

carolynvs:
Oops, it would have helped if all my changes were pushed. I've squashed and force pushed.

pkg/cnab/provider/testdata/bundle-with-param-sources.json
```diff
@@ -99,7 +99,8 @@
   },
   "definitions": {
     "connstr": {
-      "type": "string"
+      "type": "string",
```
carolynvs, line 102: Now that you validate default values, the test started failing because this didn't have a default.

vdice (approved):

PR #1619, opened 2021-06-04
feat(manifest): add parameter/output schema validation

vdice:
# What does this change
* Adds validation of the schema/definition itself for parameters and outputs

# What issue does it fix

# Notes for the reviewer
Follow-up validation now that v1 has the latest cnab-go release.

# Checklist
- [x] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/pkg/manifest/manifest.go
+++ b/pkg/manifest/manifest.go
@@ -346,7 +346,7 @@ func (pd *ParameterDefinition) Validate() error {
 
 	// Validate the Parameter Definition schema itself
 	if _, err := pdCopy.Schema.ValidateSchema(); err != nil {
-		return multierror.Append(result, err)
+		return multierror.Append(result, errors.Wrap(err, "encountered an error while validating parameter definition schema"))
 	}
 
 	if pdCopy.Default != nil {
@@ -711,7 +711,7 @@ func (od *OutputDefinition) Validate() error {
 
 	// Validate the Output Definition schema itself
 	if _, err := odCopy.Schema.ValidateSchema(); err != nil {
-		return multierror.Append(result, err)
+		return multierror.Append(result, errors.Wrap(err, "encountered an error while validating output definition schema"))
 	}
 
 	if odCopy.Default != nil {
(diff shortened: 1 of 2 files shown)


pkg/manifest/manifest.go
```diff
@@ -344,6 +344,11 @@ func (pd *ParameterDefinition) Validate() error {
 		pdCopy.ContentEncoding = "base64"
 	}
 
+	// Validate the Parameter Definition schema itself
+	if _, err := pdCopy.Schema.ValidateSchema(); err != nil {
+		return multierror.Append(result, err)
```
carolynvs, line 349: This error should be wrapped to provide more context on what failed. I think it will be confusing without saying that the schema for parameter X is invalid, or something like that.

pkg/manifest/manifest.go
```diff
@@ -704,6 +709,11 @@ func (od *OutputDefinition) Validate() error {
 		odCopy.ContentEncoding = "base64"
 	}
 
+	// Validate the Output Definition schema itself
+	if _, err := odCopy.Schema.ValidateSchema(); err != nil {
```
carolynvs, line 713: Let's wrap this error too.

carolynvs (changes requested):
Just one small change and we are all set. 👍

pkg/manifest/manifest.go
```diff
@@ -346,7 +346,7 @@ func (pd *ParameterDefinition) Validate() error {
 
 	// Validate the Parameter Definition schema itself
 	if _, err := pdCopy.Schema.ValidateSchema(); err != nil {
-		return multierror.Append(result, err)
+		return multierror.Append(result, errors.Wrap(err, "encountered an error while validating parameter definition schema"))
```
carolynvs, line 349: Oops, sorry I should have been more clear in my first review that the error should tell you which parameter it is, just like we do when validating the default value.

carolynvs (changes requested):

pkg/manifest/manifest.go
vdice: Ah, no, my fault; was not paying careful enough attention.  Definitely makes sense to include the param/output name in the error message :)

carolynvs (approved):
LGTM! :shipit:

PR #1632, opened 2021-06-11
Fix v1 build

carolynvs:
This fixes a bunch of small problems that snuck into the v1 build:
* The v1 integration tests were broken because they weren't compiling after a struct was renamed. I have updated test-unit so that it compiles (doesn't run) the integration tests so that we catch that problem quickly in a PR next time.
* Remove unused target verify
* Move test-unit into mage
* The new cnab-go validation was validating things that it shouldn't (https://github.com/cnabio/cnab-go/pull/262)
* I see dead people but not typos
* Do not use the branch name for the permalink when building pull requests, just use "dev"

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -109,7 +109,7 @@ example output
 We recommend running the following every time:
 
 ```
-make verify build test-unit
+make build test-unit
 ```
 
 If your test modified anything related to running a bundle, also run:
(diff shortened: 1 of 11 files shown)


carolynvs:
Since I was able to merge that cnab-go PR for validation, I'll do one more change to bump cnab-go to use the official repo instead of the fork I created while testing this PR.

vdice (approved):
🚀

PR #1639, opened 2021-06-14
Do not resolve prerelease versions with latest

carolynvs:
# What does this change
* Now that we will have prerelease versions of the exec mixin available, porter mixin install exec is failing, because it's resolving the mixin @ v1-canary. Latest should resolve to the most recent _stable_ version available, which excludes the -canary entries.
* I also fixed the sorting of the entries in atom.xml so that when two binaries have the same timestamp, they sort consistently by name and version.
* Fixed LoadMetadata typo
* Fixed appending the .exe file extension during cross compilation.

# What issue does it fix
As soon as I published the exec mixin with v1-canary, it broke anyone who ran `porter mixins install exec --version latest`. The workaround is to pick a specific version until a new release is cut with this fix.

# Notes for the reviewer
N/A

# Checklist
- [x] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/mage/git.go
+++ b/mage/git.go
@@ -31,8 +31,8 @@ type GitMetadata struct {
 	IsTaggedRelease bool
 }
 
-// LoadMetadatda populates the status of the current working copy: current version, tag and permalink
-func LoadMetadatda() GitMetadata {
+// LoadMetadata populates the status of the current working copy: current version, tag and permalink
+func LoadMetadata() GitMetadata {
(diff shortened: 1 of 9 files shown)


vdice (approved):

carolynvs:
@vdice I updated the PR to include the typeo fix for `LoadMetadata` after your review.

Issue #1341, opened 2020-10-23
schema should allow for mixin configuration


carolynvs:
## Describe the bug
When I declare a mixin and define configuration for it, VS Code decides that it is invalid.

![Screen Shot 2020-10-23 at 10 33 43 AM](https://user-images.githubusercontent.com/1368985/97023637-69e75280-151b-11eb-91b8-f9fc7986b135.png)

## To Reproduce

1. Create a porter.yaml
2. Declare a mixin with configuration:
    ```yaml
    mixins:
      - kubernetes:
          clientVersion: v1.15.5
    ```
3. Open it in VS Code and note the dreaded red squiggles

## Expected behavior
Ideally autocomplete for the mixin config (not supported yet). Otherwise it should at least not complain.

## Version
```
$ porter version
porter v0.29.1 (2e0f3648)
```

carolynvs:
The schema command supported by mixins needs to return the schema for configuration as well as actions.

carolynvs:
Let's do progressive enhancement on this. If the mixin doesn't support returning it's configuration schema, then we allow any properties underneath and the mixin can be a string or an object while child properties. Otherwise inject the config schema into the final schema.

PR #1649, opened 2021-06-21
feat(*): support mixin config schema

vdice:
# What does this change

* Adds support for mixin config schema
  * If a mixin already declares config schema, we validate it and inject it into Porter's schema
  * If a mixin does not declare config schema, we inject a default, which allows:
    * a config-less mixin declaration, e.g.
    ```
      mixins:
        - mymixin
     ```
    * arbitrary config for the mixin (useful for when the mixin has yet to define its own schema but does already support additional configuration), e.g.
    ```
      mixins:
        - mymixin:
            arbitraryConfig: true
    ```

See also a PR for the first non-`exec` mixin to declare its configuration: https://github.com/getporter/kubernetes-mixin/pull/32

Follow-ups: If/when this all ends up looking good, we can create issues in the rest of the mixin repos in this org to define their config schema.

# What issue does it fix
Closes https://github.com/getporter/porter/issues/1341

# Notes for the reviewer

# Checklist
- [x] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/pkg/porter/schema.go
+++ b/pkg/porter/schema.go
@@ -77,14 +77,26 @@ func (p *Porter) injectMixinSchemas(manifestSchema jsonSchema) (jsonSchema, erro
 		return nil, errors.Errorf("root porter manifest schema has invalid properties.mixins type, expected map[string]interface{} but got %T", propertiesSchema["mixins"])
 	}
 
-	mixinItemSchema, ok := mixinSchema["items"].(jsonSchema)
+	mixinItemsSchema, ok := mixinSchema["items"].(jsonSchema)
 	if !ok {
 		return nil, errors.Errorf("root porter manifest schema has invalid properties.mixins.items type, expected map[string]interface{} but got %T", mixinSchema["items"])
 	}
 
-	mixinEnumSchema, ok := mixinItemSchema["enum"].([]interface{})
+	mixinItemsAnyOfSchema, ok := mixinItemsSchema["anyOf"].([]interface{})
 	if !ok {
-		return nil, errors.Errorf("root porter manifest schema has invalid properties.mixins.items.enum type, expected []interface{} but got %T", mixinItemSchema["enum"])
+		return nil, errors.Errorf("root porter manifest schema has invalid properties.mixins.items.anyOf type, expected []interface{} but got %T", mixinItemsSchema["anyOf"])
+	}
+
+	// Build enum schema to add to mixinItemsAnyOfSchema, for populating config-less mixin item options
+	mixinEnumsSchemaStr := `{"enum":[]}`
+	mixinEnumsSchema := make(jsonSchema)
+	err := json.Unmarshal([]byte(mixinEnumsSchemaStr), &mixinEnumsSchema)
+	if err != nil && p.Debug {
+		fmt.Fprintln(p.Err, errors.Wrapf(err, "could not unmarshal mixin enums schema %q", mixinEnumsSchemaStr))
+	}
+	mixinEnumsSchemaEntries, ok := mixinEnumsSchema["enum"].([]interface{})
+	if !ok {
+		return nil, errors.Errorf("error casting mixin enums array type, expected []interface{} but got %T", mixinEnumsSchema["enum"])
 	}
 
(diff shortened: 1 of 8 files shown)


pkg/porter/schema.go
```diff
@@ -115,32 +115,65 @@ func (p *Porter) injectMixinSchemas(manifestSchema jsonSchema) (jsonSchema, erro
...
+		}
+
+		// embed the entire mixin schema in the root after allowing for default config additions
 		manifestSchema["mixin."+mixin] = mixinSchemaMap
 
 		for _, action := range coreActions {
 			actionItemSchema, ok := actionSchemas[action]["items"].(jsonSchema)
-			if err != nil && p.Debug {
-				fmt.Fprintln(p.Err, errors.Errorf("root porter manifest schema has invalid properties.%s.items type, expected map[string]interface{} but got %T", action, actionSchemas[string(action)]["items"]))
+			if !ok && p.Debug {
```
carolynvs, line 168: oops

carolynvs (changes requested):
Seems like whatever information that we are returning in the anyOf isn't enough for VS Code to provide proper autocomplete?

Screen Shot 2021-06-21 at 4 22 13 PM

vdice:
Oh, shoot, I didn't think about the VS Code autocompletion for available mixin names themselves.  I was hoping to find an example where we do so, but it looks like the actions also have similar behavior.  Must be with our use of `items`... 

Screen Shot 2021-06-21 at 3 43 41 PM

carolynvs:
Just throwing out guesses but maybe if we set title or description it will use it for display?

vdice:
> Just throwing out guesses but maybe if we set title or description it will use it for display?

Great idea!  Have yet to find the winner from title, description, $id, but I'll keep tryin'!

vdice:
Ok, so I still have yet to crack the nut on how to add identifiers/names to the vs code auto-complete for json schema arrays.  Would you be amenable to tackling this in a follow-up issue, as it affects/is relevant to other sections of the manifest (actions, etc.)?

I've updated the PR with changes to help the VS Code UX on auto-complete.  We add the enum list of mixin strings and place this first in the `anyOf` schema array for mixins, so that it offers a helpful prompt from the get-go.  Then, when users enter a mixin and continue to add optional configuration, the corresponding mixin config schema will be used for validation.  WDYT?


Screen Shot 2021-06-22 at 11 30 52 AM

Screen Shot 2021-06-22 at 11 31 06 AM

Screen Shot 2021-06-22 at 11 31 22 AM

Screen Shot 2021-06-22 at 12 05 02 PM

carolynvs:
I actually still see the same funky array drop down even with the latest changes, though if I start typing a mixin, like `exTAB` I'll get `exec`.

I'm not in a hurry to get this in, the entire point is to make VS Code not be weird. So since this PR doesn't yet get us all the way there, maybe let's work with @itowlson to figure out what we need to do to make autocomplete work the way we'd like before merging anything.

vdice:
Ah, I misunderstood the original issue.  It stated `When I declare a mixin and define configuration for it, VS Code decides that it is invalid.`  So, I focused and fixed that issue, i.e., VS Code now supports mixin configuration to be defined (it will still yell 'invalid' when it actually is invalid -- but now we actually check the mixin config schema).

I'll move this to Draft and revisit when cycles allow, since we aren't in a time crunch.

carolynvs:
Sorry it wasn't clear! Yes, the issue is all driven by the desire for someone to be able to write a porter.yaml file with VS Code or another editor helping with autocomplete. They won't need to rely on our docs and examples as much and should be able to discover what's available and write manifests more confidently and quickly.

itowlson:
@vdice Carolyn asked me if I could provide any advice on this.  Can you post an example of the schema for `mixins`?  I can't quite work it out from the PR itself I'm afraid.  That will help clarify whether the issue lies with the schema or with VS Code/RH YAML's handling of the schema.

carolynvs:
I restarted with Vaughn's changes and through the power of time, trial and error, got VS code to mostly do what I wanted. Closing in favor of #2290

PR #1656, opened 2021-07-02
Use a distroless base image

carolynvs:
# What does this change
Use a distroless base image for our porter docker images. This has less of an attack surface because it only ships the essentials to run porter, not the extra stuff that usually comes with a linux distribution.

# What issue does it fix
Part of https://github.com/getporter/operator/issues/38

# Notes for the reviewer
~This relies on #1655 and needs to be rebased after that is merged.~

# Checklist
- [x] Unit Tests
- [x] Documentation
- [ ] Schema (porter.yaml)


build/images/client/Dockerfile
```diff
@@ -19,4 +19,9 @@ RUN porter mixin install kubernetes && \
     porter plugin install azure && \ 
     porter plugin install kubernetes
 
+# Copy the porter installation into a distroless container without root
+FROM gcr.io/distroless/static:debug
+COPY --from=builder /root/.porter /root/.porter
+ENV PATH "$PATH:/root/.porter"
+
 ENTRYPOINT ["porter"]
```
Church13, line 27:
```suggestion
ENTRYPOINT ["/app/.porter/porter"]
```
Best practice to declare full path to entrypoint, even if executable is present in path.

Church13 (changes requested):
Suggestions to improve dockerfile and final image (without affecting functionality)

build/images/client/Dockerfile
```diff
@@ -1,13 +1,12 @@
...
 
 RUN mkdir -p /root/.porter/runtimes && \
     mkdir -p /root/.porter/mixins/exec/runtimes
 
 COPY bin/dev/porter-linux-amd64 /root/.porter/porter
 COPY bin/mixins/exec/dev/exec-linux-amd64 /root/.porter/mixins/exec/exec
 RUN ln -s /root/.porter/porter /root/.porter/runtimes/porter-runtime && \
-    ln -s /root/.porter/mixins/exec/exec /root/.porter/mixins/exec/runtimes/exec-runtime && \
-    ln -s /root/.porter/porter /usr/local/bin/porter
+    ln -s /root/.porter/mixins/exec/exec /root/.porter/mixins/exec/runtimes/exec-runtime
```
Church13, line 10:
```suggestion
ENV PATH "$PATH:/root/.porter"

```
Belay removal of this line -- failed to note that the porter mixin install was in the build stage.

Church13 (changes requested):

carolynvs:
@Church13 Previously the porter agent image had a script for an entry point. Will that still work?

```
FROM getporter/porter:latest

VOLUME /porter-config

COPY run.sh /

ENTRYPOINT ["/run.sh"]
```

When I try to run the image I am getting `standard_init_linux.go:228: exec user process caused: no such file or directory` which I assume is because of the shebang `#!/usr/bin/env sh` in my run.sh. Should I change this to a binary?

Church13:
> @Church13 Previously the porter agent image had a script for an entry point. Will that still work?
> 
> ```
> FROM getporter/porter:latest
> 
> VOLUME /porter-config
> 
> COPY run.sh /
> 
> ENTRYPOINT ["/run.sh"]
> ```
> 
> When I try to run the image I am getting `standard_init_linux.go:228: exec user process caused: no such file or directory` which I assume is because of the shebang `#!/usr/bin/env sh` in my run.sh. Should I change this to a binary?

Distroless containers typically do not include shells. For google’s distroless image, “sh” is not an included binary. Can you show me the contents of run.sh? Ideally, it’s functionality would be replaced with a portable application, or with a more advanced entrypoint vector with all arguments and needed commands.

carolynvs:
It's a very simple script that just copies some files where it needs to go before running. I'll make a binary and use that instead. I just wanted to confirm that was the right solution. Thanks!

Church13:
@carolynvs Moving a container to distroless is a big paradigm shift -- it's minimum level runtime, providing virtually no environment or operating system. Since you were using the `distroless/static:debug` image before, there was a busybox shell present for debugging, but this is not present in the nonroot production ready image.

From Google's GitHub distroless repo readme:
>"Distroless" images contain only your application and its runtime dependencies. They do not contain package managers, shells or any other programs you would expect to find in a standard Linux distribution.
>
>For more information, see this [talk ](https://swampup2017.sched.com/event/A6CW/distroless-docker-containerizing-apps-not-vms?iframe=no&w=100%&sidebar=yes&bg=no)([video](https://www.youtube.com/watch?v=lviLZFciDv4)).

It's not unheard of to build a shell into a distroless image (read: distroless python, powershell), and there may be a good use case for doing so here -- but there are often different (and better) ways to build the image or utilize the application such that a shell isn't necessary.

Here's an interesting golang project I found on converting UNIX shell scripts to golang binaries (https://pkg.go.dev/github.com/ganbarodigital/go_scriptish) -- could be an avenue worth pursuing. As time allows I might look into converting the [run.sh](https://github.com/getporter/porter/blob/main/build/images/agent/run.sh) script to a static golang binary using this tool.

Church13:
> It's a very simple script that just copies some files where it needs to go before running. I'll make a binary and use that instead. I just wanted to confirm that was the right solution. Thanks!

That works as well. Awesome!

carolynvs:
@Church13 I've replaced run.sh with a go binary (I have newfound respect for the humble `cp`), and updated the agent to not install mixins into the image (we needed to make that change regardless).

build/images/client/Dockerfile
carolynvs: I'm not installing mixins anymore so that's not a problem anymore.

cmd/agent/main.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"fmt"
+	"os"
+
+	"get.porter.sh/porter/pkg/agent"
+)
+
+// The porter agent wraps the porter cli,
+// handling coping config files from a mounted
```
vdice, line 11:
```suggestion
// handling copying config files from a mounted
```

pkg/agent/agent.go
```diff
@@ -0,0 +1,134 @@
...
+)
+
+// allow the tests to capture output
+var (
+	stdout io.Writer = os.Stdout
+	stderr io.Writer = os.Stderr
+)
+
+// The porter agent wraps the porter cli,
+// handling coping config files from a mounted
```
vdice, line 22:
```suggestion
// handling copying config files from a mounted
```

vdice (approved):
Great to learn all about the distroless approach and best practices.  LGTM.

PR #1668, opened 2021-07-12
Fix applyTo check for outputs in porter explain

### LINKED ISSUE #1667: `porter explain` not showing bundle outputs with applyTo array
## Describe the bug
`porter explain` doesn't appear to show bundle-level outputs that have `applyTo` sections

## To Reproduce
1. Choose a bundle with outputs that use `applyTo` (I used the `azure-terraform` bundle in the `examples` directory of this repo)
2. Run `porter explain` 
3. See that no outputs are listed even though the bundle lists 3 at the bundle-level:

```
 $ porter explain
Name: azure-terraform
Description: An example Porter Bundle using Terraform and Azure
Version: 1.1.0
Porter Version: v0.38.1

Credentials:
Name              Description   Required   Applies To
client_id                       true       All Actions
client_secret                   true       All Actions
subscription_id                 true       All Actions
tenant_id                       true       All Actions

Parameters:
Name                     Description   Type     Default            Required   Applies To
database-name                          string   porter-terraform   false      All Actions
location                               string   EastUS             false      All Actions
resource_group_name                    string   porter-terraform   false      All Actions
storage_account_name                   string   porterstorage      false      All Actions
storage_container_name                 string   portertf           false      All Actions
storage_rg                             string   porter-storage     false      All Actions

No outputs defined

No custom actions defined

No dependencies defined
``` 

## Expected behavior
I would expect the 3 bundle-level outputs to be listed (with the specific actions that they apply to as well), e.g.:

```
Outputs:
Name                          Description   Type     Applies To
cosmos-db-uri                               string   install,upgrade
eventhubs_connection_string                 string   install,upgrade
storage_account_key                         string   install
```

## Version

```
 $ porter version
porter v0.38.1 (3018b91c)
```

cc @squillace who spotted this one

carolynvs:
# What does this change
I have consolidated the logic for checking if a scoped item: parameter, credential or output, should be included in the output of explain to avoid errors when duplicating the logic for each one.

There was also a copy/paste of the appliesTo logic for ParameterDefinition (our def of a parameter in the porter manifest) that
I updated to implement Scoped so that we can hopefully avoid more problems with this type of check.

# What issue does it fix
Closes #1667

# Notes for the reviewer
N/A

# Checklist
- [x] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/pkg/manifest/manifest.go
+++ b/pkg/manifest/manifest.go
@@ -14,6 +14,7 @@ import (
 	"get.porter.sh/porter/pkg/yaml"
 	"github.com/Masterminds/semver/v3"
 	"github.com/cbroglie/mustache"
+	"github.com/cnabio/cnab-go/bundle"
 	"github.com/cnabio/cnab-go/bundle/definition"
 	"github.com/cnabio/cnab-go/claim"
 	"github.com/docker/distribution/reference"
@@ -309,6 +310,8 @@ func (pd *ParameterDefinitions) UnmarshalYAML(unmarshal func(interface{}) error)
 	return nil
 }
 
+var _ bundle.Scoped = &ParameterDefinition{}
+
 // ParameterDefinition defines a single parameter for a CNAB bundle
 type ParameterDefinition struct {
 	Name      string          `yaml:"name"`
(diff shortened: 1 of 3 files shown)


vdice (approved):
LGTM

PR #1679, opened 2021-07-24
faq.md updated

sonukushwaha403:
# What does this change
Faq ,question added 
 

# What issue does it fix
 just added , a new FAQ

 
# Checklist
- [ ] Unit Tests
- [.] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/docs/content/faq.md
+++ b/docs/content/faq.md
@@ -144,23 +144,24 @@ The Azure mixins are going to be full fledged tools that communicate with the
 Azure APIs, and provide an improved user experience. It all depends on how much
 time you want to invest, and what you are starting from.
 
-## Does the Porter has its own Registry like DockerHub, ACR,Quay etc?
+## Does the Porter has its own registry?
 
- No, Porter does have its Registry, rather use an OCI compliant artifact store. And now DockerHub is also OCI compliant registry. The list of registries that works with CNAB ie bundles [here](https://porter.sh/compatible-registries/) .
+ No, Porter does not have a dedicated registry for hosting bundles. You can use any OCI registry, such as DockerHub, with Porter. We have compiled a [list of compatible registries](https://porter.sh/compatible-registries/) that have been tested with Porter.
 
-## What is the use of Mixins?
+## What do mixins do?
 
 Mixins provide the following features:
  * Install a tool into the bundle. It handles editing your dockerfile for you.
  * Mixins can adapt imperative command line tools to work with desired state. 
-   >For example, the uninstall command should be re-runnable. I should be able to re-run porter uninstall if there was an error and it     should pick up where it left off. Most command line tools are imperative though, and don’t handle errors in a way that works well with that (e.g. returning a failure when you attempt to delete something that   is already deleted).
+ 
+    For example, the uninstall command should be re-runnable. I should be able to re-run porter uninstall if there was an error and it should pick up where it left off. Most command line tools are imperative though, and don’t handle errors in a way that works well with that (e.g. returning a failure when you attempt to delete something that   is already deleted).
  * Mixins provide rich metadata that end-users can use to both understand what a bundle will do, and companies can limit what bundles they allow based on both the mixins present and what functionality of the mi   xin is used.
  * Mixins help collect outputs from steps, and reuse those as arguments to subsequent steps. 
-   >For example, the helm mixin can create a database and generate an output with the connection string. The terraform mixin can then use that connection string as an input variable, or even expose that output to the end-user when the bundle finishes running.
+    For example, the helm mixin can create a database and generate an output with the connection string. The terraform mixin can then use that connection string as an input variable, or even expose that output to the end-user when the bundle finishes running.
  
 You never have to use mixins, other than the built-in exec. You could write a custom dockerfile and then call a bash script. But most people find that working with mixins is easier.
 
-## Are Porter Bundles and Docker-Compose the same?
+## What is the different between Porter and existing DevOps tools?
 
 Porter is never a replacement for an existing tool. Think of it as doing extra nice things on top of what those great tools already do!
  * Packages everything you use to deploy in a single artifact that can be easily distributed over registries and across air gapped networks.
@@ -172,10 +173,10 @@ Porter is never a replacement for an existing tool. Think of it as doing extra n
  * You can put together the necessary parameters and credentials so your teammates don’t need to each hunt down the proper values themselves, or even worse copy sensitive credentials into local environment variables on every machine that does the deployment. Secrets stay in a secret store and aren’t copied around.
 Those are just some of the reasons why working with bundles with the tools embedded inside is helpful vs using the same tools standalone.
 
-## Where can I use Porter in Day to Day Projects?
+## Do you have suggestions for bundles that I could make?
 
  * Create a bundle for a side project. It handles setting up any infrastructure, such as creating a VM or other cloud resources, and it then also deploys your app. Maybe pushing your serverless function as one example, or running a helm chart. 
- * This is useful because often we figure out how to do these things while working on the project and then after not having time for the project most people forget how to deploy their project. That makes it super hard to start back up on the project! So using bundles helps make the best use of your limited side project time
+   This is useful because often we figure out how to do these things while working on the project and then after not having time for the project most people forget how to deploy their project. That makes it super hard to start back up on the project! So using bundles helps make the best use of your limited side project time
 * Create a bundle for an open source project. Pick a project that you like, maybe wordpress, mysql or discourse, and then write a bundle that will deploy the software to a cloud provider. This is a fun one because other people can use your bundle!
 
 [CNAB Core Spec]: https://github.com/cnabio/cnab-spec/blob/main/100-CNAB.md
--- a/docs/content/faq.md
+++ b/docs/content/faq.md
@@ -146,15 +146,18 @@ time you want to invest, and what you are starting from.
 
 ## Does the Porter has its own Registry like DockerHub, ACR,Quay etc?
 
-NO Porter does has its own Registry, rather use an OCI compliant artifact store.And now DockerHub is also OCI compliant registry .The list of registries that works with CNAB ie bundles [here](https://porter.sh/compatible-registries/) .
+ No, Porter does have its Registry, rather use an OCI compliant artifact store. And now DockerHub is also OCI compliant registry. The list of registries that works with CNAB ie bundles [here](https://porter.sh/compatible-registries/) .
 
 ## What is the use of Mixins?
 
 Mixins provide the following features:
  * Install a tool into the bundle. It handles editing your dockerfile for you.
- * Mixins can adapt imperative command line tools to work with desired state. For example, the uninstall command should be re-runnable. I should be able to re-run porter uninstall if there was an error and it     should pick up where it left off. Most command line tools are imperative though, and don’t handle errors in a way that works well with that (e.g. returning a failure when you attempt to delete something that   is already deleted).
+ * Mixins can adapt imperative command line tools to work with desired state. 
+   >For example, the uninstall command should be re-runnable. I should be able to re-run porter uninstall if there was an error and it     should pick up where it left off. Most command line tools are imperative though, and don’t handle errors in a way that works well with that (e.g. returning a failure when you attempt to delete something that   is already deleted).
  * Mixins provide rich metadata that end-users can use to both understand what a bundle will do, and companies can limit what bundles they allow based on both the mixins present and what functionality of the mi   xin is used.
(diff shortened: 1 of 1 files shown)


DARK-art108:
Signoff the commits to fix the check @sonukushwaha403

carolynvs:
@sonukushwaha403 Your pull request does not have any changes included https://github.com/getporter/porter/pull/1679/files. I'm guessing that it got messed up when you rebased?

sonukushwaha403:
@carolynvs  changes are palced again. do check out

DARK-art108:
@sonukushwaha403 You have not added the previous Faq which is present on the Doc which is shared on the Issue Description, Do add them also.

docs/content/faq.md
```diff
@@ -144,5 +144,34 @@ The Azure mixins are going to be full fledged tools that communicate with the
...
+ * Mixins provide rich metadata that end-users can use to both understand what a bundle will do, and companies can limit what bundles they allow based on both the mixins present and what functionality of the mi   xin is used.
+ * Mixins help collect outputs from steps, and reuse those as arguments to subsequent steps. For example, the helm mixin can create a database and generate an output with the connection string. The terraform mixin can then use that connection string as an input variable, or even expose that output to the end-user when the bundle finishes running.
+You never have to use mixins, other than the built-in exec. You could write a custom dockerfile and then call a bash script. But most people find that working with mixins is easier.
+
```
DARK-art108, line 159: Mixins provide the following features:
- Install a tool into the bundle. It handles editing your dockerfile for you.
- Mixins can adapt imperative command-line tools to work with the desired state.
   > For example, the uninstall command should be re-runnable. I should be able to re-run the porter uninstall if there was an error and it should pick up where it left off. Most command-line tools are imperative though, and don’t handle errors in a way that works well with that (e.g. returning a failure when you attempt to delete something that is already deleted).

- Mixins provide rich metadata that end-users can use to both understand what a bundle will do, and companies can limit what bundles they allow based on both the mixins present and what functionality of the mixin is used.
- Mixins help collect outputs from steps and reuse those as arguments to subsequent steps. 
  > For example, the helm mixin can create a database and generate an output with the connection string. The terraform mixin can then use that connection string as an input variable, or even expose that output to the end-user when the bundle finishes running.

You never have to use mixins, other than the built-in exec. You could write a custom dockerfile and then call a bash script. But most people find that working with mixins is easier.


***Replace with this!***

DARK-art108: And use the markdown as I have added above!

docs/content/faq.md
```diff
@@ -144,5 +144,34 @@ The Azure mixins are going to be full fledged tools that communicate with the
...
+ * Reduces the operational knowledge required to manage an application. Often a deployment uses multiple tools, for example terraform, AND helm, AND kubectl and they are glued together with bash scripts. When those are put inside a bundle, it simplifies how much the end user needs to know. It’s always a single command, porter upgrade . In many cases the end user doesn’t even need to know what tools are used, or how to use those tools.
+ * Makes it easier for a team to manage an application securely. Porter remembers the actions performed previously on the bundle, the parameter values used previously, the current version of the bundle, etc. You can put together the necessary parameters and credentials so your teammates don’t need to each hunt down the proper values themselves, or even worse copy sensitive credentials into local environment variables on every machine that does the deployment. Secrets stay in a secret store and aren’t copied around.
+Those are just some of the reasons why working with bundles with the tools embedded inside is helpful vs using the same tools standalone.
+
```
DARK-art108, line 170: Porter is never a replacement for an existing tool. Think of it as doing extra nice things on top of what those great tools already do!
		 * Packages everything you use to deploy in a single artifact that can be easily distributed over registries and across air-gapped networks.
		 * Bundles can be signed and the signature verified before installing the bundle to improve supply chain security.
		 * Makes for more reliable deployments because it has the exact version of a CLI that goes with the deployment. End users don’t need to install the right versions of tools locally (sometimes different apps require different versions).
		 * Provides metadata about a deployment that you can use in many ways. End users can run porter explain to quickly see how to customize an installation with parameters, what credentials are needed by the bundle. No need to read separate installation docs to figure out how to deploy an app.
		 * Reduces the operational knowledge required to manage an application. Often a deployment uses multiple tools, for example, terraform, AND helm, AND kubectl and they are glued together with bash scripts. When those are put inside a bundle, it simplifies how much the end-user needs to know. It’s always a single command, porter upgrade. In many cases, the end-user doesn’t even need to know what tools are used, or how to use those tools.
		 * Makes it easier for a team to manage an application securely. Porter remembers the actions performed previously on the bundle, the parameter values used previously, the current version of the bundle, etc. 
		 * You can put together the necessary parameters and credentials so your teammates don’t need to each hunt down the proper values themselves, or even worse copy sensitive credentials into local environment variables on every machine that does the deployment. Secrets stay in a secret store and aren’t copied around.
		Those are just some of the reasons why working with bundles with the tools embedded inside is helpful vs using the same tools standalone.

DARK-art108: Use the above content and add suitable markdown on it like the above point!

docs/content/faq.md
```diff
@@ -144,5 +144,34 @@ The Azure mixins are going to be full fledged tools that communicate with the
...
+## Where can I use Porter in Day to Day Projects?
+
+ * Create a bundle for a side project. It handles setting up any infrastructure, such as creating a VM or other cloud resources, and it then also deploys your app. Maybe pushing your serverless function as one example, or running a helm chart. This is useful because often we figure out how to do these things while working on the project and then after not having time for the project most people forget how to deploy their project. That makes it super hard to start back up on the project! So using bundles helps make the best use of your limited side project time
+* Create a bundle for an open source project. Pick a project that you like, maybe wordpress, mysql or discourse, and then write a bundle that will deploy the software to a cloud provider. This is a fun one because other people can use your bundle!
+
```
DARK-art108, line 175: Apply Proper Markdown on it! Like above!

DARK-art108 (changes requested):

sonukushwaha403:
@DARK-art108  do checkout, done with suggested changes.

DARK-art108 (approved):
Now everything is fixed, Looks good to merge! :)

docs/content/faq.md
```diff
@@ -144,5 +144,39 @@ The Azure mixins are going to be full fledged tools that communicate with the
 Azure APIs, and provide an improved user experience. It all depends on how much
 time you want to invest, and what you are starting from.
 
+## Does the Porter has its own Registry like DockerHub, ACR,Quay etc?
+
+ No, Porter does have its Registry, rather use an OCI compliant artifact store. And now DockerHub is also OCI compliant registry. The list of registries that works with CNAB ie bundles [here](https://porter.sh/compatible-registries/) .
```
carolynvs, line 149: I realize that this is most likely a copy/paste of my own words from Slack, but I now realize that it is unclear. Let's update it to the following to avoid confusion:

```suggestion
 No, Porter does not have a dedicated registry for hosting bundles. You can use any OCI registry, such as DockerHub, with Porter. We have compiled a [list of compatible registries](https://porter.sh/compatible-registries/) that have been tested with Porter.
```

docs/content/faq.md
```diff
@@ -144,5 +144,39 @@ The Azure mixins are going to be full fledged tools that communicate with the
 Azure APIs, and provide an improved user experience. It all depends on how much
 time you want to invest, and what you are starting from.
 
+## Does the Porter has its own Registry like DockerHub, ACR,Quay etc?
```
carolynvs, line 147: Let's fix the capitalization of Registry, and remove the example registries so that the question is more straightforward.

```suggestion
## Does the Porter has its own registry?
```

docs/content/faq.md
```diff
@@ -144,5 +144,39 @@ The Azure mixins are going to be full fledged tools that communicate with the
...
+
+ No, Porter does have its Registry, rather use an OCI compliant artifact store. And now DockerHub is also OCI compliant registry. The list of registries that works with CNAB ie bundles [here](https://porter.sh/compatible-registries/) .
+
+## What is the use of Mixins?
```
carolynvs, line 151: Let's make the question more tailored to the answer provided, and make the question sound like confrontational. 😀 

```suggestion
## What do mixins do?
```

docs/content/faq.md
```diff
@@ -144,5 +144,39 @@ The Azure mixins are going to be full fledged tools that communicate with the
...
+
+Mixins provide the following features:
+ * Install a tool into the bundle. It handles editing your dockerfile for you.
+ * Mixins can adapt imperative command line tools to work with desired state. 
+   >For example, the uninstall command should be re-runnable. I should be able to re-run porter uninstall if there was an error and it     should pick up where it left off. Most command line tools are imperative though, and don’t handle errors in a way that works well with that (e.g. returning a failure when you attempt to delete something that   is already deleted).
```
carolynvs, line 156: Using a blockquote here (the `>` formatting) causes this part to look like we are quoting something, when it is an example that supports the previous sentence. Let's keep this as a normal paragraph formatted text to improve readability.

```suggestion
 * Mixins can adapt imperative command line tools to work with desired state. 
 
    For example, the uninstall command should be re-runnable. I should be able to re-run porter uninstall if there was an error and it should pick up where it left off. Most command line tools are imperative though, and don’t handle errors in a way that works well with that (e.g. returning a failure when you attempt to delete something that   is already deleted).
```

Please make sure to make the same change to other places in the faq where blockquotes were added.

docs/content/faq.md
```diff
@@ -144,5 +144,39 @@ The Azure mixins are going to be full fledged tools that communicate with the
...
+ * Mixins provide rich metadata that end-users can use to both understand what a bundle will do, and companies can limit what bundles they allow based on both the mixins present and what functionality of the mi   xin is used.
+ * Mixins help collect outputs from steps, and reuse those as arguments to subsequent steps. 
+   >For example, the helm mixin can create a database and generate an output with the connection string. The terraform mixin can then use that connection string as an input variable, or even expose that output to the end-user when the bundle finishes running.
+ 
+You never have to use mixins, other than the built-in exec. You could write a custom dockerfile and then call a bash script. But most people find that working with mixins is easier.
+
+## Are Porter Bundles and Docker-Compose the same?
```
carolynvs, line 163: I have clarified the question a bit to better match the question that is usually asked. Porter is often compared to multiple tools, not just Docker Compose and the answer addresses the more general question.

```suggestion
## What is the different between Porter and existing DevOps tools?
```

docs/content/faq.md
```diff
@@ -144,5 +144,39 @@ The Azure mixins are going to be full fledged tools that communicate with the
...
+ * Reduces the operational knowledge required to manage an application. Often a deployment uses multiple tools, for example terraform, AND helm, AND kubectl and they are glued together with bash scripts. When those are put inside a bundle, it simplifies how much the end user needs to know. It’s always a single command, porter upgrade . In many cases the end user doesn’t even need to know what tools are used, or how to use those tools.
+ * Makes it easier for a team to manage an application securely. Porter remembers the actions performed previously on the bundle, the parameter values used previously, the current version of the bundle, etc. 
+ * You can put together the necessary parameters and credentials so your teammates don’t need to each hunt down the proper values themselves, or even worse copy sensitive credentials into local environment variables on every machine that does the deployment. Secrets stay in a secret store and aren’t copied around.
+Those are just some of the reasons why working with bundles with the tools embedded inside is helpful vs using the same tools standalone.
+
+## Where can I use Porter in Day to Day Projects?
```
carolynvs, line 175: The rest of the FAQ uses normal sentence casing for the headings, not title case. Let's change it to be consistent.

```suggestion
## Do you have suggestions for bundles that I could make?
```

docs/content/faq.md
```diff
@@ -144,5 +144,39 @@ The Azure mixins are going to be full fledged tools that communicate with the
...
+
+ * Create a bundle for a side project. It handles setting up any infrastructure, such as creating a VM or other cloud resources, and it then also deploys your app. Maybe pushing your serverless function as one example, or running a helm chart. 
+ * This is useful because often we figure out how to do these things while working on the project and then after not having time for the project most people forget how to deploy their project. That makes it super hard to start back up on the project! So using bundles helps make the best use of your limited side project time
```
carolynvs, line 178: This originally was a single paragraph in the Slack comment. This doesn't stand alone as an independent list item.
```suggestion
 * Create a bundle for a side project. It handles setting up any infrastructure, such as creating a VM or other cloud resources, and it then also deploys your app. Maybe pushing your serverless function as one example, or running a helm chart. 

   This is useful because often we figure out how to do these things while working on the project and then after not having time for the project most people forget how to deploy their project. That makes it super hard to start back up on the project! So using bundles helps make the best use of your limited side project time
```

carolynvs (changes requested):
Thank you for adding these to our FAQ! 💯 There are just some small formatting and grammar changes and then this is ready to merge.

carolynvs (approved):
Thank you for updating our FAQ! 💯

PR #1701, opened 2021-08-06
Support labels

### LINKED ISSUE #1683: Support namespaces and labels
In preparation of implementing advanced dependency scenarios from [PEP #003 - Dependency Namespaces and Labels](https://github.com/getporter/proposals/pull/8), let's add the concept of namespaces and labels.

carolynvs:
# What does this change
Support applying labels to installations, parameter and credential sets. Users can use these labels to filter list results. Eventually we can use these for matching dependencies.

I have also moved when an installation is created during porter install so that we can persist the labels.

# What issue does it fix
Closes #1683 

# Notes for the reviewer

# Checklist
- [x] Unit Tests
- [x] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/cmd/porter/credentials.go
+++ b/cmd/porter/credentials.go
@@ -68,8 +68,9 @@ When you wish to install, upgrade or delete a bundle, Porter will use the
 credential set to determine where to read the necessary information from and
 will then provide it to the bundle in the correct location. `,
 		Example: `  porter credential generate
-  porter credential generate kubecred --reference getporter/porter-hello:v0.1.0 --namespace test
-  porter credential generate kubecred --reference localhost:5000/getporter/porter-hello:v0.1.0 --insecure-registry --force
+  porter credential generate kubecred --reference getporter/mysql:v0.1.4 --namespace test
+  porter credential generate kubekred --label owner=myname --reference getporter/mysql:v0.1.4
+  porter credential generate kubecred --reference localhost:5000/getporter/mysql:v0.1.4 --insecure-registry --force
   porter credential generate kubecred --file myapp/porter.yaml
   porter credential generate kubecred --cnab-file myapp/bundle.json
 `,
@@ -99,7 +100,7 @@ func buildCredentialsListCommand(p *porter.Porter) *cobra.Command {
 	opts := porter.ListOptions{}
 
 	cmd := &cobra.Command{
-		Use:     "list [QUERY]",
+		Use:     "list",
 		Aliases: []string{"ls"},
 		Short:   "List credentials",
 		Long: `List named sets of credentials defined by the user.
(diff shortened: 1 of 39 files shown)


cmd/porter/credentials.go
```diff
@@ -84,6 +84,8 @@ will then provide it to the bundle in the correct location. `,
 	f := cmd.Flags()
 	f.StringVarP(&opts.Namespace, "namespace", "n", "",
 		"Namespace in which the credential set is defined. Defaults to the global namespace.")
+	f.StringSliceVarP(&opts.Labels, "label", "l", nil,
```
vdice, line 87: Is this on the `credentials generate` command?  Should be add a corresponding example in the cmd `Example` section?

Perhaps similar for `parameters generate`, too, if applicable.

pkg/porter/list.go
```diff
@@ -16,6 +17,40 @@ import (
...
 	Namespace string
+	Name      string
+	Labels    []string
+}
+
+func (o *ListOptions) Validate(args []string) error {
+	if len(args) == 1 {
+		o.Name = args[0]
+	} else if len(args) > 1 {
+		return errors.Errorf("only one positional argument may be specified, the installation name, but multiple were received: %s", args)
```
vdice, line 28: Do we need this same check for the params/creds list commands?

e.g.
```
 $ porter params list foo foo
NAMESPACE   NAME   MODIFIED
```

vdice (commented):
My only other thought is around the use of a positional argument for the installation/paramset/credset name and how it is different from the other query options that are behind flags (--namespace and --label).  I'm wondering if we want to consider using a flag for the name query (--name), instead of consuming it via the positional arg.

However, the installation/paramset/credset name is and has been consumed via a positional arg for the corresponding `install` and `generate` commands, so in this light, the current approach is in-line with expectations.

carolynvs:
@vdice I actually ripped off the `porter mixins search [QUERY]` command with that positional argument. But that command doesn't have other filter flags. I'll update it to use --name so that all filters are flags.

pkg/porter/list.go
carolynvs: I'll switch to a --name flag so that we don't need this check at all.

carolynvs:
@vdice fixed

tests/smoke/hello_test.go
```diff
@@ -63,6 +63,20 @@ func TestHelloBundle(t *testing.T) {
 	assert.Equal(t, "hello", installations[0]["name"], "expected the hello installation to be output by porter list")
 	assert.Equal(t, "test", installations[0]["namespace"], "expected the installation to be in the test namespace")
 
+	// Search by name
+	output, err = test.Porter("list", "hello", "--namespace=*", "--output=json").Output()
```
vdice, line 67: Should this test now use `"--name=hello"`?

carolynvs: I have changed the testdata so that it fails if I mess up the the --name flag now.

vdice (approved):
LGTM!

PR #1710, opened 2021-08-10
Fix bug resolving parameter sources from claim

### LINKED ISSUE #1707: Parameter source isn't resolved when upgrading bundle
## Describe the bug
I am running `porter upgrade INSTALLATION` and am getting an error about missing a required parameter. But the parameter should have been resolved from a parameter source (and output generated during install).

## To Reproduce
1. Use a bundle that has a parameter source
2. Install the bundle
3. Upgrade the bundle, and do not provide a --reference. Use the installation name.

## Expected behavior
The parameter source should be resolved so that magic_file isn't required.

## Porter Command and Output
```
$ porter upgrade mybuns
...  TRUNCATED
Resolving parameter source...
Processed bundle extensions:
extensions.ProcessedExtensions{}
No parameter sources defined!
Error: invalid parameters: parameter "magic_file" is required
```

Note that it didn't find any parameter sources (this is new debug output I added to see this work). The problem is that Porter didn't process the required extensions on the bundle when it took the definition from the last claim.

## Version
v1.0.0-alpha.1 but I think it affects v0.38 too

carolynvs:
# What does this change
When we execute a bundle there are two code paths:
1. We load the bundle definition from a file
2. We load the bundle definition from the claim

In the latter case, we weren't processing the bundle definition to
validate it and load required extensions. So the parameter sources
extension wasn't being applied properly for upgrade when --reference
wasn't specified.

# What issue does it fix
Fixes #1707

# Notes for the reviewer
This needs to be merged into v1.

# Checklist
- [x] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)


vdice (approved):
🪄

PR #1712, opened 2021-08-11
All namespaces

### LINKED ISSUE #1709: Add --all-namespaces flag
On v1, you can specify --namespace="*" and that indicates that porter should list across all namespaces. Since bash treats * funny (you have to wrap it in quotes), this is awkward for people who are not used to this. Plus it's not as discoverable as a dedicated flag. I like keeping the * support around because it makes scripting more straightforward though.

This should be changed on v1, but can be shipped after v1.

carolynvs:
# What does this change
Add an `--all-namespaces` flag to porter installation|credential|parameter list commands. This is the equivalent of `list -n "*"`

# What issue does it fix
Closes #1709 

# Notes for the reviewer
Relies on #1701 and will need a rebase after that is merged.

# Checklist
- [x] Unit Tests
- [x] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/cmd/porter/credentials.go
+++ b/cmd/porter/credentials.go
@@ -109,20 +109,22 @@ Optionally filters the results name, which returns all results whose name contai
 The results may also be filtered by associated labels and the namespace in which the credential set is defined.`,
 		Example: `  porter credentials list
   porter credentials list --namespace prod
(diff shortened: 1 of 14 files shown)


vdice (approved):
🙌

Issue #1703, opened 2021-08-06
Support editing the underlying document for installations, credential and parameter sets


carolynvs:
For the operator, we need to have the ability to edit installation documents by loading in an existing file. This will allow porter to read in the installation CRD and react to changes. Same for the other user modifiable documents: credential and parameter sets.

```
porter installation apply FILE
porter credentials apply FILE
porter parameters apply FILE
```

Ultimately, running apply for an installation should trigger reconcilliation for that installation. However that should be done in a separate issue. 

We should support importing yaml first. If it's easy add json. I've noticed problems importing TOML for credential and parameter sets when I was implementing another PR, so we should be careful about saying we support that until it's fully tested.

PR #1715, opened 2021-08-11
Add apply command for creds and params

carolynvs:
# What does this change
* Add a new command for credential and parameter sets which allow you toapply changes (or create net new) from a file.
* Fixes how we resolve params and credentials to allow referencing a global cs/ps from within a namespace during install.

# What issue does it fix
This is part of #1703. Next up is for installations which is a bit more complex.

# Notes for the reviewer
N/A

# Checklist
- [x] Unit Tests
- [x] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/cmd/porter/parameters.go
+++ b/cmd/porter/parameters.go
@@ -13,11 +13,42 @@ func buildParametersCommands(p *porter.Porter) *cobra.Command {
 		Short:       "Parameter set commands",
 	}
 
+	cmd.AddCommand(buildParametersApplyCommand(p))
 	cmd.AddCommand(buildParametersEditCommand(p))
 	cmd.AddCommand(buildParametersGenerateCommand(p))
 	cmd.AddCommand(buildParametersListCommand(p))
 	cmd.AddCommand(buildParametersDeleteCommand(p))
 	cmd.AddCommand(buildParametersShowCommand(p))
+
+	return cmd
+}
+
+func buildParametersApplyCommand(p *porter.Porter) *cobra.Command {
+	opts := porter.ApplyOptions{}
+
+	cmd := &cobra.Command{
+		Use:   "apply FILE",
+		Short: "Apply changes to a parameter set",
+		Long: `Apply changes from the specified file to a parameter set. If the parameter set doesn't already exist, it is created.
+
+Supported file extensions: json and yaml.
+
+You can use the generate and show commands to create the initial file:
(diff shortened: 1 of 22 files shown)


cmd/porter/parameters.go
```diff
@@ -22,6 +23,36 @@ func buildParametersCommands(p *porter.Porter) *cobra.Command {
...
+		Short: "Apply changes to a parameter set",
+		Long: `Apply changes from the specified file to a parameter set. If the parameter set doesn't already exist, it is created.
+
+Supported file extensions: json and yaml.
+
+You can use the generate and show commands to create the initial file:
+  porter parameters generate myparams --reference SOME_BUNDLE
+  porter parameters show myparams --output yaml > myparams.yaml
+`,
+		Example: `  porter parameters apply --file myparams.yaml`,
```
vdice, line 40: Should be `porter parameters apply myparams.yaml`, right?  If so, same update needed for creds.

This is a good time to mention: thoughts on using a `-f/--file` flag to pass in filename vs positional arg?  I'm sure you'd thought of it.  Endorsement for the former is that it'd be natural for users of `kubectl` -- but that doesn't nec. mean it's the right way to go for Porter.  No strong opinion here.

carolynvs: I had strongly considered the -f flag (as you can see since I forgot to fix that example). I like to use a positional argument when the argument is required, is always specified, and is the object of the (command) sentence.

The file parameter fits all those characteristics and that is how the other commands in porter act as well (such as install). Having file be a positional argument is more consistent with porter itself, which I feel is more important that being consistent with kubectl.

Also there aren't a lot of other arguments that would make someone likely to be unsure what the positional argument should be. At most you are specifying `porter installation apply FILE --namespace` and to me namespace falls into the same category as say debug, or output. It's a general modifier for a bunch of commands, but it isn't the focus of the command, and therefore is unlikely to be accidentally specified as the positional argument.

If we think that people will instinctively type -f FILE from muscle memory, we can do what we do for the help text (which is supported through --help and a help subcommand). We could add a hidden file flag and bind it, so that either way works, but we document the way that matches the rest of porter.

carolynvs: Do you think that I should add that hidden flag?

vdice: Your call!  I'd say we proceed w/o but if it's trivially easy to add... maybe so :)

Issue #1704, opened 2021-08-06
Reconcile changes applied to an installation and trigger the bundle


carolynvs:
When someone runs `porter installation apply` it should cause porter to diff the desired state expressed in the applied file with the current state of that bundle. If there are differences, the bundle should be executed to bring them in sync.

To support this Porter needs to start storing additional information on the installation document:
* The desired set of parameters to apply to the installation. These are user defined and do not include sources or defaults.
* The names of parameter and credential sets to apply to the installation. They can only reference another set in the current namespace, or in the global namespace. If a set is defined with the same name in both the namespace and the global namespace, the local set wins.

Changing the credential or parameter sets, or the list of parameters on an installation triggers reconciliation. If the credentials changed, we always re-run the bundle. Parameters only trigger a re-run if the values of any parameters have changed.

carolynvs:
Fixed in #1764

PR #1720, opened 2021-08-16
Apply installation

### LINKED ISSUE #1117: Docker host access flag not passed to dependency
## Describe the bug

It appears that the [allow-docker-host-access](https://porter.sh/configuration/#allow-docker-host-access) flag/setting isn't being passed to a dependency bundle.  If that dependency bundle lists `docker` as a required extension, it will fail to install with `extension "io.cnab.docker" is required but allow-docker-host-access was not enabled`.

## To Reproduce

Here I'll use our classic [wordpress](https://github.com/deislabs/porter/blob/main/build/testdata/bundles/wordpress/porter.yaml)  bundle which uses the [mysql](https://github.com/deislabs/porter/blob/main/build/testdata/bundles/mysql/porter.yaml) bundle as a dependency.

1. Update the [mysql](https://github.com/deislabs/porter/blob/main/build/testdata/bundles/mysql/porter.yaml) bundle to declare the `docker` extension as required.  Here's the minimal yaml to do so:
```
required:
  - docker
```
2. Update the tag on that bundle to point to a personal registry
3. Build and publish the bundle
4. Update the [wordpress](https://github.com/deislabs/porter/blob/main/build/testdata/bundles/wordpress/porter.yaml) bundle to use the published tag from above for the `mysql` dependency
5. Install the bundle with e.g. `porter install -c cred --allow-docker-host-access`

## Expected behavior

The mysql bundle should receive the same `--allow-docker-host-access` flag when it is being executed, so that installation is successful.

## Porter Command and Output
```
 $ porter install -c mk --allow-docker-host-access
Executing dependency mysql...
Error: error executing dependency mysql: unable to instantiate driver: extension "io.cnab.docker" is required but allow-docker-host-access was not enabled
```

## Version

`porter v0.27.0-3-g068a6b3a (068a6b3a)`

### LINKED ISSUE #1703: Support editing the underlying document for installations, credential and parameter sets

carolynvs:
# What does this change
Adds the command `porter installation apply FILE` with the following behavior:

* The created/modified timestamps are reset
* The status from the imported file is ignored
* Same as with params and creds, the namespace in the file takes precedence over the namespace specified with config/flags.

I have made an initial step in splitting up the responsibility for working with the installation record, and the cnab bundle execution. The goal (to support reconcile) is that install/upgrade/uninstall edit the installation record and then the bundle execution is based on the values on that installation record. For now that change is just plumbing to support apply but will be expanded upon in the next PR.

I've added a new function resolveBundleReference that consolidates our logic for resolving --reference, --file, --cnab-file, installation name. This sets us up better for simplifying that mess in the future. 😀 

I also did some refactoring around how we track and pass around key bundle information. I've created BundleReference with
* the bundle definition
* the bundle reference
* the relocation mapping file

I've also added OCIReference (wrapping reference.Named) that provides useful functions on a reference. Both of these types help us parse the bundle once, pass around a common struct and avoid repeat parsing or not having all the relevant info required to work with a bundle.

For this entire epic (reconciliation for the operator), I am removing difficult to setup/manage tests that are mostly testing our implementation and not the end behavior, and adding the use cases to our smoke test. The smoke test is a few long end-to-end workflows with bundles. It's fast because we combine a bunch of test cases into a single workflow, so we aren't actually executing a bundle 1-3 times a testcase like we are now. My plan is that unit tests are great for matrix tests and key edge cases, but if a test is going to duplicate what we are doing in the smoke test, I'm removing it. This should reduce the burden of updating tests that are too deep in the weeds of our implementation/testdata setup when we update Porter.

# What issue does it fix
Closes #1703
Fixes #1117 

# Notes for the reviewer
* Reconciliation of an installation is not triggered in this PR. That is a separate issue #1704 
* I know I'm leaving a lot of unresolved TODO(carolynvs) comments. At the end of the epic, they should all be addressed or converted into issues. This is my best effort to avoid scope creek and yak shaving since there's a lot of things I'm seeing as I implement this incrementally.
* This relies on an unmerged PR https://github.com/cnabio/cnab-to-oci/pull/112. 

# Checklist
- [x] Unit Tests
- [x] Documentation
- [ ] Schema (porter.yaml)


pkg/porter/dependencies.go
```diff
@@ -248,16 +249,16 @@ func (e *dependencyExecutioner) prepareDependency(dep *queuedDependency) error {
...
-		RelocationMapping: dep.RelocationMapping,
+	depInstallation := claims.NewInstallation(e.parentOpts.Namespace,
+		extensions.BuildPrerequisiteInstallationName(e.parentOpts.Name, dep.Alias))
 
+	depArgs := cnabprovider.ActionArguments{
+		BundleReference: dep.BundleReference,
+		Action:          e.parentArgs.Action,
+		Installation:    depInstallation,
+		Driver: e.parentArgs.Driver,
+		AllowDockerHostAccess: e.parentOpts.AllowAccessToDockerHost,
```
carolynvs, line 260: This fixes #1117. We just weren't propagating the flag previously.

pkg/porter/copy.go
```diff
@@ -4,63 +4,67 @@ import (
...
-		bundleNameRef := source[strings.LastIndex(source, "/")+1:]
-		return fmt.Sprintf("%s/%s", dest, bundleNameRef)
+		srcVal := source.String()
+		bundleNameRef := srcVal[strings.LastIndex(srcVal, "/")+1:]
+		dest = fmt.Sprintf("%s/%s", dest, bundleNameRef)
 	}
-	return dest
+	destRef, err := cnab.ParseOCIReference(dest)
+	if err != nil {
+		panic(err)
```
vdice, line 51: Want to double-check that we want to keep the panic here, only b/c we usually avoid doing so (besides unit test helpers, etc.)

pkg/porter/invoke_test.go
```diff
@@ -3,10 +3,48 @@ package porter
 import (
 	"testing"
 
+	"github.com/cnabio/cnab-go/bundle"
+	"github.com/cnabio/cnab-go/bundle/definition"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
+func buildExampleBundle() bundle.Bundle {
```
vdice, line 12: nit: this function appears to not be used here, but rather in the lifecycle_integration_test... should it move there?

tests/smoke/helpers_test.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		return claims.Installation{}, err
+	}
+
+	var installation claims.Installation
+	require.NoError(t.T, json.Unmarshal([]byte(output), &installation))
+	return installation, nil
+}
+
+func (t Test) AssertInstallationExists(namespace string, name string) claims.Installation {
```
vdice, line 58: minor nit on helper functions that start w/ `Assert...` but use `require` under the hood... thoughts on updating the helper functions to `Require...`?  (Or, at least the ones that do use require.)

tests/testdata/mybuns/porter.yaml
```diff
@@ -48,6 +68,27 @@ install:
...
+status:
+  - exec:
+      description: "Print status"
+      command: echo
+      arguments:
+        - "LGTM"
+
+boom:
+  - exec:
+      description: "modify the bundle in unknowable ways"
```
vdice, line 87: bout to get exciting!

vdice (commented):
Everything looks great to me.  Well done!  Love the consolidation and simplification w/ the new BundleReference struct.  This yak is lookin 😎 !

pkg/claims/installation.go
```diff
@@ -103,24 +105,59 @@ func (i Installation) NewRun(action string) Run {
...
-	// TODO(carolynvs): set this before running the bundle, it should trigger the execution
-	/*
-		repo, _, _, err := cnab.ParseBundleReference(run.BundleReference)
-		if err == nil {
-			i.BundleRepository = repo
-			i.BundleVersion = run.Bundle.Version // Use this instead of the tag because the tag doesn't have to be the version
-			i.BundleDigest = run.BundleDigest    // Use the actual digest of the bundle executed
+	if i.BundleVersion != "" {
+		if _, err := semver.NewVersion(i.BundleVersion); err != nil {
+			return errors.Wrapf(err, "Invalid bundleVersion. Must be a valid v2 semver value.")
```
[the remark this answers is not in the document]
carolynvs: v2 means Semantic Versioning v2. That is the version of semver that bundles support. The library has it's own versioning number that isn't related to this. https://semver.org/spec/v2.0.0.html

pkg/porter/copy.go
carolynvs: This function is supposed to panic. You're right that I don't use them elsewhere in the code. This function is only for tests, so that you can parse a well known reference in a test and it will hard fail if you mess it up, without having to remember to check the err. It also simplifies the return arguments so that you can do things like this

```
assert.Equal(t, MustParseOCIReference("getporter/porter-hello:v0.1.1"), myref)
```

if we used the other function that would look like this

```
gotref, err := ParseOCIReference("getporter/porter-hello:v0.1.1")
require.NoErr(t, err)
assert.Equal(t, gotref, myref)
```

The function name starts with Must, which is a pattern for functions like this that will panic when there is an error.

carolynvs:
@vdice Updated!

pkg/porter/copy.go
vdice: Wait, but I'm looking at the following logic in `generateNewBundleRef` which is called during `CopyBundle`, i.e. non-test code, right?

```
	destRef, err := cnab.ParseOCIReference(dest)
	if err != nil {
		panic(err)
	}
```

Just want to triple-check 😉 .  I did see `MustParseOCIReference` being used in tests -- good idea.

pkg/claims/installation.go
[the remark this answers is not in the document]
vdice: Ahhh, thank you. 👍

carolynvs:
I'm looking into why the dry-run action doesn't work on windows for my smoke test.

pkg/porter/copy.go
carolynvs: MustParseOCIReference (the one that can panic) is only used in tests. The other, cnab.ParseOCIReference, is used in real code but we always check the error returned. Sorry let me know if I'm missing something, totally not trying to blow you off here!

vdice: Gotcha.  Porter will panic on corner cases of `porter copy`; just want to make sure we're okay with that.  e.g.

```
$ p copy --source localhost:5000/nfs:v0.1.0 --destination oops/
panic: invalid reference format oops/: invalid reference format

goroutine 1 [running]:
get.porter.sh/porter/pkg/porter.generateNewBundleRef(0x2e3a168, 0xc000341e40, 0x7ffeefbff39d, 0x5, 0x105344f, 0x12)
	/Users/vdice/go/src/get.porter.sh/porter/pkg/porter/copy.go:51 +0x246
get.porter.sh/porter/pkg/porter.(*Porter).CopyBundle(0xc0004d7110, 0xc0001999c0, 0x0, 0x0)
	/Users/vdice/go/src/get.porter.sh/porter/pkg/porter/copy.go:58 +0x65
main.buildBundleCopyCommand.func2(0xc00039a780, 0xc000199b80, 0x0, 0x4, 0x0, 0x0)
	/Users/vdice/go/src/get.porter.sh/porter/cmd/porter/copy.go:29 +0x33
github.com/spf13/cobra.(*Command).execute(0xc00039a780, 0xc000199b40, 0x4, 0x4, 0xc00039a780, 0xc000199b40)
	/Users/vdice/go/pkg/mod/github.com/spf13/cobra@v1.1.3/command.go:852 +0x472
github.com/spf13/cobra.(*Command).ExecuteC(0xc000473900, 0xc000473900, 0x101a801, 0x0)
	/Users/vdice/go/pkg/mod/github.com/spf13/cobra@v1.1.3/command.go:960 +0x375
github.com/spf13/cobra.(*Command).Execute(...)
	/Users/vdice/go/pkg/mod/github.com/spf13/cobra@v1.1.3/command.go:897
main.main()
	/Users/vdice/go/src/get.porter.sh/porter/cmd/porter/main.go:23 +0x65
```

vdice (approved):

pkg/porter/copy.go
carolynvs: Oh! Thanks I see what you are pointing out now. Let me fix that

carolynvs:
I've fixed the panic in copy when the destination reference is invalid. Thanks for pushing on that, I missed that I was throwing a panic outside of "MustParseOCIReference"

PR #1723, opened 2021-08-20
Always pull the invocation image

### LINKED ISSUE #1439: Image digest validation can fail when building bundles locally
## Describe the bug
I'm not sure how to perfectly reproduce this but it's happened a number of times. 

## To Reproduce
1. Build a bundle locally
2. Push the bundle
2. Install it using --reference
3. Uninstall it using the name

It seems that since the invocation image is already in the docker cache, it doesn't pull what was pushed. The local image doesn't have the repository digest. So it fails the repo digest validation.

## Expected behavior
We should update the cache with the pushed image so this doesn't happen, or detect it and try to pull it down.

## Porter Command and Output
```
/tmp$ porter uninstall porterops -c porterops
uninstalling porterops...
params: [porter-debug]
creds: [kubeconfig]
Error: 1 error occurred:
	* image digest validation failed: image ghcr.io/getporter/porter-operator-installer:v0.1.0 has no repo digests

```

## Version
v0.33.0

carolynvs:
# What does this change
Sometimes the invocation image only exists in the local docker cache but was never pulled. In that case, there isn't a repository digest for the image in the cache and it will fail validation.

Always pulling, will ensure that the repo digest is always populated to avoid unexpected errors about missing digests.

The docker driver looks for the environment variable PULL_ALWAYS

# What issue does it fix
Fixes #1439

# Notes for the reviewer
The impact of this is that you cannot use --reference when you haven't pushed the invocation image. Mostly this just closes an accidental loophole where you could be using the definition of a bundle that has been published (and therefore has a digest in the bundle.json for the invocation image) with an invocation image that is only in the local docker cache. That is wonky and we shouldn't allow that regardless.

# Checklist
- [x] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/pkg/cnab/provider/driver.go
+++ b/pkg/cnab/provider/driver.go
@@ -48,7 +48,7 @@ func (r *Runtime) newDriver(driverName string, args ActionArguments) (driver.Dri
 	}
 
 	if driverName == "docker" {
-		// Always ensure that the local docker cache has the repository digests for the
+		// Always ensure that the local docker cache has the repository digests for the invocation image
 		os.Setenv("PULL_ALWAYS", "1")
 	}
 
--- a/pkg/cnab/provider/driver.go
+++ b/pkg/cnab/provider/driver.go
@@ -1,6 +1,8 @@
 package cnabprovider
 
 import (
+	"os"
+
 	"get.porter.sh/porter/pkg/cnab/drivers"
 	"get.porter.sh/porter/pkg/cnab/extensions"
 	"github.com/cnabio/cnab-go/driver"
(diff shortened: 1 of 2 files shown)


carolynvs:
This is hard to setup and not worth adding a regression test for. I have verified that porter will now stop and required that the image have been published when you use --reference.

pkg/cnab/provider/driver.go
```diff
@@ -45,6 +47,11 @@ func (r *Runtime) newDriver(driverName string, args ActionArguments) (driver.Dri
 		return nil, err
 	}
 
+	if driverName == "docker" {
+		// Always ensure that the local docker cache has the repository digests for the
```
vdice, line 51: "for the ... "

vdice (approved):

pkg/cnab/provider/driver.go
carolynvs: FOR THE HORDE!

PR #1726, opened 2021-08-20
Allow repeating install when --force is specified

carolynvs:
# What does this change
When building out a bundle for the first time, or troubleshooting,
sometimes it is incredibly helpful to be able to repeat the install
action, even when it was successful the first time.

For example, I am iterating on the install action for a bundle, adding
new steps each time I get the bundle to install successfully. In this
case, it makes sense to allow the bundle author to repeat the install
action with --force.

# What issue does it fix
I was getting stuck authoring bundles and realized the restriction I added was ... too restrictive.

# Notes for the reviewer
I realize that --force isn't documented this way. I think the --force flag makes sense though for this scenario. It's basically "do everything you can to make this work". 😀 

# Checklist
- [x] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)


vdice (approved):
Use the force! 🧙

PR #1728, opened 2021-08-20
Always collect outputs even when bundle fails

carolynvs:
# What does this change
When a bundle fails, we still need to collect any generated outputs because they may contain state that is needed for the next run.

What I am running into is that when I install a bundle that uses terraform, and it fails, the tfstate file output is not persisted, so I can't repeat the install command again.

# What issue does it fix
I found this bug and was immediately so enraged that I had to fix it immediately

# Notes for the reviewer
N/A

# Checklist
- [x] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"path/filepath"
 
+	"github.com/hashicorp/go-multierror"
+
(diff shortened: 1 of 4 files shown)


vdice (approved):

carolynvs:
I've rebased to resolve a merge conflict and kicked the integration tests. Will merge when they pass!

PR #1729, opened 2021-08-21
Print sensitive values if show-sensitive flag is set

### LINKED ISSUE #1708: Allow running porter without hiding sensitive values
I suggest adding a config setting that disables hiding sensitive values in porter's output. It is hard as hell to debug when you can't search for key values that you are looking for because they are ***'d out. This would be a v1 change, but could be released _after_ v1.

I originally wanted to put this behind the --debug flag but we have #1429 which would remove that flag entirely anyway. Plus this will be more explicit.

ThorstenHans:
# What does this change

This PR adds the `show-sensitive` persistent flag on the root command. It's defaulting to `false`

If the flag is set to `true`, sensitive values won't be masked in the output, instead they'll be directly written to the output.

# What issue does it fix
Closes #1708 

# Checklist
- [x] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/pkg/context/context.go
+++ b/pkg/context/context.go
@@ -47,8 +47,8 @@ func New() *Context {
 		environ:    getEnviron(),
 		FileSystem: aferox.NewAferox(pwd, afero.NewOsFs()),
 		In:         os.Stdin,
-		Out:        NewCensoredWriter(os.Stdout),
-		Err:        NewCensoredWriter(os.Stderr),
+		Out:        os.Stdout,
+		Err:        os.Stderr,
 	}
 	c.defaultNewCommand()
 	c.PlugInDebugContext = NewPluginDebugContext(c)
@@ -266,7 +266,7 @@ func (c *Context) WriteMixinOutputToFile(filename string, bytes []byte) error {
 
 // SetSensitiveValues sets the sensitive values needing masking on output/err streams
 func (c *Context) SetSensitiveValues(vals []string) {
-	if len(vals) > 0 {
+	if len(vals) > 0 && !c.ShowSensitiveValues {
 		out := NewCensoredWriter(c.Out)
 		out.SetSensitiveValues(vals)
 		c.Out = out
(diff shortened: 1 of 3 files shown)


ThorstenHans:
Have to update the docs before submitting this for a review

pkg/context/context.go
```diff
@@ -265,13 +266,19 @@ func (c *Context) WriteMixinOutputToFile(filename string, bytes []byte) error {
 
 // SetSensitiveValues sets the sensitive values needing masking on output/err streams
 func (c *Context) SetSensitiveValues(vals []string) {
-	if len(vals) > 0 {
+	if len(vals) > 0 && !c.ShowSensitiveValues {
```
carolynvs, line 269: This function is used to update the censored writer with new values to *** out of the logs, when they are written later. When ShowSensitiveValues is true, we can preemptively exit the function without further changes.

The changes here will immediately print the sensitive values from the disallow list to the console, which isn't appropriate for this function.

pkg/context/context_test.go
```diff
@@ -25,3 +26,19 @@ func TestContext_EnvironMap(t *testing.T) {
 	got["c"] = "3"
 	assert.Empty(t, c.Getenv("c"), "Expected to get a copy of the context's environment variables")
 }
+
+func TestContext_SetSensitiveValues(t *testing.T) {
```
carolynvs, line 30: To test this change, we want to do make a test context like you have done below, but then we should try two cases: setting some sensitive values and then writing to the context's Out and Err fields. Based on the value of ShowSensitiveValues, the captured output should either display *** in place of the sensitive values set, or print them without alteration.

carolynvs (changes requested):

carolynvs:
@ThorstenHans No rush but I wanted to check in to see if you need some help with this, or if any of my feedback was unclear.

carolynvs:
@ThorstenHans Would you be okay with me adding some commits to your branch to finish this PR so we can merge it?

ThorstenHans:
sure @carolynvs  and sorry for letting this PR in that state 😔

carolynvs:
Hey no worries! Everyone is busy and it's always nice to collaborate with people on a feature.

kichristensen:
Closed because of lack of activity. Please create a new PR if needed.

PR #1731, opened 2021-08-23
Fix issue #1217

### LINKED ISSUE #1217: Improve auto-detect of name vs filepath for Cred/Param Sets
@vdice, The examples you provided were enough to help me figure out that my issue had indeed been with the parameter auto-detection syntax.

If I have a parameter file in my working directory, `testparams.json`, then apparently 
`porter upgrade -p testparams.json`
will interpret testparams.json as a name.  Whereas if I run
`porter upgrade -p ./testparams.json`
it will interpret that as a path,

and if I have testparams.json in a subdirectory (params) 
`porter upgrade -p params/testparams.json`
will interpret as a file path.

I think this autodetection behavior is a little surprising, but I'm satisfied that the functionality I was looking for still exists.  Thank you for the explanation.

_Originally posted by @CSberger in https://github.com/deislabs/porter/issues/1213#issuecomment-675164466_

fibonacci1729:
# What does this change

When we load a parameter set, first we check if it is a file on the local filesystem. If so load the parameter set from the file. Otherwise assume it is a parameter set name and load it normally.

# What issue does it fix
Closes #1217

# Notes for the reviewer
N/A

# Checklist
- [ ] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/pkg/porter/parameters.go
+++ b/pkg/porter/parameters.go
@@ -3,9 +3,7 @@ package porter
 import (
 	"encoding/json"
 	"fmt"
-	"path/filepath"
 	"sort"
-	"strings"
 	"time"
 
 	"get.porter.sh/porter/pkg/cnab/extensions"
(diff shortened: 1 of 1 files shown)


carolynvs:
The builds are failing because of something that I just merged into release/v1. It's not related to your change.

PR #1734, opened 2021-08-24
Only pull sometimes

carolynvs:
Fixes a regression introduced in #1723. I missed that we shouldn't try to pull the invocation image to retrieve the repository digests unless the bundle has a repository digest defined in the bundle definition (which indicates that it has been published and is "pullable"). We still want to allow people developing locally to be able to run build then install without having ever published.

This showed up only in the integration tests:

```
dependencies_test.go:81: 
        	Error Trace:	dependencies_test.go:81
        	            				dependencies_test.go:27
        	Error:      	Received unexpected error:
        	            	1 error occurred:
        	            		* Error response from daemon: manifest for localhost:5000/wordpress-installer:v0.1.4 not found: manifest unknown: manifest unknown
        	            	
        	Test:       	TestDependenciesLifecycle
        	Messages:   	install of root bundle failed

```

### DIFF
--- a/pkg/cnab/provider/driver.go
+++ b/pkg/cnab/provider/driver.go
@@ -5,6 +5,7 @@ import (
 
 	"get.porter.sh/porter/pkg/cnab/drivers"
 	"get.porter.sh/porter/pkg/cnab/extensions"
+	"github.com/cnabio/cnab-go/bundle"
 	"github.com/cnabio/cnab-go/driver"
 	"github.com/cnabio/cnab-go/driver/docker"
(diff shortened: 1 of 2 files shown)


carolynvs:
I've realized in this PR that when the tests run together (instead of you just running a single test at a time), that the pull behavior of Porter is different. Running the test singly results in Porter not trying to pull the bundle. But running the entire integration test suite at once results in a pull when it shouldn't occur.

I'm going to pause working on this PR for a bit until after KubeCon and then figure that out.

carolynvs:
I'll try again later, closing.

PR #1743, opened 2021-08-26
Add state section to porter.yaml

carolynvs:
# What does this change
* Allow bundle authors to flag files as "state" that should be persisted by the bundle, but not exposed to the user (like a parameter or output is).
* Allow file parameters to be optional. When injecting a parameter into the bundle and the type is "file" do not inject the word null into the file in the bundle, omit it. I have an open issue to correct this in upstream cnab-go https://github.com/cnabio/cnab-go/issues/266.
*  Hide porter internal from explain command. Some outputs and parameters are really porter "plumbing". They are used by porter to implement things, like state or persistent parameters, but aren't intended to be set or used by the end-user.

    When we print out how to use a bundle with porter explain, those shouldn't be displayed.

    * Parameters with sources - these will be set by Porter internally.
    * Outputs and Parameters that support bundle state.

# What issue does it fix
Implements the new state section proposed in https://github.com/getporter/porter/discussions/1672.

# Notes for the reviewer
Preview of the state docs @ https://deploy-preview-1743--porter.netlify.app/author-bundles/#state

# Checklist
- [x] Unit Tests
- [ ] Documentation TODO
- [ ] Schema (porter.yaml) TODO


pkg/cnab/config-adapter/adapter.go
```diff
@@ -179,7 +181,7 @@ func (c *ManifestConverter) generateBundleParameters(defs *definition.Definition
 		if !param.Destination.IsEmpty() {
 			p.Destination = &bundle.Location{
 				EnvironmentVariable: param.Destination.EnvironmentVariable,
-				Path:                param.Destination.Path,
+				Path:                manifest.ResolvePath(param.Destination.Path),
```
carolynvs, line 184: In a follow-up PR I want to use this function so that people can use relative paths not just for the state files, but for params and creds. That way we should see a lot less of hard-coding absolute paths like /cnab/app/mystuff, when they just use "mystuff" and assume that everything is relative to the bundle directory.

pkg/porter/build_test.go
```diff
@@ -59,7 +59,7 @@ func TestPorter_Build(t *testing.T) {
 
 	stamp, err := configadapter.LoadStamp(bun)
 	require.NoError(t, err)
-	assert.Equal(t, "d421a6249dfbdba79e26e866da7533d59590565708dfdb32423cf989f588d0ea", stamp.ManifestDigest)
+	assert.NotEmpty(t, stamp.ManifestDigest)
```
carolynvs, line 62: I've gotten super tired of checking and updating the exact manifest digest. We just update it every time we tweak the bundle which isn't helpful.

pkg/runtime/runtime-manifest.go
```diff
@@ -424,43 +434,246 @@ func (m *RuntimeManifest) ResolveStep(step *manifest.Step) error {
...
 			}
 
 			// Porter by default places parameter value into file determined by Destination.Path
 			bytes, err := m.FileSystem.ReadFile(param.Destination.Path)
 			if err != nil {
-				return fmt.Errorf("unable to acquire value for parameter %s", param.Name)
+				return errors.Wrapf(err, "unable to acquire value for parameter %s", paramName)
 			}
 
+			// TODO(carolynvs): hack around parameters ALWAYS being injected even when empty files mess things up
```
carolynvs, line 466: I logged a bug against cnab-go https://github.com/cnabio/cnab-go/issues/266 and will work on fixing this after I get this stuff in.

pkg/runtime/runtime-manifest.go
```diff
@@ -424,43 +434,246 @@ func (m *RuntimeManifest) ResolveStep(step *manifest.Step) error {
...
+		err := packStateFile(tw, s)
+		if err != nil {
+			bigErr = multierror.Append(bigErr, err)
+		}
+	}
+
+	return bigErr.ErrorOrNil()
+}
+
+// applyUnboundBundleOutputs find outputs that haven't been bound yet by a step,
```
vdice, line 627:
```suggestion
// applyUnboundBundleOutputs finds outputs that haven't been bound yet by a step,
```

vdice (approved):
Awesome UX improvement with these changes!  LGTM.

PR #1745, opened 2021-08-27
Use official mongodb driver 

carolynvs:
# What does this change
I used the mgo driver (because it's easier to use) but it doesn't support the newer version of mongodb (4+) and also doesn't support the newer connection string format used exclusively by mongodb atlas (mongo's cloud offering for mongodb, its freee 💸 soooooooooo, we are totes gonna support it).

# What issue does it fix
Now I can do my kubecon demo using a cloud database that I don't have to pay for or keep updated myself. Huzzah!

# Notes for the reviewer
Sorry for not biting the bullet and just using the official driver when I first implemented this. The driver honestly isn't as easy to use but luckily most of the changes necessary are in just a few spots.

# Checklist
- [x] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)


carolynvs:
@vdice Can you take one more look? I found a connection leak when running the tests, so I added 2 more commits.

vdice (approved):
LGTM

PR #1753, opened 2021-09-08
Refactoring before adding reconcilation

carolynvs:
# What does this change
**Only store parameter overrides on the installation record**

Installation.Parameters should reflect parameter values that the user selected, similar to --param on the command line. It shouldn't contain any resolved values, such as from parameter sets as we already track those on the installation.ParameterSets field.

**Include calculated parameters in the json/yaml output**

When we output an installation with porter installation show, the data available in the plaintext version should be comparable to what's in the json/yaml output. Now that we are displaying all resolved parameters, but only storing the user-provided parameters on the installation record, we need to output more than just the installation record for those output formats.

I have changed DisplayInstallation to better wrap an Installation record, and then put all calculated fields into _calculated so that people can easily tell which fields are associated with the record (and can be changed when imported) and which ones are informational only.

**Fix test database creation**
The timeout was set to 0 when checking if there was an existing dev database that we could reuse. I thought 0 meant no timeout, but it means "timeout immediately". So we were always skipping the existing dev database. I've set a default in the database plugin so that we can't accidentally immediately timeout again.

**Create an installation record for dependencies**

When I refactored, it causes a bunch of errors when running a bundle with dependencies because I wasn't creating an explicit installation record for the dependency.

I have updated the code so that an installation is created and labeled with the following so that later we can filter them out if we want to:

```
sh.porter.parentInstallation = installation namespace/name
```

**Fix how we resolve parameters for dependencies**
This ensures that we use the relocated resolveParameters function when running a dependency, and that the final set of parameter used for the root bundle includes resolved parameters from the dependencies.

**Print params/creds before the bundle is run**
Print the params and creds passed to a bundle, including their values. Make sure that sensitive values are censored.

I was having a hell of a time debugging and once we have #1729 merged, it's super useful.

# What issue does it fix
This is refactoring prep-work for the new reconcillation capabilities of porter installation apply. The installation record should only contain _desired state_ and not include any status values. Otherwise it's really hard to reconcile and determine when changes should be made.

# Notes for the reviewer
N/A

# Checklist
- [ ] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)


pkg/porter/cnab.go
```diff
@@ -239,6 +239,10 @@ func (o *bundleFileOptions) validateCNABFile(cxt *context.Context) error {
 // LoadParameters validates and resolves the parameters and sets. It must be
 // called after porter has loaded the bundle definition.
 func (o *sharedOptions) LoadParameters(p *Porter) error {
+	if o.combinedParameters != nil {
```
vdice, line 242: Can we add a comment here around this logic?  Do we return immediately because o.combinedParameters being non-nil means they've been processed sufficiently?

pkg/porter/testdata/show/expected-output.yaml
```diff
@@ -20,3 +21,13 @@ status:
   bundleReference: getporter/wordpress:v0.1.0
   bundleVersion: 0.1.0
   bundleDigest: sha256:88d68ef0bdb9cedc6da3a8e341a33e5d2f8bb19d0cf7ec3f1060d3f9eb73cae9
+_calculated:
+  resolvedparameters:
+    - name: logLevel
+      type: integer
+      sensitive: false
```
vdice, line 28: When a resolved parameter is sensitive, is value omitted?  Or perhaps populated with `*******`?

Do we want/need to add a param of this type for these tests?  Or covered sufficiently elsewhere?

vdice (commented):
Looks great, just a few questions.

pkg/porter/cnab.go
carolynvs: Yup, it's because I noticed that we were re-processing the parameters, so this detects if we've already done it. I'll add a comment.

pkg/porter/testdata/show/expected-output.yaml
carolynvs: in plain text output we *** out the value, in json/yaml we don't so that people can actually read and use it.

vdice (approved):

carolynvs:
I've fixed the merge conflict and found a few unit tests that I had commented out while refactoring, so I uncommented them out and re-kicked off all the builds to make sure I didn't break anything.

PR #1755, opened 2021-09-13
Properly format numbers parsed from jsonPath outputs

### LINKED ISSUE #1754: Outputs captured with jsonPath format numbers in exponential notation
I have someone using the exec mixin's jsonPath output format
and a number in the output value is 123123123 without quotes -- it's not a string
when the jsonpath is applied to the exec output, it catches the correct value and converts it to e notation

```
api mimic
{
    "url":"/api/v2/jobs/12312312/",
    "id" : 123123123
}echoing values
1.23123123e+08
```

I'm assuming this automatic conversion is documented somewhere, yes? I'm not a go jsonpath expert

https://github.com/asalbers/test_bundle demonstrates the effect

Reported by @squillace  in Slack https://cloud-native.slack.com/archives/CN8NA4F8V/p1631558110034100

carolynvs:
# What does this change
When we parse outputs for jsonPath expressions, numbers are not being formatted as expected. For example, a number like "123123123" is being output as "1.23123123e+08".

This is due to how we are parsing the jsonPath. The output is marshaled into a generic map, where json numbers are represented as float64. We use fmt.Sprintf to format the matching output value, and Go by default formats float with %f/%g, neither of which is the desired format.

I've updated our json parsing to use the Number type so that the format of the captured value matches what was used in the original output.

# What issue does it fix
Fixes #1754

# Notes for the reviewer
They who provide test data to reproduce, get fast bug fixes. ❤️ 

# Checklist
- [x] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/docs/content/mixins/exec.md
+++ b/docs/content/mixins/exec.md
@@ -88,16 +88,22 @@ For example, if the `jsonPath` expression was `$[*].id` and the command sent the
 ]
 ```
 
-Then then output would have the following contents:
+Then the output would have the following contents:
 
 ```json
 ["1085517466897181794"]
 ```
 
 When you are developing your jsonPath expression, you can specify the --debug
-flag and the full json document with your query are printed to stderr so that you
+flag, and the full json document with your query are printed to stderr so that you
 can troubleshoot and improve your query based on the real result of the mixin's
 execution.
+
+Note: Porter attempts to preserve the original format of numeric values, so if the value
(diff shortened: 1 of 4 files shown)


vdice (approved):

carolynvs:
I'm going to hold off on anymore merges until our free oss status is fixed in Azure DevOps. That's why all the builds immediately fail. 💰

PR #1770, opened 2021-09-21
Lock down file permissions

### LINKED ISSUE #1765: Lock down file permissions
Right now most files in PORTER_HOME have 0755/0750. We should be using a more strict set of permissions so that only the current user can read the files. 

Here are the desired permissions:

**PORTER_HOME** - 0700
- config.toml - 0600
- mixins - 0700
- plugins - 0700
- data directories - 0600

**.cnab** - 0700
- run.sh - 0700
- mixins - 0700
- porter.yaml - 0600
- bundle.json - 0600

UPDATE: After speaking with others, some people may be running Porter in complex conditions where their permissions need to be more open. Or Porter won't be able to change the permissions. Based on that, we should just warn and then provide a special command `porter storage fix-permisions` that they can call if they want porter to help.

~I would like to just have Porter fix the permissions when they are found to be incorrect, but since other tools don't seem to do that I'm wondering if there's a good reason why to leave files insecure and just warn. If anyone has context on that, please share!~

~If we can't have porter fix permissions automatically for existing files, we can warn and tell people to run porter storage migrate, which will then fix the permissions for all files.~

carolynvs:
# What does this change
Only give permission to the current user, never to the group or other. Only mixins/plugins and scripts should be executable.

Note that on Windows, Go can't accurately report file permissions. Since files by default on Windows are not shared with other users, I am excluding windows from the permission check.

# What issue does it fix
Closes #1765 

# Notes for the reviewer
At this point the cnab-go version that porter v0.38 uses and the one that porter v1 use have diverged quite a bit. I have a branch of cnab-go in porter's fork that lets me make small patches just for our stable branch.

I did not submit the cnab-go patch that we are using in this PR here upstream to cnab-go because that code is going to be deleted very soon, because the crud package is being removed from cnab-go entirely. There's no point it patching it upstream.

https://github.com/getporter/cnab-go/commit/59a143bd08a5be8ba9b556dccd97222d624fb4cb

# Checklist
- [x] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)


pkg/config/datastore/loader.go
```diff
@@ -71,6 +71,7 @@ func buildDataLoader(viperCfg func(v *viper.Viper)) config.DataStoreLoaderFunc {
 		}
 
 		cfg.Data = &data
+		cfg.ConfigFilePath = v.ConfigFileUsed()
```
carolynvs, line 74: I am persisting this so that when I check file permissions I know which file in PORTER_HOME was the loaded config file, since it could have been config.toml/json/yaml.

pkg/porter/build_test.go
```diff
@@ -62,7 +84,7 @@ func TestPorter_Build(t *testing.T) {
 
 	stamp, err := configadapter.LoadStamp(*bun)
 	require.NoError(t, err)
-	assert.Equal(t, "d421a6249dfbdba79e26e866da7533d59590565708dfdb32423cf989f588d0ea", stamp.ManifestDigest)
+	assert.NotEmpty(t, stamp.ManifestDigest)
```
carolynvs, line 87: This is the most awful and unhelpful check, so I've fixed it to just make sure a stamp exists, but not the exact hash. This fix is already included the release/v1 branch.

pkg/porter/porter.go
```diff
@@ -67,6 +73,79 @@ func NewWithConfig(c *config.Config) *Porter {
...
+		})
+	}
+
+	if p.ConfigFilePath != "" {
+		if err := checkFile(p.ConfigFilePath); err != nil {
+			return err
+		}
+	}
+
+	dirs := []string{"claims", "outputs", "credentials", "parameters"}
```
carolynvs, line 140: We are only going to hard stop and complain about permissions for files that could contain sensitive values. If the permissions on a mixin or the cache is a bit wider, that isn't a big deal.

carolynvs:
Oops sorry I accidentally flipped this out of draft but I'm still fixing stuff!

pkg/porter/build_integration_test.go
```diff
@@ -1,29 +1,36 @@
+// +build integration
```
carolynvs, line 1: I made this an integration test because it hits the filesystem and relies on build having been run first.

vdice (approved):
🔐 👏

PR #1772, opened 2021-09-21
Use the helm3 mixin everywhere

carolynvs:
# What does this change
Update all usages of the old helm2 mixin to helm3. I did not change the default mixin installed with Porter because we are avoiding all breaking changes on v0.38 and this would count as a breaking change, even though no one should be using helm2 anymore.

The original wordpress+mysql test bundles ... never installed properly before. This fixes how we wire the connection string from mysql to the wordpress values.

# What issue does it fix
helm2 is deprecated and shouldn't be in our example bundles or documentation.

# Notes for the reviewer
~I'm waiting for https://github.com/MChorfa/porter-helm3/issues/33 before merging as our tests rely on creating namespaces on the fly.~

# Checklist
- [x] Unit Tests
- [x] Documentation
- [ ] Schema (porter.yaml)


build/testdata/bundles/mysql/porter.yaml
```diff
@@ -40,6 +44,8 @@ install:
     set:
       db.name: "{{ bundle.parameters.database-name}}"
       db.user: "{{ bundle.parameters.mysql-user }}"
+      primary.readinessProbe.timeout: "{{ bundle.parameters.probe-timeout}}"
```
carolynvs, line 47: I added these parameters so that I can demonstrate overriding a dependency parameter in one of our tests without messing with one of the important connection string parameters.

build/testdata/bundles/wordpress/porter.yaml
```diff
@@ -33,10 +34,10 @@ parameters:
     - upgrade
 - name: namespace
   type: string
-  default: ''
+  default: 'wordpress'
```
carolynvs, line 37: Setting namespace to a usable default.

build/testdata/bundles/wordpress/porter.yaml
```diff
@@ -45,16 +46,20 @@ install:
       replace: true
       set:
         wordpressPassword: "{{ bundle.parameters.wordpress-password }}"
+        externalDatabase.host: "porter-ci-mysql.{{ bundle.parameters.namespace }}.svc.cluster.local"
+        externalDatabase.database: "wordpress"
+        externalDatabase.user: "wordpress"
         externalDatabase.password: "{{ bundle.dependencies.mysql.outputs.mysql-password }}"
         externalDatabase.port: 3306
         mariadb.enabled: false
+        service.type: NodePort
```
carolynvs, line 55: The default type is loadbalancer which doesn't work with kind.

tests/integration/dependencies_test.go
```diff
@@ -65,9 +65,7 @@ func installWordpressBundle(p *porter.TestPorter) (namespace string) {
 	installOpts.Params = []string{
 		"wordpress-password=mypassword",
 		"namespace=" + namespace,
-		"wordpress-name=porter-ci-wordpress-" + namespace,
```
carolynvs, line 68: Since we are installing into a unique namespace, we don't need to make the release name unique. This was a leftover hack from using helm2

vdice (approved):

PR #1773, opened 2021-09-22
Add installation instruction to porter explain command

### LINKED ISSUE #1511: Explain how to use a bundle for the first time
This is how we expect people to use a bundle for the first time:
1.	User knows the bundle they want to use, e.g. porter explain -r ghcr.io/squillace/kubeflow:v0.1.9
2.	They run `porter explain --reference porter explain -r ghcr.io/squillace/kubeflow:v0.1.9` which tells you lots of information about the bundle, such as what parameters it accepts, and if credentials are required.
3.	They figure out where to get those credentials or what values they want to pass as parameters using porter credentials generate or porter parameters generate.
4.	They then run porter install passing in parameters and credentials.

It would be great if the `porter explain` command told you at the end of its output which commands to run next. For example,

```console
$ porter explain --reference getporter/hello-llama:v0.1.1
Name: hello-llama
Description: An example Porter bundle with parameters
Version: 0.1.1
Porter Version: v0.38.1-32-gb76f5c1c

No credentials defined

Parameters:
Name   Description                           Type     Default   Required   Applies To
name   Name of to whom we should say hello   string   llama     false      All Actions

No outputs defined

No custom actions defined

No dependencies defined

To install this bundle run the following command, passing --param KEY=VALUE for any parameters you want to customize:
porter install --reference getporter/hello-llama:v0.1.1
```

* Any required parameters should be included in the installation command, with `--param name=TODO`
* If credentials are defined, the output should tell them to run `porter credentials generate hello-llama --reference getporter/hello-llama:v0.1.1` first and then have the installation command include `--cred hello-llama`, e.g.

```
To install this bundle, run the following commands, passing --param KEY=VALUE for any parameters you want to customize:
porter credentials generate hello-llama --reference getporter/hello-llama:v0.1.1
porter install --reference getporter/hello-llama:v0.1.1 --cred hello-llama
```

The code for this command is located [here](https://github.com/getporter/porter/blob/7f8906befde8d9cd0e9a735bcc81e7b67127f17d/pkg/porter/explain.go#L311).

joshuabezaleel:
# What does this change
Add installation instructions for `porter explain` command which includes generating credentials first and then including required parameters to the installation instruction of a particular bundle. 

image

# What issue does it fix
Closes #1511 

# Notes for the reviewer

# Checklist
- [ ] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)


joshuabezaleel:
Hi @carolynvs I've added a draft PR for this particular issue and a screenshot showing the result of running the command.
Should I also add a new unit test using a new bundle json test file for this specific case? 

Thank you so much in advance! Looking forward to the review 🙂

carolynvs:
Since you already updated the existing unit tests output, there's no need for a new unit test. 👍

pkg/porter/explain.go
```diff
@@ -326,6 +326,7 @@ func (p *Porter) printBundleExplainTable(bun *PrintableBundle) error {
 	p.printOutputsExplainBlock(bun)
 	p.printActionsExplainBlock(bun)
 	p.printDependenciesExplainBlock(bun)
+	p.printInstallationInstructionBlock(bun, bundleReference)
```
carolynvs, line 329: This is really minor but would you please move this line under 331 where we print the tools included in the bundle? That way the output looks like this

```
$ porter explain -r somebundle
// a lot of information about that bundle
// the tools in the bundle
// BIG FINISH: How to install the bundle
```

pkg/porter/explain.go
```diff
@@ -464,3 +465,38 @@ func (p *Porter) printDependenciesExplainTable(bun *PrintableBundle) error {
...
+func (p *Porter) printInstallationInstructionBlock(bun *PrintableBundle, bundleReference string) error {
+	fmt.Fprint(p.Out, "To install this bundle run the following command, passing --param KEY=VALUE for any parameters you want to customize:\n")
+
+	var bundleReferenceFlag string
+	if bundleReference != "" {
+		bundleReferenceFlag += " --reference " + bundleReference
+	}
+
+	// Generate any predefined credentials first
+	for _, credential := range bun.Credentials {
```
carolynvs, line 478: We don't need to iterate for printing out how to generate credentials because when you run generate it will make a credential set that includes all of the credentials required by the bundle. So it can be a single statement that we only print out if the length of credentials is > 0.

The final command that we will tell them to run will be:

```
porter credentials generate mycreds -r REFERENCE
```

Normally if you leave off the name of the credential set, it will default to the name of the bundle. We could try to mimic that but I just checked and ... there's actually a potential bug in that code. So let's avoid that and just use a hard coded "mycreds" for the name in this command and in porter install. 😅

pkg/porter/explain.go
```diff
@@ -464,3 +465,38 @@ func (p *Porter) printDependenciesExplainTable(bun *PrintableBundle) error {
...
+		if parameter.Required {
+			requiredParameterFlags += parameter.Name + "=TODO "
+		}
+	}
+
+	if requiredParameterFlags != "" {
+		requiredParameterFlags += " --param "
+	}
+
+	var credentialFlags string
```
carolynvs, line 494: This can be a conditional "if there are creds then include a single -c mycreds flag", otherwise don't print anything.

carolynvs (changes requested):
Thanks, this is exactly what I was looking for. I have just some small revisions but it's looking good!

carolynvs:
I just checked out the latest code and tried it out:

```console
$ porter explain -r getporter/wordpress:v0.1.4
Name: wordpress
Description:
Version: 0.1.4
Porter Version: v1.0.0-alpha.3

Credentials:
Name         Description   Required   Applies To
kubeconfig                 true       All Actions

Parameters:
Name                 Description   Type     Default               Required   Applies To
namespace                          string   wordpress             false      All Actions
wordpress-name                     string   porter-ci-wordpress   false      All Actions
wordpress-password                 string   <nil>                 true       install,upgrade

Outputs:
Name                 Description                           Type     Applies To
wordpress-password   The Wordpress installation password   string   install,upgrade

Actions:
Name   Description   Modifies Installation   Stateless
ping   ping          true                    false

Dependencies:
Alias   Reference
mysql   localhost:5000/mysql:v0.1.4

This bundle uses the following tools: exec, helm3.
To install this bundle run the following command, passing --param KEY=VALUE for any parameters you want to customize:
porter credentials generate kubeconfig --reference getporter/wordpress:v0.1.4
porter install --reference getporter/wordpress:v0.1.4wordpress-password=TODO  --param  --cred kubeconfig
```

It's super close, there are just two things that need fixing:

* Let's start the instructions on it's own line, so that it's in a separate paragraph. Right now it kind of blends in with "This bundle uses the following tools"
* The parameters are printing out incorrectly. First because there isn't a space between the bundle reference and the parameter, but also because it should be `--param wordpress-password=TODO`, right now the flag value is printed before the flag.

carolynvs:
I'm working on a new alpha release, so this will be included when it goes out next week!

joshuabezaleel:
Hey @carolynvs , sorry that I was too tired yesterday because I just got back from a trip so I left the PR kind of hanging uncomfortably in the middle of doing it, and I am also not sure why 1 one of my commits got deleted while resolving the conflicts, thus the failing tests 🤔 Thank you for the on-point and detailed review! 

Here is the screenshot of the `explain` command using the `getporter/wordpress` reference that you've used as the example earlier, also fixed the order of the `--param` flag.

Thank you so much for responding and merging the PR this fast! 😄 

image

PR #1779, opened 2021-09-25
Add command to show historical runs of porter

### LINKED ISSUE #1699: Add command to show historical runs of porter
On main, porter show includes historical run data, with the actions performed on the installation along with their claim id.

On release/v1, once we merge the installation spec support PR (#1684), that data is no longer included in show by  default because it's not "free" data anymore and would require another query.

I propose that we split the show command into two commands, porter installation show (show the details of the installation) and porter installation runs list (list the historical runs for an installation). That gives the user more control over the data that will be provided and we don't make unnecessary (expensive) remote storage calls.

```console
$ porter installation runs list [NAME] [--namespace NAMESPACE] [--output FORMAT]
RUN ID     ACTION     STARTED          STOPPED          STATUS
claimid    install    10 minutes ago   5 minutes ago    suceeded
claimid    upgrade    1 minute ago                      running
```

This requires that the command make a query to both the runs collection and the results collection, filtering by installation + namespace, and then combining the documents on the client.

phillipahereza:
### What does this change
Adds the command `porter installation runs list [NAME] [--namespace NAMESPACE] [--output FORMAT]` that lists the historical runs for an installation

Screenshot 2021-09-25 at 15 39 50

### What issue does it fix
Closes #1699

### Notes for the reviewer
N/A

### Checklist
- [x] Unit Tests
- [x] Documentation
- [ ] Schema (porter.yaml)


pkg/porter/runs.go
```diff
@@ -0,0 +1,118 @@
...
+			continue
+		}
+
+		displayRun := NewDisplayRun(run)
+
+		if len(results) > 0 {
+			displayRun.Status = results[len(results)-1].Status
+
+			switch len(results) {
+			case 2:
```
carolynvs, line 73: It is possible to have more than 2 results. You can handle that here by making this case the `default:` case. Then instead of accessing `results[1]` to get the stopped timestamp, you can access the timestamp of the last entry in the array.

pkg/porter/runs_test.go
```diff
@@ -0,0 +1,54 @@
+package porter
+
+import (
+	"get.porter.sh/porter/pkg/claims"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestPorter_ListInstallationRuns(t *testing.T) {
```
carolynvs, line 10: This is a good test. Let's add one more testcase for checking the formatted command output.

The tests for `porter installation show` in pkg/porter/show_test.go demonstrate how to write a test that checks that the printed output looks the way that you expect. It captures the output from the command, and then compares it against expected output in a file.

The `p.CompareGoldenFile` function assists with updating the contents of those test files to match the current test output (which is helpful when you intentionally change the output of those commands). Once the test files exist, you can run `mage updateTestFiles` and all of those test files will be synced with the current output.

The tricky part with those tests is making sure that timestamps do not cause the output to be different every time that you run the test. So the test uses fixed timestamps to make comparing the outputs easier.

```
i := p.TestClaims.CreateInstallation(claims.NewInstallation("dev", "mywordpress"), p.TestClaims.SetMutableInstallationValues
```

The SetMutableInstallationValues function (there is also SetMutableRunValues, and SetMutableResultValues) updates the timestamps and IDs so that the tests are reproducible.

carolynvs (changes requested):
This looks great! 💯 Once we have the performance concern addressed and a test for the output, it should be good to go.

pkg/porter/runs.go
```diff
@@ -0,0 +1,118 @@
...
+
+	var displayRuns DisplayRuns
+
+	runs, err := p.Claims.ListRuns(opts.Namespace, opts.Name)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, run := range runs {
+		results, err := p.Claims.ListResults(run.ID)
```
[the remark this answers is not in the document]
phillipahereza: Not sure I understand what you mean here; is there a function that takes a list of run IDs and returns all the associated results or a function that takes an installation and namespace as args and returns all associated Results or do I still run  a loop and list results for each run?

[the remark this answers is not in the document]
carolynvs: Sure let's take a look at how results are stored, and what is available for us to query:

https://github.com/getporter/porter/blob/release/v1/pkg/claims/result.go

Results not only store the RunID (which is how you are looking up results currently), but they also store the associated installation name and namespace. This allows us to look up all results for an installation in a single query.

The mongodb filter to retrieve all results for an installation would look like this:

```json
{
  "namespace": "INSTALLATION_NAMESPACE",
  "installation": "INSTALLATION_NAME"
}
```

If you install the [MongoDB Compass](https://www.mongodb.com/products/compass) tool (it's free), you can use it to connect to the database that Porter creates by default and you can test out queries. I find that's a lot faster way to work on database queries than editing porter and then running it to see if it worked. The connection string for the default Porter database (which runs in a docker container) is `mongodb://localhost:27018`. It does not have a username or password.

What I am proposing is that you change the existing function on ClaimStore (and the Provider interface that it implements) to return not only the Runs for an installation, but the Results as well.

```golang 
# claimstore.go

func (s ClaimStore) ListRuns(namespace string, installation string) ([]Run, map[string][]Result, error) {
	var out []Run
	opts := storage.FindOptions{
		Sort: []string{"_id"},
		Filter: bson.M{
			"namespace":    namespace,
			"installation": installation,
		},
	}
	err := s.store.Find(CollectionRuns, opts, &out)

# TODO: Use similar code as what is above to query for associated Results, changing the collection name.

# TODO: iterate over the results, and insert them into a map[string][]Result. The map key should be result.RunID

	return out, resultsMap, err
}
```

Does that help clarify?

pkg/porter/testdata/runs/expected-output.json
```diff
@@ -0,0 +1,18 @@
+[
```
carolynvs, line 1: Did you forgot to check-in the new test that uses these files? I don't see where they are being used.

Looking at the test data here, I noticed that `action` and `status` are not populated, and that there is not a testdata file for plain text output. So that may be something to look at quick when you add that test.

carolynvs (changes requested):
Everything looks great, thanks for improving the performance of that function.

We just need to get the test that checks the command output figured out and it will be all set.

pkg/porter/testdata/runs/expected-output.json
carolynvs: Sounds good, let me know when this is ready to review again!

phillipahereza: Ready to review again

pkg/porter/testdata/runs/expected-output.txt
```diff
@@ -0,0 +1,42 @@
+[
```
carolynvs, line 1: Something about how the test is being set up isn't quite right. This file should be the output that someone would see if they ran the command and didn't set the format (i.e. what you see in your screenshot). It shouldn't be json. 

Same for the yaml output test case.

carolynvs (changes requested):

PR #1780, opened 2021-09-27
Make log persistence configurable

### LINKED ISSUE #1778: Make saving the logs from a run optional
By default Porter savings the logs from a bundle execution run as an output attached to the installation named `io.cnab.invocationImageLogs`. This is configurable in cnab-go, but we always set it to true right now.

In some cases, the amount of logs generated are too large to store in azure table storage. Users should be able to disable log persistence using Porter's standard [configuration](https://porter.sh/configuration).

```
# Disable log persistence with a flag
porter install mybun --no-logs

# Disable log persistence with an environment variable
PORTER_NO_LOGS=true porter install mybun
```

Or via configuration file
```
# ~/.porter/config.toml
no-logs = true
```

carolynvs:
# What does this change
Add a flag to the bundle actions allowing the user to skip saving a copy of the logs to Porter's datastore.

```
porter install --no-logs
```

Now when they run `porter logs -i mybuns`, it will return "no logs found".

This is useful in cases where the logs are too large to store.

# What issue does it fix
Closes #1778 

# Notes for the reviewer
I will cherry-pick this into v1 after this is merged.

# Checklist
- [x] Unit Tests
- [x] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/tests/smoke/hello_test.go
+++ b/tests/smoke/hello_test.go
@@ -28,12 +28,17 @@ func TestHelloBundle(t *testing.T) {
 
 	test.RequirePorter("install", "--reference", ref)
 	test.RequirePorter("installation", "show", "mybuns")
+	test.RequirePorter("installation", "logs", "show", "-i=mybuns")
 
 	test.RequirePorter("upgrade", "mybuns")
 	test.RequirePorter("installation", "show", "mybuns")
 
-	test.RequirePorter("uninstall", "mybuns")
+	// Check that we can disable logs from persisting
+	test.RequirePorter("uninstall", "mybuns", "--no-logs")
 	test.RequirePorter("installation", "show", "mybuns")
+	_, _, err = test.RunPorter("installation", "logs", "show", "-i=mybuns")
+	require.Error(t, err, "expected log retrieval to fail")
+	require.Contains(t, err.Error(), "no logs found")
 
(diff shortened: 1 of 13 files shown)


tests/smoke/hello_test.go
```diff
@@ -27,10 +27,15 @@ func TestHelloBundle(t *testing.T) {
...
 	test.RequirePorter("install", "--reference", ref)
 	test.RequirePorter("installation", "show", "mybuns")
+	test.RequirePorter("installation", "logs", "show", "-i=mybuns")
 
 	test.RequirePorter("upgrade", "mybuns")
 	test.RequirePorter("installation", "show", "mybuns")
 
-	test.RequirePorter("uninstall", "mybuns")
+	// Check that we can disable logs from persisting
+	test.RequirePorter("uninstall", "mybuns", "--persist-logs=false")
```
vdice, line 36: `--no-logs` instead here, right?

carolynvs: Oops yeah. I just pushed a fix for the smoke tests. I had originally gone with --persist-logs but that wasn't a good flag name because in general boolean flags are easiest to use when they default to false and you only specify it when you want it to be true.

carolynvs:
@vdice All fixed, please take a look.

vdice (approved):
LGTM!

Issue #1782, opened 2021-09-28
Move database initialization overhead into an init function


carolynvs:
There are activities that we run on every CLI command, that are not zero cost:

* Check the schema of the current database and determining if a migration is needed
* Ensuring indices are created on the current database

We could keep a file in PORTER_HOME that indicates if these tasks have been performed for the current storage configuration (hashed so that if the config changes, we repeat the check), and lets us only do this setup once per storage backend. This would save us 4 db calls per CLI command, which depending on the connection speed is ~1-2 seconds.

```yaml
# ~/.porter/cache/init.json
storage:
  CONNECTION_HASH:
    schema: # insert the schema from the database
    indices: # hash of indices
# maybe check once an hour or something, a TTL on the check essentially
```

This change should be made on the v1 branch.

carolynvs:
Don't forget to repeat the cosmosdb performance tests to verify that this brings performance back to acceptable levels.

PR #1783, opened 2021-09-28
Reduce how often we hit the database

carolynvs:
# What does this change
* Consolidate creating indices into just 3 calls. We can further refine this in #1782, when we make initializing the db a single event.
* Do not ping on connect in the mongodb plugin. This does mean that the first query will be slower as it will establish the connection. So let's be aware of that and not assume the first call just has horrible performance.

I am still looking into why the first call that establishes the connection has bad performance. https://www.mongodb.com/community/forums/t/establishing-a-connection-is-slow/125826

# What issue does it fix
I was creating each index in its own call but there's actually a different function on the mongodb client that I can use to batch those calls up to improve performance.

# Notes for the reviewer
NA

# Checklist
- [x] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/docs/content/plugins/mongodb.md
+++ b/docs/content/plugins/mongodb.md
@@ -52,4 +52,4 @@ localhost at the default port 27017, using the database name "mydb".
 ### timeout
 
 Sets the timeout (in seconds) used for database queries.
-The default timeout is 2 seconds.
+The default timeout is 10 seconds.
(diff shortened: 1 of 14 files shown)


vdice (approved):
LGTM, pending CI passing.

carolynvs:
Sorry about the test, those seem to be flakes that are only failing on CI. I assume my dev machine is beefy enough that's why I'm getting different (better) timings. 

I'll flail about on this with CI until it passes consistently!

carolynvs:
Yay! Fixed some test timeouts and we are good to go. 💨

PR #1785, opened 2021-09-30
Sync operator with porter v1

carolynvs:
# What does this change
These are changes necessary to get the Porter Operator updated to work with the latest version of Porter v1

* Log which config file we loaded, since the operator injects it in non-standard spot.
* Print the used config on a separate line. If the file didn't have a newline at the end, the logs were continuing on the same line as the config output.
* Allow specifying some runtime flags (normally specified on porter install) during porter apply: runtime-driver and
allow-docker-host-access.
* Log the runtime driver used

# What issue does it fix
N/A

# Notes for the reviewer
None

# Checklist
- [ ] Unit Tests
- [x] Documentation
- [ ] Schema (porter.yaml)


magefile.go
```diff
@@ -262,13 +262,24 @@ func PublishImages() {
 	}
 }
 
+// Builds the porter-agent image and publishes it to a local test cluster with the Porter Operator.
+func LocalPorterAgentBuild() {
+	// Publish to the local registry/cluster setup by the Porter Operator.
+	os.Setenv("REGISTRY", "localhost:5000")
+	// Force the image to be pushes to the registry even though its a local dev build.
```
vdice, line 269:
```suggestion
	// Force the image to be pushed to the registry even though it's a local dev build.
```

vdice (approved):
LGTM

PR #1815, opened 2021-11-05
Fix Pushing Bundles that have a relocationMap

### LINKED ISSUE #1814: CopyBundle doesn't respect relocationMaps
## Describe the bug
When copying an image using `porter copy` or the `CopyBundle()` function, the relocationMap isn't respected, and porter tries to copy everything from the original registries used to build the bundle, to which the user may not have access. This causes copies to fail.

## To Reproduce
1. Run porter copy on any bundle that contains images that you don't have access to.

## Expected behavior
I expected porter to use relocationMap to copy the relocated images, and not originals.

## Porter Command and Output
```
TBD
```

## Version
v0.38.7 from the Porter Go library.

anjayajodha:
# What does this change
This allows users to successfully copy bundles that use source images in a repository they don't have access to.

# What issue does it fix
Closes #1814 

# Notes for the reviewer

# Checklist
- [ ] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/pkg/cnab/cnab-to-oci/registry.go
+++ b/pkg/cnab/cnab-to-oci/registry.go
@@ -114,6 +114,36 @@ func (r *Registry) PushBundle(bun bundle.Bundle, tag string, insecureRegistry bo
 	return &rm, nil
 }
 
+// Pushes a bundle while using a relocationMap to correctly locate copied images. Identical to above, except for remotes.WithRelocationMap(reloMap)
+func (r *Registry) PushBundleWithRelocationMap(bun bundle.Bundle, tag string, reloMap relocation.ImageRelocationMap, insecureRegistry bool) (*relocation.ImageRelocationMap, error) {
+	ref, err := ParseOCIReference(tag) //tag from manifest
+	if err != nil {
+		return nil, errors.Wrap(err, "invalid bundle tag reference. expected value is REGISTRY/bundle:tag")
+	}
+	var insecureRegistries []string
+	if insecureRegistry {
+		reg := reference.Domain(ref)
+		insecureRegistries = append(insecureRegistries, reg)
+	}
+
+	resolver := r.createResolver(insecureRegistries)
+
+	rm, err := remotes.FixupBundle(context.Background(), &bun, ref, resolver, remotes.WithEventCallback(r.displayEvent), remotes.WithAutoBundleUpdate(), remotes.WithRelocationMap(reloMap))
+	if err != nil {
+		return nil, errors.Wrap(err, "error preparing the bundle with cnab-to-oci before pushing")
+	}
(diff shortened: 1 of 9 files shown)


carolynvs:
@vdice I was reviewing this PR and wondering if you had any extra context on why when Porter publishes from an archive, that we don't use a relocation mapping file and preserve the original bundle.

See this very wonderfully long function comment that you wrote for context:

https://github.com/getporter/porter/blob/784d61cf283bbd5b9c05a42ea718d3d042803305/pkg/porter/publish.go#L172-L186

This PR is correcting a bug in Porter where when we copy a bundle from one registry to another, we aren't preserving the relocation map. I was checking all the code that I thought would use that function and wasn't clear on why we are essentially "re-publishing" the bundle from the archive instead of preserving the original.

pkg/cnab/cnab-to-oci/registry.go
```diff
@@ -114,6 +114,36 @@ func (r *Registry) PushBundle(bun bundle.Bundle, tag string, insecureRegistry bo
 	return &rm, nil
 }
 
+// Pushes a bundle while using a relocationMap to correctly locate copied images. Identical to above, except for remotes.WithRelocationMap(reloMap)
+func (r *Registry) PushBundleWithRelocationMap(bun bundle.Bundle, tag string, reloMap relocation.ImageRelocationMap, insecureRegistry bool) (*relocation.ImageRelocationMap, error) {
```
carolynvs, line 118: There are three cases where we call the original function `PushBundle`:

1. `porter publish`, this doesn't have a relocation map.
2. `porter publish --archive` I have pulled in @vdice to get more context on this one.
3. `porter copy`, which is what you are fixing.

Porter does not have an API compatibility guarantee at the code level, so we are free to edit the signature of the original function. Would you please update the original function `PushBundle` with the additional parameter `reloMap relocation.ImageRelocationMap`. In case #1 above, since we don't have one, we can pass in an empty relocation map.

carolynvs (changes requested):
Before we can merge this, it will require a test that verifies the new copy behavior. Since there's been a bunch of churn in our integration testing and it's not terribly straightforward to add this type of test, I will push a commit to your PR adding in an integration test to validate this scenario (copying without access to the original images).

carolynvs:
Thanks @vdice! I'll create a follow-up issue to change publish from archive to preserve the original bundle and use a relocation map.

carolynvs:
Oops, sorry about that the new test I added failed when run on CI. I'll fix that right now.

carolynvs:
@anjayajodha Thanks for the patch! I have fixed the build and will merge once it passes. 👍

carolynvs (approved):
Thanks again for calling this out and fixing it! 💯

PR #1828, opened 2021-11-17
Porter triggers autobuild when detecting that invocation image does not exist

### LINKED ISSUE #1756: Porter's autobuild check should verify the invocation image is present
## Describe the bug
It's possible to get into a situation where porter's check for if the bundle should be rebuilt before publishing passes but the publish fails because the invocation image isn't in the docker cache.

## To Reproduce
1. Build a bundle.
2. Remove the invocation image from your docker cache. People will do this when they run out of disk space with commands like `docker system prune`.
3. Run publish

## Expected behavior
Porter detects that the invocation image isn't available, (even though the hash of the bundle has not changed) and rebuilds the bundle before publishing.

## Porter Command and Output
I don't have the exact error message when you publish but it's basically a message from docker when we do the push of the invocation image to the registry "blah blah invocation image name: no such image found"

## Version
All versions (v0.38.4 and v1 branch)

[e.g. porter v0.14.1-beta.2 (72dd5df)]

joshuabezaleel:
# What does this change
Check whether the invocation image exists in the host registry and trigger autobuild of the image by setting the return value of `IsBundleUpToDate` to `false`.

# What issue does it fix
Closes #1756 

# Checklist
- [ ] Unit Tests
- [ ] Documentation
- [ ] Schema (porter.yaml)

### DIFF
--- a/pkg/cnab/cnab-to-oci/provider.go
+++ b/pkg/cnab/cnab-to-oci/provider.go
@@ -18,5 +18,6 @@ type RegistryProvider interface {
 	// Returns the image digest from the registry.
 	PushInvocationImage(invocationImage string) (digest.Digest, error)
 
-	IsInvocationImageExists(invocationImage string) (bool, error)
+	// IsImageCached checks whether a particular invocation image exists in the local image cache.
+	IsImageCached(invocationImage string) (bool, error)
 }
--- a/pkg/cnab/cnab-to-oci/provider.go
+++ b/pkg/cnab/cnab-to-oci/provider.go
@@ -17,4 +17,6 @@ type RegistryProvider interface {
 	// the expected format of the invocationImage is REGISTRY/NAME:TAG.
 	// Returns the image digest from the registry.
 	PushInvocationImage(invocationImage string) (digest.Digest, error)
+
+	IsInvocationImageExists(invocationImage string) (bool, error)
 }
--- a/pkg/cnab/cnab-to-oci/registry.go
+++ b/pkg/cnab/cnab-to-oci/registry.go
@@ -13,6 +13,7 @@ import (
 	dockerconfig "github.com/docker/cli/cli/config"
 	cliflags "github.com/docker/cli/cli/flags"
 	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/filters"
 	"github.com/docker/docker/pkg/jsonmessage"
 	"github.com/docker/docker/pkg/term"
 	"github.com/opencontainers/go-digest"
@@ -187,4 +188,31 @@ func (r *Registry) getDockerClient() (*command.DockerCli, error) {
 		return nil, err
 	}
 	return cli, nil
+}
+
+func (r *Registry) IsInvocationImageExists(invocationImage string) (bool, error) {
+	ctx := context.Background()
+
+	cli, err := r.getDockerClient()
+	if err != nil {
+		return false, err
+	}
+
+	ref, err := cnab.ParseOCIReference(invocationImage)
+	if err != nil {
+		return false, err
+	}
+
+	imageListOpts := types.ImageListOptions{All: true, Filters: filters.NewArgs(filters.KeyValuePair{Key: "reference", Value: ref.Repository()})}
+
+	imageSummaries, err := cli.Client().ImageList(ctx, imageListOpts)
+	if err != nil {
+		return false, err
+	}
+
+	if len(imageSummaries) == 0 {
+		return false, nil
+	}
+
+	return true, nil
 }
--- a/pkg/cnab/cnab-to-oci/registry.go
+++ b/pkg/cnab/cnab-to-oci/registry.go
@@ -190,7 +190,7 @@ func (r *Registry) getDockerClient() (*command.DockerCli, error) {
 	return cli, nil
 }
 
-func (r *Registry) IsInvocationImageExists(invocationImage string) (bool, error) {
+func (r *Registry) IsImageCached(invocationImage string) (bool, error) {
 	ctx := context.Background()
 
 	cli, err := r.getDockerClient()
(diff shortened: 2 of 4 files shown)


joshuabezaleel:
Hi @carolynvs ! 
This is my attempt to tackle the #1756 issue, might not be the best one we should be pursuing with so looking forward to the review! 

So basically I added the method to check whether an invocation image exists on the RegistryProvider since the docker client is only available on the RegistryProvider and use that particular method to set the `upToDate` value to `false` which will trigger the `build` process. 

Along with this comment, I attached the video demo of it. The porter of `v0.38.7` is the version that I downloaded on the release page while the `v1.0.0-alpha.5-4-gca49edf8 (ca49edf8)` is from the one that I build locally from source.

https://user-images.githubusercontent.com/7043511/142193618-d56516f7-78e3-4e67-99b1-9d2250422b03.mp4

pkg/cnab/cnab-to-oci/registry.go
```diff
@@ -188,3 +189,30 @@ func (r *Registry) getDockerClient() (*command.DockerCli, error) {
...
+	if err != nil {
+		return false, err
+	}
+
+	ref, err := cnab.ParseOCIReference(invocationImage)
+	if err != nil {
+		return false, err
+	}
+
+	imageListOpts := types.ImageListOptions{All: true, Filters: filters.NewArgs(filters.KeyValuePair{Key: "reference", Value: ref.Repository()})}
```
carolynvs, line 206:
```suggestion
	imageListOpts := types.ImageListOptions{All: true, Filters: filters.NewArgs(filters.KeyValuePair{Key: "reference", Value: ref.String()})}
```

ref.Repository will exclude the tag, but we want to know if that tag exactly is present or not in the cache. ref.String will return the full formatted string representing the image.

We could also remove the call to ParseOCIImage I think. And then pass the image name directly to the Value field here. Up to you.

pkg/cnab/cnab-to-oci/registry.go
```diff
@@ -188,3 +189,30 @@ func (r *Registry) getDockerClient() (*command.DockerCli, error) {
...
+
+	imageSummaries, err := cli.Client().ImageList(ctx, imageListOpts)
+	if err != nil {
+		return false, err
```
carolynvs, line 210: Here's a bit of background on how we are handling errors in Porter. If the error came from a third party library or the stdlibrary (as opposed to Porter's code), it really helps to wrap the error and provide context about what action was being attempted with any supplemental information (like the name of the image in this case). That way we avoid printing errors to the console like "Error: file not found" or "Error: EOF" without context about what was going on. 😀 

Otherwise, if the function you called is Porter's code, do not wrap the error, because we already wrapped the error at the source and provided context. There's no need to re-wrap the error in most cases.

So in this case, the error here should be wrapped, and then in stamp.go line 67, you would not wrap the error and just return it as-is.

Let me know if you have questions about that, I realize it's a bit of a special thing that Porter does (and not always consistently like we should).

pkg/porter/stamp.go
```diff
@@ -60,6 +60,21 @@ func (p *Porter) IsBundleUpToDate(opts bundleFileOptions) (bool, error) {
...
+		// Check whether invocation images exist in host registry.
+		for _, invocationImage := range bun.InvocationImages {
+			isInvocationImageExists, err := p.Registry.IsInvocationImageExists(invocationImage.Image)
+			if err != nil {
+				return false, errors.Wrapf(err, "error while checking for existing invocation image")
+			}
+
+			if !isInvocationImageExists {
+				if p.Debug {
+					fmt.Fprintln(p.Err, errors.New(fmt.Sprintf("Invocation image %s doesn't exist in host registry, will need to build first", invocationImage.Image)))
```
carolynvs, line 72:
```suggestion
					fmt.Fprintln(p.Err, errors.New(fmt.Sprintf("Invocation image %s doesn't exist in the local image cache, will need to build first", invocationImage.Image)))
```

carolynvs (changes requested):
This is really close. Most of my comments are about how to get the code a bit closer to fitting in with the rest of porter's conventions. Since you are interested in eventually being a maintainer, I wanted to give you more information about how we have been doing things.

pkg/cnab/cnab-to-oci/provider.go
```diff
@@ -17,4 +17,6 @@ type RegistryProvider interface {
 	// the expected format of the invocationImage is REGISTRY/NAME:TAG.
 	// Returns the image digest from the registry.
 	PushInvocationImage(invocationImage string) (digest.Digest, error)
+
+	IsInvocationImageExists(invocationImage string) (bool, error)
```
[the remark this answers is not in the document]
joshuabezaleel: Ah yes, the new name makes so much more sense, and I just realized that the previous function name was even grammatically incorrect 😅

pkg/cnab/cnab-to-oci/registry.go
joshuabezaleel: I just realized that we can directly use the `invocationImage` name and don't need to use the reference from the `ParseOCIReference` because it yields the same value, thank you!

pkg/cnab/cnab-to-oci/registry.go
joshuabezaleel: This was a really good review, a new thing that I learned. Yep, the wrapping scenario makes sense to be done and I can see how it is a good practice to do so. Thank you for the really thorough explanation, Carolyn! 🙌

joshuabezaleel:
The workflow should still work just fine as the attached video below after the recent commit even though I'm not sure why did it take so long to build the image, perhaps it was just my machine or the internet connection. 
https://user-images.githubusercontent.com/7043511/148804851-f06b0bd5-ba42-4015-9cc4-6aa420474471.mp4

Since I'm new to the open source world I have to be honest that I often feel anxious as to making beginner mistakes here and there and will burden the team/maintainer but reading your code review has always been a delight as it has always been so thoughtful. Thank you lots, @carolynvs ! 😄

carolynvs (approved):
Thanks for fixing this bug! 💖

Issue #1787, opened 2021-10-01
Porter should help me write YAML


carolynvs:
Ain't no one got time to lookup the syntax when writing a YAML file from scratch. Save me from copy/pasting from other files!

Now that Porter has resources where the end user is expected to provide a properly formatted and defined resource (usually in YAML), we should do our best to make that as easy as possible.

- [ ] A command that outputs the corresponding CRD for an installation/credential/parameter, e.g. `porter installation show -o kubernetes`
- [x] A command that just makes a blank resource in either mode so that you don't have to lookup syntax. It should ideally be commented, but I would setting for "show me all the available fields" or "what's the least number of things I can specify and get away with it". Don't make me copy paste old files because it's too damn hard to read the docs. 😀 Maybe `porter installation create -o yaml|kubernetes|json` and `porter credentials create -o yaml|kubernetes|json`, etc. There may be an existing issue around this that I can link here...
- [ ] VS Code Extension should give autocomplete for both our CRDs and our resources. We will need to change the schema Version to include the kind of resource we are working with, e.g. they all can't use 1.0.0, it should be "installation-1.0.0" or something along those lines so that VS Code can get a hint.

carolynvs:
We needs this for the operator, and it will help implement #1080. However since we aren't replacing porter credentials generate with porter credentials create, it's not a breaking change and doesn't need to be tied to v1.

carolynvs:
Closing and will follow up with a separate issue for generating yaml for the operator. It may make more sense to document how to use kubectl to achieve that instead of making a custom command in porter.

Issue #947, opened 2020-03-11
Support creating a credential set from the command line


carolynvs:
When scripting porter, the user may know the sources for all the credentials ahead of time and not need to go through the interactive prompt that `porter credentials generate` walks the user through. In that case what they really want help with is not having to write out all the encoding themselves to make a credential set.

This command accepts the credential name and its sources and uses it build a completed credential set:

```
porter credentials create NAME [--cred SOURCE_NAME=SOURCE_KEY:SOURCE_VALUE]...
```

*  `NAME` is the name of the credential set and is required.
* `--cred|c` is the credential source. It is optional and may be specified multiple times.
* The format for credential sources is mapped onto the credential set like so:

```json
{
    "name": "NAME",
    "created": "CREATED",
    "modified": "MODIFIED",
    "credentials": [
        {
            "name": "SOURCE_NAME",
            "source": {
                "SOURCE_KEY": "SOURCE_VALUE"
            }
        }
    ]
}
```

See the GenerateCredentials command for ideas for how to go about implementing this command

https://github.com/deislabs/porter/blob/9110c71d4d65c5b1d2d0705b618b035efd93c672/cmd/porter/credentials.go#L42-L47

https://github.com/deislabs/porter/blob/9110c71d4d65c5b1d2d0705b618b035efd93c672/pkg/porter/credentials.go#L126-L144

It's not exactly the same, but they will both use `p.Credentials` to manage saving the credential set which that demonstrates how to do.

This will need unit tests that exercise our stub Credentials and use the afero file system, so look at the existing credentials tests and [this bundle build test](https://github.com/deislabs/porter/blob/9110c71d4d65c5b1d2d0705b618b035efd93c672/pkg/porter/build_test.go#L15-L21) for examples of how to do that.

carolynvs:
This is similar to #689 but not the same because it's not setting up a template for the user to finish. The user is expected to know what to enter, and this just saves them to making the entire properly encoded file themselves and saving it to porter's backend. Once this is implemented, it could use this code to implement itself though perhaps.

carolynvs:
@jeremyrickard @vdice Question for y'all. I believe that any character can be encoded into the credential source name. For example, you could have a credential source name like this

```json
{
    "name": "awfulcreds",
    "created": "CREATED",
    "modified": "MODIFIED",
    "credentials": [
        {
            "name": "hell:\n%$scape-~`\"'name",
            "source": {
                "env": "hellscape"
            }
        }
    ]
}
```

[...]

PR #1838, opened 2021-11-19
Add porter credentials create command

joshuabezaleel:
# What does this change
Add `porter credentials create FILE [--output yaml|json]` to create a blank resource defining a credential set.

# What issue does it fix
Closes 1 task out of 3 from #1787 

# Notes for the reviewer

# Checklist
- [x] Unit Tests
- [x] Documentation
- [x] Schema (porter.yaml)


joshuabezaleel:
Here I attached the demo of the `porter credentials create` command, looking forward to the review! 😄 


https://user-images.githubusercontent.com/7043511/142559319-723d7e5b-4368-4ea3-a724-4c3885739881.mp4

pkg/templates/templates.go
```diff
@@ -78,3 +78,13 @@ func (t *Templates) GetDockerfile() ([]byte, error) {
 	tmpl := fmt.Sprintf("templates/build/%s.Dockerfile", t.Data.BuildDriver)
 	return t.fs.ReadFile(tmpl)
 }
+
+// GetCredentialSetJSON returns a credential-set.schema.json template file to define new credential set.
+func (t *Templates) GetCredentialSetJSON() ([]byte, error) {
+	return t.fs.ReadFile("templates/credentials/create/credential-set.schema.json")
```
carolynvs, line 84: There is a bit of an inconsistency here between what the command does when json is requested (returning a schema) instead of yaml (returning a usable file with placeholders). The original issue was looking for the latter behavior, placeholders.

The schema for credential sets is stored in https://github.com/getporter/porter/blob/release/v1/pkg/schema/credential-set.schema.json and we point people to it from this documentation page: https://release-v1.porter.sh/reference/file-formats/

pkg/templates/templates/credentials/create/credential-set.schema.json
```diff
@@ -0,0 +1,89 @@
+{
```
carolynvs, line 1: Sorry for the confusion around what this template should return! If you have any improvements over the existing schema https://github.com/getporter/porter/blob/release/v1/pkg/schema/credential-set.schema.json, definitely add them to that file!

This template should be just the json version of the yaml file (obviously without comments):

```json
{
  "schemaType": "CredentialSet",
  "schemaVersion": "1.0.0",
  "name": "NAME",
  "namespace": "NAMESPACE",
  "labels": {
    "LABEL": "LABEL_VALUE"
  },
  "credentials": [
    {
      "name": "CREDENTIAL_NAME",
      "source": {
        "env": "CREDENTIAL_VALUE"
      }
    }
  ]
}
```

pkg/templates/templates/credentials/create/credential-set.yaml
```diff
@@ -0,0 +1,13 @@
...
+# Omit the namespace to define the credential set in the current current namespace
+namespace: NAMESPACE
+# Optionally define labels which can be used for filtering. Labels must be string values.
+labels:
+  LABEL: "LABEL_VALUE"
+credentials:
+  - name: CREDENTIAL_NAME
+    source:
+      # Replace the value field below with any of the following: secret, env, path, command or value
+      env: CREDENTIAL_VALUE
```
carolynvs, line 13: What do you think about having multiple credentials included in the template, each demonstrating a popular source: value, env, secret perhaps? Would that be helpful?

carolynvs (changes requested):
Sorry for all the feedback! The overall form of this is on-track. I'm just making sure that the user experience is consistent with the rest of Porter.

cmd/porter/credentials.go
```diff
@@ -214,3 +215,28 @@ func buildCredentialsShowCommand(p *porter.Porter) *cobra.Command {
...
+	opts := porter.CredentialCreateOptions{}
+
+	cmd := &cobra.Command{
+		Use:   "create",
+		Short: "Create a Credential",
+		Long:  "Create a new blank resource for the definition of a Credential Set.",
+		Example: `
+		porter credentials create FILE [--output yaml|json]
+		porter credentials create credential-set.json
+		porter credentials create credential-set --output yaml`,
```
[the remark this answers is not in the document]
joshuabezaleel: If we are discarding the possibilities of input with `porter credentials create credName --output outputType`, what will be the scenario of the user using the `--output` flag? Since the extension can be parsed from the `FILE` input (`FILE.{json|yaml}`) and which I assume will be the default usecase of the command.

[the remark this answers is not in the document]
joshuabezaleel: Since you mentioned that the default format will be `yaml`, will `porter credentials create creds` save a `creds.yaml` file to disk or return an error since the user didn't specify any format on the command input?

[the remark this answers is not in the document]
carolynvs: Sorry my feedback was a bit piecemeal and I didn't collect it all together to give a single recommendation about how the command should look.

What I am suggesting is the following:

```
porter credentials create [FILENAME] [--output OUTPUT]
```

* When a filename is not specified, we write to standard out in specified output.
* When --output is not specified, and the filename is set, try to use the file extension.
* When --output is not specified, and the filename is not yet, use yaml (and write to stdout).

Here are some examples showing how it looks
```
$ porter credentials create
schemaVersion: 1.0.0
name: NAME
# Omit the namespace to define the credential set in the current current namespace
namespace: NAMESPACE
# Optionally define labels which can be used for filtering. Labels must be string values.
labels:
    LABEL: "LABEL_VALUE"
credentials:
- name: CREDENTIAL_NAME
  source:
    # Replace the value field below with any of the following: secret, env, path, command or value
    value: CREDENTIAL_VALUE

$ porter credentials create mycreds.yaml
# writes yaml formatted creds to mycreds.yaml

$ porter credentials create mycreds.yml
# writes yaml formatted creds to mycreds.yml

$ porter credentials create mycreds.json
# writes json formatted creds to mycreds.json

$ porter credentials create creds
could not detect the file format from the file extension (.txt). Specify the format with --output.

$ porter credentials create creds --output json
# writes json formatted creds to mycreds
```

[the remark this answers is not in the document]
joshuabezaleel: Ah sorry, I should be the one listing all the possibilities of the command. Thank you for doing it, Carolyn 🙂 

> I am a bit concerned about this use case here, where they specify the file name but not the extension. It doesn't quite fit the existing behavior of porter and could easily result in people misunderstanding how the command works and wondering where the file went.
> 
> Let's remove this logic and just have Porter use the filename that they pass in as an argument without modification. It will be easier for users to predict what porter will do and in general one of our design principle's is to not surprise the user.

I apologize in advance if I still don't fully get it. On the first comment you reference to the use case of use specifying file name without the extension (`porter credentials create credential-set --output yaml` on my code) and said to remove the logic, but you mentioned the similar scenario on the last example on your list (`porter credentials create creds --output json` on your code snippet). How are those two different and which logic were you referring to that should be removed?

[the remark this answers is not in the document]
carolynvs: In all examples the positional argument is a file path. `porter credentials create creds -o json` should be interpreted as "generate the credentials in json format and save to a file named creds (without any extension)". I used that example specifically to help clarify that we shouldn't assert in code that the file extension MUST match the --output flag.

When no --output is specified, we try to guess the format from the file extension. If we can't we error out and tell them to provide --output (see second to last example). 

Does that help clarify? The problem I have with what is currently implemented is that we enforce that the --output and the file extension match.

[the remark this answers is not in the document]
joshuabezaleel: I literally had my eureka moment and said "OOH" out loud after understanding what you meant. First and foremost, I am really sorry that it took me some time get it and sorry if it frustrated you that you need to explain these multiple times. Thank you so much for the patience, @carolynvs ! Really aprpeciate it. 🙂 

There are still few things that I want to clarify before proceeding with the fixes:
1. At the examples that you listed above, you mentioned that the command of `porter credentials create creds` would yield error since we can't deduce the output format both from the filename because it didn't supply the extension and also because there is no `--output` flag input. But since you suggested [here](https://github.com/getporter/porter/pull/1838#discussion_r754508807) to use the default value of `yaml` for the `--output` flag, shouldn't the command `porter credentials create creds` then write the credentials in `yaml` format and save it to a file in disk name `creds` since the output flag with no input will automatically be filled with `yaml` as the flag default value?
2. As you mentioned [here](https://github.com/getporter/porter/pull/1838#discussion_r754503628) to use the `switch` primitive for checking the file format and since the `Marshal` example also mentioned it, do we also want the `.toml` file as one of the option for the `--output` format?

pkg/templates/templates/credentials/create/credential-set.schema.json
joshuabezaleel: Ah no, this should be my fault, I am sorry. Looking at it now, I am not sure why did I use the placeholders for the `.yaml` template but using the different format for the `.json` file back then 🤔 Thank you for the really good catch, Carolyn! Will fix it at the next commits.

pkg/templates/templates/credentials/create/credential-set.yaml
joshuabezaleel: This is a really great input which will definitely be more helpful to the user to provide more examples that encompass all of the possibilities of credentials. Thank you lots for the much needed reminder! 😄

cmd/porter/credentials.go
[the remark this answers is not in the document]
carolynvs: I actually suggested this when I was saying use yaml as a default. Note that no file was passed, so we are just printing to the console and not saving to a file.

```
$ porter credentials create
# defaults --output to yaml and outputs it to stdout
```

Maybe I've made this too confusing and users would have trouble picking up on the subtle differences between the ways to call this command though... 🤔 Like we could just not help write the file at all and say people have to use the redirect operators like `porter credentials create --output json > mycreds.json`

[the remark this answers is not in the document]
joshuabezaleel: @carolynvs The command example that you mentioned of having no file input passed (`porter credentials create`) was numbered 1 on my screenshot below while the one that I was referring was the command number 2 (`porter credentials create creds`), where user inputs the file (`creds`). 

image

It was mentioned that it should have yield error since we couldn't deduce the file format from either of the file input or the `--output` flag. But shouldn't it write YAML format output to the `creds` file at the disk since the `--output` flag will have a default value of YAML?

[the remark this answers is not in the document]
carolynvs: Sometimes the default value for a flag isn't a static value, and that's what I was suggesting. i.e. instead of setting the output default directly with cobra when we define the flag, the Validate function for the create options struct should default it with more complex logic.

1. When no filename is specified, default output to yaml.
2. When a file name is specified, default output using the file extension.

joshuabezaleel:
https://user-images.githubusercontent.com/7043511/152088874-11966097-48c8-45ff-aff4-cf23407d347b.mp4

Thank you so much for patiently answering my questions, @carolynvs , I've learned a lot. Hope this could answer the previous concerns. Looking forward to another round of review! :D

carolynvs:
I'm out sick for a few days, sorry I won't be able to review this until later next week!

joshuabezaleel:
@carolynvs Oh no, I'm sorry to hear that... No worries, your health should be the number one thing on your priorities. Hope you will get well really soon, Carolyn! Sending good thoughts and much love.

pkg/templates/templates/credentials/create/credential-set.yaml
```diff
@@ -0,0 +1,22 @@
...
+      command: echo 'credential command'
+  - name: credential-env
+    source:
+      env: CREDENTIAL_ENV
+  - name: credential-value
+    source:
+      value: credentialvalue
+  - name: credential-secret
+    source:
+      value: credential-secret
```
carolynvs, line 22: I was hoping that we could have inline comments that would help someone know what to do with this file.
```suggestion
# See the Credential Set file format documentation at https://release-v1.porter.sh/reference/file-formats/#credential-set
schemaType: CredentialSet
schemaVersion: 1.0.0
# Name of the credential set.
name: NAME
# The namespace in which the credential set is defined.
# Omit the namespace to define the credential set in the current current namespace.
namespace: NAMESPACE
# Optionally define labels which can be used for filtering. Labels must be string values.
labels:
  LABEL: LABEL_VALUE
# Credentials specify how Porter can resolve credentials when executing a bundle.
# Allowed sources are: value, path, env, secret, and command.
credentials:
  - name: credential-path
    source:
      # Resolve the credential value from the contents of the specified file.
      path: /path/to/credential-path-file.txt
  - name: credential-command
    source:
      # Resolve the credential value from the output of the specified command.
      command: echo 'credential command'
  - name: credential-env
    source:
      # Resolve the credential value from the value of the specified environment variable.
      env: CREDENTIAL_ENV
  - name: credential-value
    source:
      # Use the specified value for the credential. Do not use "value" for sensitive data.
      value: credentialvalue
  - name: credential-secret
    source:
      # Resolve the credential value from the secret store specified in your Porter configuration file.
      secret: credential-secret
```

carolynvs: On line 34, it should be "secret", not "value".

pkg/templates/templates/credentials/create/credential-set.schema.json
```diff
@@ -0,0 +1,41 @@
...
```
carolynvs, line 1: The file name is still confusing because it has "schema" in it (but this isn't a schema). Can you please rename it to "credential-set.json"

carolynvs (changes requested):
This looks great! Thanks for sticking with it as we worked out how we want the command to behave.

I have two small changes to the template and then we are ready to merge.

pkg/templates/templates/credentials/create/credential-set.yaml
joshuabezaleel: Ah, sorry for the "secret" typo. And thank you for adding the recommendations for the inline comments as well, Carolyn!

joshuabezaleel: The documentation site that we put in the inline comment for the people to refer to is the `release-v1` one (https://release-v1.porter.sh/reference/file-formats/#credential-set), should we add a TODO here to change it back to just porter.sh after the v1 got merged in the future? Since I assume that we will deprecate the `release-v1.porter.sh` site after it got merged (?)

joshuabezaleel:
@carolynvs Thank **_YOU_** for the really kind, thorough, and thoughtful guide and comment along the way of this particular and also other issues, Carolyn. Will always be grateful for it.

joshuabezaleel:
I assume this PR is also related to issues #947 (?).

pkg/templates/templates/credentials/create/credential-set.yaml
carolynvs: Instead of adding TODO's everywhere in the code (which is easy to forget!) I've added an issue and added it to the v1 milestone https://github.com/getporter/porter/issues/1949

carolynvs:
> I assume this PR is also related to issues https://github.com/getporter/porter/issues/947 (?)

Thanks for pointing that out. I've closed that issue as it is replaced by #1787.

carolynvs (approved):
Looks great, let's get this merged! 🚀

PR #1846, opened 2022-01-04
Allow mixins to ignore a failed command

### LINKED ISSUE #592: exec mixin should allow certain commands to have non-zero exit codes
Sometimes a command is allowed to have a non-zero exit code. For example when deleting a resource that is already gone, our mixins check for the "not found" text on stdout and then continue gracefully.

* Mixin authors should be able to define in the delete action text that means that the resource was already deleted, so that the action succeeds.
* Bundle authors should be able to define text for mixins extending exec (and for exec itself) where if that text is in stdout/err, then the exit code can be ignored and the command considered successful.

**Ignore a non-zero exit code when stderr/stdout contains a message**
```yaml
- exec:
    command: helm
    arguments:
    - delete
    - myRelease
    ignoreError:
      output:
        contains: ["not found", "not exist"]
```

**Ignore a non-zero exit code when stderr/stdout matches a regular expression**
```yaml
- exec:
    command: helm
    arguments:
    - delete
    - myRelease
    ignoreError:
      output:
        regex: ["release: .* not found"]
```

**Ignore some exit codes**
```yaml
- exec:
    command: helm
    arguments:
    - delete
    - myrelease
    ignoreError:
      exitCodes: [1, 127]
```

**Ignore all non-zero exit codes**
```yaml
- exec:
    command: rm
    arguments:
    - foo
    ignoreError:
      all: true
```

If we make the change to the our mixin framework directly (which a lot of mixins use), then other mixins besides the exec mixin can update and get this feature for free too

https://github.com/getporter/porter/blob/5efddb87b292184baceaf64a3301639e3949e4d1/pkg/exec/builder/execute.go#L145-L148

This should include test cases for gracefully handling errors, either in execute_test.go or in the exec mixins tests, whichever is easier to add a new test for. I don't believe we have an existing test that you can modify.

It's totally fine to just implement a single use case listed here. The best one to start with is ignoring all errors (last use case).

carolynvs:
# What does this change
Add the HasErrorHandling interface to the shared pkg/exec/builder
package used by all the mixins. When implemented, a mixin can examine a
failed command and optionally handle it.

The exec mixin now supports ignoring failed commands, which demonstrates
how to use builder.IgnoreErrorHandler struct to use the same error
handling behavior.

Here is the updated doc for the exec mixin: https://deploy-preview-1846--porter.netlify.app/mixins/exec/. I have added comments to the sample yaml config, and a section about ignoring errors.

# What issue does it fix
Closes #592 

# Notes for the reviewer
I am working on a PR for the az mixin that uses this to make resource groups and handle errors when it already exists. https://github.com/getporter/az-mixin/pull/34

# Checklist
- [x] Unit Tests
- [x] Documentation
- [x] Schema (porter.yaml)


pkg/context/helpers.go
```diff
@@ -71,6 +71,8 @@ func NewTestCommand(c *Context) CommandBuilder {
 		cmd.Env = []string{
 			fmt.Sprintf("%s=true", test.MockedCommandEnv),
 			fmt.Sprintf("%s=%s", test.ExpectedCommandEnv, c.Getenv(test.ExpectedCommandEnv)),
+			fmt.Sprintf("%s=%s", test.ExpectedCommandExitCodeEnv, c.Getenv(test.ExpectedCommandExitCodeEnv)),
```
carolynvs, line 74: I added these so that we can simulate a command failing in our unit tests.

vdice (approved):

PR #1851, opened 2022-01-06
Support uninstall through the installation apply command

carolynvs:
# What does this change
Allow someone to manage the entire installation lifecycle with a file, and not have to jump back to using uninstall. That causes problems when people are scripting porter calls based on what's in a repo, for example, they need to detect deleted files and treat them differently. Also if the bundle is already uninstalled, calling uninstall twice results in an error, whereas apply does the right thing.

Also as part of this I moved two fields, created and modified, that are really status and not part of the installation's desired state.

# What issue does it fix
Part of https://github.com/getporter/operator/issues/27

# Notes for the reviewer
I'd like your feedback on the new field `active`. If you have ideas for a more clear name, let me know.

Also I realize that maybe this sets people up for tripping over "I made an installation file (but forget to set active: true) and now it's not installing and I don't know why".  Perhaps there is a better name that we can use where defaulting to false, yields the more obvious behavior. So that if you omit this field, it still installs normally.

I could flip it to `inactive: true` so that you have to explicitly add that field to tell porter to delete it? But I'm not sure that people will get what inactive means... Suggestions welcome! 😅 

# Checklist
- [x] Unit Tests
- [x] Documentation
- [x] Schema (resource schema files)

### DIFF
--- a/docs/content/quickstart/desired-state.md
+++ b/docs/content/quickstart/desired-state.md
@@ -105,6 +105,7 @@ Create a file named installation.yaml, paste the following definition into the f
 ```yaml
 schemaVersion: 1.0.0
 name: desired-state
+active: true
 bundle:
   repository: getporter/credentials-tutorial
   version: 0.2.0
@@ -255,12 +256,34 @@ The installation is out-of-sync, running the upgrade action...
 Skipping bundle execution because --dry-run was specified
 ```
 
-## Cleanup
+## Uninstall
 
-To clean up the resources installed from this QuickStart, use the `porter uninstall` command.
+Installations have a field named **active** that control if the installation should be installed and managed by Porter.
+After a bundle has been installed, set active to false on the installation to uninstall it.
 
+Edit installation.yaml, set active to false, and then save the file.
+
+```yaml
+active: false
 ```
-porter uninstall desired-state
+
+The installation.yaml file should look like this:
+
+```yaml
+schemaVersion: 1.0.0
+name: desired-state
+active: false
+# remaining fields are not relevant to uninstalling
+```
+
+Now, apply the installation.yaml file to trigger an uninstall:
+
+```console
+$ porter installation apply installation.yaml
+Updated quickstart/desired-state installation
+Triggering because the installation is now inactive
+The installation is out-of-sync, running the uninstall action...
+# bundle output truncated for brevity
 ```
 
 ## Next Steps
--- a/docs/content/quickstart/desired-state.md
+++ b/docs/content/quickstart/desired-state.md
@@ -105,7 +105,6 @@ Create a file named installation.yaml, paste the following definition into the f
 ```yaml
 schemaVersion: 1.0.0
 name: desired-state
-active: true
 bundle:
   repository: getporter/credentials-tutorial
   version: 0.2.0
@@ -258,13 +257,13 @@ Skipping bundle execution because --dry-run was specified
 
 ## Uninstall
 
-Installations have a field named **active** that control if the installation should be installed and managed by Porter.
-After a bundle has been installed, set active to false on the installation to uninstall it.
+Installations have a field named **uninstalled** that control if the installation should be uninstalled.
+After a bundle has been installed, set uninstalled to true on the installation to uninstall it.
 
-Edit installation.yaml, set active to false, and then save the file.
+Edit installation.yaml, set uninstalled to true, and then save the file.
 
 ```yaml
-active: false
+uninstalled: true
 ```
 
 The installation.yaml file should look like this:
@@ -272,7 +271,7 @@ The installation.yaml file should look like this:
 ```yaml
 schemaVersion: 1.0.0
 name: desired-state
-active: false
+uninstalled: true
 # remaining fields are not relevant to uninstalling
 ```
 
@@ -281,7 +280,7 @@ Now, apply the installation.yaml file to trigger an uninstall:
 ```console
 $ porter installation apply installation.yaml
 Updated quickstart/desired-state installation
-Triggering because the installation is now inactive
+Triggering because installation.uninstalled is true
 The installation is out-of-sync, running the uninstall action...
 # bundle output truncated for brevity
 ```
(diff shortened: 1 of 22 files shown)


pkg/porter/reconcile_test.go
```diff
@@ -0,0 +1,174 @@
...
+			Parameters: map[string]interface{}{
+				"my-second-param": "spring-music-demo",
+			},
+		}
+		upgradeOpts := NewUpgradeOptions()
+		upgradeOpts.bundleRef = &cnab.BundleReference{Definition: bun}
+		insync, err := p.IsInstallationInSync(p.RootContext, i, &run, upgradeOpts)
+		require.NoError(t, err)
+		assert.True(t, insync)
+		// Nothing is printed out in this case, the calling function will print "up-to-date" for us
```
VinozzZ, line 66: I wonder if it's still a good idea to assert this behavior so we know when it changes in the future

pkg/porter/reconcile.go
```diff
@@ -122,12 +129,40 @@ func (p *Porter) ReconcileInstallation(ctx context.Context, opts ReconcileOption
...
+		}
+
+		// Otherwise ignore inactive
+		log.Info("Ignoring because the installation is inactive")
+		return true, nil
+	}
+
+	// We want to upgrade but we don't have values to compare against
+	// This shouldn't happen but check just in case
+	if lastRun == nil {
```
VinozzZ, line 160: Maybe I'm missing something here but I can't find a situation that this code path being executed

docs/content/quickstart/desired-state.md
```diff
@@ -255,12 +256,34 @@ The installation is out-of-sync, running the upgrade action...
...
+## Uninstall
 
-To clean up the resources installed from this QuickStart, use the `porter uninstall` command.
+Installations have a field named **active** that control if the installation should be installed and managed by Porter.
+After a bundle has been installed, set active to false on the installation to uninstall it.
 
+Edit installation.yaml, set active to false, and then save the file.
+
+```yaml
+active: false
```
VinozzZ, line 267: I wonder if it would be helpful to have separate config values for controlling `Uninstall` and `Managed Install`. The default behavior would be auto-installation and not uninstall. I feel having them being separate helps me understand what each of the config values mean
```yaml
autoInstall: false
uninstall: false
```

pkg/porter/reconcile_test.go
carolynvs: Other tests check for the "up-to-date" message that is printed. We can't check for it in this unit test because the function that prints it isn't called.

Here is my sad attempt at visualizing the call tree and who prints what
```
ApplyInstallation()
  -> IsInstallationInSync()
          prints if it should be triggered
  performs extra logic and prints if the installation is up to date or what action will be executed
```

pkg/porter/reconcile.go
carolynvs: That's why my comment included "This shouldn't happen but check just in case". 😀  It's just a quick check to prevent a panic in case our assumptions about how lastRun is passed changes.

docs/content/quickstart/desired-state.md
carolynvs: What I am having trouble with when it comes to naming the field is that the installation document is intended to represent the desired state of the installation, and shouldn't contain fields that represent a specific action. I originally thought to go with something like `uninstalled` but felt it didn't fit well with that approach.

I'm not sure what `dry-run` is supposed to indicate? Is that for when you create an installation that has active=false, which causes porter to just ignore the installation entirely?

carolynvs: I could be persuaded to use uninstalled though. I think we could make that work and remove the oddball case of "making a new installation with active=false should do nothing" which is a scenario I'm not sure we care to actually support. It would be pretty weird for someone to make a new installation and set "uninstalled" to true, since something can't be uninstalled until it's installed first.

I'm confusing myself! 😝

VinozzZ: > I'm not sure what dry-run is supposed to indicate? Is that for when you create an installation that has active=false, which causes porter to just ignore the installation entirely?

Yes, that's exactly what I meant by `dry-run`. My thoughts on the use case for this is maybe someone want to use the installation document to manage the state of an install action, instead of the `--dry-run` flag. But I agree that this is a somewhat odd case. 


In terms of the `uninstalled` state, my thinking is that if there's nothing installed then that means it's already the desired state since that's the result of `uninstalled: true`

pkg/porter/reconcile_test.go
VinozzZ: I see. That make sense. What about asserting that nothing is printed out, so we can confirm this expected behavior in the test?
```go
assert.Contains(t, p.TestConfig.TestContext.GetError(), "")
```

docs/content/quickstart/desired-state.md
carolynvs: So for uninstalled, it would default to false correct? If I apply an installation that doesn't exist yet with uninstalled=true, would porter do nothing? Store the record? That is how it works in this PR currently, and I'm trying to understand if we just want to rename active to uninstalled or if the behavior needs to change too.

carolynvs: To be clear I really don't like that right now active defaults to false, it kind of goes against how Porter does things, so this is probably the way we want to go. I'm just making sure I understand your feedback. 😀

carolynvs: I'm going to tweak this PR to use "uninstalled" instead of "active". If someone makes an installation with uninstalled = true, then we will not install the bundle.

pkg/porter/reconcile_test.go
carolynvs: When I went to add that check I discovered that other functions end up printing to stderr. I don't think it's worth asserting that nothing is printed in this case, it would just make the test more fragile.

I was checking for the existence of the "triggering" and "up-to-date" because they help give the user context into what's going on, and why things happened or didn't. But checking for the lack of printing something is something that is much easier to break when we make changes, like adding more log statements elsewhere. And elsewhere in our tests we do check for "up-to-date", so I feel like we have this code path covered pretty well already.

VinozzZ:
LGTM
One nit about the commit message. Should we update it to reference `uninstall` instead of `active` in there as well?

vdice (approved):

Issue #1857, opened 2022-01-11
Trace bundle execution


carolynvs:
Now that tracing is supported in #1831 the full bundle execution path should be supported. Right now we have a trace when a bundle starts/stops. But we don't have any insight into the bundle's execution. It would be great if we could see each mixin step's execution and runtime logs from porter-runtime.

If we passthrough the standard open telemetry environment variables that will work great for a listener that is _not_ on localhost. For example, if the listener is on a public ip or domain name, then the container can reach it. For localhost listeners though, we will need to get creative since "localhost" won't resolve to the host from the container. This is mainly a concern for local development though, so if we just make it work for public listeners, that would be a great start.

carolynvs:
TIL about `host.docker.internal` which lets you refer back to the docker host from inside a container. 🎉

PR #1864, opened 2022-01-19
pkg/{porter,tracing}, cmd/porter: add tracing instrumentation for list

VinozzZ:
command

# What does this change

This PR adds instrumentation for list command. It also adds support for
automatically setting span name during creation by the caller function
name.
Some fixes for unclosed spans is included as well

# What issue does it fix
related to #1857 

# Notes for the reviewer

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/build/buildkit/buildx.go
+++ b/pkg/build/buildkit/buildx.go
@@ -61,7 +61,7 @@ func (l unstructuredLogger) Write(p []byte) (n int, err error) {
 
 func (b *Builder) BuildInvocationImage(ctx context.Context, manifest *manifest.Manifest) error {
 	log := tracing.LoggerFromContext(ctx)
-	ctx, log = log.StartSpan("BuildInvocationImage", attribute.String("image", manifest.Image))
+	ctx, log = log.StartSpan(attribute.String("image", manifest.Image))
 	defer log.EndSpan()
 
 	log.Info("Building invocation image")
@@ -103,7 +103,7 @@ func (b *Builder) BuildInvocationImage(ctx context.Context, manifest *manifest.M
 
 	out := ioutil.Discard
 	if b.IsVerbose() || b.Config.IsFeatureEnabled(experimental.FlagStructuredLogs) {
-		ctx, log = log.StartSpan("buildkit", attribute.String("source", "porter.build.buildkit"))
+		ctx, log = log.StartSpan(attribute.String("source", "porter.build.buildkit"))
 		defer log.EndSpan()
 		out = unstructuredLogger{log}
 	}
--- a/pkg/build/buildkit/buildx.go
+++ b/pkg/build/buildkit/buildx.go
@@ -103,7 +103,7 @@ func (b *Builder) BuildInvocationImage(ctx context.Context, manifest *manifest.M
 
 	out := ioutil.Discard
 	if b.IsVerbose() || b.Config.IsFeatureEnabled(experimental.FlagStructuredLogs) {
-		ctx, log = log.StartSpan(attribute.String("source", "porter.build.buildkit"))
+		ctx, log = log.StartSpanWithName("buildkit", attribute.String("source", "porter.build.buildkit"))
 		defer log.EndSpan()
 		out = unstructuredLogger{log}
 	}
--- a/pkg/porter/list.go
+++ b/pkg/porter/list.go
@@ -124,7 +124,7 @@ func NewDisplayRun(run claims.Run) DisplayRun {
 
 // ListInstallations lists installed bundles.
 func (p *Porter) ListInstallations(ctx context.Context, opts ListOptions) ([]claims.Installation, error) {
-	_, log := p.Log.StartSpanNamedByCaller(ctx)
+	_, log := p.Log.StartSpan(ctx)
 	defer log.EndSpan()
 
 	installations, err := p.Claims.ListInstallations(opts.GetNamespace(), opts.Name, opts.ParseLabels())
@@ -132,7 +132,7 @@ func (p *Porter) ListInstallations(ctx context.Context, opts ListOptions) ([]cla
 }
 
 // PrintInstallations prints installed bundles.
-func (p *Porter) PrintInstallations(ctx context.Context, opts ListOptions) (err error) {
+func (p *Porter) PrintInstallations(ctx context.Context, opts ListOptions) error {
 	installations, err := p.ListInstallations(ctx, opts)
 	if err != nil {
 		return err
--- a/pkg/porter/list.go
+++ b/pkg/porter/list.go
@@ -10,7 +10,6 @@ import (
 	"get.porter.sh/porter/pkg/claims"
 	"get.porter.sh/porter/pkg/cnab"
 	"get.porter.sh/porter/pkg/printer"
-	"get.porter.sh/porter/pkg/tracing"
 	dtprinter "github.com/carolynvs/datetime-printer"
 	"github.com/pkg/errors"
 )
@@ -125,8 +124,7 @@ func NewDisplayRun(run claims.Run) DisplayRun {
 
 // ListInstallations lists installed bundles.
 func (p *Porter) ListInstallations(ctx context.Context, opts ListOptions) ([]claims.Installation, error) {
-	log := tracing.LoggerFromContext(ctx)
-	ctx, log = log.StartSpanNamed()
+	_, log := p.Log.StartSpanNamedByCaller(ctx)
 	defer log.EndSpan()
 
 	installations, err := p.Claims.ListInstallations(opts.GetNamespace(), opts.Name, opts.ParseLabels())
--- a/pkg/porter/list.go
+++ b/pkg/porter/list.go
@@ -1,6 +1,7 @@
 package porter
 
 import (
+	"context"
 	"fmt"
 	"sort"
 	"strings"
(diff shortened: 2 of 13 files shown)


pkg/tracing/scopedLogger.go
```diff
@@ -2,13 +2,16 @@ package tracing
...
+	"runtime"
+	"strings"
 
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
 )
 
 type ScopedLogger interface {
 	StartSpan(op string, attrs ...attribute.KeyValue) (context.Context, ScopedLogger)
+	StartSpanNamed(attrs ...attribute.KeyValue) (context.Context, ScopedLogger)
```
carolynvs, line 14: When should someone use StartSpan and explicitly name the op vs. using StartSpanNamed and have the op named after the current function?

Also it's not clear to me what StartSpanNamed will do. How about we tweak the name a bit so it's clear that the span will be named after the function, like `StartSpanFromFunction` or `StartSpanNamedByFunction` or something else like that?

pkg/tracing/scopedLogger.go
```diff
@@ -92,3 +99,17 @@ func (l scopedTraceLogger) Error(err error, attrs ...attribute.KeyValue) error {
 func (l scopedTraceLogger) Errorf(msg string, args ...interface{}) error {
 	return l.rootLogger.Errorf(l.span, msg, args...)
 }
+
+func callerFunc(frames int) string {
```
carolynvs, line 103: While we don't need unit tests for the logs themselves, this is a great function to write a unit test for.

carolynvs (changes requested):
Looks good, I just have some comments to help improve readability

pkg/tracing/scopedLogger.go
VinozzZ: From my perspective, a span is for a logical unit that usually contains side affect on a state. That often happens within a function. It also helps to create a trace that easily relate back to the code. We will know exactly which function created a span. I also changed the name to `StartSpanNamedByCaller`. Let me know what do you think?

pkg/tracing/scopedLogger.go
VinozzZ: Good idea! I will work on it in my next commit

pkg/tracing/scopedLogger.go
carolynvs: When would someone use StartSpan vs StartSpanByCaller? Or should we remove StartSpan and never give a custom name to a span?

pkg/tracing/logger.go
```diff
@@ -15,6 +17,7 @@ import (
 
 type RootLogger interface {
 	StartSpan(ctx context.Context, op string, attrs ...attribute.KeyValue) (context.Context, ScopedLogger)
+	StartSpanNamedByCaller(ctx context.Context, attrs ...attribute.KeyValue) (context.Context, ScopedLogger)
```
carolynvs, line 20: Thanks, this name is more clear to me what it will do! 👍

pkg/tracing/logger.go
```diff
@@ -50,21 +53,37 @@ func NewLogger(logger *zap.Logger, tracer trace.Tracer, attrs ...attribute.KeyVa
...
-	childSpan.SetAttributes(attrs...)
+
+	span.SetAttributes(attrs...)
 
 	isRootSpan := true
-	if s, ok := childSpan.(HasParentSpan); ok {
+	if s, ok := span.(HasParentSpan); ok {
 		isRootSpan = !s.ParentSpanID().IsValid()
 	}
+	// TODO: why do we set root span attributes onto child span as well?
```
carolynvs, line 82: That's not what this is doing. When we create a TraceLogger, there are some attributes that we want set on the root span (the very first span that is created by the cli). This checks if we just made the first span (root span) and sets attributes on the root. If this is a child span, nothing is set on it.

In particular, I am generating a correllation id that should be set on the root span, so that when we create a span in another process (like inside the invocation image) it can use the same correllation id so we can trace across system.

pkg/tracing/logger.go
```diff
@@ -50,21 +53,37 @@ func NewLogger(logger *zap.Logger, tracer trace.Tracer, attrs ...attribute.KeyVa
...
 
 func (l traceLogger) StartSpan(ctx context.Context, op string, attrs ...attribute.KeyValue) (context.Context, ScopedLogger) {
+	childCtx, childSpan := l.Tracer.Start(ctx, op)
+
+	l.addAttributesToSpan(childCtx, childSpan, attrs...)
+
+	return NewScopedLogger(childCtx, childSpan, l)
+}
+
+func (l traceLogger) StartSpanNamedByCaller(ctx context.Context, attrs ...attribute.KeyValue) (context.Context, ScopedLogger) {
```
carolynvs, line 63: The implementation for this function looks like it could be reduced down to the following, no? Though I'm still trying to figure out when someone would use StartSpan vs StartSpanByCaller?

```
return  l.StartSpan(ctx, callerFunc(0), attrs...)
```

pkg/tracing/logger.go
```diff
@@ -195,6 +214,12 @@ func (n consoleLogger) StartSpan(ctx context.Context, op string, attrs ...attrib
 	return NewScopedLogger(childCtx, childSpan, n)
 }
 
+func (n consoleLogger) StartSpanNamedByCaller(ctx context.Context, attrs ...attribute.KeyValue) (context.Context, ScopedLogger) {
```
carolynvs, line 217: This shouldn't re-implement StartSpan and instead we can reuse StartSpan and supply the op name from the caller

```
return n.StartSpan(ctx, callerFunc(0), attrs...)
```

pkg/tracing/logger.go
```diff
@@ -231,3 +256,28 @@ func (n consoleLogger) Error(span trace.Span, err error, attrs ...attribute.KeyV
...
+	}
+	// translate the PC into function information
+	frame, _ := runtime.CallersFrames(pc[:]).Next()
+	if frame.Function == "" {
+		return "unknown"
+	}
+
+	// extract function name from the return value which contains the complete
+	// import path for a function
+	// for example: "github.com/getporter/poter.ListInstallations"
```
carolynvs, line 277:
```suggestion
	// for example: "github.com/getporter/porter.ListInstallations"
```

pkg/porter/list.go
```diff
@@ -122,14 +123,17 @@ func NewDisplayRun(run claims.Run) DisplayRun {
...
+	defer log.EndSpan()
+
 	installations, err := p.Claims.ListInstallations(opts.GetNamespace(), opts.Name, opts.ParseLabels())
 	return installations, errors.Wrap(err, "could not list installations")
 }
 
 // PrintInstallations prints installed bundles.
-func (p *Porter) PrintInstallations(opts ListOptions) error {
-	installations, err := p.ListInstallations(opts)
+func (p *Porter) PrintInstallations(ctx context.Context, opts ListOptions) (err error) {
```
carolynvs, line 135: In Porter, I avoid using named return arguments. Since this function isn't using the named return argument err here (it's being immediately redefined below, can you switch it back to just `) error {`?

tests/integration/claim_migration_test.go
```diff
@@ -29,7 +31,7 @@ func TestClaimMigration_List(t *testing.T) {
 	err := p.MigrateStorage()
 	require.NoError(t, err, "MigrateStorage failed")
 
-	installations, err := p.ListInstallations(porter.ListOptions{})
+	installations, err := p.ListInstallations(context.TODO(), porter.ListOptions{})
```
carolynvs, line 34: This should be updated to `context.Background()` as well.

carolynvs (changes requested):
Let's think about when a developer would use StartSpan vs StartSpanByCaller and consider if we should just have a single function that uses the callerFunc. Right now I think other devs working on Porter will be unsure when to use each function.

pkg/tracing/scopedLogger.go
VinozzZ: I was thinking maybe someone would want multiple spans within a complex function and then they will need to manually provide names for them.

carolynvs:
I just checked out the failed build, and it's not from your changes. I think it's a flakey test and the next time it runs, it should pass.

pkg/build/buildkit/buildx.go
```diff
@@ -103,7 +103,7 @@ func (b *Builder) BuildInvocationImage(ctx context.Context, manifest *manifest.M
 
 	out := ioutil.Discard
 	if b.IsVerbose() || b.Config.IsFeatureEnabled(experimental.FlagStructuredLogs) {
-		ctx, log = log.StartSpan("buildkit", attribute.String("source", "porter.build.buildkit"))
+		ctx, log = log.StartSpan(attribute.String("source", "porter.build.buildkit"))
```
carolynvs, line 106: This seems like a good spot to use StartSpanWithName because we are creating a span within a function, right?

carolynvs (commented):
Everything looks great, I just have one more question (below).

pkg/tracing/logger_test.go
```diff
@@ -0,0 +1,21 @@
...
+		expected string
+		ok       bool
+	}{
+		{"", "", false},
+		{"porter/", "", false},
+		{"porter/v.", "", false},
+		{"github.com/getporter/porter/tracing.StartSpan", "StartSpan", true},
+	} {
+		fn, ok := extractFuncName(test.input)
+		if fn != test.expected || ok != test.ok {
```
carolynvs, line 17: You don't need to change this test, it's fine. But I did want to make you aware that we use github.com/stretchr/testify a lot to help with our test asserts. It helps craft the error message of a failed test so that you don't have to, and has a lot of useful built in assertions.

carolynvs (approved):
This is great thank you for setting up some patterns for us to follow as we instrument the rest of Porter! 💯

PR #1865, opened 2022-01-20
Print the porter version from the agent

carolynvs:
# What does this change
The porter agent used to print the porter version before executing the requested command. When I rewrote the bash script as a go program, that regressed and it stopped printing it out.

This adds that back.

# What issue does it fix
I noticed when testing the operator that this broke. It's mostly useful for troubleshooting since the image in the porter agent pod is by digest, not a tag that you could figure out the version from.

# Notes for the reviewer
N/A

# Checklist
- [x] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/pkg/errors"
 	"golang.org/x/sync/errgroup"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/agent/agent.go
```diff
@@ -63,9 +64,13 @@ func Execute(porterCommand []string, porterHome string, porterConfig string) (er
 	}
 
 	// Remind everyone the version of Porter we are using
+	fmt.Fprintf(stderr, "porter version\n")
 	cmd := exec.Command(porter, "version")
-	cmd.Stdout = stdout
+	cmd.Stdout = stderr // send all non-bundle output to stderr
```
VinozzZ, line 69: Just curious, why do we want to send all non-bundle output to stderr?

carolynvs: In general, I've learned that it helps to decide what a user should expect to see in stdout vs stderr up-front. For example, outputting a mix of debug into and the actual command output (like a json document) to stdout makes it impossible to parse. We are still correcting that in porter in some places (as part of our instrumentation efforts).

Thinking about what a user of the operator would expect to see in stdout for the porter agent container, I am assuming that they want the _output of the requested porter command_. So even though the agent is doing a few extra things, like here we are printing porter version for debugging, it makes it easier for someone to separate out debug info vs the actual command output if we send `porter version` to stderr.

Let me know if you see a problem with sending the output of version to stderr for this container!

VinozzZ (approved):

PR #1868, opened 2022-01-24
Porter explain warns for docker host access

### LINKED ISSUE #1830: explain should warn about a bundle needing docker host access
Right now when you run a bundle that needs the `--allow-docker-host-access` flag (because it has `require: ["docker"]`in the porter.yaml), you don't realize it until it fails to run.

The porter explain command should give people a heads up that this flag is needed, and that it accesses the host, which is a security concern and only trusted bundles should be allowed to do that. 

- [ ] Warn about docker host access
- [ ] Generate the right install command that specifies that flag

joshuabezaleel:
# What does this change
When the bundle has docker as one of the required mixins, `porter explain` command warns for the host access in advance and also include `--allow-docker-host-access` flag for the installation command at the end.

I attached the screenshot for running the `porter explain` command for the `tabbycat-demo` bundle with `docker` added as one of the required mixins and also for the `whalesay` bundle.
image

# What issue does it fix
Closes #1830 

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/porter/explain.go
+++ b/pkg/porter/explain.go
@@ -161,17 +161,17 @@ func (p *Porter) Explain(o ExplainOpts) error {
 	if err != nil {
 		return errors.Wrap(err, "unable to print bundle")
 	}
-	return p.printBundleExplain(o, pb)
+	return p.printBundleExplain(o, pb, bundleRef.Definition)
 }
 
-func (p *Porter) printBundleExplain(o ExplainOpts, pb *PrintableBundle) error {
+func (p *Porter) printBundleExplain(o ExplainOpts, pb *PrintableBundle, bun cnab.ExtendedBundle) error {
 	switch o.Format {
 	case printer.FormatJson:
 		return printer.PrintJson(p.Out, pb)
 	case printer.FormatYaml:
 		return printer.PrintYaml(p.Out, pb)
 	case printer.FormatPlaintext:
-		return p.printBundleExplainTable(pb, o.Reference)
+		return p.printBundleExplainTable(pb, o.Reference, bun)
 	default:
 		return fmt.Errorf("invalid format: %s", o.Format)
 	}
@@ -312,7 +312,7 @@ func generateApplyToString(appliesTo []string) string {
 
 }
 
-func (p *Porter) printBundleExplainTable(bun *PrintableBundle, bundleReference string) error {
+func (p *Porter) printBundleExplainTable(bun *PrintableBundle, bundleReference string, extendedBundle cnab.ExtendedBundle) error {
 	fmt.Fprintf(p.Out, "Name: %s\n", bun.Name)
 	fmt.Fprintf(p.Out, "Description: %s\n", bun.Description)
 	fmt.Fprintf(p.Out, "Version: %s\n", bun.Version)
@@ -329,21 +329,13 @@ func (p *Porter) printBundleExplainTable(bun *PrintableBundle, bundleReference s
 
 	fmt.Fprintf(p.Out, "This bundle uses the following tools: %s.\n", strings.Join(bun.Mixins, ", "))
 
-	// Check whether the bundle requires docker mixin and warn for host access.
-	var requireDocker bool = false
-	for _, mixin := range bun.Mixins {
-		if mixin == "docker" {
-			requireDocker = true
-		}
-	}
-
-	if requireDocker {
+	if extendedBundle.SupportsDocker() {
 		fmt.Fprintln(p.Out, "") // force a blank line before this block
 		fmt.Fprintf(p.Out, "🚨 This bundle will grant docker access to the host, make sure the publisher of this bundle is trusted.")
 		fmt.Fprintln(p.Out, "") // force a blank line after this block
 	}
 
-	p.printInstallationInstructionBlock(bun, bundleReference)
+	p.printInstallationInstructionBlock(bun, bundleReference, extendedBundle)
 	return nil
 }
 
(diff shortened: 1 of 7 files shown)


pkg/porter/explain.go
```diff
@@ -328,6 +328,21 @@ func (p *Porter) printBundleExplainTable(bun *PrintableBundle, bundleReference s
 	p.printDependenciesExplainBlock(bun)
 
 	fmt.Fprintf(p.Out, "This bundle uses the following tools: %s.\n", strings.Join(bun.Mixins, ", "))
+
+	// Check whether the bundle requires docker mixin and warn for host access.
+	var requireDocker bool = false
+	for _, mixin := range bun.Mixins {
```
carolynvs, line 334: Using the docker mixin isn't the problem we need to warn about. What we are looking for is when a bundle declares that it requires docker host access:

https://porter.sh/author-bundles/#required

```
required:
  - docker # this says inject the docker socket into the bundle
```

The preferred way to check for that is with `bun.SupportsX`, like `bun.SupportsFileParameters`, which checks if a CNAB required extension is defined on the bundle. There _should_ be a function named `bun.SupportsDocker` but I guess that function never made.

It should look like this:

https://github.com/getporter/porter/blob/6dab64a39100ff6cbf1aecbeaaf58b377f69a7fa/pkg/cnab/extensions/parameter_sources.go#L198

which should be added to https://github.com/getporter/porter/blob/main/pkg/cnab/extensions/docker.go

carolynvs (changes requested):
Can you also add a test case for the warning that should appear when the bundle requires the docker extension? Here's where we have a similar test already: https://github.com/getporter/porter/blob/6dab64a39100ff6cbf1aecbeaaf58b377f69a7fa/pkg/porter/explain_test.go#L123

pkg/porter/explain.go
joshuabezaleel: I just realized that I didn't stumble upon this one since the directory of `pkg/cnab/extensions` doesn't exist on `release/v1` branch but only exists on `main` branch. Does this mean that the PR for this particular issue should be made upon the `main` branch?

carolynvs: Oops, sorry I shouldn't have linked to the main branch. This PR should target release/v1 but the code that I wanted to show you as an example moved a bit: https://github.com/getporter/porter/blob/a93a44a1bbf1e43bf96a80310b34e33a90de52f2/pkg/cnab/parameter_sources.go#L204-L206.

joshuabezaleel:
@carolynvs Also I found this particular TODO line while working on this issue. Does this mean that we have to address this before `v1.0`? And if the answer to the previous question is yes, do we already have a corresponding issue for this one? 

https://github.com/getporter/porter/blob/release/v1/pkg/cnab/file_parameter.go#L39-L42

joshuabezaleel:
Wait why can't my link for line of code on GitHub be rendered as a snippet like yours 🤔

carolynvs:
> Also I found this particular TODO line while working on this issue. Does this mean that we have to address this before v1.0? And if the answer to the previous question is yes, do we already have a corresponding issue for this one?

oof, that is a good question. If we removed that and relied 100% on if the bundle says that it supports file parameters, then old bundles would stop working in porter v1.0.0. I don't really see a good reason to do that, and it's just a single line to keep around to keep old bundles working. We should just remove that todo and not remove that line of code.

pkg/porter/explain.go
```diff
@@ -494,10 +501,18 @@ func (p *Porter) printInstallationInstructionBlock(bun *PrintableBundle, bundleR
...
 	var credentialFlags string
 	if len(bun.Credentials) > 0 {
-		credentialFlags += "--cred mycreds"
+		credentialFlags += " --cred mycreds"
 	}
 
-	fmt.Fprintf(p.Out, "porter install%s%s%s\n", bundleReferenceFlag, requiredParameterFlags, credentialFlags)
+	porterInstallCommand := fmt.Sprintf("porter install%s%s%s", bundleReferenceFlag, requiredParameterFlags, credentialFlags)
+
+	// Check whether the bundle requires docker mixin and add flag for host access for install command.
```
carolynvs, line 509: This comment is a bit misleading since we aren't checking for the docker mixin, but if the bundle requires the docker socket to be mounted into the bundle.

pkg/porter/explain_test.go
```diff
@@ -38,14 +38,37 @@ func TestExplain_generateTable(t *testing.T) {
...
+	opts := ExplainOpts{}
+	opts.RawFormat = "plaintext"
+
+	err = opts.Validate([]string{}, p.Context)
+	require.NoError(t, err)
+
+	err = p.printBundleExplain(opts, pb, b)
+	assert.NoError(t, err)
+	gotOutput := p.TestConfig.TestContext.GetOutput()
+	expected, err := ioutil.ReadFile("testdata/explain/expected-table-output-docker.txt")
```
carolynvs, line 67: We have a test helper that makes it easier to compare command output with the contents of a file and keep it in sync when the command output changes.

```
import "get.porter.sh/porter/pkg/test"

test.CompareGoldenFile(t, "testdata/explain/expected-table-output-docker.txt", gotOutput)
```

If you run `mage updateTestFiles` it will find any tests whose output has changed and updates the testdata file automatically so the test passes.

We aren't using that new helper function everywhere in Porter yet, but I'd like new tests to use it so they are easier to maintain.

carolynvs (commented):
This looks great! I just had some small suggestions. Once the unit tests are passing (there's a build failure in the tests), this is ready to go! 🚀

joshuabezaleel:
@carolynvs Sorry for the failing test, my bad. Should be all green by now. Thank you Carolyn! 😄

carolynvs (approved):
I love all the improvements to porter explain! It really makes it a much better experience for new users and helping them learn how to use bundles. 👍

joshuabezaleel:
I really hope the `explain` command will come useful for people using it! 😄

PR #1875, opened 2022-01-27
Only include required install params in explain command

carolynvs:
# What does this change
When porter explain prints a suggested installation command, it should
only include parameters that are required _and_ apply to the install
command. Parameters can be limited to only apply to a particular action
and if it doesn't apply to install, we don't want to tell people to
specify in for porter install.

# What issue does it fix
I found this when running porter explain on the operator bundle, which has a parameter that only applies to the configureNamespace action.

# Notes for the reviewer
N/A

# Checklist
- [x] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/porter/explain.go
+++ b/pkg/porter/explain.go
@@ -84,6 +84,7 @@ type PrintableDependency struct {
 }
 
 type PrintableParameter struct {
+	param       *bundle.Parameter
 	Name        string      `json:"name" yaml:"name"`
 	Type        interface{} `json:"type" yaml:"type"`
 	Default     interface{} `json:"default" yaml:"default"`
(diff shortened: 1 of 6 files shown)


vdice (approved):

pkg/porter/explain.go
```diff
@@ -230,6 +231,7 @@ func generatePrintable(bun cnab.ExtendedBundle, action string) (*PrintableBundle
 	sort.Sort(SortPrintableCredential(pb.Credentials))
 
 	for p, v := range bun.Parameters {
+		v := v // Go closures are funny like that
```
joshuabezaleel, line 234: @carolynvs May I know the reason behind this one? Might be something that I haven't known yet about Go 😕

carolynvs: Yeah it's super obscure and a really annoying part of how Go handles "closures". If you search for "golang for loop closure" you'll find a lot of articles on it.

Without this hack, when I save off a reference to `v` on the display parameter, it ends up with all parameters having a reference to the same address instead of each display parameter getting a reference to its proper parameter.

joshuabezaleel: Ah, I just realized they mentioned it on the official repo. It's this [one](https://github.com/golang/go/wiki/CommonMistakes#using-reference-to-loop-iterator-variable), right? Thank you for the nugget of knowledge, Carolyn! Appreciate it 😄

carolynvs: Yup! That is the problem I was working around. 👍

Issue #316, opened 2019-05-03
Porter's Template Language


carolynvs:
# Problem Background

Let's say a step generates an output file, and the next step wants to use it how would that look?

The next step needs to say, "I want to use this output as a file"

```yaml
install:
  - aks:
      outputs:
      - name: kubeconfig
  - helm:
      kubeconfig: "{{ path bundle.outputs.kubeconfig }}"
```

One problem that I found is that we are using mustache and that doesn't support helper functions, so we would need to move to handlebars. Not a huge deal.

But as I thought about this further, I realized that we have a problem representing int and bool types right now. The problem with int/bool is a problem with json,toml and hcl too, not just yaml. **Any templating on top of markup languages where we need to read the markup before parsing the template is** a minefield. Yaml is the least problematic because it doesn't require strings to be quoted.

Consider the following:

 ```
 port: "{{ bundle.outputs.port }}"
 ```

We had to have those quotes in order to load and use the manifest. Unfortunately, there is no way for the templating engine to get rid of the wrapping quotes, and no way for porter to either without awkward hacks. So when a mixin reads in and validates `port`, it is going to fail because its schema required an int (or bool) not a string.

# Porter's Own Template Language

As we found above, mustache (the template engine we use currently) doesn't allow for helper functions, so all my plans include moving to handlebars. Handlebars is a superset of mustache that supports helper functions. This will allow users to transform outputs which is going to be critical for porter.

NOTE: Since we plan to use handlebars under the hood there is nothing stopping someone from quoting properly and using handlebars up the wazoo. What I am proposing here is just nicer syntax which that also fixes the problems outlined above with int and bool.

## Option 1: Ugly Handlebars
```yaml
install:
  - aks:
      outputs:
      - name: kubeconfig
  - helm:
      kubeconfig: <% path bundle.outputs.kubeconfig %>
      set:
        kubeport: <% bundle.outputs.kubeport %>
        db: "<% bundle.params.dbhost %>:<% bundle.outputs.dbport %>"
```

1. Replace `<% %>` with `{{ }}`
1. Run handlebars

Pros:
* Let's people use their existing knowledge/docs of handlebars
* Workaround the problem with requiring surrounding quotes for {{ }}

Cons:
* Not so easy on the eyes, verbose

## Option 2: Transparent Handlebars

```yaml
install:
  - aks:
      outputs:
      - name: kubeconfig
  - helm:
      kubeconfig: bundle.outputs.kubeconfig
      set:
[...]

PR #1900, opened 2022-02-08
Enable setting custom values during porter build.

### LINKED ISSUE #1789: Suggestion: Set custom values during porter build
I would like to be able to use the custom section as variables essentially. Here's one thing that I often want to do: tie a bundle directly to a version of an application. It's really hard to do without scripting out editing the dockerfile and porter.yaml so injecting like this where porter supports such a common case explicitly would make it much easier to package an app automatically when new versions are released.

```
$ porter build --custom version=0.2.0
```

```yaml
# porter.yaml
custom:
  version: "placeholder"

install:
  - helm3:
      chart: ./charts/myapp
      version: "{{ bundle.custom.version }}"
```

```Dockerfile
# Dockerfile

# Install the desired version of the chart
ARG CUSTOM_VERSION=TODO
RUN curl -O https://example.com/charts/$CUSTOM_VERSION/myapp.tgz
RUN tar -xzf myapp.tgz /cnab/app/charts/ && rm myapp.tgz
```

This change should be made on the release/v1 branch.

joshuabezaleel:
# What does this change
Enable user to use `--custom` flag for setting custom values when building a particular bundle using `porter build` command.

https://user-images.githubusercontent.com/7043511/153036317-3033fc6b-d4f4-46b1-9916-a114477fceb5.mp4

# What issue does it fix
#1789 

# Notes for the reviewer

# Checklist
- [x] Did you write tests?

# Reviewer Checklist


joshuabezaleel:
Currently this PR is only able to cater for `string` type value on the `porter.yaml` manifest. This is because since our templating language needs encapsulating double quote to instantiate the variable value for example `"{{ bundle.custom.customKey1 }}"` which the value will always be surrounded by double quote e.g. `"editedCustomValue1"` no matter whether we put double quote (to denote string) or not (to denote boolean or integer) when we pass the custom value flag on `porter build` command.

The type issue on the templating language is addressed separately on this [issue](https://github.com/getporter/porter/issues/316).

carolynvs:
I wholeheartedly approve of demonstrating functionality through whalesay 🐳

cmd/porter/bundle.go
```diff
@@ -81,6 +82,8 @@ Porter uses the docker driver as the default build driver, an alternate driver m
 		"Path to the build context directory where all bundle assets are located.")
 	f.StringVar(&opts.Driver, "driver", porter.BuildDriverDefault,
 		fmt.Sprintf("Experimental. Driver for building the invocation image. Allowed values are: %s", strings.Join(porter.BuildDriverAllowedValues, ", ")))
+	f.StringSliceVar(&opts.Customs, "custom", nil,
+		"Define an individual key-value pair for custom section in the form of NAME=VALUE. May be specified multiple times.")
```
carolynvs, line 86: Maybe we should include how to set a nested value?
```suggestion
		"Define an individual key-value pair for the custom section in the form of NAME=VALUE. Use dot notation to specify a nested custom field.  May be specified multiple times.")
```

pkg/build/buildkit/buildx.go
```diff
@@ -207,3 +219,28 @@ func (b *Builder) TagInvocationImage(ctx context.Context, origTag, newTag string
 	}
 	return nil
 }
+
+func convertMap(mapInput map[string]interface{}) (map[string]string, error) {
```
carolynvs, line 223: We can define this in a new package, maybe pkg/build/driverutil, and then call it something a bit closer to what it is really doing, like "FlattenMap". Then you can call it from both places with `driverutil.FlattenMap(...)`

pkg/build/docker/docker.go
```diff
@@ -30,6 +31,20 @@ func NewBuilder(cxt *portercontext.Context) *Builder {
 }
 
 func (b *Builder) BuildInvocationImage(ctx context.Context, manifest *manifest.Manifest) error {
+	buildArgs := make(map[string]*string)
```
carolynvs, line 34: Since we have this same logic in a couple places, how about we put this in a new package, pkg/build/driverutil, and then call it something like `driverutil.GetBuildArgs`

pkg/porter/generateManifest.go
```diff
@@ -41,5 +41,13 @@ func (p *Porter) generateInternalManifest(opts BuildOptions) error {
 		}
 	}
 
+	if opts.parsedCustoms != nil {
```
carolynvs, line 44: You can leave off this if check. The for loop won't do anything if the map is nil.

pkg/porter/build_integration_test.go
```diff
@@ -230,3 +230,24 @@ func TestBuildOptions_Defaults(t *testing.T) {
 		assert.Equal(t, config.BuildDriverDocker, opts.Driver)
 	})
 }
+
+func TestPorter_BuildWithCustomValues(t *testing.T) {
+	p := NewTestPorter(t)
+	defer p.Teardown()
+
+	p.TestConfig.TestContext.AddTestFile("./testdata/porter-with-custom-values.yaml", config.Name)
```
carolynvs, line 238: Instead of making a new test porter.yaml file for this test, how about we edit an existing testdata porter.yaml file and add in the custom section? I've learned the hard way that having lots of test porter.yaml files is a lot of work to maintain any time we change the format of the file.

pkg/build/docker/docker.go
joshuabezaleel: The docker one expects a `map[string]*string` for its image's buildArgs while buildkit expects a `map[string]string`. can we achieve this with different expected output type? Should we use the build driver as a parameter and let the function determine the type of the map value from it? Example as follows.

```
func GetBuildArgs(custom manifest.CustomDefinitions, buildDriver string) (interface{}, error) {
   switch buildDriver {
    case "docker":
           output := make(map[string[*string)
           // flatten customDefinitions and put it in the map
           return output, nil
    case "buildkit"
           output := make(map[string[string)
           // flatten customDefinitions and put it in the map
           return output, nil
    default:
         return nil, errors.New("build driver is not expected: %+v", buildDriver)
    }
}
```

carolynvs: What you have in the PR now is fine 👍

carolynvs (changes requested):
Everything looks great but we still need documentation for the website explaining your new feature so that people can use it.

Let's update the https://release-v1.porter.sh/author-bundles/#custom section and show people how to build a bundle and set custom metadata.

pkg/build/driverutil/driverutil.go
```diff
@@ -0,0 +1,33 @@
...
+				return nil, err
+			}
+			for innerKey, innerValue := range tmp {
+				out[key+"."+innerKey] = innerValue
+			}
+		case map[string]string:
+			for innerKey, innerValue := range v {
+				out[key+"."+innerKey] = innerValue
+			}
+		default:
```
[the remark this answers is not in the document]
joshuabezaleel: Ah, sorry that I totally missed this possibility of array type! Thank you for pointing this one out.

Previously [you](https://github.com/getporter/porter/issues/1789#issuecomment-945754819) mentioned we can use the example of `--param` flag on `install` command on retrieving the flag's values, which I used in this [one](https://github.com/getporter/porter/pull/1900/files#diff-80399e63a47b8bb575b1c832b91a42ad97100f0356a3635d70033185cfe436f1R81) using `ParseVariableAssignments` like how it was used on this [one](https://github.com/getporter/porter/blob/release/v1/pkg/porter/cnab.go#L265-L272) for the `install`'s `param`. But when I tried to supply the value using comma separated one e.g. `key="value1,value2"` or `key=value1,value2`, it yield the error of `invalid parameter (%s), must be in name=value format` from this [particular line](https://github.com/getporter/porter/blob/release/v1/pkg/parameters/parameters.go#L24-L27), like my screenshot below.
image

Would it be okay to add handling case of multiple comma separated values for this `ParseVariableAssignments` method as well (which will still be used by `install`'s `--param`)?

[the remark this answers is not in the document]
carolynvs: The way `StringSliceVar` works in cobra (which is how you defined the flag) is to allow the user to specify the parameter multiple times. Like so, "--param A=1 --param B=2", then cobra binds that to the slice like so `[]string{"A=1", "B=2"}`. So there's no need to do further parsing work to support it.

carolynvs:
Yup, --build-arg and --custom have overlap but I decided it was worth it because in one case someone isn't thinking about the porter bundle itself and is just trying to use docker features that they know (--build-arg) to parameterize their dockerfile. While as you said, custom is intended to help you use that value in your bundle as well.

I'm glad you asked about this because when both --custom and --build-arg have the same key, e.g. --custom MYKEY=1 --build-arg MYKEY=2, we should have the --custom flag take precedence (so that the build arg and the bundle.custom have the same value when used in porter.yaml

Yeah, per https://github.com/getporter/porter/issues/1789#issuecomment-1020119736 lets just do strings for now.

joshuabezaleel:
@carolynvs I've fixed the file conflicts and added the documentation on https://release-v1.porter.sh/author-bundles/#custom . I'll file the issue to track for other data type after this PR got successfully merged. Again, I hope you're not tired hearing me saying thank you since I definitely wouldn't be able to finish any without your tremendous help and enormous patience. Thank you, Carolyn!

cmd/porter/bundle.go
```diff
@@ -87,6 +88,9 @@ func buildBundleBuildCommand(p *porter.Porter) *cobra.Command {
 		"Secret file to expose to the build (format: id=mysecret,src=/local/secret). May be specified multiple times.")
 	f.BoolVar(&opts.NoCache, "no-cache", false,
 		"Do not use the Docker cache when building the bundle's invocation image.")
+	f.StringSliceVar(&opts.Customs, "custom", nil,
```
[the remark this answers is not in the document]
joshuabezaleel: Got it, I assume it was similar to this one https://github.com/getporter/porter/pull/1917#issuecomment-1105720433 referring to `--param` flag on install command that is addressed on #1931.

carolynvs (approved):
I'm looking forward to having this functionality in porter finally! 🎉

Issue #1869, opened 2022-01-24
How we push invocation images is confusing


carolynvs:
Right now when we publish a bundle, it creates two copies of the invocation image:

* getporter/porter-hello-installer:v0.1.1
* getporter/porter-hello@sha256:somedigest

The first one should not be used, only the second one embedded in the bundle repository is the correct image. The first is an awkward side effect that confuses people and causes extra security configuration on some registries like Amazon's because they have to allow for access to two different repositories.

Seems like the OCI distribution spec supports [deleting tags](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#deleting-tags). What if we pushed the invocation image directly to a tag in the bundle repository (such as getporter/porter-hello:installer-IMAGEID), retrieved the digest, and then after we publish the bundle, remove the tag.

That way we don't get dangling images or tags that people inevitably ask questions about.

@vdice What do you think?

vdice:
👍 I'm in favor of the simplification (unification into one repository, helping the AWS case and perhaps others) and the ability to clean up during publishing.

jasmdk:
Then what about when you publish a bundle from an archive, then you get even more very confusing repositories which are just temporary, but never removed. They may even conflict as different bundles risk publish to the same tag if they contain images with the same alias.
```
porter archive whalegap.tgz --reference getporter/whalegap:v0.1.1
porter publish --archive whalegap.tgz --reference <myregistry>/library/whalegap:v0.1.1
```
This produces 3 different repositories:
- library/whalegap
- library/whalegap-installer
- library/**whalesayd**

carolynvs:
@jasmdk Thanks for bringing that up as well. The plan would be to not "litter" extra images for the invocation image and referenced images as well. So in the whalegap example, porter would push whalegap-installer and whalesayd into the target bundle registry under a temporary tag as well.

carolynvs:
@vdice I'll look into how much trouble it would be to make this happen with changes in porter and cnab-to-oci. Right now I'm not sure if this is an ugly big effort or some naming changes and a cleanup after we publish.

VinozzZ:
After some investigation, I found that the tag deletion spec is still pretty new. Not many registry client has this new feature implemented. For now, I implemented the randomized tag for invocation image. Maybe we can wait to implement the deletion until registry client libraries to catch up with the spec

PR #1901, opened 2022-02-09
pkg: push invocation image under the same repo as the bundle

VinozzZ:
# What does this change
Currently, invocation image is published to a separate repository from
the bundle repo. It creates some confusion when users see it. This PR
changes the behavior to publish the invocation image under the same repo
as the bundle with a randomly generated tag based on bundle information.

# What issue does it fix
part of the fix for #1869 

# Notes for the reviewer

Due to the randomness of the registry port number in the smoke test, we can't predict what the hash of the invocation image tag would be in the test. Currently, I just removed the check for the tag from the test. Is there better way to handle this?

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist


carolynvs:
@VinozzZ Looks like this is almost ready? There is one failing smoke test but it looks like that is because the test used to assert the old behavior and just needs to be updated to match the changes made in this PR.

Let me know if you need any help getting this ready to review!

pkg/common/util.go
```diff
@@ -0,0 +1,22 @@
...
+	"encoding/binary"
+	"math/rand"
+)
+
+// RandomString generates a hash string with fixed length from "a-zA-Z0-9" using
+// a seed string and provided length.
+func RandomString(seed string, length int) string {
+	h := sha256.New()
+	h.Write([]byte(seed))
+	rand.Seed(int64(binary.BigEndian.Uint64(h.Sum(nil))))
```
carolynvs, line 14: The function name is a bit misleading since this isn't a random function, the output is deterministic based on the input string.

Is the goal to just hash the input string into a consistent length? Elsewhere in the code we use [md5 sums to make strings like that](https://github.com/getporter/porter/blob/784d61cf283bbd5b9c05a42ea718d3d042803305/pkg/cache/cached_bundle.go#L49-L54). I think that would be a bit more straightforward if what you are looking for is a hash. What do you think?

pkg/manifest/manifest.go
```diff
@@ -895,11 +896,11 @@ func (m *Manifest) SetInvocationImageAndReference(ref string) error {
...
 
-	imageName, err := cnab.ParseOCIReference(bundleRef.Repository() + "-installer")
-	if err != err {
-		return errors.Wrapf(err, "could not set invocation image to %q", bundleRef.Repository()+"-installer")
+	imageName, err := cnab.ParseOCIReference(bundleRef.Repository())
+	if err != nil {
+		return errors.Wrapf(err, "could not set invocation image to %q", bundleRef.Repository())
 	}
-	imageRef, err := imageName.WithTag(dockerTag)
+	imageRef, err := imageName.WithTag(common.RandomString(bundleRef.Named.String()+dockerTag, 10))
```
carolynvs, line 903: I suggest using `bundleRef.Named.WithTag(dockerTag)` instead of concatenating. Then the input string to your hash function will be a real docker reference, e.g. "example/mybundle:mytag" instead of "example/mybundlemytag".

Which seed string we use doesn't impact the solution much but when we are maintaining this code in the future, I think it would help to use that as a seed instead of something that _looks_ like a reference but isn't.

pkg/porter/publish_test.go
```diff
@@ -84,63 +84,63 @@ func TestPublish_validateTag(t *testing.T) {
 
 func TestPublish_getNewImageNameFromBundleReference(t *testing.T) {
 	t.Run("has registry and org", func(t *testing.T) {
-		newInvImgName, err := getNewImageNameFromBundleReference("localhost:5000/myorg/apache-installer", "example.com/neworg/apache:v0.1.0")
+		newInvImgName, err := getNewImageNameFromBundleReference("localhost:5000/myorg/apache-installer", "example.com/neworg/apache:v0.1.0", true)
```
carolynvs, line 87: I am wondering if we actually need the `isInvocationImage` flag for this function? Seems like since Porter is always passing in the invocation image name from the bundle, e.g. "example/mybundle-installer:v0.1.1", we can just use that value same as we use the image names of referenced images.

Want to try that out and see if it works?

pkg/porter/stamp.go
```diff
@@ -63,7 +68,13 @@ func (p *Porter) IsBundleUpToDate(opts bundleFileOptions) (bool, error) {
 
 		// Check whether invocation images exist in host registry.
 		for _, invocationImage := range bun.InvocationImages {
-			isImageCached, err := p.Registry.IsImageCached(invocationImage.Image)
+			// if the invovationImage is built before using a random string tag,
```
carolynvs, line 71: I'm having trouble following what this is trying to do. My best guess is that you are trying to account for bundles that were built from before this change?

If that's the case, I don't believe it's necessary since part of how we check if a bundle is out of date is make sure it was built with the same version of porter that is currently running. Since your change will be in a new version, that will automatically invalidate any old bundles and they will be rebuilt before publishing.

pkg/common/util.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"crypto/sha256"
+	"encoding/binary"
+	"math/rand"
+)
+
+// RandomString generates a hash string with fixed length from "a-zA-Z0-9" using
+// a seed string and provided length.
+func RandomString(seed string, length int) string {
```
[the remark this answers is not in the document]
VinozzZ: That's a great suggestion!

pkg/porter/stamp.go
VinozzZ: Yes, your best guess is right on point! I ran into issue while doing local testing and turns out it's due to old invocation image built before this change. Thanks for catching it and explain how the check works

pkg/porter/publish_test.go
carolynvs: I don't quite understand why it's different for the invocation image, when the referenced images also are going to be in the old location after archiving. I would have thought that they would both have the same problem then. Do you have an example of what the names look like?

VinozzZ: Yes, for example in this test, https://github.com/getporter/porter/blob/cb68d13594854ed8286c0a38591115ede2147bf0/pkg/porter/publish_test.go#L114. 
Do we want the new image to be placed under the bundle repo or keep it under the old location?
the difference would be `getporter.azurecr.io/newor/whalesayd:497b3e01ed1153e542a19df184607866` or `getporter.azurecr.io/neworg/whalegap:497b3e01ed1153e542a19df184607866`

carolynvs: With this feature all images (both referenced and invocation images) should be stored in the same bundle repository.

pkg/manifest/manifest.go
carolynvs: It looks like the hash is only including the repository portion of the image reference, "example/myrepo", instead of the full reference "example/myrepo:v1.0.0". Your PR originally was including the dockerTag variable in the hash, and I was just suggesting that you alter how you concatenate the repository with the tag so that it is "example/myrepo:v1.0.0" instead of (example/myrepov1.0.0). 

Is there a reason it was dropped or can we switch both invocation images and referenced images over to generating the hash using the full image reference?

pkg/porter/publish_test.go
VinozzZ: I see. I updated the code to reflect this

pkg/porter/publish.go
```diff
@@ -360,15 +360,13 @@ func getNewImageNameFromBundleReference(origImg, bundleTag string, isInvocationI
 		return image.EmptyName, errors.Wrapf(err, "unable to parse bundle tag %q into domain/path components", bundleTag)
 	}
 
-	// Use the image name with the bundle location to generate a randomized tag
-	imgName := path.Join(path.Dir(bundleName.Name()), path.Base(origName.Name()))
-	nameHash := md5.Sum([]byte(imgName))
-
+	// Use the original image name with the bundle location to generate a randomized tag
+	source := path.Join(path.Dir(bundleName.Name()), path.Base(origName.Name()))
```
carolynvs, line 364: Can we update this line as well so that the image includes the tag, just like you are doing above with the invocation image?

I just checked the code in cnab-to-oci and first it pushes images and then it pulls it down, _using the tag_, to determine its repository digest. I want to avoid a race condition where two people are pushing a bundle (maybe two different versions which is possible!) and then having them use the wrong repo digests because the temporary tag names we picked weren't unique.

VinozzZ: Do you mean something like this? 
```
source := path.Join(path.Dir(bundleName.Name()), path.Base(origName.Name())) + ":" +bundleName.Tag()

```

carolynvs: Yes, thanks. 👍

carolynvs:
I was testing out your change and realized there is a bug in v1 that I need to fix before we merge your PR. Otherwise I don't feel confident that I've verified your change.

https://github.com/getporter/porter/issues/1908

Everything looks good at this point but I'll hold off on merging until I've had a chance to see your PR work end to end in the cli (which this bug is blocking me from doing).

PR #1907, opened 2022-02-14
pkg/{config/context}: set log correlation id through env variable

### LINKED ISSUE #1873: Set correllation id from an environment variable
When the operator is instrumented, it will create a porter agent, and it would be great to use the same correlation id from the operator to porter. Porter currently generates a new correlation id for each command executed. Porter should look for an environment variable, `PORTER_CORRELATION_ID` and use that when present.

VinozzZ:

# What does this change

This PR allows setting logging correlation ID through `PORTER_CORRELATION_ID` env variable

# What issue does it fix
Closes #1873 

# Notes for the reviewer

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -33,6 +33,10 @@ const (
 	// EnvDEBUG is a custom porter parameter that signals that --debug flag has been passed through from the client to the runtime.
 	EnvDEBUG = "PORTER_DEBUG"
 
+	// EnvCORRELATION_ID is the name of the environment variable containing the
+	// id to correlate logs with a workflow.
+	EnvCorrelationID = "PORTER_CORRELATION_ID"
+
 	// CustomPorterKey is the key in the bundle.json custom section that contains the Porter stamp
 	// It holds all the metadata that Porter includes that is specific to Porter about the bundle.
 	CustomPorterKey = "sh.porter"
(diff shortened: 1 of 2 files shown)


carolynvs:
When you manually tested this (when the env var is set) and look at the logs, do you see any log entries with the original correlation id that is generated in Context.New? I'm concerned that we could end up with some log entries using the generated correlation id and then after the configure logging method is called, it switches to the new correlation id.

I was hoping we could put the logic to either use the env var or generate a new one in a single spot so that the problem outlined above isn't possible.

VinozzZ:
> When you manually tested this (when the env var is set) and look at the logs, do you see any log entries with the original correlation id that is generated in Context.New? I'm concerned that we could end up with some log entries using the generated correlation id and then after the configure logging method is called, it switches to the new correlation id.
> 
> I was hoping we could put the logic to either use the env var or generate a new one in a single spot so that the problem outlined above isn't possible.

My local test doesn't have log entries with the original correlation id showing up. My understanding is that the logging configuration method is always called at the beginning of any porter command. Is that not true?

VinozzZ:
> I wasn't 100% sure if there weren't some code paths where we output logs before we call ConfigureLogging, as we do load configuration files and things before we have all the information needed to finish configuring logging/tracing.
> 
> I'd check the log file for that command in ~/.porter/logs and just double check that we only see a single correlation id, and that the file is named after the correlation id.

After some local testing, I believe we have come to the conclusion that we only see a single correlation id and the file is named after that single id

carolynvs (approved):
🚢

PR #1916, opened 2022-02-18
pkg/context: add function to construct user agent string for porter

VinozzZ:

# What does this change

This PR create a way to retrieve a Porter user agent string in the format of `porter/<current-build-version>` from portercontext.
This will allow any mixins or plugins to add this information into it's user agent. 

# What issue does it fix
This is related to this [issue]( https://github.com/getporter/az-mixin/issues/38)

# Notes for the reviewer
The usage of this change is in this PR https://github.com/getporter/az-mixin/pull/39

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/context/context.go
+++ b/pkg/context/context.go
@@ -11,7 +11,6 @@ import (
 	"strings"
 	"time"
 
-	"get.porter.sh/porter/pkg"
 	"get.porter.sh/porter/pkg/tracing"
 	"github.com/carolynvs/aferox"
 	cnabclaims "github.com/cnabio/cnab-go/claim"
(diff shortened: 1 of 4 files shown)


pkg/context/context.go
```diff
@@ -444,3 +445,19 @@ func (c *Context) SetSensitiveValues(vals []string) {
 		c.Err = err
 	}
 }
+
+// UserAgent returns a string that can be used as a user agent for porter.
+func (c *Context) UserAgent() string {
```
carolynvs, line 450: What do you think about making this a function in pkg/version.go? Nothing about this relies on the context, so this moves the function closer to the data it depends upon.

pkg/context/context.go
```diff
@@ -444,3 +445,19 @@ func (c *Context) SetSensitiveValues(vals []string) {
...
+	product := "porter"
+
+	if pkg.Commit == "" && pkg.Version == "" {
+		return product
+	}
+
+	v := pkg.Version
+	if len(v) == 0 {
```
carolynvs, line 458: You don't have to change this logic here but I wanted you to know that version unless you run a build using `go build` manually, all the build scripts always set both a version and a commit. The version will be the output of running `git describe` so that even untagged commits can express a version value.

carolynvs (approved):
LGTM

PR #1917, opened 2022-02-19
Enabling JSON object as parameter

### LINKED ISSUE #1508: Direct way of consuming parameters/credentials as JSON objects?
Is there a direct way of consuming parameters or credentials in `porter.yaml` as JSON objects (i.e. use their sub-properties as inputs to commands)?  Currently one might feed a parameter/credential into an `exec` command, which calls a `bash` script, which simply `echo`s the value, which can then be teased into individual outputs using `jsonPath`.  Is there a better way?

joshuabezaleel:
# What does this change
Consuming JSON object from a file as parameter with the type of "object".

# What issue does it fix
Closes #1508 

# Notes for the reviewer

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/runtime/runtime-manifest.go
+++ b/pkg/runtime/runtime-manifest.go
@@ -4,8 +4,10 @@ import (
 	"archive/tar"
 	"compress/gzip"
 	"encoding/base64"
+	"encoding/json"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"reflect"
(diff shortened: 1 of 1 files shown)


joshuabezaleel:
This PR is a work in progress thus marked as a draft. Pushed it as a PR so reviewers can take a look at the WIP code.

https://user-images.githubusercontent.com/7043511/154788806-9e15f740-e190-483d-b34c-a32bd5fedb16.mp4

Hi @carolynvs , I am sorry in advance for pushing a not-finished PR since I guess there are some things that I missed on how things work here.

I am not sure why does the contents of the file from the root folder (which should be `{"KEY": "VALUE"}`) got replaced by the default value (`{"color": "red"}`) inside the `/cnab/app` directory when it was referenced as one of the parameter? 

In the video, I tried to print all the contents of the file inside the `/cnab/app` directory. 
From the first example we can see that if we define it as different file name, `other-config.json` in the root directory and `config.json` on the parameter, the `other-config.json` still has the value of ` `{"KEY": "VALUE"}`` and `config.json` has the value of `{"color": "red"}`, which was expected since we didn't define the content of the `config.json` in advance. 
But when we define the `config.json` in the root directory and also referenced it on the parameters, its content get replaced by the default value.
And when I didn't define the default value on the third example, it yields an error saying that the parameter is required. 

Thank you lots in advance! 😄

carolynvs:
Here's how to set a parameter value with quotes in it (wrap the entire flag value in the single quotes)

```
porter install --param 'name={"foo":"bar"}'
```

That still won't work though because there is a bug in how we are defining the --param flag, which is being fixed in [#space](https://github.com/getporter/porter/pull/1931). To work around it change the following line to use StringArrayVar

https://github.com/getporter/porter/blob/9d4c06772ac6115da7a385450c11435adc07d6b8/cmd/porter/bundle.go#L168

carolynvs:
Yeah don't worry about it. Feel free to close this PR.

PR #1918, opened 2022-02-19
enable steps to define environment variables

devigned:
# What does this change
This change add the ability for steps to add environment variables to be added to the env when executing the command for a step.

This change is related to https://github.com/getporter/az-mixin/issues/38. With this change, the mixin could add the user-agent env var to the command.

# Notes for the reviewer
This is my first contribution here, so please let me know if I missed / misunderstood something.

# Checklist
- [x] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/exec/builder/execute_test.go
+++ b/pkg/exec/builder/execute_test.go
@@ -4,13 +4,16 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
+	"strings"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"get.porter.sh/porter/pkg"
 	"get.porter.sh/porter/pkg/context"
 	"get.porter.sh/porter/pkg/test"
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
 )
 
 type TestAction struct {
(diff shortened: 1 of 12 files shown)


pkg/exec/builder/execute_test.go
```diff
@@ -156,3 +157,24 @@ func TestExecuteStep_SpecifiesCustomWorkingDirectory(t *testing.T) {
...
+			Command:         "env",
+			EnvironmentVars: map[string]string{"SOME_VAR_123": "foo"},
+		},
+	}
+
+	c.Setenv(test.ExpectedCommandEnv, "env")
+
+	_, err := ExecuteStep(c.Context, step)
+	require.NoError(t, err, "Execute Step failed")
+	assert.Contains(t, fmt.Sprintln("SOME_VAR_123=foo"), c.GetOutput())
```
carolynvs, line 179: This assert isn't checking what you think here. It's saying assert "SOME_VAR_123=foo" contains "", which will always pass. After changing this line to the following:

```suggestion
	assert.Contains(t, c.GetOutput(), fmt.Sprintln("SOME_VAR_123=foo"))
```

the test fails because context.NewCommand is using a mock for unit tests and isn't actually calling env.

Here's how to get your env command to run:

```
// Skip on windows because env is mac/linux only
if runtime.GOOS == "windows" {
  t.Skip()
}

c := context.NewTestContext(t)
c.UseFilesystem() // Don't use a mocked file system or NewCommand function, actually run commands against the host
```

carolynvs (changes requested):
This looks great! Just a couple things need to be addressed first:

* main is for our stable release, v0.38, and we are only merging high priority patches into that branch. Pull requests with new features should target release/v1.
* This adds support for allowing mixins to set environment variables but at this point a user couldn't actually try out the feature yet. What we normally do is have the exec mixin implement new features added to the builder package, demonstrating how it works and vetting the implementation.
* We also need documentation showing how it works, this is usually what the mixin authors will copy when they look at the exec mixin and incorporate the changes necessary to support a new feature (like setting env vars). Here's the [exec mixin page](https://release-v1.porter.sh/mixins/exec/) that should be updated to include how to set an env var in its supported syntax section.

carolynvs (approved):
Thanks for the fix!

pkg/exec/schema/exec.json
```diff
@@ -72,6 +72,13 @@
             "type": "string"
           }
         },
+        "envs": {
```
carolynvs, line 75: A million brownie points for remembering to update the schema that VS Code uses to do autocomplete 💯

carolynvs (approved):

carolynvs:
You can fix the broken unit test by running `mage UpdateTestFiles`

devigned:
> You can fix the broken unit test by running `mage UpdateTestFiles`

Yep. Thank you for having the prompt in test failure output. That was super helpful.

Issue #1000, opened 2020-04-17
Param parsing when commas are involved


vdice:
## Describe the bug
String parameters including `,` characters seem to break the parser.

## To Reproduce
1. Run `porter install --param msg="testing 1, 2, 3"` (on any bundle... `msg` doesn't need to even be a valid param; it will fail before validation)
2. See error:
```
Error: invalid parameter ( 2), must be in name=value format`
````

Other special characters (`.`, `!`, etc.)  in the string parse fine:

```
 $ porter install --param msg="testing 1! 2! 3!"
installing whalesay...
...
 $ porter instance show -o json | jq -r '.parameters'
{
  "msg": "testing 1! 2! 3!",
  "porter-debug": false
}
```

## Expected behavior
A supplied string parameter containing `,` characters should parse fine.
## Porter Command and Output
```
 $ porter install --param msg="testing 1, 2, 3"
Error: invalid parameter ( 2), must be in name=value format
```

## Version
Copy the output of `porter version` below

```
 $ porter version
porter v0.26.0-beta.1-19-g7c6eb425 (7c6eb425)
```

ei09010:
Hi, I wouldn't mind fixing this bug if no one else is doing it.

carolynvs:
@ei09010 That would be great, thanks! The PR should target the main branch.

carolynvs:
I think this is a bug in cobra. Let's confirm and submit a patch and use our fork until it's merged.

https://github.com/spf13/cobra/issues/661#issuecomment-902406803

sleepypioneer:
We have inadvertently added the fix for this in this PR https://github.com/getporter/porter/pull/1931 (FYI @Simpcyclassy) I hope this has not caused any inconvenience to @ei09010

carolynvs:
No that's fine! 👍

PR #1931, opened 2022-02-23
test for parameters with space

### LINKED ISSUE #1862: Parameter with space isn't treated properly
## Describe the bug
I have a parameter "name" and gave it a value of "demo time". When the bundle printed out the name it only printed demo. Somewhere in Porter seems like we are not quite handling parameters with spaces correctly?

## To Reproduce

1. Install the getporter/hello-llama:v0.1.1 bundle with the name parameter set to a value with a space in it, e.g. demo time

note that it only prints "Hello, demo"

## Porter Command and Output
This is output from the operator
```
$porter install -r getporter/hello-llama:v0.1.1 --param name="demo time"
executing install action from hello-llama (installation: farts)
Hello, demo
```

## Version
v1.0.0-alpha.7

Simpcyclassy:
# What does this change
This change will allow for parameters containing a space to be parsed in their entirety.

# What issue does it fix
Closes #1862 

# Notes for the reviewer
It was observed that the wrong method was being used from the `cobra` library, which resulted in parameters with space being split. We switched to using the array method and we added a test to confirm this function acts as expected.

# Checklist
- [x] Did you write tests?

# Reviewer Checklist


workshop/hello-llama/porter.yaml
```diff
@@ -17,18 +17,18 @@ install:
       command: ./helpers.sh
       arguments:
         - install
-        - "{{ bundle.parameters.name }}"
+        - "'{{ bundle.parameters.name }}'"
```
Simpcyclassy, line 20: @carolynvs 
We investigated this further and concluded it has to do with how the string is escaped in YAML. This probably requires an update to documentation but additionally requires code changes to existing bundles.

For the scope of this issue, would a change of documentation and a code change to the llama bundle be sufficient or do you see an alternative route?
Our proposal is to extend this [documentation](https://porter.sh/quickstart/parameters/#create-a-parameter-set) to include a note regarding how to handle parameters that include a space. We are unsure about the functionality behind the `porter parameters generate` command and whether an update here is required.
Lastly, what we see as being out of scope is updating all the other places where YAML variables could include space and therefore could face a similar problem.

Would you agree?

carolynvs: Hmm, I wasn't expecting the wrapping quotes to matter here, since we are parsing this and converting it into a an array of arguments in Go, e.g. ["./helpers.sh", "install", "my parameter"]. I would have expected calling a bash script in Go and passing an argument with a space to work, without the quotes. If you can look into that more, that would be great, otherwise I'll try to when I have time.

Simpcyclassy: From our investigation, we see that the strings get passed separately in the bash script, for example when we passed $1, $2, and so on in the `/helpers.sh` that was as many words as we could get. We also tried $@ which we did not think was very optimal because this would catch all arguments even when not unique to the command. It looks like the values get read in the YAML file in the same manner.
I have had a couple of other engagements of recent and will investigate this further. In the meantime, if you get a chance to look at this and give me more pointers, that will be great. We came across https://github.com/getporter/porter/blob/release/v1/pkg/exec/builder/execute.go#L102, especially the `splitCommand(args)` line, we will look at this next.

Simpcyclassy: We looked further into the `[splitCommand(args)](https://github.com/getporter/porter/blob/release/v1/pkg/exec/builder/execute.go#L128)` function which was added in this [commit](https://github.com/getporter/porter/pull/931). This does look like a possible culprit but not something we will like to change. We are potentially missing some context, is there a chance that you @carolynvs could give us some context? Thanks in advance!

sleepypioneer: Adding a note here, that I spotted that the escaping in the `porter.yaml` was being done in these [example bundles](https://github.com/getporter/example-bundles/blob/main/azure-keyvault/porter.yaml#L20) for flags, still not sure if they are handled differently somehow

sleepypioneer: I can confirm that this `splitCommand(args)` function is responsible for dividing the argument (I commented it out and rebuilt and now see the full name as expected) Perhaps there is somewhere earlier in the code where we can enforce this string is not escaped and therefore split when this function runs.

carolynvs: Thanks for narrowing down where the bug is!

sleepypioneer: Awesome congrats @Simpcyclassy :tada: on your awesome work here and be along the way in solving your first open-source bug :bug: :raised_hands: In our next session we can look at what propositions we can make for a solution.

carolynvs: I just ran into this problem with a different flag in Porter. It sounds like we are using cobra incorrectly and that's why it's splitting on the comma

https://github.com/spf13/cobra/issues/661#issuecomment-377684634

carolynvs: Let me know if you need some help figuring out what needs to be changed in the cmd/porter package files. I believe switching any use of StringSliceVar to StringArrayVar fix it.

cmd/porter/bundle.go
```diff
@@ -163,9 +163,9 @@ For example, the 'debug' driver may be specified, which simply logs the info giv
 		"Path to the porter manifest file. Defaults to the bundle in the current directory.")
 	f.StringVar(&opts.CNABFile, "cnab-file", "",
 		"Path to the CNAB bundle.json file.")
-	f.StringSliceVarP(&opts.ParameterSets, "parameter-set", "p", nil,
+		f.StringArrayVarP(&opts.ParameterSets, "parameter-set", "p", nil,
 		"Name of a parameter set for the bundle. It should be a named set of parameters and may be specified multiple times.")
-	f.StringSliceVar(&opts.Params, "param", nil,
+	f.StringArrayVar(&opts.Params, "param", nil,
```
sleepypioneer, line 168: :+1:  great, and you already saw we have different commands that use this, so we need this for them also, for example invoke https://github.com/getporter/porter/pull/1931/files#diff-9fc541206435fd5b66cd057759dc4fd2ccf36aed05594a186148f90cb57caf75R282

sleepypioneer:
@Simpcyclassy our test is failing here with the same error we saw on my computer:

```
Invalid schema version
```

You can see it here: https://dev.azure.com/getporter/porter/_build/results?buildId=3568&view=logs&j=c892508d-390a-5765-0ad0-b341c7fd0dd4&t=99b6391b-efca-5f28-e88f-b44afecfc894&l=10900

cmd/porter/bundle.go
```diff
@@ -163,9 +163,9 @@ For example, the 'debug' driver may be specified, which simply logs the info giv
 		"Path to the porter manifest file. Defaults to the bundle in the current directory.")
 	f.StringVar(&opts.CNABFile, "cnab-file", "",
 		"Path to the CNAB bundle.json file.")
-	f.StringSliceVarP(&opts.ParameterSets, "parameter-set", "p", nil,
+		f.StringArrayVarP(&opts.ParameterSets, "parameter-set", "p", nil,
```
sleepypioneer, line 166: Watch the indentation here, you need to remove the tab in the front.

sleepypioneer:
> * Did you change porter.yaml or a storage document record? Update the corresponding schema file.

So I got curious after reading the checklist in the PR `Did you change porter.yaml or a storage document record? Update the corresponding schema file.` 

I therefore checked the difference between our tests [porter.yaml](https://github.com/getporter/porter/blob/5ca9a754f10414f914b74cf5e8f8575fbc339420/tests/integration/testdata/bundles/bundle-with-string-params/porter.yaml) and another tests [porter.yaml](https://github.com/getporter/porter/blob/5ca9a754f10414f914b74cf5e8f8575fbc339420/tests/integration/testdata/bundles/bundle-with-file-params/porter.yaml#L1) If you add the `schemaVersion` this will pass the tests :)  Seems this was added after we first started working on this #1946

workshop/hello-llama/porter.yaml
sleepypioneer: @carolynvs we went with the approach of switching our StringSliceVar to StringArrayVar but it appears to not have made the change we imagined. We will keep working on this but we need to take a break this next week as @Simpcyclassy is traveling. We will for sure pick it up when she is back. We feel so close!

carolynvs: @sleepypioneer @Simpcyclassy I just realized that I completely messed up and linked to the StringArray cobra fix on the wrong pull request! I'm so sorry! 

Both (parsing the comma and handling whitespace) were things that needed to be fixed, but I messed up and commented on the wrong pull request. I really didn't mean to derail you.

It sounds like you had a fix working for the space problem too? I think you were on the right track with `splitCommand`. You really are close, I just accidentally threw at you a second (unreleated) bug on this pull request.

You can either fix both on this PR, or just skip the one failing test (tests/integration/install_test.go) and we can merge what you have here (which does fix the comma bug), then fix the parameter space bug in a second pull request. Whatever works easier for you.

tests/integration/install_test.go
```diff
@@ -111,3 +111,26 @@ func TestInstall_withDockerignore(t *testing.T) {
...
+
+	p := porter.NewTestPorter(t)
+	defer p.Teardown()
+	p.SetupIntegrationTest()
+	p.Debug = false
+
+	p.AddTestBundleDir("testdata/bundles/bundle-with-string-params", false)
+
+	installOpts := porter.NewInstallOptions()
+	installOpts.Params = []string{"name=Demo Time"}
```
sleepypioneer, line 126:
```suggestion
	installOpts.Params = []string{"name=\"Demo Time\""}
```

sleepypioneer: @Simpcyclassy try this, the problem I believe is that our test doesn't have quotes around the parameter. Rereading the issue this is expected https://github.com/getporter/porter/issues/1862

As this should be the given command: `$porter install -r getporter/hello-llama:v0.1.1 --param name="demo time"`

At least for me locally this passes the tests :tada: and from reading this again as I understand this is the intended behavior. Using single and double quotes catches both this behavior but also when no quotes were passed so that is incorrect as that would catch potentially other flags passed to the command.

sleepypioneer: Hmm on further inspection perhaps this is not a good solution. I am running the code locally and it is not passing through "demo time" only "demo" :(

workshop/hello-llama/porter.yaml
sleepypioneer: Oh good to know, I didn't catch that there was a second bug for commas. We can look again at the `splitCommand` function, we stopped looking there, after seeing about the cobra method misuse :see_no_evil: Thanks for your patience, we are super excited to be contributing. We only meet once a week so it takes us a little time but as you said we are really close! Thanks for writing and letting us know!!

tests/integration/install_test.go
sleepypioneer: :warning:  ignore this

workshop/hello-llama/porter.yaml
sleepypioneer: @carolynvs this is the issue for the comma https://github.com/getporter/porter/issues/1000 I will write there that this PR will add the fix.

I am not too sure how we can best change the behavior, from reading the code for `splitCommand`, it is working as intended. We either need to wrap our param in extra quotes or we have to skip this processing for parameters but I am concerned that is undoing the reason this was implemented in the first place?

sleepypioneer: Adding a little extra context.

If we leave the `splitCommand(args)` in this is the output we see for first a param with quotes around it and secondly a param without quotes:

```
greenmachine@pop-os:/tmp/hello-world$ porter install --param name="demo time" --force
executing install action from porter-hello (installation: /porter-hello)
Install Hello World
Hello demo

greenmachine@pop-os:/tmp/hello-world$ porter install --param name=demo time --force
executing install action from porter-hello (installation: /time)
Install Hello World
Hello demo
```

Now the same with the call to `splitCommand` removed:

```
greenmachine@pop-os:/tmp/hello-world$ porter install --param name="demo time" --force
executing install action from porter-hello (installation: /porter-hello)
Install Hello World
Hello demo time

greenmachine@pop-os:/tmp/hello-world$ porter install --param name=demo time --force
executing install action from porter-hello (installation: /time)
Install Hello World
Hello demo
```
We see that removing the `splitCommand` call makes things behave as expected, but I am still unsure if there is an edge case we miss that makes it unwise to remove this. We could move the call up and run it only for flags (I haven't dug into how flags use this or are expected to behave) It seems at least to me that what this command was trying to solve is no longer an issue for params.

tests/integration/install_test.go
```diff
@@ -111,3 +111,26 @@ func TestInstall_withDockerignore(t *testing.T) {
...
+
+	err := installOpts.Validate(context.Background(), []string{}, p.Porter)
+	require.NoError(t, err)
+
+	err = p.InstallBundle(context.Background(), installOpts)
+	require.NoError(t, err)
+
+	output := p.TestConfig.TestContext.GetOutput()
+	require.Contains(t, output, "Hello, Demo Time", "expected action output to contain provided file contents")
+}
```
sleepypioneer, line 136: @Simpcyclassy We should add a test after this one for checking the second bug :lady_beetle: ie `TestInstall_stringParamWithCommas` we can essentially use the same bundle and copy most of the test but change the param name to `testing 1, 2, 3` instead of `demo time`.

workshop/hello-llama/porter.yaml
sleepypioneer: I am also thinking that in this case arguments != parameters, so we may want to only have an effect on parameters that are passed through as other arguments may still require this behavior from `splitCommand` looking to see where we could do that.

carolynvs (approved):
Thank you for the fix! Someone needed the fix on our stable (main) branch, so I'm helping to get it merged today and backported to v0.38).

I have added a commit to your PR so that I can merge what we have working now (The fix for StringSliceVar).

* Ran `mage build` so that the generated cli documenation was updated with your fix
* Skipped the failing test case for parameters with spaces. We can uncomment it when fixed.

sleepypioneer:
Congrats @Simpcyclassy on your first contribution 🎉 👏👏👏

8970ed7648df814c-animated-confetti-gif-www-imgkid-com-the-image-kid-has-it

Thank you @carolynvs for helping get this merged :sparkling_heart: we had to take a small break due to personal commitments but we are back now and starting again to fix the parameter with a space issue :smile:

PR #1946, opened 2022-02-27
Add schemaVersion to porter.yaml

### LINKED ISSUE #1566: Include compatible porter version in porter.yaml
The porter.yaml file should include a field that indicates which versions of porter it is compatible with. We can have `porterVersion: PORTER_SEMVER_RANGE`, for example `schemaVersion: 1.x` would say that it is written for porter versions 1-2. `schemaVersion: 1.5.x` would translate to 1.5-2.

For versions above 1.0 this works well because we only make breaking schema changes on major versions.

For versions < 1.0 (i.e. today) it's a bit dicier because we will break things before v1. In those cases you can specify a range, e.g .`porterVersion: 0.37-0.45.0`

If your version of porter doesn't match the porter.yaml, porter should tell you to use a version of porter within the supported range.

If no version is specified, it should warn you. After our 1.0 changes, it should be a hard failure.

carolynvs:
# What does this change
This adds a document schema version to the Porter manifest (porter.yaml) similar to what we already have on credential and parameter sets. The schema of the any document should not change within a major version.

I've added a page that lists which schema versions are used by various porter versions.

Preview of doc changes at https://deploy-preview-1946--porter.netlify.app/reference/file-formats/

# What issue does it fix
Closes #1566 

# Notes for the reviewer
I've created #1956 as a follow-up so that we start using the schemaVersion set.

# Checklist
- [x] Did you write tests?

# Reviewer Checklist


pkg/templates/templates.go
```diff
@@ -68,7 +67,7 @@ func (t *Templates) GetRunScript() ([]byte, error) {
 }
 
 // GetSchema returns the template manifest schema for the porter manifest.
-// Note that is is incomplete and does not include the mixins' schemas.ß
+// Note that it is incomplete and does not include the mixins' schemas.ß
```
vdice, line 70:
```suggestion
// Note that it is incomplete and does not include the mixins' schemas.
```

docs/content/reference/file-formats.md
```diff
@@ -1,20 +1,238 @@
...
+      command: ./status.sh
+```
+
+| Field                            | Required | Description                                                                                                                                    |
+|----------------------------------|----------|------------------------------------------------------------------------------------------------------------------------------------------------|
+| schemaType                       | false    | The type of document. This isn't used by Porter but is included when Porter outputs the file, so that editors can determine the resource type. |
+| schemaVersion                    | true     | The version of the Bundle schema used in this file.                                                                                            |
+| name                             | true     | The name of the bundle.                                                                                                                        |
+| version                          | false    | The version of the bundle, must adher to [semver v2].                                                                                          |
```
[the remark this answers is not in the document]
vdice: Should we mention what value will be used for the bundle's tag if version isn't provided?  Just tested and it appears to be a SHA of some sort (which isn't in itself semver, right?)

carolynvs:
> One thing that I'm not sure is when a dev makes a schema change, do we have a way to let them know that they also need to increase the supported schema version?

That's a great idea. I'll add someone about breaking changes to CONTRIBUTING.MD, as I think it's a general topic. Then also add some comments inline so people know "If you change this data structure, here's what you have to do next..."

pkg/cnab/config-adapter/stamp_test.go
```diff
@@ -27,7 +25,7 @@ func TestConfig_GenerateStamp(t *testing.T) {
 	a := NewManifestConverter(c.Context, m, nil, nil)
 	stamp, err := a.GenerateStamp()
 	require.NoError(t, err, "DigestManifest failed")
-	assert.Equal(t, simpleManifestDigest, stamp.ManifestDigest)
+	assert.NotEmpty(t, stamp.ManifestDigest)
```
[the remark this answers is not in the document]
carolynvs: Every time we make even the slightest change to our test data or porter, the digest changes. There isn't really any value in asserting that the manifest equals this exact constant. We just care that it is set and the extra effort caused by this test failing and having to update the constant isn't justified.

docs/content/reference/file-formats.md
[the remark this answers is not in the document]
carolynvs: I fixed the docs to indicate that version is required, though people can set it programmatically during build thanks to the feature you added a while back. So it's not possible to avoid setting version.

The bundle still uses a tag defaulted using the version unless you set it with the reference field, or --tag when publishing.

What you are probably noticing is Yingrong's recent fix for how we first push the invocation image (which used to be named BUNDLENAME-installer), retrieved its repository digest and then patched the bundle.json before publishing the bundle. We now push the invocation image to a tag (a hash) in the same repository as the bundle. We still do the same trick to update the repository digest in the bundle before publishing. The difference is that we don't create additional repositories that are unused (i.e. the -installer image isn't actually used by the bundle, we use the one in BUNDLENAME@sha256:abc123). This makes it more clear to people that we are generating stuff that they shouldn't try to reference by name and that the image is distributed with the bundle.

Untagging so that we don't leave stuff around to confuse people isn't supported yet but is on the backlog.

carolynvs:
@vdice Updated with everyone's review feedback. Please take another look when you can!

CONTRIBUTING.md
```diff
@@ -574,6 +575,20 @@ we send regular command output to `Out` and debug information to `Err`. It
 allows us to then run the command and see the debug output separately, like so
 `porter schema --debug 2> err.log`.
 
+## Breaking Changes
+
+Some changes in Porter break our compatibility with previous versions of Porter.
+When that happens, we need to release that change a new major version number to indicate to users that it contains breaking changes.
```
vdice, line 581:
```suggestion
When that happens, we need to release that change with a new major version number to indicate to users that it contains breaking changes.
```

docs/content/reference/file-formats.md
```diff
@@ -158,70 +158,70 @@ status:
...
+| schemaVersion                    | true     | The version of the Bundle schema used in this file.                                                                                                                                    |
+| name                             | true     | The name of the bundle.                                                                                                                                                                |
+| version                          | true     | The version of the bundle, must adhere to [semver v2].<br/>The bundle tag defaults to the version with a v prefix, e.g. mybundle:v1.0.0. Use --tag or the reference field to override. |
+| description                      | false    | A description of the bundle.                                                                                                                                                           |
+| registry                         | true*    | The OCI registry to use when the bundle is published.<br/> /*Either registry or reference must be specified.                                                                           |
```
vdice, line 168: Wasn't sure if the `/*` prefix was intended (instead of `*`), but thought I'd mention/show it:

Screen Shot 2022-03-11 at 10 29 49 AM

vdice (approved):
LGTM!

docs/content/reference/file-formats.md
carolynvs: Oops, wrong direction for the slash. I was trying to escape the * in the markdown so that it would show up as an * and not a bullet point. Fixing!

PR #1950, opened 2022-02-28
mage: add checksum for published assets

### LINKED ISSUE #1797: Publish checksums
We need to publish checksums along with our release binaries.

One thing to look into is if our build would work with goreleaser yet. When I last tried to move over to goreleaser I ran into like 100 things that we do that doesn't work with it. From the go build, to things like our latest and canary tags. It could be worth cataloging everything that we do that isn't standard and figure out how we could use goreleaser, and still make things like installing the latest release, or the canary easier.

VinozzZ:

# What does this change
Publish checksums for all assets so users can validate whether the downloaded file has been tempered

# What issue does it fix
Part of the fix for #1797 

# Notes for the reviewer

Since only using checksum doesn't prevent the case when checksum file itself has been tempered, another PR will address that issue by using signing using a private key 

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/mage/releases/publish.go
+++ b/mage/releases/publish.go
@@ -2,6 +2,7 @@ package releases
 
 import (
 	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -254,13 +255,13 @@ func GenerateChecksum(data io.Reader, path string) (string, error) {
 	}
 	sum := hash.Sum(nil)
 
-	return AppendDataPath(sum, path)
+	return AppendDataPath(sum, path), nil
 }
 
-func AppendDataPath(data []byte, path string) (string, error) {
+func AppendDataPath(data []byte, path string) string {
 	// write the checksum and file name to the checksum file so it can be
 	// verified by tools like `shasum`
-	return fmt.Sprintf("%x  %s", data[:], filepath.Base(path)), nil
+	return hex.EncodeToString(data) + "  " + filepath.Base(path)
 }
 
 func createChecksumFile(contentPath string, checksumFile string) error {
--- a/mage/releases/publish.go
+++ b/mage/releases/publish.go
@@ -192,11 +192,17 @@ func AddFilesToRelease(repo string, tag string, dir string) {
 
 	checksumFiles := make([]string, len(files))
 	for i, file := range files {
-		cf, err := generateChecksumFile(file)
-		if err != nil {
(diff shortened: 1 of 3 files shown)


mage/releases/publish.go
```diff
@@ -217,3 +231,35 @@ func listFiles(dir string) []string {
 
 	return names
 }
+
+func generateChecksumFile(path string) (string, error) {
+	if filepath.Ext(path) == "sha256sum" {
```
carolynvs, line 236: Wouldn't the extension be ".sha256sum"? 🤔

mage/releases/publish.go
```diff
@@ -217,3 +231,35 @@ func listFiles(dir string) []string {
...
+	if _, err := io.Copy(hash, data); err != nil {
+		return "", err
+	}
+	sum := hash.Sum(nil)
+
+	// write the checksum and file name to the checksum file so it can be
+	// verified by tools like `shasum`
+	_, err = f.WriteString(fmt.Sprintf("%x  %s", sum[:], filepath.Base(path)))
+	if err != nil {
+		return "", err
```
carolynvs, line 261: Even though this is just our build script, it's still super helpful to wrap the errors from low level calls like what's in this function with `errors.Wrap` and provide context into what we were trying to do when it failed. It helps avoid a build failing with "EOF" or "file not found" without any clue what we were trying to do. 😁

mage/releases/publish.go
```diff
@@ -217,3 +231,35 @@ func listFiles(dir string) []string {
 
 	return names
 }
+
+func generateChecksumFile(path string) (string, error) {
```
carolynvs, line 235: Let's add some unit tests for this function so that we can verify that it works. Getting build scripts related to publishing can be tricky and it's no fun to cut a release only to realize that there's a bug in the publish logic. Ask me how I know. 😉

carolynvs (changes requested):
This looks right all my comments are just to help us bring our build scripts up to production code level which is a bit of technical debt that we have been trying to address over time.

mage/releases/publish.go
VinozzZ: You are absolutely right!! Thanks for catching it

mage/releases/publish.go
VinozzZ: I broke down the function into more modular pieces. Let me know if that helps

carolynvs (approved):
Looks good! I'm going to run the test release pipeline on this to see how it goes

Issue #1956, opened 2022-03-02
Commands that read the manifest should check the schema version 


carolynvs:
Now that we will soon have a schemaVersion set on manifests (#1946), that lets us add checks to any command that reads a porter manifest to verify that it knows how to read the file, e.g. porter build, porter explain, etc.

- [x] porter explain should work exactly the same in either case
- [x] porter inspect should work exactly the same in either case
- [x] porter lint should print a warning (so build gets it for free)
- [x] porter publish should print a warning (publish calls build if needed, so again we get it for free)
- [x] porter installation commands should go off of the cnab schema version, i.e. do we support that cnab spec? So this won't break until we implement advanced dependencies. We shouldn't enforce the manifest schemaVersion matching.

All warnings should be put behind a feature flag so we can quickly toggle if it's a hard error or a warning. I think it will be useful to have that always.

PR #1974, opened 2022-03-15
Do not assume that all bundles have a supported porter.yaml file

carolynvs:
# What does this change
These are changes necessary to have `porter explain` and `porter inspect` work with bundles created with an unsupported porter version.

**Do not require a supported porter.yaml file in the bundle cache**

It's important that we should be able to pull a bundle and fall back to working with it as a vanilla cnab bundle when it was built with an unsupported porter version. This removes trying to load the manifest from the bundle cache and each calling command will need to decide how to gracefully handle the bundle it pulls from the cache (i.e. some commands have to require a supported version, like build, but others like install can still work without a manifest).

**Remove Porter.Manifest field**
The porter struct has always had a Manifest field that is great at causing nil pointer exceptions (which is why I also fixed Porter to print out panic stack traces) because it's essentially a global variable and not all code paths populate it.

Since we need to support bundles that may not be able to populate a manifest, i.e. it's an unsupported manifest schema, or a vanilla CNAB bundle, I've removed it and made any commands that need the manifest to load it themselves instead of relying on this global.

# What issue does it fix
Part of #1956 

# Notes for the reviewer
N/A

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/cmd/porter/main.go
+++ b/cmd/porter/main.go
@@ -5,6 +5,7 @@ import (
 	_ "embed"
 	"fmt"
 	"os"
+	"runtime/debug"
 	"strings"
 
 	"get.porter.sh/porter/pkg/cli"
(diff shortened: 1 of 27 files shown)


vdice (approved):
LGTM

PR #1990, opened 2022-03-29
Update to go 1.18

### LINKED ISSUE #1989: CVE-2022-27191
Go 1.7.8 is affected by CVE-2022-27191. This isn't a vulnerability that affects Porter (we don't use that library/function) but it causes our IronBank CI pipeline to fail.

VinozzZ:
# What does this change
  update go version to 1.18

# What issue does it fix
Closes #1989 

# Notes for the reviewer

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module get.porter.sh/porter
 
-go 1.17
+go 1.18
 
 replace (
 	// See https://github.com/hashicorp/go-plugin/pull/127 and
--- a/magefile.go
+++ b/magefile.go
@@ -41,7 +41,7 @@ import (
 
 const (
 	PKG       = "get.porter.sh/porter"
-	GoVersion = ">=1.18"
+	GoVersion = ">=1.17"
 )
 
 var must = shx.CommandBuilder{StopOnError: true}
--- a/magefile.go
+++ b/magefile.go
@@ -41,7 +41,7 @@ import (
 
 const (
 	PKG       = "get.porter.sh/porter"
-	GoVersion = ">=1.17"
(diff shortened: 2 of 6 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module get.porter.sh/porter
 
-go 1.17
+go 1.18
```
carolynvs, line 3: I see that this caused the go.sum file to change, but at least it doesn't impact the go.mod file like the upgrade to 1.17 did. We may run into the sum file changing for no reason depending on which version of go a developer is running but hopefully that doesn't cause errors like when you add a go mod entry and forget to run tidy. We'll see I guess. 😁

magefile.go
```diff
@@ -41,7 +41,7 @@ import (
 
 const (
 	PKG       = "get.porter.sh/porter"
-	GoVersion = ">=1.17"
+	GoVersion = ">=1.18"
```
carolynvs, line 44: If we only want to use Go 1.18 on the build server but allow people to use go 1.17 locally, then this line should be reverted.

carolynvs (approved):
Thanks for vetting the upgrade!

carolynvs:
Let's keep an eye out for where we think the code base could take advantage of generics and create issues to tackle them. I think our mixin library (pkg/exec/builder) could use generics to help all the mixins with unmarshaling custom commands.

https://github.com/getporter/porter/blob/55d92c0f283c1e234a8477ad9d49a5e8c3f08107/pkg/exec/builder/action.go#L13-L64

Issue #1027, opened 2020-05-05
Store sensitive data in a secret store, not Porter's database


carolynvs:
Claims store sensitive data (parameters) and should not be stored in the database. We should come up with a way to store sensitive data from claims and other tables in the user's secret store, then when we retrieve the data, look up the sensitive values if needed to get a full document.

For some calls, maybe we don't need the data, we are just listing claims, but others like retrieving outputs and showing the values, will need to get the real values from the secret store.

carolynvs:
We can also just encrypt the sensitive bits
claim:
name: foo
parameters: ENCRYPTED # key in porter config

VinozzZ:
The current plan is to allow users to specify a destination for porter output through the same secrete plugin they use for resolving secrete values that needed to run bundle.
It would look something like this
`
default-secrets = "mysecrets"
   
[[secrets]]
name = "mysecrets"
plugin = "azure.keyvault"
   
[secrets.config]
input-vault = "myinputvault"
output-vault = "myoutputvault"
`

After checking the limitations on Azure key vault and Hashicorp vault, it looks like the upper limit for number of keys stored in a vault is only affected by the underlying storage space. However, Hashicorp does have a[ per object size limit](https://www.vaultproject.io/docs/internals/limits#storage-related-limits). Therefore, I plan to store each installation output separately using installation id or run id.

carolynvs:
This sounds like a good plan, looking forward to your PR! 🚀

VinozzZ:
I will start work on the Quickstart document

VinozzZ:
Started on the Task document

VinozzZ:
After looking at the example output from azure keyvault, we probably should add an dash in the secret key to make it more friendly to users to see

carolynvs:
Is this okay to close?

PR #1992, opened 2022-03-29
Save sensitive params into secret store

### LINKED ISSUE #1795: Do not allow repeats of credential or parameter sets on an installation
## Describe the bug
I don't know how the data got into this state. I guess I can install a bunch even when the last install worked.

Now when I look at the installation it has repeated copies of the credential sets used:

```yaml
schemaVersion: 1.0.0
name: operator
namespace: operator
created: 2021-10-11T11:48:29.598255-05:00
modified: 2021-10-11T11:48:29.598255-05:00
bundleRepository: localhost:5000/porter-operator
bundleTag: canary
parameters:
  namespace: test
credentialSets:
  - kind
  - kind
  - kind
  - kind
  - kind
  - kind
  - kind
  - kind
  - kind
  - kind
  - kind
  - kind
  - kind
  - kind
  - kind
  - kind
  - kind
  - kind
  - kind
status:
  runId: 01FHX42QBPE8KNTYVD677WE8DH
  action: install
  resultId: 01FHX42RSJV5KD4W3RQ5RJT5CA
  resultStatus: failed
  installationCompleted: true
  bundleReference: localhost:5000/porter-operator:canary
  bundleVersion: 0.2.0-19-g0519cf2
  bundleDigest: sha256:26e6b58ad3fc4138afe46fb2bd0f6db73c9ab182bc95ccadfa241c836ec652be
_calculated:
  resolvedparameters: []
```
## To Reproduce
1. Run `...`
2. Use this porter.yaml from this repository or gist
3. Run this porter command `...`

## Expected behavior
Porter should cleanup the parameter and credential set list on an installation so that only unique values are kept.

## Version
```
porter v1.0.0-alpha.3-48-g6266cf4f (6266cf4f)
```

VinozzZ:
# What does this change
This PR implements a new feature to handle sensitive data in porter.
Prior to this change, porter saves all sensitive parameters and outputs into its own database. This created security concerns.
In this PR, porter will have the ability to store all sensitive data into a secret plugins configured by users and retrieve the data when needed.

The way that I implemented this functionality is by changing the `Parameters` field on `Installation` record to be a parameter set. This allows us to reuse code from parameter set store to handle secret data.
For outputs,  I introduced a new `Key` field on the `Output` record to store the secret keys that's used to reference a sensitive value from a secret store.

all secret key is associated with a run record so we can preserve different values for each run.

# What issue does it fix
Part 1 work for #1027 
Closes #1795 #2012

Docs update will be a separate PR

# Notes for the reviewer

# Checklist
- [x] Did you write tests?

# Reviewer Checklist


go.mod
```diff
@@ -3,6 +3,8 @@ module get.porter.sh/porter
 go 1.18
 
 replace (
+	// TODO: remove before merging
+	github.com/cnabio/cnab-go v0.23.1 => github.com/vinozzz/cnab-go v0.23.2-0.20220408154326-6a3da71c53b6
```
carolynvs, line 7: Go ahead and open a PR to cnab-go, as draft, and I can review those changes too

pkg/claims/installation.go
```diff
@@ -158,18 +168,70 @@ func (i *Installation) SetLabel(key string, value string) {
 	i.Labels[key] = value
 }
 
+// EncodeSensitiveParameter encodes parameters that contains sensitive data so
+// it can be associated with the current installation record.
+func (i Installation) EncodeSensitiveParameter(param secrets.Strategy) secrets.Strategy {
```
carolynvs, line 173: nit: Let's try to come up with a more descriptive name for this function. Encode implies that we are changing the data, like HTML encoding, URL encoding, etc. As I understand it, this function is making the sensitive parameter unique to the installation when saved into the secret store right?

pkg/claims/output.go
```diff
@@ -37,6 +39,24 @@ func (o Output) GetSchema(b cnab.ExtendedBundle) (definition.Schema, bool) {
 	return definition.Schema{}, false
 }
 
+// FormatSensitive transforms an output value to a secret key that can be used
+// to store/retrieve the sensitive output value from a secret store.
+// It returns the transformed output.
+func (o Output) FormatSensitive() Output {
```
carolynvs, line 45: nit: This name implies that it's formatting the output, but it's actually generating a secretkey, let's update the name to better matches what it's doing (the actual output value isn't actually involved at all).

pkg/claims/result.go
```diff
@@ -66,3 +69,21 @@ func (r Result) NewOutput(name string, data []byte) Output {
 		Value:         data,
 	}
 }
+
+func (r Result) FilterSensitiveOutputs(result driver.OperationResult, bun bundle.Bundle, store secrets.Store) []Output {
```
carolynvs, line 73: Since this is only used in SaveOperationResult, let's move this function closer to the code that's using it.

pkg/claims/result.go
```diff
@@ -66,3 +69,21 @@ func (r Result) NewOutput(name string, data []byte) Output {
...
+	o := make([]Output, 0, len(result.Outputs))
+	for name, value := range result.Outputs {
+		sensitive, err := bun.IsOutputSensitive(name)
+		ot := r.NewOutput(name, []byte(value))
+		if err != nil || !sensitive {
+			o = append(o, ot)
+			continue
+		}
+
+		ot = ot.FormatSensitive()
```
carolynvs, line 83: It would be really easy to accidentally call FormatSensitive function before you saved the original value of the output (before overwriting the value with the secret key).

I would prefer that we don't have Output.Value mean two things depending on the context: either it's the actual output value or it's a key.

It seems that you are persisting the output with the value set to the secret key, correct? How about instead we added a field to Output, maybe `Output.SecretKey` with the location of the secret, so that we don't use Value for two different purposes. Then change the function name to something like "FlagAsSecret", so it's more clear what it is doing.

pkg/claims/result.go
```diff
@@ -66,3 +69,21 @@ func (r Result) NewOutput(name string, data []byte) Output {
 		Value:         data,
 	}
 }
+
+func (r Result) FilterSensitiveOutputs(result driver.OperationResult, bun bundle.Bundle, store secrets.Store) []Output {
+	o := make([]Output, 0, len(result.Outputs))
+	for name, value := range result.Outputs {
+		sensitive, err := bun.IsOutputSensitive(name)
+		ot := r.NewOutput(name, []byte(value))
+		if err != nil || !sensitive {
```
carolynvs, line 78: How come this is appending the output if there's an error?

pkg/claims/result.go
```diff
@@ -66,3 +69,21 @@ func (r Result) NewOutput(name string, data []byte) Output {
...
+			o = append(o, ot)
+			continue
+		}
+
+		ot = ot.FormatSensitive()
+		o = append(o, ot)
+		store.Create(secrets.SourceSecret, string(ot.Value), string(value))
```
carolynvs, line 85: The error isn't being handled here.

Since this function is persisting sensitive outputs, the name of the function should be changed to make it clear that it is going to do that. Otherwise it sounds like it's just filtering and is safe to call anytime.

pkg/claims/run.go
```diff
@@ -70,15 +79,53 @@ func (r Run) DefaultDocumentFilter() interface{} {
...
+	bun := cnab.ExtendedBundle{r.Bundle}
+	for name, value := range params {
+
+		stringVal, err := bun.WriteParameterToString(name, value)
+		if err != nil {
+			return err
+
+		}
+
+		strategy := parameters.DefaultStrategy(name, stringVal)
```
carolynvs, line 106: I wasn't expecting to need a new Parameter Set resolution strategy for this feature. Can you explain how DefaultStrategy works? It would really help to get a high level walk through of how and where it's used.

pkg/secrets/plugins/secrets_protocol.go
```diff
@@ -11,4 +11,6 @@ type SecretsProtocol interface {
 	// - keyName=key, keyValue=conn-string
 	// - keyName=path, keyValue=/tmp/connstring.txt
 	Resolve(keyName string, keyValue string) (string, error)
+
+	Create(keyName string, keyValue string, value string) error
```
carolynvs, line 15: Don't forget to add a docstring here. Maybe explain why only create is needed.

pkg/claims/run.go
```diff
@@ -56,7 +59,13 @@ type Run struct {
...
 	// Parameters is the full set of resolved parameters stored on the claim.
 	// This includes internal parameters, resolved parameter sources, values resolved from parameter sets, etc.
-	Parameters map[string]interface{} `json:"parameters" yaml:"parameters" toml:"parameters"`
+	ResolvedParameters map[string]interface{} `json:"-" yaml:"-" toml:"-"`
+
+	// Parameters is the full set of parameters that's being used during the
+	// current run.
+	// This includes internal parameters, parameter sources, values from parameter sets, etc.
+	// Any sensitive data will be sannitized before saving to the database.
+	Parameters parameters.ParameterSet `json:"parameters,omitempty" yaml:"parameters,omitempty" toml:"parameters,omitempty"`
```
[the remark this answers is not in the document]
VinozzZ: The ParameterOverrides is only storing the value that user pass in through the command line. We still need to store the full parameter values that includes default and internal parameter values, like `porter-state`

pkg/porter/testdata/show/expected-output.txt
```diff
@@ -10,13 +10,13 @@ Bundle:
 Labels:
   io.cnab/app: wordpress
   io.cnab/appVersion: v1.2.3
-
```
carolynvs, line 13: Please undo the changes to the output formatting so that the display of the expected outputs rename unchanged (unless we are showing new/different data, the whitespace and headers should not change).

pkg/secrets/strategy.go
```diff
@@ -64,9 +64,6 @@ type Source struct {
 }
 
 func (s Source) MarshalRaw() interface{} {
-	if s.Key == "" {
```
carolynvs, line 67: It's not immediately clear to me why this was removed and I'm concerned it may change how things are serialized. Did something make you need to remove it?

tests/smoke/desiredstate_test.go
```diff
@@ -65,9 +65,12 @@ func TestDesiredState(t *testing.T) {
 	output, stderr, err := test.RunPorter("installation", "apply", "mybuns.yaml", "--namespace", "operator")
 	require.NoError(t, err)
 	require.Contains(t, stderr, "The installation is out-of-sync, running the install action")
-	require.Contains(t, stderr, "Triggering because the installation has not completed successfully yet")
-	test.RequireInstallationExists("operator", "mybuns")
+	require.Contains(t, stderr, "Triggering because the installation has not completed successfully yet", stderr)
```
carolynvs, line 68: Did you intend to have stderr  added as the last argument?

tests/smoke/desiredstate_test.go
```diff
@@ -94,9 +97,13 @@ func TestDesiredState(t *testing.T) {
 	require.NoError(t, err)
 	tests.RequireOutputContains(t, output, "The installation is out-of-sync, running the upgrade action")
 
-	installation, err := test.ShowInstallation("operator", "mybuns")
+	displayInstallation, err := test.ShowInstallation("operator", "mybuns")
 	require.NoError(t, err)
-	require.Equal(t, float64(3), installation.Parameters["log_level"])
+	for name, value := range displayInstallation.Parameters {
```
carolynvs, line 102: Since Parameters is still a map, I think you can keep what was there originally instead of iterating now?

carolynvs (commented):
All done reviewing! It looks great, most of my questions and comments are about keeping the changes consistent with the rest of Porter (or avoiding my own past mistakes! 😅) Let me know if you have any questions.

pkg/claims/result.go
VinozzZ: Because I'm returning the outputs that doesn't contain any sensitive data back to the caller. I will add documentation about that behavior

pkg/porter/helpers.go
```diff
@@ -226,6 +232,56 @@ func (p *TestPorter) CompareGoldenFile(goldenFile string, got string) {
 	p.TestConfig.TestContext.CompareGoldenFile(goldenFile, got)
 }
 
+// CreateInstallation saves an installation record into claim store and store
+// sensitive parameters into secret store.
+func (p *TestPorter) CreateInstallation(i claims.Installation, bun cnab.ExtendedBundle) claims.Installation {
+	return p.TestClaims.CreateInstallation(i, func(i *claims.Installation) {
```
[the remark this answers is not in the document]
VinozzZ: I have updated the function to use the newly created `sanitizer` package

pkg/parameters/parameters.go
```diff
@@ -37,3 +41,12 @@ func ParseVariableAssignments(params []string) (map[string]string, error) {
 
 	return variables, nil
 }
+
+// DefaultStrategy is the strategy used to store non-sensitive parameters
+func DefaultStrategy(name string, value string) secrets.Strategy {
+	return secrets.Strategy{
+		Name:   name,
+		Source: secrets.Source{Key: DEFAULT_STRATEGY, Value: value},
```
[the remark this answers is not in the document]
VinozzZ: Maybe I have a misunderstanding here regarding different strategies. I thought `host.SourceValue` is only used when the secret plugin is configured to use the host plugin. If a user specifies other plugins, does that mean the plugin implementation needs to know how to resolve `host.SourceValue` too?

pkg/parameters/parameters.go
```diff
@@ -4,13 +4,17 @@ import (
...
+	"get.porter.sh/porter/pkg/secrets"
 	"github.com/cnabio/cnab-go/schema"
 )
 
 const (
 	// SchemaVersion represents the version associated with the schema
 	// for parameter set documents.
 	SchemaVersion = schema.Version("1.0.1")
+
+	//
```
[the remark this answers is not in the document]
VinozzZ: I added comments in the code

pkg/porter/parameters.go
```diff
@@ -572,7 +572,7 @@ func (p *Porter) resolveParameters(installation claims.Installation, bun cnab.Ex
 	for key, unconverted := range mergedParams {
 		param, ok := bun.Parameters[key]
 		if !ok {
-			return nil, fmt.Errorf("parameter %s not defined in bundle", key)
+			return nil, fmt.Errorf("resolvedParameters:  %s not defined in bundle", key)
```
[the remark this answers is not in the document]
VinozzZ: Oops, I changed it back. It was only changed for debugging purpose

pkg/porter/reconcile.go
```diff
@@ -126,7 +133,7 @@ func (p *Porter) ReconcileInstallation(ctx context.Context, opts ReconcileOption
 
 // IsInstallationInSync determines if the desired state of the installation matches
 // the state of the installation the last time it was modified.
-func (p *Porter) IsInstallationInSync(ctx context.Context, i claims.Installation, lastRun *claims.Run, action BundleAction) (bool, error) {
+func (p *Porter) IsInstallationInSync(ctx context.Context, i *claims.Installation, lastRun *claims.Run, action BundleAction) (bool, error) {
```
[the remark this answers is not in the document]
VinozzZ: I changed it back. I think it was a leftover code from previous implementation

pkg/porter/show_test.go
```diff
@@ -78,23 +87,56 @@ func TestPorter_ShowBundle(t *testing.T) {
 					},
 				},
 			}
-			i := p.TestClaims.CreateInstallation(claims.NewInstallation("dev", "mywordpress"), p.TestClaims.SetMutableInstallationValues, func(i *claims.Installation) {
+
+			i := p.TestClaims.CreateInstallation(newInstallation("01FZVC5AVP8Z7A78CSCP1EJ604", "dev", "mywordpress", tm, tm), p.TestClaims.SetMutableInstallationValues, func(i *claims.Installation) {
```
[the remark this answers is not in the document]
VinozzZ: I changed how I implement the new helper function. Let me know if it makes more sense now

pkg/porter/show_test.go
```diff
@@ -106,3 +148,16 @@ func TestPorter_ShowBundle(t *testing.T) {
 		})
 	}
 }
+
+func newInstallation(id string, namespace string, name string, created, modified time.Time) claims.Installation {
```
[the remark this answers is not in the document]
VinozzZ: Yea..I didn't realize the `SetMutableRunValues` already does this

pkg/secrets/strategy.go
VinozzZ: I changed it back

tests/smoke/desiredstate_test.go
VinozzZ: Oops, I removed it

go.mod
VinozzZ: https://github.com/cnabio/cnab-go/pull/282/files

pkg/cnab/provider/action.go
```diff
@@ -140,9 +140,16 @@ func (r *Runtime) Execute(ctx context.Context, args ActionArguments) error {
...
 	currentRun.BundleReference = args.BundleReference.Reference.String()
 	currentRun.BundleDigest = args.BundleReference.Digest.String()
-	currentRun.Parameters = args.Params
+
+	currentRun.Parameters.Parameters, err = r.sanitizer.CleanRawParameters(args.Params, cnab.ExtendedBundle{b.Bundle}, currentRun.ID)
+	if err != nil {
+		return err
+	}
+
+	currentRun.EncodeParameterOverrides()
```
carolynvs, line 149: Can we move this method onto the sanitizer so that it's with the other sensitive data logic in one place?

pkg/cnab/provider/action.go
```diff
@@ -232,6 +243,10 @@ func (r *Runtime) SaveOperationResult(opResult driver.OperationResult, installat
 
 	for outputName, outputValue := range opResult.Outputs {
 		output := result.NewOutput(outputName, []byte(outputValue))
+		output, err = r.sanitizer.CleanOutput(output, cnab.ExtendedBundle{run.Bundle})
+		if err != nil {
+			bigerr = multierror.Append(bigerr, errors.Wrapf(err, "error sanitizing sensitive %s output for %s run of installation %s\n%#v", output.Name, run.Action, installation, output))
```
carolynvs, line 248: To avoid leaking sensitive data, let's not print the output value as well, just the other fields.

pkg/parameters/parameterset.go
```diff
@@ -65,6 +69,15 @@ func NewParameterSet(namespace string, name string, params ...secrets.Strategy)
 	return ps
 }
 
+// NewParameterSet creates a new ParameterSet with the required fields initialized.
+func NewInternalParameterSet(namespace string, name string, params ...secrets.Strategy) ParameterSet {
+	return NewParameterSet(namespace, INTERNAL_PARAMETERER_SET+"-"+name, params...)
+}
+
+func (s ParameterSet) IsInternalParameterSet() bool {
```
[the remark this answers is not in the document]
carolynvs: This still need to be removed

pkg/cnab/provider/action.go
VinozzZ: I believe the value field of the output after calling `CleanOutput()` will always be cleared out. It shouldn't have leakage issue

VinozzZ: Good point! I changed the `CleanOutput` to always clear out the value field when there's an error

pkg/porter/show.go
```diff
@@ -119,24 +119,20 @@ func (p *Porter) ShowInstallation(ctx context.Context, opts ShowOptions) error {
 			}
 		}
 
-		// Print parameters, if any
 		if len(displayInstallation.Parameters) > 0 {
-			fmt.Fprintln(p.Out)
```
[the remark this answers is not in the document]
VinozzZ: oops, moved it back

pkg/porter/testdata/show/expected-output.json
```diff
@@ -1,6 +1,7 @@
 {
   "schemaType": "Installation",
   "schemaVersion": "1.0.0",
+  "id": "01FZVC5AVP8Z7A78CSCP1EJ604",
```
carolynvs, line 4: This is just a heads up that we have a follow-up issue to expose this value as an environment variable and template value in the bundle. I think it will be useful for people to have a (pretty) random value to use with their installation when setting resource names.

pkg/secrets/plugins/host/plugin.go
```diff
@@ -22,6 +24,10 @@ func (p Plugin) Close() error {
 	return nil
 }
 
+func (p Plugin) Create(keyName string, keyValue string, value string) error {
+	return errors.New("not implemented")
```
carolynvs, line 28: Are there any code flows that would hit this error? I'm concerned that this error message if we did accidentally hit it would not be helpful to the user. Perhaps we can say "The secrets.porter.host plugin does not support saving secrets. Please configure a secret plugin."

tests/smoke/desiredstate_test.go
```diff
@@ -94,7 +97,7 @@ func TestDesiredState(t *testing.T) {
 	require.NoError(t, err)
 	tests.RequireOutputContains(t, output, "The installation is out-of-sync, running the upgrade action")
 
-	installation, err := test.ShowInstallation("operator", "mybuns")
+	displayInstallation, err := test.ShowInstallation("operator", "mybuns")
```
carolynvs, line 100: Did you mean to change this variable name? Subsequent asserts aren't looking at the correct variable (see two lines down).

tests/smoke/desiredstate_test.go
```diff
@@ -66,8 +66,11 @@ func TestDesiredState(t *testing.T) {
 	require.NoError(t, err)
 	require.Contains(t, stderr, "The installation is out-of-sync, running the install action")
 	require.Contains(t, stderr, "Triggering because the installation has not completed successfully yet")
-	test.RequireInstallationExists("operator", "mybuns")
+	installation := test.RequireInstallationExists("operator", "mybuns")
+	require.Equal(t, "succeeded", installation.Status.ResultStatus)
 
+	_, err = test.ShowInstallation("operator", "mybuns")
```
carolynvs, line 72: It's not clear what you were testing when you added the call to show installation here?

tests/tester/helpers.go
```diff
@@ -46,22 +47,27 @@ func (t Tester) MakeTestBundle(name string, ref string) {
...
 
 func (t Tester) RequireInstallationExists(namespace string, name string) claims.Installation {
-	installation, err := t.ShowInstallation(namespace, name)
+	di, err := t.ShowInstallation(namespace, name)
 	require.NoError(t.T, err)
-	require.Equal(t.T, name, installation.Name, "incorrect installation name")
-	require.Equal(t.T, namespace, installation.Namespace, "incorrect installation namespace")
+	require.Equal(t.T, name, di.Name, "incorrect installation name")
+	require.Equal(t.T, namespace, di.Namespace, "incorrect installation namespace")
+	installation, err := di.ConvertToInstallation()
```
carolynvs, line 68: This function `RequireInstallationExists` is meant to check that we persisted the installation, so I'd prefer that we don't add checks for specific features here.

Instead of asserting this functionality across all the smoke tests, let's add a test case to one of the existing smoke tests to verify that sensitive data was persisted correctly. At the smoke test level I'm much more concerned about if we persisted a secret to the secret store, and that the installation renders properly.

This check here probably belongs in a unit test in the same file where ConvertToInstallation is defined with accompanying asserts that it was converted properly.

pkg/claims/run.go
```diff
@@ -99,20 +103,40 @@ func (r Run) ShouldRecord() bool {
...
 		BundleReference: r.BundleReference,
-		Parameters:      r.Parameters,
+		Parameters:      r.TypedParameterValues(),
 		Custom:          r.Custom,
 	}
 }
 
+// TypedParameterValues returns parameters values that have been converted to
+// its typed value based on its bundle definition.
+func (r Run) TypedParameterValues() map[string]interface{} {
```
carolynvs, line 122: Can you add a unit test in run_test.go for this function?

pkg/cnab/provider/action.go
carolynvs: I really like that 👍 It should help prevent us from accidentally persisting it if the clean failed.

pkg/parameters/parameterset.go
```diff
@@ -80,5 +93,28 @@ func (s ParameterSet) Validate() error {
 }
 
 func (s ParameterSet) String() string {
-	return fmt.Sprintf("%s/%s", s.Namespace, s.Name)
+	if s.Namespace != "" {
```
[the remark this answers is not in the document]
carolynvs: Looks like you accidentally marked this as resolved without reverting the changes to this function.

pkg/porter/cnab.go
```diff
@@ -268,10 +270,34 @@ func (o *sharedOptions) parseParams() error {
 	if err != nil {
 		return err
 	}
+
 	o.parsedParams = p
 	return nil
 }
 
+func (o *sharedOptions) populateInternalParameterSet(p *Porter, bun cnab.ExtendedBundle, i *claims.Installation) error {
```
carolynvs, line 278: Can you add a unit test for this new function?

pkg/porter/install.go
```diff
@@ -103,32 +103,35 @@ func (p *Porter) applyActionOptionsToInstallation(i *claims.Installation, opts *
...
-		for _, existing := range i.ParameterSets {
-			if existing == cs {
-				continue
-			}
+	i.CredentialSets = append(i.CredentialSets, Unique(i.CredentialSets, opts.CredentialIdentifiers...)...)
+
+	return nil
+}
+
+func Unique(existings []string, n ...string) []string {
```
carolynvs, line 121: I've marked this PR as fixing #1795. Please add a unit test for this function, and maybe another unit or smoke test that checks that calling say "upgrade" twice doesn't result in the names of the parameter sets having duplicates.

pkg/porter/list.go
```diff
@@ -61,28 +64,116 @@ func parseLabels(raw []string) map[string]string {
...
+		ParameterSets:  installation.ParameterSets,
+		Status:         installation.Status,
+	}
+
+	return di
+}
+
+// ConvertToInstallationClaim transforms the data from DisplayInstallation into
+// a Installation record.
+func (d DisplayInstallation) ConvertToInstallation() (claims.Installation, error) {
```
carolynvs, line 132: Can you add a unit test for the new functions in this file?

carolynvs (commented):
If I wanted to try this out locally, which secret plugin should I use? Is there a branch of the azure secret plugin with create support or should I wait for the filesystem secret plugin?

tests/smoke/desiredstate_test.go
VinozzZ: Oops, I fixed it now

tests/tester/helpers.go
VinozzZ: This change is required here is due to the fact that now `ShowInstallation()` returns the type of `DisplayInstallation` rather than `claims.Installation`. However, it seems that this function expects a claim record to be returned. Maybe that's not necessary, I will double check

VinozzZ:
> If I wanted to try this out locally, which secret plugin should I use? Is there a branch of the azure secret plugin with create support or should I wait for the filesystem secret plugin?

You should be able to try it out with the new filesystem secret plugin now!

pkg/parameters/parameterset.go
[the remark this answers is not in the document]
VinozzZ: Oops, I changed it back

pkg/secrets/plugins/host/plugin.go
```diff
@@ -22,6 +24,10 @@ func (p Plugin) Close() error {
 	return nil
 }
 
+func (p Plugin) Create(keyName string, keyValue string, value string) error {
+	return fmt.Errorf("the %s plugin does not support saving secrets. Please configure a secret plugin", PluginKey)
```
carolynvs, line 28: This error message doesn't give the user enough information to know how to fix the problem. They would need to then dig around on the site and hope that somewhere we explained what to do.

Whenever we know how to resolve an error, we should always link to the relevant documentation page. Or if there isn't a doc page (yet), tell them how to fix it in the message like "Edit your porter config file and set default-storage-plugin to X to use the insecure filesystem plugin. Do not use this plugin for production data."

pkg/secrets/plugins/file-system/plugin.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"os"
+
+	"get.porter.sh/porter/pkg/secrets"
+	"get.porter.sh/porter/pkg/secrets/plugins"
+	"github.com/hashicorp/go-hclog"
+)
+
+const PluginKey = secrets.PluginInterface + ".porter.file-system"
```
carolynvs, line 11: Let's remove the hyphen (it's not really needed for this word anyway, its a single word. Just "filesystem" will work.

pkg/secrets/plugins/file-system/store.go
```diff
@@ -0,0 +1,120 @@
...
+
+	pathPrefix := s.config.PathPrefix
+	if pathPrefix == "" {
+		pathPrefix = SECRET_FOLDER
+
+	}
+
+	secretDir := filepath.Join(porterHomeDir, pathPrefix)
+
+	if err := os.MkdirAll(secretDir, os.ModePerm); err != nil && !errors.Is(err, os.ErrExist) {
```
carolynvs, line 64: Heads up that you shouldn't use `os.ModePerm` anywhere in Porter. We've learned the hard way that it easily leads to giving the wrong set of permissions to files, which in this case when we are dealing with sensitive data, is an even larger problem.

```
$ ls -la ~/.porter/secrets
total 16
drwxr-xr-x   4 carolynvs  staff  128 Apr 22 10:35 .
drwxr-x---  10 carolynvs  staff  320 Apr 22 10:35 ..
-rw-------   1 carolynvs  staff    5 Apr 22 10:35 01G18YMRP1EXPG0D19SKB9RA7Edbpassword
-rw-------   1 carolynvs  staff    5 Apr 22 10:35 01G18YMRP69SWANWY4C64W6SNRdbpassword
```

Right now we have named permission sets in pkg/version.go to avoid accidentally typing the wrong value in hex but none of them are appropriate in this case. 

All files in the directory (including the directory itself) should be `0600` with only the current user having access, not even the group. Let's make some more constants for this use case and indicate its for sensitive files/directories, e.g. `FileModeSensitiveDirectory`, `FileModeSensitiveWritable`.

tests/smoke/desiredstate_test.go
VinozzZ: Should be fixed now!

pkg/secrets/plugins/file-system/store.go
carolynvs: Can you add a unit test that verifies that the files are being created with the appropriate permissions? In general this plugin needs unit tests anyway but let's make sure that the permissions are included when you add them.

pkg/secrets/plugins/host/plugin.go
VinozzZ: I have updated the error message. Let me know if it helps!

pkg/porter/install.go
VinozzZ: I have added a unit test and a test case in the hello_test.go in smoke test

pkg/secrets/plugins/host/plugin.go
VinozzZ: Thanks for the recommended error message! I have updated it in the code

pkg/secrets/plugins/file-system/store.go
VinozzZ: That makes sense. Thanks for pointing it out! I have added tests

tests/tester/helpers.go
carolynvs: I haven't looked in the code, but I think it's safe to return the display installation here instead of claims.Installation

tests/testdata/mybuns/porter.yaml
```diff
@@ -19,6 +19,10 @@ parameters:
     minimum: 1
     maximum: 11
     default: 5
+  - name: password
```
carolynvs, line 22: It would help to define a default value for this new parameter. That way existing tests that were not testing scenarios related to secrets do not need to be changed (like the uninstall integration test). This will make it easier to maintain these tests over time as we need more test cases to this bundle.

carolynvs (changes requested):

pkg/claims/run.go
VinozzZ: added!

pkg/secrets/plugins/filesystem/store.go
```diff
@@ -0,0 +1,156 @@
...
+const (
+	SECRET_FOLDER              = "porter-secrets"
+	FileModeSensitiveDirectory = 0700
+	FileModeSensitiveWritable  = 0600
+	FileModeSensitiveReadable  = 0400
+)
+
+// Config contains information needed for creatina a new store.
+type Config struct {
+	PathPrefix    string `json:"path_prefix"`
```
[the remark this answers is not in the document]
VinozzZ: That sounds good to me. I wanted to give some flexibility but I agree that it may give users the wrong idea

pkg/secrets/plugins/filesystem/store.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+// Resolve implements the Resolve method on the secret plugins' interface.
+func (s *Store) Resolve(keyName string, keyValue string) (string, error) {
+	// check if the keyName is secret
+	if keyName != secrets.SourceSecret {
+		return s.hostStore.Resolve(keyName, keyValue)
+	}
+
+	f, err := s.OpenSecretFile(keyValue, os.O_RDONLY, FileModeSensitiveReadable)
```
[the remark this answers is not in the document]
VinozzZ: That's a great idea. I implemented your suggestion

pkg/secrets/pluginstore/store.go
```diff
@@ -55,11 +58,43 @@ func (s *Store) Resolve(keyName string, keyValue string) (string, error) {
...
+		if err != nil {
+			return nil, err
+		}
+
+		homeDir, err := s.GetHomeDir()
+		if err != nil {
+			return nil, err
+		}
+		cfg := filesystemsecret.NewConfig(s.DebugPlugins, homeDir)
+		err = json.Unmarshal(data, &cfg)
```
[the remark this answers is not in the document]
VinozzZ: I removed the config based on your suggestion

tests/testdata/mybuns/porter.yaml
VinozzZ: It definitely made smoke test cleaner!

tests/integration/testdata/bundles/bundle-with-sensitive-data/Dockerfile.tmpl
```diff
@@ -0,0 +1,16 @@
+# syntax=docker/dockerfile-upstream:1.4.0
+FROM debian:stretch-slim
+
+ARG BUNDLE_DIR
+
+RUN rm -f /etc/apt/apt.conf.d/docker-clean; echo 'Binary::apt::APT::Keep-Downloaded-Packages "true";' > /etc/apt/apt.conf.d/keep-cache
+RUN --mount=type=cache,target=/var/cache/apt --mount=type=cache,target=/var/lib/apt \
+    apt-get update && apt-get install -y ca-certificates
+
+ENV CUSTOM_VAR=boop
```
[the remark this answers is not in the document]
VinozzZ: i deleted the file since it's not needed

pkg/porter/list_test.go
```diff
@@ -88,3 +90,61 @@ func TestPorter_ListInstallations(t *testing.T) {
...
+	require.Equal(t, i.Bundle.Digest, convertedInstallation.Bundle.Digest, "invalid installation bundle")
+
+	require.Equal(t, len(i.Labels), len(convertedInstallation.Labels))
+	for key := range di.Labels {
+		require.Equal(t, i.Labels[key], convertedInstallation.Labels[key], "invalid installation lables")
+	}
+
+	require.Equal(t, i.Custom, convertedInstallation.Custom, "invalid installation custom")
+
+	require.True(t, reflect.DeepEqual(convertedInstallation.CredentialSets, i.CredentialSets), "invalid credential set")
```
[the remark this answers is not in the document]
VinozzZ: Changed!

pkg/porter/cnab_test.go
```diff
@@ -355,3 +358,51 @@ func Test_bundleFileOptions(t *testing.T) {
...
+	i := claims.NewInstallation("", bun.Name)
+
+	err = opts.populateInternalParameterSet(p.Porter, bun, &i)
+	require.NoError(t, err)
+
+	require.Len(t, i.Parameters.Parameters, 2)
+
+	// there should be no sensitive value on installation record
+	for _, param := range i.Parameters.Parameters {
+		if bun.IsSensitiveParameter(param.Name) {
```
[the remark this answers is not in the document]
VinozzZ: It makes perfect sense! I changed it to check for the correct data instead of relying on `IsSensitiveParameter`. I didn't make it into a table test since there isn't multiple test cases. Let me know if it works

pkg/claims/run_test.go
```diff
@@ -84,3 +88,69 @@ func TestRun_ShouldRecord(t *testing.T) {
...
+
+	for _, param := range params {
+		v, ok := typed[param.Name]
+		require.True(t, ok)
+		if param.Name == "porter-state" {
+			require.Nil(t, v)
+			continue
+		}
+
+		stringVal, err := cnab.WriteParameterToString(param.Name, v)
```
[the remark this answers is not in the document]
VinozzZ: Based on your other comments, I have changed the test to check the result of the typed value

pkg/claims/run_test.go
```diff
@@ -84,3 +88,69 @@ func TestRun_ShouldRecord(t *testing.T) {
...
+	}
+
+	run := NewRun("dev", "mybuns")
+	run.Bundle = bun
+	run.Parameters = parameters.NewParameterSet(run.Namespace, run.Bundle.Name)
+	params := []secrets.Strategy{
+		parameters.ValueStrategy("baz", "baz-test"),
+		parameters.ValueStrategy("name", "porter-test"),
+		parameters.ValueStrategy("porter-state", ""),
+		{Name: "foo", Source: secrets.Source{Key: secrets.SourceSecret, Value: "runID"}, Value: "supersecret"},
```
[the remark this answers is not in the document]
VinozzZ: fixed it

carolynvs (approved):
:shipit: Thanks for sticking with it through such a large change! Next time we'll do a proposal and design process so that this doesn't end up taking us by surprise at how big it is.

Well done! 👏

Issue #1996, opened 2022-03-31
Set bundle.json schemaVersion to 1.2.0


carolynvs:
Porter has always supported [CNAB 1.2.0](https://github.com/cnabio/cnab-spec/tree/cnab-core-1.2.0) since it was released but we forgot to mark that on the bundles we build. Let's set that correctly on our bundles going forward.

Also we need to validate that we are checking the bundle schema version and allowing a range of acceptable values: 1.x. That can be in a separate PR.

carolynvs:
Right now Porter does not check if the bundle schemaVersion is compatible. We only check if Porter supports all the required extensions listed (such as dependencies or file parameters).

When we bump our supported version of the bundle schema to 1.2, we need to improve how we check this value when working with a bundle. Working could me: porter explain, porter install, etc. Anything that reads a bundle.json should respect that schemaVersion value.

Bundles could be old, and use a previously supported version, like 1.0.0, 1.0.1, 1.1 or now 1.2. When we add validation to the bundle schema version it should allow for a list or range of known CNAB Core Specification schema versions. We should use the new schema-check configuration setting that is introduced in the new template delimiter PR that is coming later today. That will give us the validation function to use, and the config setting.

We should add an integration test that checks that the current version of porter can run a bundle from 1.0 of the spec.

PR #1997, opened 2022-03-31
Check manifest schemaVersion in Porter commands before executing

### LINKED ISSUE #1956: Commands that read the manifest should check the schema version

carolynvs:
# What does this change
**Check the manifest schema before build-time commands**
The following commands should check the manifest schemaVersion before
attempting to work with a manifest:

* lint
* build
* publish

Really all commands chain off of eachother, so as long as we have
consistent error handling when loading a manifest, all the commands will
have the same behavior.

I've added a config flag "schema-check", that indicates how the user
wants schema checks to occur. By default, we check that the schemaVersion
exactly matches the supported schema. The user may change that handling
using schema-check or PORTER_SCHEMA_CHECK (there is no flag). Its mostly
intended to help people move from one version of Porter to another and
migrate files without a hard error that may make that difficult.

Available values for schema-check are:

* exact - the entire version must match
* major - the MAJOR portion of the version must match
* minor - the MAJOR.MINOR portion of the version must match
* none - Just warn but don't stop execution regardless of how poor the
  match is.

In all cases, when the versions don't match exactly, we will always
print a warning. The config setting only controls if the porter command
immediately fails or if the user is allowed to continue past the
warning.

**Do no enforce manifest schemaVersion at runtime**

After a bundle has been built, if the author loosened the schema check
level then the runtime should respect that. So we will warn if it
doesn't match, but will always try to run it.

# What issue does it fix
Closes #1956 

# Notes for the reviewer
I have created a follow-up issue to handle the bundle.json schemaVersion which is a separate thing even though they have the same name. #1996 

# Checklist
- [x] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/docs/content/configuration.md
+++ b/docs/content/configuration.md
@@ -347,9 +347,9 @@ Therefore, it does not work with the Azure Cloud Shell driver.
 
 ### Schema Check
 The schema-check configuration file setting controls Porter's behavior when the schemaVersion of a resource does not match [Porter's supported version](/reference/file-formats/#supported-versions).
-By default, Porter requires that a resources schemaVersion field exactly matches the supported version.
+By default, Porter requires that a resource's schemaVersion field exactly matches the supported version.
 In some cases, such as when migrating to a new version of Porter, it may be helpful to use a less strict version comparison.
-Allows values are:
+Allowed values are:
 
 * exact - Default behavior. Require that the schemaVersion on the resource exactly match Porter's supported version.
   If it doesn't match, the command will fail.
--- a/docs/content/configuration.md
+++ b/docs/content/configuration.md
@@ -343,3 +343,22 @@ You should trust any bundles that you execute with this setting enabled as it gi
 
 ⚠️️ This configuration setting is only available when you are in an environment that provides access to the local docker daemon.
 Therefore, it does not work with the Azure Cloud Shell driver.
+
+
+### Schema Check
+The schema-check configuration file setting controls Porter's behavior when the schemaVersion of a resource does not match [Porter's supported version](/reference/file-formats/#supported-versions).
+By default, Porter requires that a resources schemaVersion field exactly matches the supported version.
+In some cases, such as when migrating to a new version of Porter, it may be helpful to use a less strict version comparison.
+Allows values are:
+
+* exact - Default behavior. Require that the schemaVersion on the resource exactly match Porter's supported version.
+  If it doesn't match, the command will fail.
+* minor - Require that the MAJOR.MINOR portion of the schemaVersion on the resource match Porter's supported version.
+  For example, a bundle with a schemaVersion of 1.2.3 would work even though the supported version is 1.2.5.
+* major - Require that the MAJOR portion of the schemaVersion on the resource match Porter's supported version.
+  For example, a bundle with a schemaVersion of 1.2.3 would work even though the supported version is 1.3.0.
+* none - Only print a warning when the schemaVersion does not exactly match the supported version.
+
+Porter can only guarantee correct parsing of the file when the schemaVersion exactly matches.
+Depending on what has changed between schema versions, you can make a judgement call on if those changes are relevant to your situation.
+
--- a/pkg/cnab/config-adapter/adapter_test.go
+++ b/pkg/cnab/config-adapter/adapter_test.go
@@ -1,6 +1,7 @@
 package configadapter
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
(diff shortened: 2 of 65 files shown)


docs/content/configuration.md
```diff
@@ -343,3 +343,22 @@ You should trust any bundles that you execute with this setting enabled as it gi
 
 ⚠️️ This configuration setting is only available when you are in an environment that provides access to the local docker daemon.
 Therefore, it does not work with the Azure Cloud Shell driver.
+
+
+### Schema Check
+The schema-check configuration file setting controls Porter's behavior when the schemaVersion of a resource does not match [Porter's supported version](/reference/file-formats/#supported-versions).
+By default, Porter requires that a resources schemaVersion field exactly matches the supported version.
+In some cases, such as when migrating to a new version of Porter, it may be helpful to use a less strict version comparison.
+Allows values are:
```
vdice, line 352:
```suggestion
Allowed values are:
```

docs/content/configuration.md
```diff
@@ -343,3 +343,22 @@ You should trust any bundles that you execute with this setting enabled as it gi
 
 ⚠️️ This configuration setting is only available when you are in an environment that provides access to the local docker daemon.
 Therefore, it does not work with the Azure Cloud Shell driver.
+
+
+### Schema Check
+The schema-check configuration file setting controls Porter's behavior when the schemaVersion of a resource does not match [Porter's supported version](/reference/file-formats/#supported-versions).
+By default, Porter requires that a resources schemaVersion field exactly matches the supported version.
```
vdice, line 350:
```suggestion
By default, Porter requires that a resource's schemaVersion field exactly matches the supported version.
```

pkg/cnab/config-adapter/adapter_test.go
```diff
@@ -799,10 +800,10 @@ func TestManifestConverter_generateDefaultAction(t *testing.T) {
 		}},
 		{
 			"help", bundle.Action{
-				Description: "Print an help message to the standard output",
-				Modifies:    false,
-				Stateless:   true,
-			}},
+			Description: "Print an help message to the standard output",
```
vdice, line 803:
```suggestion
			Description: "Print a help message to the standard output",
```

pkg/schema/check-strategy.go
```diff
@@ -0,0 +1,86 @@
...
+type CheckStrategy string
+
+const (
+	// CheckStrategyExact requires that resource schemaVersion values exactly match the supported schema version.
+	CheckStrategyExact CheckStrategy = "exact"
+
+	// CheckStrategyMinor requires that resource schemaVersion values match the MAJOR.MINOR of the supported schema version.
+	CheckStrategyMinor CheckStrategy = "minor"
+
+	// CheckStrategyMajor requires that resource schemaVersion values exactly MAJOR of the supported schema version.
```
vdice, line 22:
```suggestion
	// CheckStrategyMajor requires that resource schemaVersion values exactly match the MAJOR portion of the supported schema version.
```

pkg/schema/check-strategy.go
```diff
@@ -0,0 +1,86 @@
...
+// CheckStrategy is an enum of values for handling schemaVersion
+// comparisons of two resources. Allowed values are: CheckStrategyExact,
+// CheckStrategyMinor, CheckStrategyMajor, CheckStrategyNone.
+type CheckStrategy string
+
+const (
+	// CheckStrategyExact requires that resource schemaVersion values exactly match the supported schema version.
+	CheckStrategyExact CheckStrategy = "exact"
+
+	// CheckStrategyMinor requires that resource schemaVersion values match the MAJOR.MINOR of the supported schema version.
```
vdice, line 19:
```suggestion
	// CheckStrategyMinor requires that resource schemaVersion values match the MAJOR.MINOR portion of the supported schema version.
```

pkg/schema/check-strategy.go
```diff
@@ -0,0 +1,86 @@
...
+	CheckStrategyMinor CheckStrategy = "minor"
+
+	// CheckStrategyMajor requires that resource schemaVersion values exactly MAJOR of the supported schema version.
+	CheckStrategyMajor CheckStrategy = "major"
+
+	// CheckStrategyNone ignores the resource schemaVersion. Errors will most likely ensue but have fun!
```
vdice, line 25:
```suggestion
	// CheckStrategyYolo ignores the resource schemaVersion. Errors will most likely ensue but have fun!
```

vdice: jk

vdice (approved):

PR #1999, opened 2022-04-01
Support additional docker flags and remove legacy docker build driver

### LINKED ISSUE #1769: Support buildkit flags
When I added buildkit support, I didn't expose all the necessary flags that docker build has so that the user can pass in information to the build during `porter build` that buildkit supports. When we build a bundle that uses buildkit in the template, we should support passing in extra arguments such as ssh, mounts and secrets.

https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/syntax.md

```
porter build --ssh default=$SSH_AUTH_SOCK
porter build --secret id=aws,src=$HOME/.aws/credentials
porter build --build-arg VERSION=1.2.3
```

The scenarios we want to support are:
* Run a command when building the invocation image inside a private repository that we don't want included in the final image. The user should have already checked out the repository into the bundle directory.
* Pull a private helm chart secured by a secret that is passed into porter build
* Specify a variable, like the application version, during build and use it.

### LINKED ISSUE #1941: Use docker system config
We have seen a few issues with [build](927), publish and "pull" where the user has customizations such as a proxy or custom certificate that works with Docker but not with Porter. Let's review how we are loading the docker configuration so that it includes docker's configuration file.

### LINKED ISSUE #1954: Make buildkit the default driver in porter v1
The performance improvements and feature sets in buildkit are useful enough that I'd like to see us make buildkit the default build driver when we release v1.

- [x] Make build drivers not an experimental feature anymore, it's been tested long enough to know it's working
- [ ] Change the default driver from docker to buildkit
- [x] Have a good think on if we even need to support just plain docker without buildkit. I'm not sure there is a reason at this point?

carolynvs:
# What does this change
**Support additional docker flags**

This adds support for the following docker flags:

--build-arg: Pass build arguments that can be used in the template dockerfile
--ssh: Provide a ssh configuration to the container while building
--secret: Provide a secret to the container while building
--no-cache: Build the image and do not use cached layers.

It also fixes how we call the docker buildx plugin so that user configuration, such as a proxy, is used.

After upgrading to a new version of buildx, I was also able to pick up a fix for pretty printing the progress to stderr, while capturing the plaintext output to the logs.

The dockerfile syntax 1.4.0 is officially released, so I've also edited our templates to use the official version `# syntax=docker/dockerfile-upstream:1.4.0`

**Remove experimental flag for build drivers**
Default to buildkit only and remove the docker build driver. Keep the build driver config option, in case we ever get creative in the future

# What issue does it fix
Closes #1769 (support new buildkit flags)
Closes #1954 (make buildkit the default driver and get rid of legacy docker)
Closes #1941 (load docker config)

# Notes for the reviewer
~🚨 This must be rebased before merging, it relies on #1998~

I kept the configuration infra in place in case we ever add a new build driver in the future. But the flag --driver is now hidden for the build command since there's no reason to set it at the moment. I also kept the experimental config documentation and just let people know it's not needed anymore. I felt that people who were used to the flag would be confused to have its doc disappear without explanation.

Long term maybe we need a page/section of "promoted" experimental features.

# Checklist
- [x] Did you write tests?

# Reviewer Checklist


vdice (approved):
So much goodness in this one!  The added example is great as well.  LGTM.

carolynvs:
@vdice The example sort of snuck in 😂 It's going to be the source for an upcoming blog post telling people about the new buildkit features. I'll add a follow-up to make it a proper example linked from our examples page too though! https://release-v1.porter.sh/examples.

PR #2001, opened 2022-04-01
Add installation.namespace template var

### LINKED ISSUE #2000: Make installation namespace and name available to actions
Currently, the template variable `installation.name` is set directly from the CNAB_INSTALLATION_NAME env variable. This is 
of the form `namespace/bundle_name` or just `/bundle_name` if no namespace is provided.

That form is not so useful as a substitution when writing action definitions, where it is much more helpful to be able to reference the namespace and bundle name separately. In particular, when defining a kubernetes action, it is clearly beneficial to be able to access the namespace component and also the bundle name can be used to generate an installation-specific resource name. Neither can be done at present using the composite form, without some ugly parsing from a previous exec action.

The proposal is to replace `installation.name` with the bundle name (stripped of namespace and /) and add `installation.namespace` to carry the namespace. This seems more consistent with how name and namespace are handled just about everywhere else.

Note: the CNAB_INSTALLATION_NAME env variable will not change: only the template variables derived from it. If you want the composite form, you can use either `{{ env.CNAB_INSTALLATION_NAME }}` or `{{ installation.namespace }}/{{ installation.name }}`.

An example of an action definition that might use these new variables:
```yaml
k8saction:
  - kubernetes:
      description: "Create some k8s resources"
      namespace: "{{ installation.namespace }}"
      manifests:
        - my-manifests
      wait: true
      outputs:
        - name: SVC_IP_ADDRESS
          namespace: "{{ installation.namespace }}"
          resourceType: service
          resourceName: "svc-{{ installation.name }}"
          jsonPath: "{.spec.clusterIP}"
```

**Additional Context**
I believe this only affects the porter Manifest.

**Risks/Concerns**
The proposal redefines `installation.name`, which may break some existing bundles. The alternative is to keep that as is (composite form) and create a new variable. But I feel that would be more confusing in the long run.

Do we need to vet this proposal to the community as a breaking change?

**Implementation details**
Pretty trivial and isolated changes.

# Checklist
- [ ] An announcement of this proposal has been sent to the Porter mailing list: https://porter.sh/mailing-list

Jemgoss:
Signed-off-by: Jeremy Goss <jeremy.goss@hpe.com>

# What does this change
Adds a new template variable `installation.namespace` and strips namespace/ from `installation.name`.

# What issue does it fix
Closes #2000.

# Notes for the reviewer
As far as I can tell, none of the example apps assume installation.name contains a namespace or / character, so it should be safe to make this change.

# Checklist
- [x ] Did you write tests?
- [x ] Did you write documentation?
- [ x] If this is your first pull request, please add your name to the bottom of our [Contributors][contributors] list. Thank you for making Porter better! 🙇‍♀️

# Reviewer Checklist

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -62,3 +62,4 @@ and we will add you. **All** contributors belong here. 💯
 * [Mahendra Bishnoi](https://github.com/mahendrabishnoi2)
 * [Yingrong Zhao](https://github.com/VinozzZ)
 * [Saksham Sharma](https://github.com/sakkshm26)
+* [Jeremy Goss](https://github.com/Jemgoss)
--- a/pkg/runtime/runtime-manifest.go
+++ b/pkg/runtime/runtime-manifest.go
@@ -88,8 +88,20 @@ func (m *RuntimeManifest) loadBundle() error {
 	return nil
 }
 
+func (m *RuntimeManifest) GetInstallationNamespace() string {
+	before, _, found := strings.Cut(m.Getenv(config.EnvInstallationName), "/")
+	if found {
+		return before
+	}
+	return ""
+}
+
 func (m *RuntimeManifest) GetInstallationName() string {
-	return m.Getenv(config.EnvInstallationName)
+	before, after, found := strings.Cut(m.Getenv(config.EnvInstallationName), "/")
+	if found {
+		return after
+	}
+	return before
 }
 
 func (m *RuntimeManifest) loadDependencyDefinitions() error {
@@ -225,6 +237,7 @@ func (m *RuntimeManifest) buildSourceData() (map[string]interface{}, error) {
 
 	inst := make(map[string]interface{})
 	data["installation"] = inst
+	inst["namespace"] = m.GetInstallationNamespace()
 	inst["name"] = m.GetInstallationName()
 
 	bun := make(map[string]interface{})
--- a/pkg/runtime/runtime-manifest.go
+++ b/pkg/runtime/runtime-manifest.go
@@ -89,19 +89,11 @@ func (m *RuntimeManifest) loadBundle() error {
 }
 
 func (m *RuntimeManifest) GetInstallationNamespace() string {
-	before, _, found := strings.Cut(m.Getenv(config.EnvInstallationName), "/")
-	if found {
-		return before
-	}
-	return ""
+	return m.Getenv(config.EnvPorterInstallationNamespace)
 }
 
 func (m *RuntimeManifest) GetInstallationName() string {
-	before, after, found := strings.Cut(m.Getenv(config.EnvInstallationName), "/")
-	if found {
-		return after
-	}
-	return before
+	return m.Getenv(config.EnvPorterInstallationName)
 }
 
 func (m *RuntimeManifest) loadDependencyDefinitions() error {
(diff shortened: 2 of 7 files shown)


CONTRIBUTORS.md
```diff
@@ -61,4 +61,5 @@ and we will add you. **All** contributors belong here. 💯
 * [Avinash Upadhyaya](https://github.com/avinashupadhya99)
 * [Mahendra Bishnoi](https://github.com/mahendrabishnoi2)
 * [Yingrong Zhao](https://github.com/VinozzZ)
-* [Saksham Sharma](https://github.com/sakkshm26)
\ No newline at end of file
+* [Saksham Sharma](https://github.com/sakkshm26)
+* [Jeremy Goss](https://github.com/Jemgoss)
```
carolynvs, line 65: 🎉

pkg/runtime/runtime-manifest.go
```diff
@@ -88,8 +88,20 @@ func (m *RuntimeManifest) loadBundle() error {
 	return nil
 }
 
+func (m *RuntimeManifest) GetInstallationNamespace() string {
+	before, _, found := strings.Cut(m.Getenv(config.EnvInstallationName), "/")
```
carolynvs, line 92: I think it would be helpful to make it easy for people to access the name and namespace without having to use the template language or parse the CNAB_INSTALLATION environment variable.

How about we introduce `PORTER_INSTALLATION_NAME` and `PORTER_INSTALLATION_NAMESPACE` that have the values we want, so that at runtime we don't need to parse the CNAB_INSTALLATION variable?

The place to populate those environment variables is in this function below. You can make a new function similar to `r.AddFiles` which can then set `op.Env` using `args.Installation`:

https://github.com/getporter/porter/blob/fe0146b9c8a330eeb63a265576b1e7f7eeba6a4e/pkg/cnab/provider/action.go#L51-L57

Jemgoss: Yes, this makes sense. Then we don't have to change the existing definition of installation.name. And as you mention, the values are available directly to action scripts without having to pass them in through template vars. I'll go ahead and update this PR, although its name is no longer accurate. Should I close this and create a new PR?

Jemgoss: I've updated the PR with those suggestions. As it stands, the relevant env in the runtime is:
```
CNAB_INSTALLATION_NAME=inst_namespace/inst_name
PORTER_INSTALLATION_NAMESPACE=inst_namespace
PORTER_INSTALLATION_NAME=inst_name
```
And template variables installation.namespace and installation.name are derived from PORTER_INSTALLATION_NAMESPACE and PORTER_INSTALLATION_NAME.

carolynvs:
Hey @Jemgoss, it looks like you had some trouble updating your branch with the latest changes in the release/v1 branch (there are commits from release/v1 showing up in this PR as changes that you made which isn't right). Let me know if you need help getting your branch fixed up!

Jemgoss:
Yeah, don't know why that happened, except that I was trying to run units tests with the latest changes from release/v1. Maybe best to rebase off release/v1.

carolynvs (approved):
Looks great, thank you for spotting the problem and submitting a PR! 👍 

I have created a follow-up issue, #2006, to provide users with a replacement if they were using CNAB_INSTALLATION to generate unique resource names.

Issue #2009, opened 2022-04-06
Switch from pkg/errors package to Go's errors package


carolynvs:
Now that go supports erorrs.Is and other nice functionality that we previously relied on pkg/errors to do for us, let's get off of that and use the built-in errors package.

As I work on new features, I'll update as I go but it would be great if we could do a sweep get everything consistent.

joshuabezaleel:
1. The functionality that I found so far on the codebase using `pkg/errors` that can be substituted with `errors` package are `Is` and `New`. Does this mean that we will be using the built-in errors package for those two and will still be using `pkg/errors` for other that hasn't supported e.g. `Wrap()` and `WithStack()`? 
2. If the answer to no 1 is yes, supposed that we have a file where there are `Is()` function which we will use built-in `errors` and `Wrap()` function which we will use `pkg/errors`. Since those two have the same name, is it safe to alias the `pkg/errors` to other on import e.g.
```
import (
   "errors"
   
   pkgerrors "github.com/pkg/errors"
)

func someFunction() {
   ...
   errors.Is(
   ...
   pkgerrors.Wrap(
}
```

carolynvs:
@joshuabezaleel For now, please hold off on taking on any more issues until your current PRs are merged. I want to give other people a chance to take on these types of issues since they are great for people new to Porter. Plus having PRs open for a long period of time really increases the likelihood that the merge conflicts will get real bad.

joshuabezaleel:
@carolynvs Ah yes, I was getting carried away since the issue was on the top recent and it piqued my interest. I thought at first it will already go for Jeremy to tackle on. I'll get back and try to finish the other of my PRs by this week. So sorry and thank you for the kind reminder, Carolyn!

PR #2026, opened 2022-04-20
Instrument plugins

### LINKED ISSUE #1984: Trace plugin execution
Currently, during a bundle execution, there are several components that the code pass through, such as Porter CLI, Porter Mixin, Porter Plugin, etc. Even though Porter isn't an distributed system, it would be helpful to see which component that  traces are being generated from.
We can define each component as a service in tracing. This will allow us to not only have clear view in the UI which layer a trace is coming from, it also gives us jaeger built-in filtering and sorting ability for traces.

carolynvs:
# What does this change
[Plumb context.Context for the plugins](https://github.com/getporter/porter/pull/2026/commits/1c5e3edd8cd5b1f3c062ba585dfd330f6f6d9633)

I wanted to add tracing to the plugins before working on the migration
(which will do being extreme plugin shenanigans so we'll need the extra
logs there). This is a first pass to add context.Context to all
functions that lead to the plugins.

[Fix integration test use of namespace](https://github.com/getporter/porter/pull/2026/commits/33c4676da036e376acb3817bb919d208bbf62cfc)

The test wasn't setting namespace and was creating data in the global
namespace.

[Add plugin adapter for secrets](https://github.com/getporter/porter/pull/2026/commits/e8253b977c04cbfe6c66138ca63985eddc235b75)

[Instrument plugin execution](https://github.com/getporter/porter/pull/2026/commits/961ceb0fa163e9e133c753a6604a66530b68799b)

This has lots of semi-related fixes for everything required to get a
plugin instrumented.

* Add support for creating secondary tracers with a different service name
set when interacting with the plugins. In Jaeger, you will see a span
tree for "porter", and now another span tree for each plugin that it
interacts with, e.g. "mongodb", or "azure.storage.blob".
* Support grpc in the storage plugin. Not sure yet if we are keeping
  net/rpc. Only grpc plugins can be instrumented, net/rpc doesn't have a
  mechanism for propogating the context.
* Change storage protocol to include context.Context, and to use
  concrete types instead of interface{} when sending filters, sorts,
  etc.
* Run all plugins through the hashicorp/go-plugin framework. I was
  running internal plugins in-process which meant that I didn't notice
  when we broke external storage plugins. It also was more work to
  instrument a new separate service name when I had to do it both for
  external and internal plugins. This makes the code paths the same
  regardless of who wrote the plugin.
* Improve error handling when serving internal plugins. All errors must
  be logged with hclog, and we must not write to stdout/err as that
  causes the connection to the plugin to fail.
* Dump logs from the plugins into a span as a fallback mechanism to
  collect info from an uninstrumented plugin.
* Set porter into "plugin mode" and have porter log with hclog

# What issue does it fix
Closes #1984

# Notes for the reviewer
Definitely go through this commit by commit so you can quickly walk through all the files that added context.Context without your eyes glazing over.

# Checklist
- [x] Did you write tests?

# Reviewer Checklist


pkg/secrets/plugins/filesystem/store.go
```diff
@@ -1,138 +1,132 @@
...
-}
-
 // Store implements an file system secrets store for testing and local
 // development.
 type Store struct {
-	config    Config
-	logger    hclog.Logger
-	hostStore cnabsecrets.Store
-	storage   aferox.Aferox
+	config    *config.Config
```
carolynvs, line 28: I reworked this a bit to be consistent with the other plugins:

1. Each method is responsible for calling connect
2. Moved the creation of the secrets folder to connect
3. Passed a config struct in the constructor to simplify what we are passing (it already has context, filesystem, logs, etc)

pkg/secrets/plugins/in-memory/store.go
```diff
@@ -36,36 +28,13 @@ func NewStore() *Store {
 	return s
 }
 
-func (s *Store) Connect() error {
```
carolynvs, line 39: I realized that a lot of this code was unused and leftover from when it was repurposed from the older storage protocol.

pkg/storage/plugin_adapter.go
```diff
@@ -21,88 +22,61 @@ var _ Store = PluginAdapter{}
...
-
-	return a.plugin.Count(opts.ToPluginOptions(collection))
+func (a PluginAdapter) Count(ctx context.Context, collection string, opts CountOptions) (int64, error) {
+	return a.plugin.Count(ctx, opts.ToPluginOptions(collection))
 }
 
-func (a PluginAdapter) Find(collection string, opts FindOptions, out interface{}) error {
-	err := a.Connect()
+func (a PluginAdapter) Find(ctx context.Context, collection string, opts FindOptions, out interface{}) error {
+	rawResults, err := a.plugin.Find(ctx, opts.ToPluginOptions(collection))
```
carolynvs, line 60: All structs that have Connect are responsible for calling it themselves. It's just the caller who is responsible for calling Close() if defined.

pkg/storage/plugins/mongodb/mongodb.go
```diff
@@ -192,58 +229,102 @@ func (s *Store) buildFindOptions(opts plugins.FindOptions) *options.FindOptions
...
 
-	return query
+	return query, nil
 }
 
-func (s *Store) Insert(opts plugins.InsertOptions) error {
-	cxt, cancel := context.WithTimeout(context.Background(), s.timeout)
-	defer cancel()
+func (s *Store) Insert(ctx context.Context, opts plugins.InsertOptions) error {
+	ctx, span := tracing.StartSpan(ctx)
```
carolynvs, line 236: The plugins don't strictly have to create spans, the grpc server will create them automatically. We can tweak this to either include more info or just add attributes to the parent span later.

pkg/storage/plugins/storage_plugin.go
```diff
@@ -1,15 +1,11 @@
...
-const PluginInterface = "storage"
-
-// StoragePlugin is the interface used to wrap a storage plugin.
-// It is not meant to be used directly.
-type StoragePlugin interface {
-	plugins.Plugin
-	StorageProtocol
-}
+	// PluginProtocolVersion is the currently supported plugin protocol version for storage.
+	PluginProtocolVersion = 3
```
carolynvs, line 10: This change incremented the protocol version for both plugins

pkg/storage/plugins/storage_protocol.go
```diff
@@ -121,11 +123,11 @@ type PatchOptions struct {
 
 	// Query is a query filter document
 	// See https://docs.mongodb.com/manual/core/document/#std-label-document-query-filter
-	QueryDocument interface{}
+	QueryDocument bson.M
 
 	// Transformation is set of instructions to modify matching
 	// documents.
-	Transformation interface{}
+	Transformation bson.D
```
carolynvs, line 130: I switched to concrete types so that it was easier to write converters for the grpc client/server.

pkg/storage/plugins/mongodb_docker/store.go
```diff
@@ -154,15 +230,15 @@ func EnsureMongoIsRunning(c *portercontext.Context, container string, port strin
...
 		default:
 			conn := mongodb.NewStore(c, mongoPluginCfg)
 			err := conn.Connect(ctx)
 			if err == nil {
 				return conn, nil
-			} else if c.Debug {
-				fmt.Fprintln(c.Err, errors.Wrapf(err, "mongodb isn't ready yet"))
 			}
+
+			time.Sleep(time.Second)
```
VinozzZ, line 241: Why do we need a Sleep here?

pkg/cnab/provider/action.go
```diff
@@ -49,10 +49,10 @@ type ActionArguments struct {
 	PersistLogs bool
 }
 
-func (r *Runtime) ApplyConfig(args ActionArguments) action.OperationConfigs {
+func (r *Runtime) ApplyConfig(args ActionArguments, ctx context.Context) cnabaction.OperationConfigs {
```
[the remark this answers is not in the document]
carolynvs: oops! I'm so glad you caught that. I used my IDE to refactor and add that parameter everywhere and must have missed bumping it to the first parameter.

pkg/storage/plugins/mongodb_docker/store.go
```diff
@@ -65,6 +65,8 @@ func (s *Store) Close() error {
 }
 
 func EnsureMongoIsRunning(c *portercontext.Context, container string, port string, dataVol string, dbName string, timeoutSeconds int) (*mongodb.Store, error) {
+	ctx := context.TODO()
```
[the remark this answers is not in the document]
carolynvs: context.Background() works well when you don't have a context and don't anticipate having one in the future. context.TODO is used when later you _may_ get a context, and this helps you find places where you can do further "context plumbing" or adding context to a chain of functions.

Later I can search for context.TODO and find more places that need additional work.

So main will use context.Background, along with any tests, but the rest of the code should always be passing a context, and not creating one themselves.

pkg/plugins/pluggable/loader.go
```diff
@@ -25,138 +30,164 @@ type PluginLoader struct {
...
+	// Start the plugin
+	span.Debug("Connecting to plugin", attribute.String("plugin-command", strings.Join(pluginCommand.Args, " ")))
 	rpcClient, err := client.Client()
 	if err != nil {
 		cleanup()
 		if stderr := errbuf.String(); stderr != "" {
 			err = errors.Wrap(errors.New(stderr), err.Error())
 		}
-		return nil, nil, errors.Wrapf(err, "could not connect to the %s plugin", l.SelectedPluginKey)
+		return PluginConnection{}, span.Error(errors.Wrapf(err, "could not connect to the %s plugin", l.SelectedPluginKey))
```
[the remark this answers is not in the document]
carolynvs: I am calling cleanup (which closes the pipe's reader/writer) two lines above here.

pkg/plugins/server.go
```diff
@@ -1,18 +1,66 @@
...
+// ServeMany plugins that the client will select by named interface.
+func ServeMany(c *portercontext.Context, pluginMap map[int]plugin.PluginSet) {
 	plugin.Serve(&plugin.ServeConfig{
-		HandshakeConfig: HandshakeConfig,
-		Plugins:         pluginMap,
+		HandshakeConfig:  HandshakeConfig,
+		VersionedPlugins: pluginMap,
+		GRPCServer: func(opts []grpc.ServerOption) *grpc.Server {
+			opts = append(opts,
+				// These handlers are called from left to right. The right-most handler is the one that calls the actual implementation
```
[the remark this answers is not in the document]
carolynvs: Yeah same thing really since it's a comma separated list of handlers, it's just on each line separately because it was getting long.

pkg/porter/internal_plugins.go
```diff
@@ -0,0 +1,147 @@
...
+		return errors.New("Multiple positional arguments were specified but only one, KEY is expected")
+	}
+
+	o.Key = args[0]
+	return nil
+}
+
+func (o *RunInternalPluginOpts) Validate() error {
+	if o.Key == "" {
+		return fmt.Errorf("no plugin key was specified")
```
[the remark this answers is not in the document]
carolynvs: We have an open issue ( #2009 ) to move off of the pkg/errors package and move to native errors. As I am working on the codebase, as I spot errors that don't rely on pkg/errors.Cause, I am switching them to native errors so that we can move to errors.Is eventually throughout.

We'll still need a sweep eventually of the entire codebase, but in the meantime I don't want to add to the tech debt.

pkg/secrets/plugins/in-memory/store.go
```diff
@@ -37,36 +23,13 @@ func NewStore() *Store {
...
-
-func (s *Store) Resolve(keyName string, keyValue string) (string, error) {
+func (s *Store) Resolve(ctx context.Context, keyName string, keyValue string) (string, error) {
 	_, ok := s.Secrets[keyName]
 	if !ok {
 		s.Secrets[keyName] = make(map[string]string, 1)
 	}
 
-	if keyName == secrets.SourceSecret {
+	if keyName == "secret" {
```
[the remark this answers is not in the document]
carolynvs: It creates a circular dependency unfortunately

pkg/storage/plugins/mongodb/plugin.go
```diff
@@ -1,28 +1,38 @@
...
 
-func NewPlugin(cxt *portercontext.Context, pluginConfig interface{}) (plugins.StoragePlugin, error) {
+func NewPlugin(c *portercontext.Context, rawCfg interface{}) plugin.Plugin {
 	cfg := PluginConfig{
 		Timeout: 10,
 	}
-	if err := mapstructure.Decode(pluginConfig, &cfg); err != nil {
-		return nil, errors.Wrapf(err, "error decoding %s plugin config from %#v", PluginKey, pluginConfig)
+	if err := mapstructure.Decode(rawCfg, &cfg); err != nil {
+		panic(fmt.Errorf("error reading plugin configuration: %w", err))
```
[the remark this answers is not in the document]
carolynvs: The RunInternalPlugin function validates the plugin input before creating the plugins. So there isn't really a path that would lead to this error. I figured I would panic instead of ignoring the error entirely.

There isn't a way to handle the error if we run a plugin and send malformed config on STDIN. I could change the entire call path to bubble up the error, but we are already handling panics in main, so it wouldn't change the behavior anyway?

pkg/storage/plugins/mongodb_docker/store.go
carolynvs: I thought that it probably wasn't great to have a tight loop without any yields constantly checking the progress of mongodb starting up, which takes cycles away from ... mongo starting up. Perhaps a second isn't the right amount of time to sleep, but I do think we should sleep here for some amount of time.

pkg/plugins/plugins.go
```diff
@@ -12,11 +13,11 @@ import (
 type Plugin interface {
 	// Connect establishes a connection to the plugin.
 	// Safe to call multiple times, the existing connection is reused.
-	Connect() error
+	Connect(ctx context.Context) error
 
 	// Close the connection to the plugin.
 	// Safe to call multiple times.
-	Close() error
+	Close(ctx context.Context) error
```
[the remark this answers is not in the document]
carolynvs: Thanks for calling this out. I went through and standardized how I'm using Close.

pkg/secrets/plugins/in-memory/store.go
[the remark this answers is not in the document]
carolynvs: I'm not going to change it now but here's some context about how I made this mess in Porter. :joy: 

I made it easier to create circular dependencies because I am putting test helpers in official packages. The test helpers often reference a particular implementation of an interface, which can lead to a circular dep between that pkg with the implementation and the pkg with the helper. 

If I had created a subpackage for test helpers e.g. pkg/secrets/tests that would have helped avoid this I think (I haven't tried that out though).

I originally didn't make a separate package because in enough cases, the test helper was accessing unexported fields on the original type. If at some point we wanted to review how we write tests, and try out some improvements, I'd totally be up for that.

pkg/storage/plugins/mongodb/plugin.go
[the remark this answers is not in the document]
carolynvs: yeah okay this was annoying me too, and this was the push I needed to fix it properly. I've updated the code to return errors

pkg/plugins/pluggable/loader.go
```diff
@@ -25,138 +30,164 @@ type PluginLoader struct {
...
 	}
 
-	cleanup, err = l.setUpDebugger(client, cleanup)
+	cleanup, err = l.setUpDebugger(ctx, client, cleanup)
 	if err != nil {
-		return nil, nil, errors.Wrap(err, "could not set up debugger for plugin")
+		cleanup()
+		return PluginConnection{}, span.Error(errors.Wrap(err, "could not set up debugger for plugin"))
```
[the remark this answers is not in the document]
carolynvs: Same as before, cleanup closes the pipe

pkg/storage/pluginstore/grpc.go
```diff
@@ -0,0 +1,407 @@
...
+}
+
+// ConvertFloatToInt works around a weirdness in how numbers are represented
+// by structpb.Value, where integer values are stored in float64. When we
+// deserialize from protobuf, this walks the specified value, finds ints
+// that were encoded as floats, and converts them back to ints.
+func ConvertFloatToInt(src interface{}) interface{} {
+	switch tv := src.(type) {
+	case float64:
+		intVal := int(tv)
```
VinozzZ, line 296: I wonder if it's better to use int64 here just in case for some reason the float64 value doesn't fit into a int32

pkg/storage/query_test.go
```diff
@@ -68,7 +69,18 @@ func TestFindOptions_ToPluginOptions(t *testing.T) {
...
+
+func TestFoo(t *testing.T) {
+	x := bson.D{
+		{Key: "foo", Value: "bar"},
+		{Key: "a", Value: "b"},
+	}
+
+	result, err := bson.Marshal(x)
+	require.NoError(t, err)
+	fmt.Println(string(result))
```
VinozzZ, line 85: Is this intended to just print out the value?

tests/integration/pluggable_integration_test.go
```diff
@@ -51,11 +53,9 @@ func TestPlugins_CatchStderr(t *testing.T) {
...
 		}}
 
-		createInternalPlugin := func(string, interface{}) (plugins.Plugin, error) {
-			return nil, nil
-		}
-		ll := NewPluginLoader(c.Config, createInternalPlugin)
-		_, _, err = ll.Load(cfg)
+		ll := pluggable.NewPluginLoader(c.Config)
+		conn, err := ll.Load(context.Background(), cfg)
+		conn.Close()
```
VinozzZ, line 58: Should this be in a defer instead?

pkg/storage/plugins/mongodb_docker/store.go
VinozzZ: Maybe using `time.Ticker` is better suited for implementing a backoff here? I know this is minor but using a time.Sleep here may add delay for the 10 second time out we set for trying to connect to the db

pkg/secrets/plugins/in-memory/store.go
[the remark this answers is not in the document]
VinozzZ: Thanks for the clarification! Yes, I have noticed our tests are in the same package as the actual code. It may help us to catch more things if we have a separate test package too. I would love to slowly get to there as well

pkg/storage/grpc_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+// TODO: test helpers in grpc.go
+func TestNewStructList(t *testing.T) {
+	t.Skip()
+}
+
+func TestGRPCTelemetry(t *testing.T) {
```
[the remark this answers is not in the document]
carolynvs: oops that file was not meant to be added. I was just using it to run stuff through the debugger at one point, I'll remove it.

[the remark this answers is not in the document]
carolynvs: I hope you enjoyed my todo list in that same file! 😂 


```
/*
1. Finish tracing in mongo plugin
2. ensure we are tracing which plugin is running, including version
3. check that all functions trace (e.g. find, ensure index)
4. look for todo comments
5. check for unnecessary changes to our contracts/files
6. add grpc tests to make sure data goes around properly?
7. add otel test with mock tracer?
8. add logger for plugin that emits hclog
9. emit the correlation id for the traces emitted in a plugin before it handles  message
10. how will net/rpc work? do we keep it? (less to test if we don't)
11. update existing plugins to use grpc, telemetry
12. do secrets
make sure that the secrets plugins for azure and hashicorp are updated
fork mchora's helm3
remove net/rpc
look for context.TODO()
make sure code gen for protobuf is happening, and tools automatically installed
*/

```

pkg/storage/pluginstore/grpc.go
```diff
@@ -0,0 +1,407 @@
...
+	case bson.M:
+		raw := make(map[string]interface{}, len(t))
+		for k, v := range t {
+			raw[k] = ConvertBsonToPrimitives(v)
+		}
+		return raw
+	}
+
+	// Hope and pray we aren't sending something that protobuf doesn't like
+	// If we do, add another case above to handle that type
```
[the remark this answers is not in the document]
carolynvs: I'll remove that comment as it's misleading. Types like float, bool, string, etc all fall through to that and are returned unconverted. I was more referring to some other permutation of list, map, etc that I didn't think of that we'd need to traverse and convert.

pkg/storage/query_test.go
carolynvs: Sorry that wasn't supposed to be committed, I was just trying something out and wrote a test to help me see what it serialized as. 🤦‍♀️

tests/integration/pluggable_integration_test.go
carolynvs: We aren't doing anything with the connection in the test so there's no need to put off closing the connection.

carolynvs:
Fixed! 🤞

pkg/storage/grpc_test.go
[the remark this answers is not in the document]
VinozzZ: Ahaha, I was thinking that how nice we are "committing" to this todo list

pkg/storage/plugins/mongodb_docker/store.go
carolynvs: I'm not sure I'm following why ticker would work fundamentally different than the loop + sleep here?

It's only sleeping 50 milliseconds at each iteration of the loop (to give mongo time to start up). During sleep, the goroutine yields so it's not keeping the CPU busy. Is there something I'm missing?

pkg/plugins/pluggable/loader.go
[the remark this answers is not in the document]
VinozzZ: You are right. Idk why I didn't see it the first time around

pkg/storage/plugins/mongodb_docker/store.go
VinozzZ: Yes, if I understand correctly, if the time out is about to happen right before we hit the sleep call. Then we would be waiting for another 50 milliseconds. If we used ticker in the `select` statement, the select would immediately return since the done channel would be closed. The code would look like this
```golang
d := time.NewTicker(50 * time.millisecond)
defer d.Stop()
for {
     select {
	case <-timeout.Done():
		return nil, span.Error(errors.New("timeout waiting for local mongodb daemon to be ready"))
         case <-d.C:
		conn := mongodb.NewStore(c, mongoPluginCfg)
		err := conn.Connect(ctx)
		if err == nil {
			return conn, nil
		}
      }
 }
```
The performance here probably isn't critical. I would be fine to merge as it is

carolynvs: Oh! Thanks for explaining. Yeah I can switch to a ticker so that we absolutely stop waiting when we hit our deadline.

pkg/storage/plugins/mongodb_docker/store.go
```diff
@@ -222,19 +222,19 @@ func EnsureMongoIsRunning(ctx context.Context, c *portercontext.Context, contain
 		Timeout: timeoutSeconds,
 	}
 	timeout, cancel := context.WithTimeout(ctx, 10*time.Second)
+	tick := time.NewTicker(50 * time.Millisecond)
+
```
VinozzZ, line 226:
```suggestion
defer tick.Stop()
```

VinozzZ (approved):
Yahoo! looks good to me

PR #2043, opened 2022-04-28
Gracefully handle non-string custom values

carolynvs:
# What does this change
When converting custom values set on the command-line during build into build args, use CNAB's string representation for non-string types, like we do for parameters.

# What issue does it fix
This is follow-up to #1900 

# Notes for the reviewer
N/A

# Checklist
- [x] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/build/buildkit/buildx.go
+++ b/pkg/build/buildkit/buildx.go
@@ -3,11 +3,14 @@ package buildkit
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
+
+	"get.porter.sh/porter/pkg/cnab"
 
 	"get.porter.sh/porter/pkg/build"
 	"get.porter.sh/porter/pkg/config"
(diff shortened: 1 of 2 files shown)


carolynvs:
@joshuabezaleel I had an idea for how to deal with converting custom values to strings when we turn them into build arguments, let me know what you think!

VinozzZ (approved):

joshuabezaleel:
@carolynvs Appreciate the shoutout on this particular PR and the fixes, Carolyn! The `WriteParameterToString` function was neat, I should've found it earlier. 

While we were able to particularly parse the other datatype value to `string`, don't we still need to wait #316 for the value to be able to be used because it will still be held back by the template if we want to use the value for its original datatype e.g. `boolean`? Since it will still be rendered surrounded by double quote in the end (`"true"` and `"false"` for what should only be `true` and `false`). 

I'm using the `whalesay` bundle as example (as always 🐳 ) and tried to inject the `rm` boolean property on `docker run` with custom data to demonstrate the question above.
image
image
image

The parsing part is solve but I a assume we still need to specify to the user that the `--custom` flag still only works with only `string` datatype.

carolynvs:
I agree this doesn't fix the existing templating with yaml problem in that other issue. It was never meant to fix that problem.

This patch fixes a problem with how we are parsing custom values and converting them to build arguments to pass to docker when building the bundle. I didn't quite catch this during the review but immediately noted it when using the patch after merging.

```
$ porter build --custom favorite-number=6
Defaulting the type of output passwd to file
Copying porter runtime ===>
Copying mixins ===>
Copying mixin exec ===>

Writing Dockerfile =======>
2022-04-29T10:11:18.424-0500	info	Building invocation image
2022-04-29T10:11:18.428-0500	error	unable to build CNAB invocation image: Unknown type 6: %!t(int=6)
```

You get the same error if you even define a custom value that isn't a string (even if you don't pass it on the command-line).

```yaml
name: mybundle

custom:
  favorite-number: 5
```

With this patch we can make that work without the panic, which I really should have caught that this PR introduced a regression in defining custom values.

joshuabezaleel:
Ah I see the difference now, got it. Thank you lots for the fixes, Carolyn! 🎉

PR #2049, opened 2022-05-02
Revert "Use a symlink in a linux install"

### LINKED ISSUE #2048: Can't update porter on linux with install script anymore for stable (v0.38)
We recently changed the linux installation script to create a symlink from porter to the runtime in #2032. This caused a regression in our installation scripts though because it only works for installing in a fresh PORTER_HOME. When the user tries to upgrade, they get the following error:

```
# curl -L https://cdn.porter.sh/canary/install-linux.sh | bash
Installing porter@canary to /root/.porter from https://cdn.porter.sh
ln: failed to create symbolic link '/root/.porter/runtimes/porter-runtime': File exists
```

The old script overwrote any existing binaries which is why it worked previously.

carolynvs:
This reverts commit 9a1934665d8b995277be0a5a1b49c5a2987cb459.

Closes #2048

If we want to make this change, let's redo it on the release/v1 branch, and make sure to run the install script tests with it.

### DIFF
--- a/scripts/install/install-linux.sh
+++ b/scripts/install/install-linux.sh
@@ -19,7 +19,7 @@ mkdir -p $PORTER_HOME/runtimes
 
 curl -fsSLo $PORTER_HOME/porter $PORTER_MIRROR/$PORTER_PERMALINK/porter-linux-amd64
 chmod +x $PORTER_HOME/porter
-ln -s $PORTER_HOME/porter $PORTER_HOME/runtimes/porter-runtime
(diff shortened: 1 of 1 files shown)


VinozzZ (approved):

carolynvs:
I had trouble running the install check using comments, so here is the test run that's failing on main that I manually kicked for this branch:

https://dev.azure.com/getporter/porter/_build/results?buildId=3812&view=results

PR #2052, opened 2022-05-03
Consolidate storage-related packages into pkg/storage

carolynvs:
# What does this change
This consolidates the pkg/credentials pkg/parameters, and pkg/sanitizer packages into the storage package.

This helps avoid a lot of circular dependencies that get introduced because these packages all rely on each other in
various ways (they are just views on top of the storage layer).

# What issue does it fix
I'm working on the data migration and keep running into circular dependencies. I don't think these packages really warrant being split up, and it will make it easier for us to move this code around.

# Notes for the reviewer

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist


carolynvs:
I spoke with Yingrong and the sanitizer package may better fit in the pkg/secrets package.

carolynvs:
@VinozzZ I checked and moving sanitizer into the secrets package creates a circular dependency between pkg/storage and pkg/secrets. I get that the sanitizer moves data into secrets, but I think of it as belonging with storage because it's sanitizing the data we store.

pkg/porter/list_test.go
```diff
@@ -4,19 +4,19 @@ import (
...
 	"get.porter.sh/porter/pkg/secrets"
+	"get.porter.sh/porter/pkg/storage"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
 func TestNewDisplayInstallation(t *testing.T) {
 	t.Run("installation has been installed", func(t *testing.T) {
-		cp := claims.NewTestClaimProvider(t)
+		cp := storage.NewTestClaimProvider(t)
```
VinozzZ, line 16: Do we want to also change the `NewTestClaimProvider` to `NewTestStorageProvider` ?

pkg/storage/claim_storage_provider.go
```diff
@@ -1,9 +1,11 @@
...
 
-import "context"
+import (
+	"context"
+)
 
-// Provider is an interface for interacting with Porter's claim data.
-type Provider interface {
+// ClaimProvider is an interface for interacting with Porter's claim data.
+type ClaimProvider interface {
```
VinozzZ, line 8: Do you mind to explain why we use `claim` here?

pkg/storage/credential_store.go
```diff
@@ -1,32 +1,43 @@
...
-var _ Provider = &CredentialStore{}
+var _ CredentialSetProvider = &CredentialStore{}
 
 const (
 	CollectionCredentials = "credentials"
 )
 
-// CredentialStore provides access to credential sets by instantiating plugins that
-// implement CRUD storage.
+// CredentialStore provides access to work with Porter
```
VinozzZ, line 21: Maybe something is missing after the world `access` ?

pkg/storage/credential_store.go
```diff
@@ -1,32 +1,43 @@
...
+// credential sets, usually referred to as "credentials" as a shorthand.
+//
+// Credential Sets define mappings from a credential needed by a bundle to where
+// to look for it when the bundle is run. For example: Bundle needs Azure
+// storage connection string and it should look for it in an environment
+// variable named `AZURE_STORATE_CONNECTION_STRING` or a key named `dev-conn`.
+//
+// Porter discourages storing the value of the credential directly, though it
+// it is possible. Instead Porter encourages the best practice of defining
```
VinozzZ, line 30:
```suggestion
// is possible. Instead Porter encourages the best practice of defining
```

pkg/storage/credential_store.go
```diff
@@ -1,32 +1,43 @@
...
 const (
 	CollectionCredentials = "credentials"
 )
 
-// CredentialStore provides access to credential sets by instantiating plugins that
-// implement CRUD storage.
+// CredentialStore provides access to work with Porter
+// credential sets, usually referred to as "credentials" as a shorthand.
+//
+// Credential Sets define mappings from a credential needed by a bundle to where
```
VinozzZ, line 24: I wonder if this documentation applies better with the `CredentialSet` struct instead of here

VinozzZ (commented):
LGTM, just some couple nit questions

pkg/porter/list_test.go
carolynvs: That is in the follow-up PR #2053

pkg/storage/claim_storage_provider.go
VinozzZ: That makes perfect sense. I just wanted to make sure I understand the scope of this PR. Thanks for clarifying it!

PR #2060, opened 2022-05-06
pkg/storage: fix connection leak

VinozzZ:
# What does this change

After running `porter installation list`, I was not able to run `mage build install` successfully due to `Error: open /home/yingrong/.porter/porter: text file busy`
I checked running processes on my machine and found that the storage plugin is still running after porter has exited.
This PR fixes the Close function on the store to match with the storage.Store interface

_For example if it introduces a new command or modifies a commands output, give an example of you running the command and showing real output here._

# What issue does it fix

# Notes for the reviewer

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/storage/plugins/mongodb/mongodb.go
+++ b/pkg/storage/plugins/mongodb/mongodb.go
@@ -201,6 +201,7 @@ func (s *Store) Find(ctx context.Context, opts plugins.FindOptions) ([]bson.Raw,
 	if err != nil {
 		return nil, span.Error(err)
 	}
+	defer cur.Close(ctx)
 
 	var results []bson.Raw
 	for cur.Next(cxt) {
--- a/pkg/storage/pluginstore/store.go
+++ b/pkg/storage/pluginstore/store.go
@@ -3,6 +3,7 @@ package pluginstore
 import (
 	"context"
 	"fmt"
+	"io"
 
 	"get.porter.sh/porter/pkg/config"
 	"get.porter.sh/porter/pkg/plugins/pluggable"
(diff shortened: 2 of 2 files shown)


pkg/storage/pluginstore/store.go
```diff
@@ -11,7 +12,12 @@ import (
 	"go.mongodb.org/mongo-driver/bson"
 )
 
-var _ plugins.StorageProtocol = &Store{}
+type pluginCloser interface {
```
carolynvs, line 15: Since we aren't using this interface anywhere other than to assert that we are implementing close, how about we remove it and add an extra interface check like so

```go
var _ io.Closer = &Store{}
var _ plugins.StorageProtocol = &Store{}
```

pkg/storage/plugins/mongodb/mongodb.go
```diff
@@ -201,6 +201,7 @@ func (s *Store) Find(ctx context.Context, opts plugins.FindOptions) ([]bson.Raw,
 	if err != nil {
 		return nil, span.Error(err)
 	}
+	defer cur.Close(ctx)
```
carolynvs, line 204: I wonder if there's an IDE plugin that will catch when I forgot to call close on variables that I create? This needs to exist. 😁

carolynvs (changes requested):

pkg/storage/pluginstore/store.go
VinozzZ: Oh, that's a good idea. I changed it to your suggestion

pkg/storage/plugins/mongodb/mongodb.go
VinozzZ: maybe someone wrote a linter for it but I can see how it can be hard to detect it

carolynvs: I found two linters for specific libraries that check if Close was called, but not a general purpose "did I close things that I created" check

https://github.com/golangci/awesome-go-linters

carolynvs (approved):
I'm so glad that you caught this! 💯

PR #2061, opened 2022-05-06
Fix enabling telemetry

carolynvs:
# What does this change
* The if statement was inverted so we weren't ever using the tracer.
* I was accidentally closing the tracer instead of ending the span in one spot.

# What issue does it fix
Fixing a regression caused by #2026 when I applied some review feedback and got the boolean logic flipflopped. I've added an integration test to verify that the tracer is a real one.

# Notes for the reviewer

# Checklist
- [x] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/tests/integration/telemetry_test.go
+++ b/tests/integration/telemetry_test.go
@@ -1,0 +1,50 @@
+//go:build integration
+
+package integration
+
+import (
+	"context"
+	"os"
+	"testing"
+	"time"
+
+	"get.porter.sh/porter/pkg/porter"
+	"get.porter.sh/porter/tests/tester"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// Validate that we can configure a live connection to a telemetry endpoint
+func TestTelemetrySetup(t *testing.T) {
+	test, err := tester.NewTest(t)
+	defer test.Close()
+	require.NoError(t, err, "test setup failed")
+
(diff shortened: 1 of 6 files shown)


tests/integration/telemetry_test.go
```diff
@@ -0,0 +1,50 @@
...
+	defer test.Close()
+	require.NoError(t, err, "test setup failed")
+
+	ctx := context.Background()
+	_, _, err = test.RunPorter("install", "otel-jaeger", "-r=ghcr.io/getporter/examples/otel-jaeger:v0.1.0", "--allow-docker-host-access")
+	require.NoError(t, err)
+	defer test.RunPorter("uninstall", "otel-jaeger", "--allow-docker-host-access")
+
+	// Wait until the collection should be up
+	time.Sleep(10 * time.Second)
```
VinozzZ, line 29: I don't have a immediate solution in mind but this maybe flaky in some situations

VinozzZ (approved):

tests/integration/telemetry_test.go
carolynvs: Agreed! I wanted to get _something_ to test that it's working or not for now and if I come up with a better way I'll update the test.

I tried doing a net dial but that passes just fine and then later the trace setup fails because the otel service isn't 100% ready. I feel like the best fix would be to have the porter bundle that installs the otel collection have logic and wait until the collector is up.

PR #2073, opened 2022-05-18
Fix plugin tracing again 😭

### LINKED ISSUE #2057: Add a test that checks telemetry setup is working
Let's add an integration test that runs an otel collector container, and checks that when we create a new porter instance, that it detects it when configured and creates a real tracer (not a no-op tracer).

carolynvs:
# What does this change
When we fixed recursive plugin loading in #2071, we did so by disabling config loading in the plugin. This was _almost_ right but not quite. While Porter does resolve the relevant config for the plugin and pass it on stdin, this does not provide some of the general configuration settings that can be used by the plugin such as logging and telemetry.

Specifically, our change caused a regression where the plugins didn't load the telemetry configuration and stopped sending trace data.

I have updated how configuration data is loaded for plugins:

1. When we start a plugin binary we grab relevant configuration settings and pass them as environment variables to the plugin.
2. When loading configuration for a plugin, we skip reading just the config file, but we still load configuration using environment values.
3. Added a new telemetry setting that redirects traces from the standard open telemetry collector to write them to a file instead (in PORTER_HOME/traces). This helped me finally write a regression test for sending trace data so that I don't accidentally break this again.

While I was working on this I found that we aren't handling close errors when we close porter's log file and we were missing a race condition where the plugin closes the log file before porter is done with it. I've split up the log and trace files for the plugins and porter so that they write to CORRELLATIONID.json for porter and
CORRELLATIONID-PLUGINKEY.json for the plugins.

# What issue does it fix
Restores plugin telemetry.

Closes #2057

# Notes for the reviewer

# Checklist
- [x] Did you write tests?

# Reviewer Checklist


vdice (approved):
Thanks for the context and breakdown of changes.  LGTM!

PR #2078, opened 2022-05-18
Support the interrupt signal SIGINT

### LINKED ISSUE #950: Graceful handling of canceled commands
We need to add plumbing and tweak some of our code to gracefully cancel when someone cancels a command, or an error occurs and we know immediately that we should stop executing other commands (for example when querying mixins and one of them fails but we need all of their results to continue).

Ideally this should take advantage of Porter's context that is available everywhere, and not proliferate `context.Context` to the signature of all of our methods if we can help it.

This isn't a single PR but an epic to consider how to refactor our code to accommodate this in multiple locations throughout the codebase and for many tasks (CTRL+C, early breaks when we realize we should stop, etc).

carolynvs:
# What does this change
[Add context.Context to a bunch of functions](https://github.com/getporter/porter/commit/75eee4e8c109295b9a74818ff232168f7a054d98)

Refactor before implementing SIGINT handling

[Always use a context when running commands](https://github.com/getporter/porter/commit/a5601ff9323936946ddcc9653c3d47d3c6f90138)

When we run a command, like calling a plugin or mixin, always use a context so that if it's cancelled the process is closed automatically

[Support the interrupt signal SIGINT](https://github.com/getporter/porter/commit/c7ed6d384db65d1e31706e8a16980bde37d006b9)

When the user presses CTRL+C after running a porter command, we should attempt to stop as quickly and gracefully as possible.

I have added context.Context in a bunch of places so that any processes that we start are started with a context. When SIGINT is received, we cancel the context and the os/exec package handles killing any running processes.

I've patched our fork of hashicorp/go-plugin to support context cancellation. I've found that it's still possible to deadlock their Start and Kill functions because of how locks are used. So the loader still only gives the plugin only so long to close before intervening so that we close cleanly.

[Check if canceled before running a bundle](https://github.com/getporter/porter/pull/2078/commits/825e5a03c96b9fd0e0767eb93091b174478ba5ea)

We should always check ctx.Done before running anything that is long or blocking.

[Move plugin logic into PluginConnection](https://github.com/getporter/porter/pull/2078/commits/7f298135af368eaec5918779475a4518ed11ab5e)

I've split up the plugin loader and moved a bunch of the hashicorp/go-plugin specific logic and workarounds into the
PluginConnection struct so that its easier to understand the logic and manage resources that need to be released when porter closes.

# What issue does it fix
Closes #950

# Notes for the reviewer
I recommend reviewing this commit by commit so that you can separate out the refactoring from the more interesting changes.

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist


vdice (approved):
SIGLGTM!

PR #2083, opened 2022-05-19
Fix bson.D conversion in GRPC

### LINKED ISSUE #2063: porter installation output list command fails
## Describe the bug
When running `porter installation output list` with the new grpc version of plugin implementation, it always returns an error, ` rpc error: code = Unknown desc = (Location15973) the $sort key specification must be an object`

## To Reproduce
1. Run ``porter installation output list`

## Expected behavior
The command should return a list of output or an empty list

## Porter Command and Output
```
$ porter installation output list
 2022-05-10T17:37:13.433-0400	error	rpc error: code = Unknown desc = (Location15973) the $sort key specification must be an object
        	           	2022-05-10T17:37:13.433-0400	error	rpc error: code = Unknown desc = (Location15973) the $sort key specification must be an object
```

## Version
porter v1.0.0-alpha.19-34-g699cc38f (699cc38f)

VinozzZ:
# What does this change

MongoDB expects to receive `bson.D` for `$sort` and `$group` in a aggregation pipeline. 
Prior to this PR, after the grpc server receives data from a client, it does not convert the data from []interface{} back to bson.D, which caused the query to fail.
This PR fixes this issue and adds tests to check for proper conversion.

# What issue does it fix
Closes #2063 

# Notes for the reviewer

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist


pkg/storage/grpc_test.go
```diff
@@ -66,4 +66,43 @@ func TestRoundTripDataOverGRPC(t *testing.T) {
...
+			}}},
+			// Group them by output name and select the last value for each output
+			{{"$group", bson.D{
+				{"_id", "$name"},
+				{"lastOutput", bson.M{"$first": "$$ROOT"}},
+			}}},
+		},
+	})
+	require.NoError(t, err)
+	require.Len(t, results1, 0)
```
carolynvs, line 107: Can you insert two outputs and change this assert to check that they were aggregated correctly based on the query? This checks our grpc conversion doesn't completely fail, but misses if we are correctly converting the pipeline.

pkg/storage/pluginstore/grpc.go
```diff
@@ -284,11 +284,12 @@ func ConvertBsonToPrimitives(src interface{}) interface{} {
...
 }
 
-// ConvertFloatToInt works around a weirdness in how numbers are represented
+// ConvertPrimitivesToBson converts go primitive to bson primitive.
+// it also works around a weirdness in how numbers are represented
 // by structpb.Value, where integer values are stored in float64. When we
 // deserialize from protobuf, this walks the specified value, finds ints
 // that were encoded as floats, and converts them back to ints.
-func ConvertFloatToInt(src interface{}) interface{} {
+func ConvertPrimitivesToBson(src interface{}) interface{} {
```
carolynvs, line 292: Thanks this is a much better name! 👍

pkg/storage/pluginstore/grpc.go
```diff
@@ -297,13 +298,20 @@ func ConvertFloatToInt(src interface{}) interface{} {
 		}
 		return tv
 	case []interface{}:
+		toBson := make(bson.D, 0)
```
carolynvs, line 301: Let's size this based on the length of tv
```suggestion
		toBson := make(bson.D, len(tv))
```

carolynvs (changes requested):
The CI build is failing but this looks like it's going in the right direction

pkg/storage/grpc_test.go
VinozzZ: Yes! I just added it

pkg/storage/pluginstore/grpc.go
```diff
@@ -358,10 +392,17 @@ func NewStruct(src map[string]interface{}) *structpb.Struct {
 }
 
 // AsMap converts a protobuf struct into its original representation, bson.M.
-func AsMap(src *structpb.Struct) bson.M {
+func AsMap(src *structpb.Struct, c ...converter) bson.M {
 	dest := src.AsMap()
+	converts := []converter{ConvertFloatToInt}
+	if c != nil {
+		converts = append(converts, c...)
+	}
```
carolynvs, line 400: If you always want to run ConvertFloatToInt and then append extras that are passed, I think that you can simplify this a bit
```suggestion
	c = append(c, ConvertFloatToInt)
```

Then just use c instead of converts (or maybe rename c to converts since it's a better (more descriptive) parameter name.

pkg/storage/pluginstore/grpc.go
```diff
@@ -374,12 +415,17 @@ func AsMapList(src []*structpb.Struct) []bson.M {
...
+type converter func(src interface{}) interface{}
+
 // AsOrderedMap converts an array of protobuf structs into its original
 // representation, bson.D.
-func AsOrderedMap(src []*structpb.Struct) bson.D {
+func AsOrderedMap(src []*structpb.Struct, c ...converter) bson.D {
 	dest := make(bson.D, 0, len(src))
+	if c == nil {
+		c = []converter{ConvertFloatToInt}
+	}
```
carolynvs, line 426: I don't think that it's possible for c to be nil here? If you call the function without passing any optional arguments the value is an empty array, not nil.

I noticed that in AsMap you always use the ConvertFloatToInt converter, even when they specify the optional custom converter parameter. Since you are passing c through to AsMap, I believe that this results in ConvertFloatToInt being used twice in AsMap when c is nil (which is never as I said above). 

I'm pretty sure that you can delete these lines here and it will work exactly the same.

```suggestion
	
```

carolynvs (changes requested):

pkg/storage/pluginstore/grpc.go
VinozzZ: I simplified it to always run ConverFloatToInt before running passed in converters

pkg/storage/grpc_test.go
```diff
@@ -106,10 +108,12 @@ func TestRoundTripDataOverGRPC(t *testing.T) {
 			// Group them by output name and select the last value for each output
 			{{"$group", bson.D{
 				{"_id", "$name"},
-				{"lastOutput", bson.M{"$first": "$ROOT"}},
+				{"lastOutput", bson.M{"$first": "$$ROOT"}},
```
carolynvs, line 111: Was there a problem with this line? It's logic we have been using in Porter, so if this test here was wrong, then InstallationStore has the same bug.

carolynvs (commented):
Just one thing to clear up and then this is ready to merge! 🚀

pkg/storage/grpc_test.go
VinozzZ: I think it was a copy/paste problem. The one in `installation_store.go` is already `$$ROOT`

carolynvs (approved):
🤞

PR #2088, opened 2022-05-21
Do not print "tools used" section on explain

### LINKED ISSUE #1892: Explain command doesn't work for non-porter bundles
The explain command should work for all valid CNAB bundles, not just those made by porter. We should detect if it's not a porter bundle and then print a more limited set of data that doesn't include mixins and other porter specific information.

```console
$ porter explain -r getporter/porter.mixins.exec:v1.0.0-alpha.1
Name: exec
Description: Run arbitrary commands and scripts
Version: 1.0.0-alpha.1

Parameters:
------------------------------------------------------------------------------------------------------
  Name                        Description                     Type    Default  Required  Applies To
------------------------------------------------------------------------------------------------------
  porter-mixin-configuration  Global mixin configuration      object  <nil>    false     All Actions
                              snippet from porter.yaml.
                              Porter internal parameter that
                              should not be set manually.

Actions:
--------------------------------------------------------------------------------------
  Name              Description                     Modifies Installation  Stateless
--------------------------------------------------------------------------------------
  sh.porter.invoke  Handles the porter invoke       true                   true
                    command

This bundle uses the following tools: .

To install this bundle run the following command, passing --param KEY=VALUE for any parameters you want to customize:
porter install --reference getporter/porter.mixins.exec:v1.0.0-alpha.1
```

The example above is the explain output for a non-porter bundle. We shouldn't list the mixins since that's porter specific. Since len(mixins) is 0, it should be skipped.

```
This bundle uses the following tools: .
```

joshuabezaleel:
# What does this change
Explain command will not print `This bundle uses the following tools:` section for (1) non-Porter bundle (the one with no `custom.sh` section) and (2) a bundle with no mixins used.

# What issue does it fix
Closes #1892 

# Notes for the reviewer

# Checklist
- [x] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/porter/explain.go
+++ b/pkg/porter/explain.go
@@ -331,7 +331,9 @@ func (p *Porter) printBundleExplainTable(bun *PrintableBundle, bundleReference s
 	p.printActionsExplainBlock(bun)
 	p.printDependenciesExplainBlock(bun)
 
-	fmt.Fprintf(p.Out, "This bundle uses the following tools: %s.\n", strings.Join(bun.Mixins, ", "))
+	if extendedBundle.IsPorterBundle() && len(bun.Mixins) > 0 {
+		fmt.Fprintf(p.Out, "This bundle uses the following tools: %s.\n", strings.Join(bun.Mixins, ", "))
+	}
 
 	if extendedBundle.SupportsDocker() {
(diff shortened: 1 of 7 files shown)


carolynvs (approved):
Perfect! Thanks for the fix! :shipit:

PR #2096, opened 2022-05-24
Fix performance for porter list

### LINKED ISSUE #2040: porter list should not make individual calls for each installation
After I instrumented the plugins, I found a bug in the list command. We are calling out to the database and the secrets plugin for each installation returned by list.

Screen Shot 2022-04-27 at 10 34 00 AM

That entire long tail of calls to storage and secrets should not be there. We should only make enough calls to display the information required for `porter list --output plaintext`. This means that even though the json/yaml outputs _could theoretically_ provide more information, we shouldn't make extra calls to grab that info.

I'm not sure why yet that we are making all these calls, they are appropriate for `porter installation show` for example, but for a list command the performance / chattiness is a problem. I'm guessing we are retrieving the latest run for each installation? If needed we can optimize how we are retrieving the data so that it's a single call to installations, and then another single call to retrieve all runs, and then join the data in the client.

carolynvs:
# What does this change
Do not retrieve the full installation definition for each installation returned from list. We only need that information
for the show command. Right now porter is making multiple calls to both the storage and secrets plugin for each installation in the list results.

# What issue does it fix
Closes #2040

# Notes for the reviewer
Here is what the trace looks like with this fix (compared to the original trace which was huge)

Screen Shot 2022-05-24 at 3 01 32 PM

# Checklist
- [x] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/porter/list.go
+++ b/pkg/porter/list.go
@@ -214,30 +214,31 @@ func NewDisplayRun(run storage.Run) DisplayRun {
 }
 
 // ListInstallations lists installed bundles.
-func (p *Porter) ListInstallations(ctx context.Context, opts ListOptions) ([]storage.Installation, error) {
+func (p *Porter) ListInstallations(ctx context.Context, opts ListOptions) (DisplayInstallations, error) {
 	ctx, log := tracing.StartSpan(ctx)
 	defer log.EndSpan()
 
 	installations, err := p.Installations.ListInstallations(ctx, opts.GetNamespace(), opts.Name, opts.ParseLabels())
-	return installations, errors.Wrap(err, "could not list installations")
-}
-
-// PrintInstallations prints installed bundles.
-func (p *Porter) PrintInstallations(ctx context.Context, opts ListOptions) error {
-	installations, err := p.ListInstallations(ctx, opts)
 	if err != nil {
-		return err
+		return nil, errors.Wrap(err, "could not list installations")
 	}
 
 	var displayInstallations DisplayInstallations
 	for _, installation := range installations {
-		di, err := p.generateDisplayInstallation(ctx, installation)
-		if err != nil {
-			return err
-		}
+		di := NewDisplayInstallation(installation)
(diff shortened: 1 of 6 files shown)


pkg/porter/list.go
```diff
@@ -214,31 +214,32 @@ func NewDisplayRun(run storage.Run) DisplayRun {
...
+		return nil, errors.Wrap(err, "could not list installations")
 	}
 
 	var displayInstallations DisplayInstallations
 	for _, installation := range installations {
-		di, err := p.generateDisplayInstallation(ctx, installation)
-		if err != nil {
-			return err
-		}
+		di := NewDisplayInstallation(installation)
```
carolynvs, line 228: I moved the conversion of installations to display installations so that we could test these functions. In other parts of Porter we do the same thing but I guess this function is old enough it was doing the conversion in the print function still.

pkg/porter/show.go
```diff
@@ -163,27 +171,23 @@ func (p *Porter) ShowInstallation(ctx context.Context, opts ShowOptions) error {
...
 }
 
-func (p *Porter) generateDisplayInstallation(ctx context.Context, installation storage.Installation) (DisplayInstallation, error) {
-	run, err := p.Installations.GetRun(ctx, installation.Status.RunID)
-	if err != nil {
-		return DisplayInstallation{}, err
-	}
-
+func (p *Porter) NewDisplayInstallationWithSecrets(ctx context.Context, installation storage.Installation, run *storage.Run) (DisplayInstallation, error) {
 	bun := cnab.ExtendedBundle{run.Bundle}
```
VinozzZ, line 175: What if the run is nil here?

carolynvs: Haha you caught that at the same time I did, I was fixing that right before our meeting this morning. I just pushed it up now.

carolynvs:
@VinozzZ Okay I think I've fixed everything, please take another look!

VinozzZ (approved):

PR #2120, opened 2022-05-31
Remove group from StorageProtocol.FindOptions

carolynvs:
# What does this change
When I was adding tests for the [FindInstallations function](#2119), I realized that I'd made a mistake earlier when defining the StorageProtocol.FindOptions. It had erroneously included Group (which is only for Aggregate). In addition, our wrapper around FindOptions wasn't exposing the Select field (that's not a plugin protocol problem though, since select was always on the underlying StorageProtocol definition).

I've also updated our StorageProtocol wrapper to use bson.D/M for consistency and ensure we are always passing types that will work with our gRPC conversion logic that specifically handles bson.D.

There aren't any non-internal storage plugins to test against to determine if this is a breaking change. We could probably just get away with removing it and pretending it never existed. However for practice with making changes to a published gRPC protocol, I've removed the field and marked it as reserved [per the gRPC versioning document](https://docs.microsoft.com/en-us/aspnet/core/grpc/versioning?view=aspnetcore-6.0#binary-breaking-changes).

> Removing a field - Values from a removed field are deserialized to a message's [unknown fields](https://developers.google.com/protocol-buffers/docs/proto3#unknowns). This isn't a gRPC protocol breaking change, but the client needs to be updated if it upgrades to the latest contract. It's important that a removed field number isn't accidentally reused in the future. To ensure this doesn't happen, specify deleted field numbers and names on the message using Protobuf's [reserved](https://developers.google.com/protocol-buffers/docs/proto3#reserved) keyword.

# What issue does it fix
Remove an oopsie before we release v1

# Notes for the reviewer
I'd appreciate some extra thinking about if this could have unintended consequences that would break things. I think I caught it early enough that it won't be an issue but let's double check that!

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist


carolynvs:
@VinozzZ Do you have any comments that needs to be addressed or is this okay to merge?

VinozzZ:
> @VinozzZ Do you have any comments that needs to be addressed or is this okay to merge?

Yes! I'm good to merge it

PR #2123, opened 2022-06-01
Improve Porter's tagline

carolynvs:
@VinozzZ suggested that we switch from saying "versioned bundle" to "installer" since it better fits with how people think about what Porter actually does.

I've tightened up the wording a bit so that its shorter and hopefully a little bit clearer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,12 +4,8 @@
 
 # Porter
 
-Package your application artifact, client tools, configuration and deployment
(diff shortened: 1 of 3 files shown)


VinozzZ (approved):
Awesome!

carolynvs:
I'll leave this open for a couple days so that others can weigh in since this is a change that we should have consensus on (how we explain porter).

PR #2126, opened 2022-06-02
Fix race condition in plugin logs routine

carolynvs:
# What does this change
When we run a plugin, we start a go routine that captures logs from the plugin and forwards them to our traceLogger. I've found that there is a race condition where:

1. We close the plugin
2. The plugin emits a message saying that its closing
3. We cancel the context for the log go routine but ctx.Done doesn't immediately trigger on the go routine.
4. Instead the plugin connections's close function returns before the routine is stopped.
5. Porter moves on and closes our logger.
6. The log go routine then tries to forward the message to the logger which is now closed.

I've added a wait group that ensures that the log go routine finishes before the plugin connection returns from Close. This ensures that all resources managed by the plugin connection are cleaned up before Close returns.

# What issue does it fix
If someone is running in debug mode, a good percentage of the time, there would be a debug message saying that we couldn't write to the close log file at the end of the porter command. This removes that noise and ensures that if there is an error emitted by the plugin on close that we capture it.

# Notes for the reviewer
N/A

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/plugins/pluggable/connection.go
+++ b/pkg/plugins/pluggable/connection.go
@@ -285,6 +285,7 @@ func (c *PluginConnection) setupLogCollector(ctx context.Context) {
 	c.logsReader, c.logsWriter = io.Pipe()
 	ctx, c.cancelLogCtx = context.WithCancel(ctx)
 
+	c.logsWaitGroup.Add(1)
 	go c.collectPluginLogs(ctx)
 }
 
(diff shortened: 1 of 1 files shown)


pkg/plugins/pluggable/connection.go
```diff
@@ -286,6 +293,9 @@ func (c *PluginConnection) setupLogCollector(ctx context.Context) {
 // The best way to get that information is to instrument the plugin itself. This is mainly a fallback mechanism to
 // collect logs from an uninstrumented plugin.
 func (c *PluginConnection) collectPluginLogs(ctx context.Context) {
+	c.logsWaitGroup.Add(1)
```
VinozzZ, line 296: What do u think moving this line to line 287? 
That way we know where the goroutine is started

carolynvs:
@VinozzZ Fixed!

VinozzZ (approved):
Yahoo!!

PR #2137, opened 2022-06-06
Adding pagination for installation, parameter, and credential list result using skip and limit option

### LINKED ISSUE #2051: Add paging options to porter list and other commands that list installation data
We have support in the underlying plugin protocol to page data when listing, but we still don't expose flags for that in the CLI. At some point we should add those flags in so that people with a significant number of installations, or long lived often updated installations don't run into performance problems.

Adding this post 1.0 won't be a breaking change because the plugins already support it.

joshuabezaleel:
# What does this change
User that use the Porter CLI can use --skip and --limit flag for paginating their list output.

# What issue does it fix
Closes #2051 

# Notes for the reviewer

# Checklist
- [x] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/storage/credential_store.go
+++ b/pkg/storage/credential_store.go
@@ -110,10 +110,13 @@ func (s CredentialStore) InsertCredentialSet(ctx context.Context, creds Credenti
 	return s.Documents.Insert(ctx, CollectionCredentials, opts)
 }
 
-func (s CredentialStore) ListCredentialSets(ctx context.Context, namespace string, name string, labels map[string]string) ([]CredentialSet, error) {
+func (s CredentialStore) ListCredentialSets(ctx context.Context, listOptions ListOptions) ([]CredentialSet, error) {
 	var out []CredentialSet
 	opts := FindOptions{
-		Filter: CreateListFiler(namespace, name, labels),
+		Sort:   []string{"namespace", "name"},
+		Filter: CreateListFiler(listOptions),
+		Skip:   listOptions.Skip,
+		Limit:  listOptions.Limit,
 	}
 	err := s.Documents.Find(ctx, CollectionCredentials, opts, &out)
 	return out, err
--- a/pkg/storage/credential_store.go
+++ b/pkg/storage/credential_store.go
@@ -112,13 +112,7 @@ func (s CredentialStore) InsertCredentialSet(ctx context.Context, creds Credenti
 
 func (s CredentialStore) ListCredentialSets(ctx context.Context, listOptions ListOptions) ([]CredentialSet, error) {
 	var out []CredentialSet
-	opts := FindOptions{
-		Sort:   []string{"namespace", "name"},
-		Filter: CreateListFiler(listOptions),
-		Skip:   listOptions.Skip,
-		Limit:  listOptions.Limit,
-	}
-	err := s.Documents.Find(ctx, CollectionCredentials, opts, &out)
+	err := s.Documents.Find(ctx, CollectionCredentials, listOptions.ToFindOptions(), &out)
 	return out, err
 }
 
--- a/pkg/storage/credential_store_test.go
+++ b/pkg/storage/credential_store_test.go
@@ -20,35 +20,17 @@ func TestCredentialStorage_CRUD(t *testing.T) {
 
 	require.NoError(t, cp.InsertCredentialSet(context.Background(), cs))
 
-	creds, err := cp.ListCredentialSets(context.Background(), ListOptions{
-		Namespace: "dev",
-		Name:      "",
-		Labels:    nil,
-		Skip:      0,
-		Limit:     0,
-	})
+	creds, err := cp.ListCredentialSets(context.Background(), ListOptions{Namespace: "dev"})
 	require.NoError(t, err)
 	require.Len(t, creds, 1, "expected 1 credential set")
 	require.Equal(t, cs.Name, creds[0].Name, "expected to retrieve sekrets credentials")
 	require.Equal(t, cs.Namespace, creds[0].Namespace, "expected to retrieve sekrets credentials")
 
-	creds, err = cp.ListCredentialSets(context.Background(), ListOptions{
-		Namespace: "",
-		Name:      "",
-		Labels:    nil,
-		Skip:      0,
-		Limit:     0,
-	})
+	creds, err = cp.ListCredentialSets(context.Background(), ListOptions{})
 	require.NoError(t, err)
 	require.Len(t, creds, 0, "expected no global credential sets")
 
-	creds, err = cp.ListCredentialSets(context.Background(), ListOptions{
-		Namespace: "*",
-		Name:      "",
-		Labels:    nil,
-		Skip:      0,
-		Limit:     0,
-	})
+	creds, err = cp.ListCredentialSets(context.Background(), ListOptions{Namespace: "*"})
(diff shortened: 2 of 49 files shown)


joshuabezaleel:
@carolynvs Should we also make this `--skip` and `--limit` option available for `porter credentials list` and `porter parameters list` commands as well?

pkg/storage/installation_store.go
```diff
@@ -63,14 +63,16 @@ func (s InstallationStore) Initialize(ctx context.Context) error {
 	return span.Error(err)
 }
 
-func (s InstallationStore) ListInstallations(ctx context.Context, namespace string, name string, labels map[string]string) ([]Installation, error) {
+func (s InstallationStore) ListInstallations(ctx context.Context, namespace string, name string, labels map[string]string, skip int64, limit int64) ([]Installation, error) {
```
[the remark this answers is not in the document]
joshuabezaleel: @carolynvs Actually been thinking about that as well! Will add it on the next commits.

Also, what do you think of making the `--skip` and `--limit` options available to `porter credentials list` and `porter parameters list` commands as well? If yes, I can also do the process of changing the function argument to option struct for those two as well.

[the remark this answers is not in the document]
joshuabezaleel: @carolynvs should've been ready for another round of review! I made the option struct a general one (`ListOptions`) so that it can also be used for ListParameters and ListCredentialSets but looking forward to hearing your thoughts!

pkg/storage/credential_store.go
```diff
@@ -110,10 +110,13 @@ func (s CredentialStore) InsertCredentialSet(ctx context.Context, creds Credenti
 	return s.Documents.Insert(ctx, CollectionCredentials, opts)
 }
 
-func (s CredentialStore) ListCredentialSets(ctx context.Context, namespace string, name string, labels map[string]string) ([]CredentialSet, error) {
+func (s CredentialStore) ListCredentialSets(ctx context.Context, listOptions ListOptions) ([]CredentialSet, error) {
 	var out []CredentialSet
 	opts := FindOptions{
-		Filter: CreateListFiler(namespace, name, labels),
+		Sort:   []string{"namespace", "name"},
```
carolynvs, line 116: Thanks for adding the sort!

pkg/storage/credential_store_test.go
```diff
@@ -20,17 +20,35 @@ func TestCredentialStorage_CRUD(t *testing.T) {
...
 	require.NoError(t, cp.InsertCredentialSet(context.Background(), cs))
 
-	creds, err := cp.ListCredentialSets(context.Background(), "dev", "", nil)
+	creds, err := cp.ListCredentialSets(context.Background(), ListOptions{
+		Namespace: "dev",
+		Name:      "",
+		Labels:    nil,
+		Skip:      0,
+		Limit:     0,
+	})
```
carolynvs, line 29: So the nice thing about using an options struct is that you can now omit fields where you aren't changing the default. This allows us to set a better default in the future, and limit how often code has to change when it's about fields that it doesn't care about.

Let's remove the labels, skip and limit fields from here and anywhere else that we aren't setting a non-default value.

pkg/storage/credential_store_test.go
```diff
@@ -20,17 +20,35 @@ func TestCredentialStorage_CRUD(t *testing.T) {
...
 	require.NoError(t, err)
 	require.Len(t, creds, 1, "expected 1 credential set")
-	require.Equal(t, cs.Name, creds[0].Name, "expected to retrieve secreks credentials")
-	require.Equal(t, cs.Namespace, creds[0].Namespace, "expected to retrieve secreks credentials")
+	require.Equal(t, cs.Name, creds[0].Name, "expected to retrieve sekrets credentials")
```
carolynvs, line 32: 😂

pkg/storage/credential_store_test.go
```diff
@@ -20,17 +20,35 @@ func TestCredentialStorage_CRUD(t *testing.T) {
...
+		Labels:    nil,
+		Skip:      0,
+		Limit:     0,
+	})
 	require.NoError(t, err)
 	require.Len(t, creds, 0, "expected no global credential sets")
 
-	creds, err = cp.ListCredentialSets(context.Background(), "*", "", nil)
+	creds, err = cp.ListCredentialSets(context.Background(), ListOptions{
+		Namespace: "*",
```
carolynvs, line 46: So in this case you can remove all of the fields except namespace.

pkg/storage/migrations/manager_test.go
```diff
@@ -177,7 +213,13 @@ func TestCredentialStorage_NoMigrationRequiredForEmptyHome(t *testing.T) {
 	testSecrets := secrets.NewTestSecretsProvider()
 	credStore := storage.NewTestCredentialProviderFor(t, mgr, testSecrets)
 
-	names, err := credStore.ListCredentialSets(context.Background(), "", "", nil)
+	names, err := credStore.ListCredentialSets(context.Background(), storage.ListOptions{
```
carolynvs, line 216: In this case you can just pass an empty struct.

pkg/storage/parameter_store.go
```diff
@@ -100,10 +100,13 @@ func (s ParameterStore) InsertParameterSet(ctx context.Context, params Parameter
 	return s.Documents.Insert(ctx, CollectionParameters, opts)
 }
 
-func (s ParameterStore) ListParameterSets(ctx context.Context, namespace string, name string, labels map[string]string) ([]ParameterSet, error) {
+func (s ParameterStore) ListParameterSets(ctx context.Context, listOptions ListOptions) ([]ParameterSet, error) {
 	var out []ParameterSet
 	opts := FindOptions{
-		Filter: CreateListFiler(namespace, name, labels),
+		Sort:   []string{"namespace", "name"},
```
carolynvs, line 106: suggestion: You don't have to do this but one way we can standardize all three of these list functions and make it easier to maintain going forward is to change CreateListFilter to a function like `listOptions.ToFindOptions` and it would handle setting the sort, filter, skip and limit.

pkg/storage/query.go
```diff
@@ -292,16 +292,36 @@ func convertToRawJsonDocument(in interface{}, raw interface{}) error {
 // * matching namespace
 // * name contains substring
 // * labels contains all matches
-func CreateListFiler(namespace string, name string, labels map[string]string) map[string]interface{} {
+// func CreateListFiler(namespace string, name string, labels map[string]string) map[string]interface{} {
```
carolynvs, line 295: Can you remove this commented out function signature?

joshuabezaleel: Ah, sorry that this got left out. Thank you for pointing this out!

pkg/storage/credential_store_test.go
joshuabezaleel: I am also actually not sure why did I do that either... I think my brain automatically did that since seeing the previous function parameter input that lists out all of the default primitive values so I did that to the struct as well 😂 Fixed!

pkg/storage/parameter_store.go
joshuabezaleel: That also crossed my mind since we write all of the same kind of input to the FindOptions to credential, parameter, and installation! Updated 🙂

carolynvs (approved):
Oh this looks just great now! Thanks for the extra refactoring. 👍

PR #2142, opened 2022-06-08
Use user specified directory for resolving file path

### LINKED ISSUE #1447: porter install --file fails with path to file outside pwd
## Describe the bug
Executing porter install with --file specifying a manifest in a directory outside the current working directory fails.

## To Reproduce
1. create a directory a
2. place any valid porter.yaml file in this directory
3. Run this porter command `porter install --file a\porter.yaml`

## Expected behavior
Bundle should be built and installed

## Porter Command and Output
```
$ porter install -f tests/testdata/porter.yaml
 Building bundle ===>
Error: unable to generate manifest: could not read the manifest at "porter.yaml": open /home/simon/github/simongdavies/kubernetes-plugins/porter.yaml: no such file or directory
```
## Version

porter v0.33.0 (991a5a23)

VinozzZ:
# What does this change
Currently, porter assumes current working directory is `pwd()` even if users have specified a directory through `--dir`.
Therefore  using `--file` flag requires users to always provide the absolute file path to porter manifest file, or porter sill use `pwd()` as the root directory when a relative path is provided.

example:
```
$ pwd
$ /home/test
$ porter build --file hello/porter.yaml --dir /home/test/example
```
The resolved porter manifest path would be `/home/test/hello/porter.yaml` 

This PR changes the behavior to use the directory specified by `--dir` as the root directory when resolving relative path provided by `--file` .

example:
```
$ pwd
$ /home/test
$ porter build --file hello/porter.yaml --dir /home/test/example
```
The resolved porter manifest path would be `/home/test/example/hello/porter.yaml` 

# What issue does it fix
Closes #1447 

# Notes for the reviewer

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist


cmd/porter/bundle.go
```diff
@@ -80,7 +80,7 @@ The docker driver builds the bundle image using the local Docker host. To use a
 	f.StringVar(&opts.Name, "name", "", "Override the bundle name")
 	f.StringVar(&opts.Version, "version", "", "Override the bundle version")
 	f.StringVarP(&opts.File, "file", "f", "",
-		"Path to the Porter manifest. Defaults to `porter.yaml` in the current directory.")
+		"Path to the Porter manifest. When build context directory is specified, relative path will be resolved using the build context directory as the root directory. Defaults to `porter.yaml` in the current directory.")
```
carolynvs, line 83: I think we can simplify this a bit since the build context directory defaults to the working directory.

```suggestion
		"Path to the Porter manifest. The path is relative to the build context directory. Defaults to porter.yaml in the current directory.")
```

pkg/porter/cnab.go
```diff
@@ -68,6 +59,19 @@ func (o *bundleFileOptions) Validate(cxt *portercontext.Context) error {
 		o.Dir = cxt.FileSystem.Abs(o.Dir)
 	}
 
+	if o.File != "" {
+		if o.Dir != "" && !filepath.IsAbs(o.File) {
+			o.File = cxt.FileSystem.Abs(filepath.Join(o.Dir, o.File))
```
carolynvs, line 64: Isn't o.Dir always set in the if statement above? I think we can simplify this to the following
```suggestion
		if !filepath.IsAbs(o.File) {
			o.File = cxt.FileSystem.Abs(filepath.Join(o.Dir, o.File))
```

pkg/porter/cnab.go
```diff
@@ -164,13 +168,17 @@ func (o *bundleFileOptions) defaultBundleFiles(cxt *portercontext.Context) error
 	} else if o.CNABFile != "" { // --cnab-file
 		// Nothing to default
 	} else {
-		manifestExists, err := cxt.FileSystem.Exists(config.Name)
+		defaultPath := config.Name
+		if o.Dir != "" {
```
carolynvs, line 172: Again, isn't o.Dir always set?

pkg/porter/cnab.go
```diff
@@ -164,13 +168,17 @@ func (o *bundleFileOptions) defaultBundleFiles(cxt *portercontext.Context) error
...
+			defaultPath = filepath.Join(o.Dir, config.Name)
+		}
+		manifestExists, err := cxt.FileSystem.Exists(defaultPath)
 		if err != nil {
 			return errors.Wrap(err, "could not check if porter manifest exists in current directory")
 		}
```
carolynvs, line 178:
```suggestion
		if err != nil {
			return errors.Wrapf(err, "could not find a porter manifest at %s", defaultpath)
		}
```

carolynvs (commented):
This is looking good! I just have some small suggestions

pkg/porter/cnab.go
VinozzZ: I thought the `o.Dir` is only resolved if it's not empty in the if statement above?

carolynvs: Ah okay thanks I thought that if it was empty, we were setting it to the absolute path of the current working directory. If we did that then we wouldn't need a bunch of defensive checks later in the code since we know it was initialized properly. What do you think?

cmd/porter/bundle.go
VinozzZ: I wanted to make sure that users understand that if a absolute path is passed in, then it won't be relative to the build context directory. How about we say
`When a relative path is provided, the path is relative to the build context directory"

carolynvs: The build context directory (--dir) defaults to the current directory. So when you say "relatative to the build context directory" I think that covers both cases, no? 

I just double checked and --dir doesn't say "Defaults to the current directory". How about we add that (it should have been included so that's an oversight)? I'm trying to avoid complex explanations, and explaining the --dir flag on the helptext for the --file flag.

carolynvs:
Sorry about all the nits. I'm just trying to reduce the "if this, then this, otherwise that" aspect of this feature that was baked into the code. It adds unnecessary complexity for us and the user, and I'm hoping we can make this easier for everyone.

VinozzZ:
> Sorry about all the nits. I'm just trying to reduce the "if this, then this, otherwise that" aspect of this feature that was baked into the code. It adds unnecessary complexity for us and the user, and I'm hoping we can make this easier for everyone.

No worries at all! I would love to make the helper message more clear

carolynvs (approved):
Thanks for your patience while we hammered out how this should work. This looks great and I think will be easier for people to use. 👍

PR #2149, opened 2022-06-09
use relocation map for pulling images during archive

### LINKED ISSUE #2050: Archive isn't pulling images from the published bundle
When porter archive is run, it should copy all of the bundle and any referenced images into a tgz file. The images should be copied directly from the bundle's repository. The bug is that we are instead pulling referenced images from their original location, instead of from the bundle's repository.

So if a bundle is using the nginx image, porter archive should pull the nginx image from the bundle repo (e.g. getporter/porter-hello:UGLY_DIGEST, instead we are pulling from docker.io/library/nginx.

This causes a problem because now someone could delete the original image, and cause porter archive to fail when we actually had all the info we needed to succeed in the bundle repository.

## Steps to Reproduce

1. Make a copy of the whalesayd image and push it to your own registry that you can create and delete images from. Replace YOURNAME with your docker registry.
   ```
   $ docker pull carolynvs/whalesayd:v0.1.1
   $ docker tag carolynvs/whalesayd:v0.1.1 YOURNAME/whalesayd:latest
   $ docker push YOURNAME/whalesayd:latest
    The push refers to repository [docker.io/carolynvs/whalesayd]
    fcb12292dbbb: Layer already exists
    5f70bf18a086: Mounted from carolynvs/porter-hello
    d061ee1340ec: Layer already exists
    d511ed9e12e1: Layer already exists
    091abc5148e4: Layer already exists
    b26122d57afa: Layer already exists
    37ee47034d9b: Layer already exists
    528c8710fd95: Layer already exists
    1154ba695078: Layer already exists
    latest1: digest: sha256:adf33c54558425e772c5a4fd026cf59b4e0cdbf3934270ab07b20cc3f2c5cbf6 size: 2613
    ```
1. Save the digest that was printed out during push.
1.  Clone the example bundles repository `git clone https://github.com/getporter/examples.git`
1. `cd examples/airgap`
1. Open porter.yaml in your editor and change the repository and digest fields to match your copy of the whalesayd image
   ```yaml
    images:
      whalesayd:
        description: "Whalesay as a service"
        imageType: "docker"
        repository: "YOURNAME/whalesayd" # Change to your copy
        digest: "sha256:YOUR_DIGEST" # Change to your digest from the docker push command
   ```
1. Build and publish the bundle to your registry by running. Change the registry flag to match your registry location.
   ```
   porter build
   porter publish --reference YOURNAME/whalegap:latest
   ```
1. Go to your registry and delete your copy of the whalesayd image. This triggers the bug.
1. Archive your bundle with `porter archive whalegap.tgz --reference YOURNAME/whalegap:latest` and it will fail because it isn't pulling the whalesayd image from the bundle repository and instead is trying to pull the image that you just deleted.

## Expected Behavior
Porter archive should succeed because it pulled the whalesayd image from the bundle's repository.

I believe the relevant code is in https://github.com/getporter/porter/blob/main/pkg/porter/archive.go. This is hard to test but if you wanted to make a regression test you can do so using a similar trick as tests/smoke/copy_test.go, which spins up multiple registries. You would push the image to one registry, publish your bundle to another, and then delete the first registry, and then run archive. I'll accept a PR without that regression test though. 😀 

_Originally reported by @jasmdk in https://github.com/getporter/porter/issues/1357#issuecomment-1114579391_

VinozzZ:
# What does this change

This PR does the following two things:
- adds `relocation-mapping.json` into archived bundle
- updates `porter archive` to use relocation map when pulling images instead of the original reference used in bundle.js

# What issue does it fix
Closes #2050 #2152 

# Notes for the reviewer

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist


pkg/porter/archive.go
```diff
@@ -44,34 +46,41 @@ func (o *ArchiveOptions) Validate(ctx context.Context, args []string, p *Porter)
 // any referenced images locally (if needed), export them to individual layers, generate a bundle.json and
 // then generate a gzipped tar archive containing the bundle.json and the images
 func (p *Porter) Archive(ctx context.Context, opts ArchiveOptions) error {
+	ctx, log := tracing.StartSpan(ctx)
```
carolynvs, line 49: Thank you for adding tracing in all your PRs! 💖

pkg/porter/archive_test.go
```diff
@@ -51,3 +54,45 @@ func TestArchive_Validate(t *testing.T) {
...
+		{"image successfully added", relocation.ImageRelocationMap{"image:v0.1.0": "image@sha256:123"}, "image:v0.1.0", "image@sha256:123", false},
+	}
+
+	for _, tc := range tests {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			baseImage := bundle.BaseImage{Image: tc.inputImg, Digest: "digest"}
+			ex := exporter{relocationMap: tc.relocationMap, imageStore: mockImageStore{t: t, expected: tc.expectedImg}}
+			err := ex.addImage(baseImage)
+			require.Equal(t, tc.hasError, err != nil)
```
carolynvs, line 80: Can you switch from just checking that an error occurred to checking for a substring in the error message You can use the tests.RequireErrorContains helper. That way we can be sure that it failed for the reason we were expecting.

tests/integration/archive_test.go
```diff
@@ -83,3 +88,21 @@ func getHash(p *porter.TestPorter, path string) string {
...
+	require.NoError(p.T(), err)
+	defer p.FileSystem.RemoveAll(tmpDir)
+
+	source := p.FileSystem.Abs(path)
+	l := loader.NewLoader()
+	imp := packager.NewImporter(source, tmpDir, l)
+	err = imp.Import()
+	require.NoError(p.T(), err, "opening archive failed")
+
+	_, err = p.FileSystem.Stat(filepath.Join(tmpDir, strings.TrimSuffix(filepath.Base(source), ".tgz"), "relocation-mapping.json"))
```
carolynvs, line 103: I feel like this isn't giving us as much test coverage as we need here. For example, the test would pass if relocation-mapping.json is empty right? 

Can you add a check that it has the expected images in it as well?

carolynvs (changes requested):

tests/smoke/archive_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+	// Publish the bundle to one location
+	origRef := fmt.Sprintf("localhost:%s/orig-mydb:v0.1.1", tempRegistryPort)
+	test.MakeTestBundle(testdata.MyDb, origRef)
+
+	archiveFilePath := filepath.Join(test.TestDir, "archive-test.tgz")
+	test.RequirePorter("archive", archiveFilePath, "--reference", origRef)
+	stopTempRegistry()
+
+	newRef := "localhost:5000/copy-new-mydb:v0.1.1"
```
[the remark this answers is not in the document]
VinozzZ: Thank you for the suggestion. I made the change accordingly

carolynvs (approved):
Thanks for updating the smoke test so that it's a regression test for this bug! Looks great and I'm excited to see us closing all these little gaps in getting airgapped bundles to work as intended. 💯

PR #2150, opened 2022-06-10
Support migration of data from v0 to v1

### LINKED ISSUE #1700: Migrate data from v0.38 format to v1.0
The entire data storage format and allowed backends is changing for Porter v1. We are moving from flat file storage (without query support) to mongodb. So both the backend plugins are changing AND the storage format.

🚨 **Migrations between alpha versions are not supported. Do not use the v1 alpha of Porter with production data.**

After #1684 is merged, we cannot release a prerelease of v1 until we have a migration for existing data.

* The migration is one way and does not alter your existing data.
* We need to pick new storage schema versions for our data, and decide where to encode the porter schema version vs the cnab schema version. 

One option is to store the porter schema version on the document, and then always generate the cnab version in the new ToCNAB() functions on our document types.

For this situation I think we will need two commands to move old data to the new format:

1. `porter storage copy --source myazuretables --destination myazuremongo` - Copies the data AS IS, from one storage backend to another. This gets the data out of azure tables and into a mongo database.
2. `porter storage migrate` - Executes a data migration on the data in the currently configured storage backend. Data in the old format that has been migrated is deleted during the migration, leaving a clean set of documents.

We also require a documentation page and blog post announcing EARLY before the v1 release how the data migrations will work.

carolynvs:
# What does this change
The porter storage migrate command now supports migrating from v0.38.* (not previous versions) to v1.0.0-beta.1 and higher.

You must keep your old PORTER_HOME directory around and then create a new PORTER_HOME that has v1 installed. The storage migrate command reads the data from the old home, and copies it into the new database for the v1 version of porter.

Data in the old porter home is not modified at all (and our interface to interact with it is a readonly subset of the original plugin protocol to avoid any oopsies).

# What issue does it fix
Closes #1700 

# Notes for the reviewer
N/A

# Checklist
- [x] Did you write tests?

# Reviewer Checklist


carolynvs:
I am still writing the user documentation but the code is ready to review

pkg/storage/migrations/migration.go
```diff
@@ -399,11 +445,157 @@ func convertOutput(result storage.Result, outputKey string, data []byte) (storag
...
+	if err != nil {
+		return span.Error(err)
+	}
+
+	dest, err := convertCredentialSet(m.opts.NewNamespace, data)
+	if err != nil {
+		return span.Error(err)
+	}
+
+	err = m.destStore.Insert(ctx, storage.CollectionCredentials, storage.InsertOptions{Documents: []interface{}{dest}})
```
VinozzZ, line 484: Would this result in an error if some old record has been migrated? If so, should we just ignore duplicated key error?

pkg/storage/migrations/migration_test.go
```diff
@@ -194,5 +215,55 @@ func TestMigration_MigrateInstallations(t *testing.T) {
...
+	test.CompareGoldenFile(t, wantFile, string(logsOutput.Value))
+}
+
+func validateMigratedCredentialSets(ctx context.Context, t *testing.T, destStore storage.TestStore, opts storage.MigrateOptions) {
+	store := storage.NewCredentialStore(destStore, nil)
+	credentialSets, err := store.ListCredentialSets(ctx, storage.ListOptions{Namespace: opts.NewNamespace})
+	require.NoError(t, err, "could not list credentialSets in the destination database")
+	assert.Len(t, credentialSets, 1, "expected 1 credential set to be migrated")
+
+	// Validate that the installation as migrated correctly
```
VinozzZ, line 229:
```
	// Validate that the Credential Set as migrated correctly
```

pkg/storage/migrations/migration_test.go
```diff
@@ -194,5 +215,55 @@ func TestMigration_MigrateInstallations(t *testing.T) {
...
+	assert.Equal(t, "GITHUB_TOKEN", cred.Source.Value, "incorrect credential source value")
+}
+
+func validateMigratedParameterSets(ctx context.Context, t *testing.T, destStore storage.TestStore, opts storage.MigrateOptions) {
+	store := storage.NewParameterStore(destStore, nil)
+	parameterSets, err := store.ListParameterSets(ctx, storage.ListOptions{Namespace: opts.NewNamespace})
+	require.NoError(t, err, "could not list parameterSets in the destination database")
+	assert.Len(t, parameterSets, 1, "expected 1 parameter set to be migrated")
+
+	// Validate that the installation as migrated correctly
```
VinozzZ, line 253: // Validate that the Parameter Set as migrated correctly

pkg/storage/migrations/source_schema.go
```diff
@@ -0,0 +1,12 @@
+package migrations
+
+import (
+	"github.com/cnabio/cnab-go/schema"
+)
+
+// SourceSchema represents the file format of Porter's v0.38 schema document
+type SourceSchema struct {
+	Claims      schema.Version `json:"claims"`
```
VinozzZ, line 9: Just curious, why does this schema only supports `json` format vs `SourceParameterSet`

tests/integration/migration_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+	migrationhelpers "get.porter.sh/porter/pkg/storage/migrations/testhelpers"
+	testhelpers "get.porter.sh/porter/pkg/test"
+	"get.porter.sh/porter/tests/tester"
+	"github.com/stretchr/testify/require"
+)
+
+// Do a migration. This also checks for any problems with our
+// connection handling which can result in panics :-)
+func TestMigration(t *testing.T) {
```
VinozzZ, line 16: Should we also make sure sensitive data is correctly migrated?

docs/content/storage-migrate.md
```diff
@@ -2,81 +2,180 @@
...
-  
-NOTE: Only storage plugins have data that requires migration, not secret plugins.
+$ porter list
+-------------------------------------------------------------------------------------
+  NAMESPACE  NAME                  CREATED     MODIFIED    LAST ACTION  LAST STATUS
+-------------------------------------------------------------------------------------
+```
 
-### Azure Plugin
+## 5. Migrate Your Data
```
VinozzZ, line 91: Do you think we need to call out that if their old data contains any sensitive data, they also need to configure a secret store?

VinozzZ (commented):
I'm just done with my first round of the review. I would like to spend some time to set up a local environment to try out this new feature tomorrow!

cmd/porter/storage.go
```diff
@@ -25,21 +25,34 @@ func buildStorageCommand(p *porter.Porter) *cobra.Command {
...
-	flags.StringVarP(&opts.Source, "src", "s", "",
-		"Name of the source storage account defined in your Porter config file")
-	flags.StringVarP(&opts.Destination, "dest", "d", "",
-		"Name of the destination storage account defined in your Porter config file")
+	flags.StringVar(&opts.OldHome, "old-home", "",
+		"Path to the old PORTER_HOME directory where the previous version of Porter is installed")
+	flags.StringVar(&opts.OldStorageAccount, "old-account", "",
+		"Name of the storage account in the old Porter configuration file containing the data that should be migrated. If unspecified, the default storage account is used.")
+	flags.StringVarP(&opts.Namespace, "namespace", "n", "",
+		"Destination namespace where the migrated data should be saved.")
```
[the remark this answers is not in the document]
carolynvs: The default is a bit complicated. I tried to explain it a bit more in the storage-migrate.md document but I'll include some of that context here too.

pkg/storage/migrations/migration.go
carolynvs: I'll switch to upsert to avoid the problem.

pkg/storage/migrations/source_schema.go
carolynvs: These source data structs are copied directly from porter v0.38. The schema document only ever supported json, while parameter sets and credential sets supported json and yaml (since we read them from the user on commands like porter install). Porter v0.38 only ever stored parameter and credential sets as json files though when using the old filesystem storage plugin.

carolynvs:
@VinozzZ Thanks for the review! I have a few items fixed already but some need a bit more time. I'll let you know when this is ready to look at again.

pkg/storage/migrations/migration.go
```diff
@@ -0,0 +1,424 @@
...
+	}
+
+	// Use the most recent claim timestamp as the modified date of the installation
+	if inst.Status.Modified.IsZero() || inst.Status.Modified.Before(run.Created) {
+		inst.Status.Modified = run.Created
+	}
+
+	// Sanitize sensitive values on the source claim
+	bun := cnab.ExtendedBundle{Bundle: run.Bundle}
+	run.Parameters.Parameters, err = m.sanitizer.CleanParameters(ctx, run.Parameters.Parameters, bun, run.ID)
```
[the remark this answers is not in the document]
carolynvs: That's a great question. I think that the secrets protocol should define that create will essentially do a Create or Update (upsert). There are other scenarios where it's possible that Porter could retry an operation that would result in create being called on the same secret key.

pkg/storage/migrations/migration.go
```diff
@@ -0,0 +1,424 @@
...
+	}
+
+	output, err := convertOutput(result, outputKey, data)
+	if err != nil {
+		return span.Error(err)
+	}
+
+	// Sanitize sensitive outputs
+	bun := cnab.ExtendedBundle{Bundle: run.Bundle}
+	output, err = m.sanitizer.CleanOutput(ctx, output, bun)
```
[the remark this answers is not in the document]
carolynvs: I'd rather save to the secret store first, and possibly end up with a "dangling" secret than the flip side which would mean that we saved a reference to an external secret that we later find out we couldn't persist, leaving out database in a bad state. If we go with the secret protocol saying that Create really means CreateOrUpdate (upsert) then is this (dangling secrets) still a problem?

If we want to change how the secrets protocol / sanitizer works, I think it makes sense to articular the edge cases that we don't feel are well supported now and then change it in a separate PR so that it gets the consideration it requires.

carolynvs:
@VinozzZ I've created a new PR with some of the fixes for the problems you identified. Once that is merged I'll reset the more recent commits on this PR (since they are really unrelated to the migration). So please hold off on reviewing new commits until I've cleaned this one up.

pkg/storage/migrations/migration.go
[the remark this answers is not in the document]
VinozzZ: I think if the secret plugins is doing upsert, this is probably ok

cmd/porter/storage.go
```diff
@@ -52,7 +52,7 @@ This upgrades the data to the current storage schema, and does not change the da
 	flags.StringVar(&opts.OldStorageAccount, "old-account", "",
 		"Name of the storage account in the old Porter configuration file containing the data that should be migrated. If unspecified, the default storage account is used.")
 	flags.StringVarP(&opts.Namespace, "namespace", "n", "",
-		"Destination namespace where the migrated data should be saved.")
+		"Destination namespace where the migrated data should be saved. By default, Porter migrates your data into the current namespace as configured by environment variables and your config file, and otherwise the global namespace is used.")
```
VinozzZ, line 55: `and otherwise the global namespace is used.` --> `, otherwise the global namespace is used.~

docs/content/storage-migrate.md
```diff
@@ -35,7 +36,20 @@ The first version of Porter v1 that supports migrating old data from v0 is v1.0.
 You may choose to set the PORTER_HOME environment variable to change the default location, or omit setting that environment variable to install Porter v1 in the default directory **~/.porter**.
 The instructions on the installation page have you set PORTER_HOME to avoid accidentally overwriting your v0 directory but once you have moved your old directory it's safe to use the default installation location for Porter v1.
 
-## 3. Configure a Storage Plugin
+## 3. Configure a Secrets Plugin
+
+Porter v1 requires that you configure a secrets plugin.
+For non-production environments, you may use the [filesystem] secrets plugin.
```
VinozzZ, line 42: Is the filesystem meant to be a link to the doc  here?

tests/integration/testdata/migration/installation-show-sensitive-data-output.json
```diff
@@ -0,0 +1,39 @@
...
+    "bundleReference": "",
+    "bundleVersion": "0.0.1",
+    "bundleDigest": ""
+  },
+  "_calculated": {
+    "resolvedParameters": [
+      {
+        "name": "name",
+        "type": "string",
+        "sensitive": false,
```
VinozzZ, line 26: Should this be true if it's a sensitive data?

pkg/storage/migrations/migration.go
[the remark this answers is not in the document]
carolynvs: I'm fixing the k8s plugin to upsert 👍

tests/integration/testdata/migration/installation-show-sensitive-data-output.json
carolynvs: The name parameter is not defined as sensitive, just the password parameter.

https://github.com/getporter/examples/blob/571103890795923800762b132ac7b22afb28ff28/sensitive-data/porter.yaml#L7-L13

tests/integration/testdata/migration/secrets/01G6K8CZ08T78WXTJYHR0NTYBS-name
```diff
@@ -0,0 +1 @@
+example-bundle
```
VinozzZ, line 1: If the name isn't defined as sensitive, I'm a bit confused on why the secret plugin still created a secret for this parameter

carolynvs: There is both a parameter "name" and an output "name". The parameter is not sensitive. The output is though

https://github.com/getporter/examples/blob/571103890795923800762b132ac7b22afb28ff28/sensitive-data/porter.yaml#L18-L21

So we have one sensitive parameter, "password" and one sensitive output "name" which is why we got 2 secrets in the filesystem plugin.

Maybe the example bundle could be improved for clarity if we changed the output to be admin-password (instead of name) and have it persist the password parameter value? I think part of the problem may be that the parameter isn't sensitive but the output is and they both have the same name.

VinozzZ (approved):
💯 🥇

Issue #2081, opened 2022-05-19
Resolve lint and vet errors


carolynvs:
Once we have lint and vet targets from #2080, we need to identify which lint rules we want to follow. Not all will make sense for our project and desired coding style. So some may end up being ignored via staticcheck's [config](https://staticcheck.io/docs/configuration/) file.

Whoever grabs this should try to fix what makes sense, and then if the required fixes for some parts doesn't make the code better or makes it hard to read, then ignore those rules.

I don't think vet is configurable so we will need to just fix all of those.

See our [Contributing Tutorial](https://release-v1.porter.sh/contribute/tutorial) and [New Contributor Guide](https://github.com/getporter/porter/blob/release/v1/CONTRIBUTING.md) for help getting started contributing to Porter.

tchaudhry91:
@carolynvs 
I've got #2080 working locally, while I'm waiting for the requisite merge to happen there,
I've started fixing some of these, just confirming, this development should happen against the release/v1 branch?

carolynvs:
Thanks for your patience! I was busy last week getting a new release of Porter and the plugins out and didn't have time for reviews.

I've merged your magefiles PR and tagged a new release [v0.3.0](https://github.com/getporter/magefiles/releases/tag/v0.3.0) that you can use in the go.mod file.

tchaudhry91:
@carolynvs 
Can we reopen this?
Looks like a bunch of new Lint errors have also emerged after some other merges.
It may make sense to only close this once we also close #2082

carolynvs:
I tried swapping out the term package and it seems to work fine! 👍

PR #2153, opened 2022-06-12
Fix Vet Errors

tchaudhry91:
# What does this change
This PR is meant to eliminate the go vet warnings/errors in the codebase.
Present State:
1 big commit which fixes the most common recurring vet errors (unkeyed fields in composite literals). 
There are 2 major culprits here:
`ExtendedBundle` which has an embedded `Bundle` struct. Relatively straightforward fix here to add `Bundle` before supplying the embedded struct. Even I didn't know this was a thing, always thought embedded structs had to be anonymous :) 
Found the solution from here: https://stackoverflow.com/questions/49461354/go-vet-composite-literal-uses-unkeyed-fields-with-embedded-types

The other one is in the Primitive.E type in the mongo driver. We are generally passing Key and Value as a `{"some_key", "some_value"}` in all places. The recommended way is to go with `{Key: "some_key", Value: "some_value"}`. This feels a little extra verbose to me, but as highlighted in the link below, it offers good protection against changes in the library, so I suggest we keep it. Also fixes vet warnings :)
https://stackoverflow.com/questions/54548441/composite-literal-uses-unkeyed-fields

Update:
I have added another commit to resolve all the Struct Tag issues. Only minor typos here (although, it's surprising functionality wasn't broken in a few places). Nevertheless, all resolved now.

Update 2:
I have also renamed the example tests to use suffixes according to (https://pkg.go.dev/testing#hdr-Examples).
The old values were incompatible with Vet.

# What issue does it fix
Makes incremental progress on  #2081 

# Notes for the reviewer
This is a major chunk of the go vet warnings, but not all. I will resolve the rest via a separate PR. I kept this one segregated because of the theme.
Update: Also added a commit to resolve struct tag issues.

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist


pkg/porter/list.go
```diff
@@ -103,7 +103,7 @@ type DisplayInstallation struct {
 }
 
 type DisplayInstallationMetadata struct {
-	ResolvedParameters DisplayValues `json:"resolvedParameters", yaml:"resolvedParameters"`
+	ResolvedParameters DisplayValues `json:"resolvedParameters" yaml:"resolvedParameters"`
```
tchaudhry91, line 106: @carolynvs 
This was broken earlier and hence the tag wasn't being used properly earlier. I think this has resulted in a broken test. I'm not sure what the best way to resolve this would be. 
We could regenerate the golden file, but I don't know how "breaking" this is.

tchaudhry91:
I have added a comment regarding the broken unit-test. The malformed struct tag for yaml meant the file wasn't using the correct name. Will defer to someone with more knowledge to get the best way to fix this. Long term, we probably want to keep the correct tag and regenerate the golden comparison file. Not sure how much that will impact current consumers. Alternatively, we could use the "wrong" value in the tag if we want. Let me know.

pkg/porter/list.go
VinozzZ: Thanks for fixing it! We can use `mage updateTestFiles` to regenerate the golden files for tests

VinozzZ:
Thank you for submitting the PR to fix lint and vet errors!

> I have added a comment regarding the broken unit-test. The malformed struct tag for yaml meant the file wasn't using the correct name. Will defer to someone with more knowledge to get the best way to fix this. Long term, we probably want to keep the correct tag and regenerate the golden comparison file. Not sure how much that will impact current consumers. Alternatively, we could use the "wrong" value in the tag if we want. Let me know.

Thank you for putting up the PR for fixing lint and vet errors.  From my perspective,, we should fix all the schema issues while we are still in alpha. but we probably should wait for  @carolynvs to confirm it

tchaudhry91:
Hey all,
This PR is now complete from my point of view, here is a summary for the reviewers:
1. Fixes all the unkeyed fields in the composite literals.
2. Fixes the struct tags. This is a bit more complicated as this will technically be a breaking change? This is because the old tags were incorrect and hence ignored. So far, this has been detected in one place in a test file, but I assume others would also be impacted. For now, I have regenerated that test file and the tests pass now.
3. Renamed ExampleTests to use PackageName_suffix format as dictated by Vet.

The unit test failure in the PR seems to be due to a connection reset, i.e flaky. Can someone re-run it? I have confirmed that unit tests run fine on multiple machines.

carolynvs (commented):
Thank you for going over the codebase and getting all the vet errors fixed! Let me know if you'd like to add that constructor function and I'll hold off on merging.

carolynvs:
I restarted the test run, it was a temporary glitch in the matrix. 😀

tchaudhry91:
@carolynvs I'll add the constructor in this PR only. Let's hold off on merging till then. Shouldn't be long

pkg/cache/cache_test.go
```diff
@@ -196,7 +196,7 @@ func TestStoreManifest(t *testing.T) {
 	}{
 		{
 			name: "embedded manifest",
-			bundle: cnab.ExtendedBundle{bundle.Bundle{
+			bundle: cnab.ExtendedBundle{Bundle: bundle.Bundle{
```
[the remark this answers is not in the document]
tchaudhry91: Sounds good! I'll make this change.

tchaudhry91:
@carolynvs I have added the constructor and made the replacements in all the non-zero initializations.
For consistency it's probably better to come up with a something to replace all the `ExtendedBundle{}` symbols too. Although, these are almost exclusively used to return zero values along with an error, in which case this value should never be referenced, however, for readability, I suppose a mix on `NewBundle` and `ExtendedBundle{}` might be a bit confusing. Nevertheless, if we want to replace those too, I can do in a new PR. This one is getting a bit unwieldy already :)

PR #2154, opened 2022-06-13
Sanitize archive folder name

### LINKED ISSUE #2151: archive command fails when the bundle name has a slash
## Describe the bug
When I try to archive a bundle that has a slash in the bundle name, archive fails.

## To Reproduce
```console
$ porter archive whalegap.tgz --reference ghcr.io/getporter/examples/whalegap:v0.2.0
Writing logs to /Users/carolynvs/.porter/logs/01G57B1W09P2FB11DQHJYRPBQV.json
Writing logs to /Users/carolynvs/.porter/logs/01G57B1W09P2FB11DQHJYRPBQV.json
Writing logs to /Users/carolynvs/.porter/logs/01G57B1W09P2FB11DQHJYRPBQV.json
Writing logs to /Users/carolynvs/.porter/logs/01G57B1W09P2FB11DQHJYRPBQV.json
mkdir /var/folders/mv/cx7sy9c54wn741p4x6v0_t100000gn/T/examples/whalegap-0.2.0350843934: no such file or directory
```

## Expected behavior
The command should succeed. Seems like the problem is that we are making a directory with the name of the bundle and it's causing a missing parent directory.

Let's fix this by cleaning up the bundle name before using it as a directory name. There may be other allowed characters in a bundle name by the CNAB spec that could also cause trouble when put in a path. Maybe just do a string replace for anything that isn't a-z0-9-_. and turn it into underscores?

## Version
1.0.0-alpha.20

VinozzZ:
# What does this change

Currently, porter uses bundle name as part of the temporary archive folder for storing all artifacts and metadata that's needed for archiving a bundle.
The bundle name may contain invalid path characters, which causes `porter archive` to fail.

This PR fixes above issue by sanitizing bundle name to remove non-standard path characters before using it as part of the archive folder path.
If a path separator is present in a bundle name, we will replace them with a dash, `-`

# What issue does it fix
Closes #2151 

# Notes for the reviewer

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/porter/archive.go
+++ b/pkg/porter/archive.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"get.porter.sh/porter/pkg"
@@ -185,22 +186,18 @@ func (ex *exporter) addImage(image bundle.BaseImage) error {
 
 // createArchiveFolder set up a temporary directory for storing all data needed to archive a bundle.
 // It sanitizes the name and make sure only the current user has full permission to it.
-// If the name contains a path separator, all transit paths are created.
+// If the name contains a path separator, all path separators will be replaced with "-".
 func (ex *exporter) createArchiveFolder(name string) (string, error) {
-	cleanedPath := afero.UnicodeSanitize(name)
-	tmpDir := filepath.Join(os.TempDir(), filepath.Dir(cleanedPath))
-	if err := ex.fs.MkdirAll(tmpDir, 0744); err != nil {
-		return "", err
-	}
-
-	archiveDir, err := ex.fs.TempDir(tmpDir, filepath.Base(cleanedPath))
+	cleanedPath := strings.ReplaceAll(afero.UnicodeSanitize(name), string(os.PathSeparator), "-")
+	archiveDir, err := ex.fs.TempDir("", cleanedPath)
 	if err != nil {
-		return "", err
-	}
-	if err := ex.fs.Chmod(archiveDir, 0744); err != nil {
-		return "", err
+		return "", fmt.Errorf("can not create a temporary archive folder: %w", err)
 	}
 
+	err = ex.fs.Chmod(archiveDir, pkg.FileModeDirectory)
+	if err != nil {
+		return "", fmt.Errorf("can not change permission for the temporary archive folder: %w", err)
+	}
 	return archiveDir, nil
 }
 
--- a/pkg/porter/archive.go
+++ b/pkg/porter/archive.go
@@ -16,6 +16,7 @@ import (
 	"github.com/cnabio/cnab-go/imagestore/construction"
 	"github.com/docker/docker/pkg/archive"
 	"github.com/pkg/errors"
+	"github.com/spf13/afero"
 )
 
 // ArchiveOptions defines the valid options for performing an archive operation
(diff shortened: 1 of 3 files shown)


pkg/porter/archive.go
```diff
@@ -186,6 +183,27 @@ func (ex *exporter) addImage(image bundle.BaseImage) error {
...
+	tmpDir := filepath.Join(os.TempDir(), filepath.Dir(cleanedPath))
+	if err := ex.fs.MkdirAll(tmpDir, 0744); err != nil {
+		return "", err
+	}
+
+	archiveDir, err := ex.fs.TempDir(tmpDir, filepath.Base(cleanedPath))
+	if err != nil {
+		return "", err
+	}
+	if err := ex.fs.Chmod(archiveDir, 0744); err != nil {
```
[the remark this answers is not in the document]
VinozzZ: I changed the handling logic and it should make more sense now

pkg/porter/archive.go
```diff
@@ -186,6 +183,27 @@ func (ex *exporter) addImage(image bundle.BaseImage) error {
 	return checkDigest(image, dig)
 }
 
+// createArchiveFolder set up a temporary directory for storing all data needed to archive a bundle.
+// It sanitizes the name and make sure only the current user has full permission to it.
+// If the name contains a path separator, all transit paths are created.
+func (ex *exporter) createArchiveFolder(name string) (string, error) {
+	cleanedPath := afero.UnicodeSanitize(name)
+	tmpDir := filepath.Join(os.TempDir(), filepath.Dir(cleanedPath))
```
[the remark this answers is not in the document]
VinozzZ: The reason why I couldn't directly use the `TmpDir` is because it doesn't create nested folders. I changed the function to replace all path separators to a dash. Let me know what do u think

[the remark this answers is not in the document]
carolynvs: That sounds like a good way to deal with it, thanks! 👍

pkg/porter/archive_test.go
```diff
@@ -51,3 +52,18 @@ func TestArchive_Validate(t *testing.T) {
...
+	p := NewTestPorter(t)
+	defer p.Close()
+	ex := exporter{
+		fs: p.FileSystem,
+	}
+
+	dir, err := ex.createArchiveFolder("examples/test-bundle-0.2.0")
+	require.NoError(t, err)
+
+	info, err := ex.fs.Stat(dir)
```
carolynvs, line 66: Can you add one more assert that checks the name created? That will help check for replacing the path seperator with a dash.

carolynvs (changes requested):
Looks great! I just have one suggestion for better test coverage and we are all set!

pkg/porter/archive_test.go
```diff
@@ -51,3 +52,18 @@ func TestArchive_Validate(t *testing.T) {
...
+	require.NoError(t, err)
+	require.Equal(t, pkg.FileModeDirectory.String(), info.Mode().Perm().String(), dir)
```
carolynvs, line 68: We also have a test helper that gives you a better error message whe comparing file modes 

```
import "get.porter.sh/porter/tests"

... 

tests.AssertDirectoryPermissionsEqual(...)
```

carolynvs: You don't have to change the code to use it, I just wanted you to be aware of the helpers since they are spread out. We really should consolidate them eventually! 😅

VinozzZ: Oh, I did not notice we have this nice helper function. Thank your for mentioning it. I just changed to use this helper funcion cuz it's so nice

carolynvs (approved):
Thanks for fixing this! 🎉

PR #2176, opened 2022-06-20
staticcheck lint fixes

### LINKED ISSUE #2081: Resolve lint and vet errors

tchaudhry91:
# What does this change
A bunch of staticcheck fixes that I felt were straightforward. Some maybe debatable and can be "ignored" or worked around. Those can be discussed in the PR.

# What issue does it fix
#2081 - Incremental Progress

# Notes for the reviewer
Some fairly straightforward things unchecked errors etc have been fixed. A couple of potential bugs as well (checking err instead of !ok).

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -17,10 +17,9 @@ import (
 )
 
 var (
-	kahn1dot0Hash  = "887e7e65e39277f8744bd00278760b06"
-	kahn1dot01     = cnab.MustParseOCIReference("deislabs/kubekahn:1.0")
-	kahnlatestHash = "fd4bbe38665531d10bb653140842a370"
-	kahnlatest     = cnab.MustParseOCIReference("deislabs/kubekahn:latest")
+	kahn1dot0Hash = "887e7e65e39277f8744bd00278760b06"
+	kahn1dot01    = cnab.MustParseOCIReference("deislabs/kubekahn:1.0")
+	kahnlatest    = cnab.MustParseOCIReference("deislabs/kubekahn:latest")
 )
 
 func TestFindBundleCacheExists(t *testing.T) {
@@ -65,6 +64,7 @@ func TestFindBundleBundleCached(t *testing.T) {
 	expectedCacheCNABDirectory := filepath.Join(expectedCacheDirectory, "cnab")
 	expectedCacheFile := filepath.Join(expectedCacheCNABDirectory, "bundle.json")
 	foundIt, err := cfg.Config.FileSystem.Exists(expectedCacheFile)
+	require.NoError(t, err, "the cache dir should exist, no error should have happened")
 	require.True(t, foundIt, "test data not loaded")
 	c := New(cfg.Config)
 
@@ -95,6 +95,7 @@ func TestCacheWriteNoCacheDir(t *testing.T) {
 
 	c := New(cfg.Config)
 	cb, err := c.StoreBundle(cnab.BundleReference{Reference: kahn1dot01, Definition: bun})
+	assert.NoError(t, err, "storing bundle should have succeeded")
 
 	home, err := cfg.Config.GetHomeDir()
 	require.NoError(t, err, "should have had a porter home dir")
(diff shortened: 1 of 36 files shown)


pkg/cache/cache_test.go
```diff
@@ -181,6 +181,7 @@ func TestStoreRelocationMapping(t *testing.T) {
 			assert.Equal(t, tc.wantedReloPath, cb.RelocationFilePath, "didn't get expected path for store")
 
 			cb, _, err = c.FindBundle(tc.tag)
+			assert.NoError(t, err, fmt.Sprintf("didn't expect find bundle error for test %s", tc.tag))
```
carolynvs, line 184: You don't have to change this but I wanted you to be aware that the assert.NoError function will handle formatting for you. So this line can be simplified to 
```suggestion
			assert.NoError(t, err, "didn't expect find bundle error for test %s", tc.tag)
```

pkg/pkgmgmt/client/filesystem.go
```diff
@@ -45,6 +45,9 @@ type FileSystem struct {
 
 func (fs *FileSystem) List() ([]string, error) {
 	parentDir, err := fs.GetPackagesDir()
+	if err != nil {
+		return nil, errors.Wrap(err, "could not get package directory")
```
carolynvs, line 49: Can you switch this to use fmt.Errorf and the "%w" placeholder instead of the pkg/errors package? We are trying to stop using the errors package (#2009).

pkg/pkgmgmt/feed/generate.go
```diff
@@ -167,6 +167,9 @@ func (feed *MixinFeed) Save(opts GenerateOptions) error {
 	tmplData["Updated"] = entries[0].Updated()
 
 	atomXml, err := mustache.Render(string(feedTmpl), tmplData)
+	if err != nil {
+		return errors.Wrap(err, "error rendering template")
```
carolynvs, line 171: Can you use fmt.Errorf + "%w" instead of the errors package here?

pkg/porter/upgrade.go
```diff
@@ -72,6 +72,9 @@ func (p *Porter) UpgradeBundle(ctx context.Context, opts UpgradeOptions) error {
 	}
 
 	err = p.applyActionOptionsToInstallation(ctx, &i, opts.BundleActionOptions)
+	if err != nil {
+		return errors.Wrap(err, "could not apply options to installation")
```
carolynvs, line 76: Can you use fmt.Errorf + "%w" instead of the errors package here?

carolynvs (changes requested):
Looks good! I just have a couple small changes that need to be made and this is ready to merge.

tchaudhry91:
@carolynvs completed changes as requested!

carolynvs (approved):
Thanks for cleaning up all the lint errors! This is really helpful and we can finally enforce no lint errors going forward in our builds. 💪

tchaudhry91:
Hehe. Thanks @carolynvs but there are a few staticcheck fixes still left before we enforce lint in the CI. The remaining changes may require your input though or atleast I felt like there was something not totally straightforward with those.

carolynvs:
@tchaudhry91 oops, I must have missed that. Are we tracking that problem in another issue or pull request?

tchaudhry91:
@carolynvs we can track it in #2081 only. I'll update it with the remaining fixes.

PR #2186, opened 2022-06-21
Update relocation map during porter publish

### LINKED ISSUE #1819: Publish from archive should use a relocation map
I was reviewing [this PR](https://github.com/getporter/porter/pull/1815) and that we don't use a relocation mapping file and preserve the original bundle.

See this very wonderfully long function comment for context:

https://github.com/getporter/porter/blob/784d61cf283bbd5b9c05a42ea718d3d042803305/pkg/porter/publish.go#L172-L186

This PR is correcting a bug in Porter where when we copy a bundle from one registry to another, we aren't preserving the relocation map. I was checking all the code that I thought would use that function and wasn't clear on why we are essentially "re-publishing" the bundle from the archive instead of preserving the original.

_Originally posted by @carolynvs in https://github.com/getporter/porter/issues/1815#issuecomment-963296356_

Vaughn confirmed that we should update the implementation of publish from archive to preserve the original bundle and use a relocation mapping file.

VinozzZ:
# What does this change

Currently, when publishing from a bundle, porter generates a new bundle that don't preserve content digests. 
This PR updates the logic to use relocation maps in the original bundle to associate the old images with the newly copied images.

# What issue does it fix
Closes #1819 

# Notes for the reviewer

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/porter/publish.go
+++ b/pkg/porter/publish.go
@@ -231,7 +231,7 @@ func (p *Porter) publishFromArchive(ctx context.Context, opts PublishOptions) er
 	// Push updated images (renamed based on provided bundle tag) with same digests
 	// then update the bundle with new values (image name, digest)
 	for _, invImg := range bundleRef.Definition.InvocationImages {
-		relocMap, err := p.updateRelocationMapping(bundleRef.RelocationMap, layout, invImg.Image, opts.Reference)
+		relocMap, err := p.relocateImage(bundleRef.RelocationMap, layout, invImg.Image, opts.Reference)
 		if err != nil {
 			return err
 		}
@@ -239,7 +239,7 @@ func (p *Porter) publishFromArchive(ctx context.Context, opts PublishOptions) er
 		bundleRef.RelocationMap = relocMap
 	}
 	for _, img := range bundleRef.Definition.Images {
-		relocMap, err := p.updateRelocationMapping(bundleRef.RelocationMap, layout, img.Image, opts.Reference)
+		relocMap, err := p.relocateImage(bundleRef.RelocationMap, layout, img.Image, opts.Reference)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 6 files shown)


carolynvs:
@VinozzZ Sorry it took me a bit to review this and in the meantime there are merge conflicts. Can you resolve them and then I'll do a review?

pkg/porter/publish.go
```diff
@@ -416,6 +357,31 @@ func (p *Porter) rewriteBundleWithInvocationImageDigest(ctx context.Context, m *
 	return bun, nil
 }
 
+func (p *Porter) updateRelocationMapping(relocationMap relocation.ImageRelocationMap, layout registry.Layout, originImg string, newReference string) (relocation.ImageRelocationMap, error) {
```
carolynvs, line 360: The function name made me initially assume that this function only updated the relocation mapping file, and it never ocurred to me that it would also push the image. What do you think about changing the function name so that it's more obvious what it does?

```suggestion
func (p *Porter) relocateImage(relocationMap relocation.ImageRelocationMap, layout registry.Layout, originImg string, newReference string) (relocation.ImageRelocationMap, error) {
```

pkg/porter/publish_test.go
```diff
@@ -145,51 +147,56 @@ func TestPublish_getNewImageNameFromBundleReference(t *testing.T) {
...
 	require.NoError(t, err, "should have successfully created a digest")
 
-	// update invocation image
-	newInvImgName, err := getNewImageNameFromBundleReference(bun.InvocationImages[0].Image, tag)
-	require.NoError(t, err, "should have successfully derived new image name from bundle tag")
+	testcases := []struct {
+		description   string
+		relocationMap relocation.ImageRelocationMap
+		layout        registry.Layout
+		hasError      bool
```
carolynvs, line 163: Let's change this to use `wantError string` so that we can check that it failed for the expected reason.

tests/integration/archive_test.go
```diff
@@ -77,6 +77,8 @@ func TestArchive_StableDigest(t *testing.T) {
 
 	// Use a fixed bundle to work with so that we can rely on the registry and layer digests
 	const reference = "ghcr.io/getporter/examples/whalegap:v0.2.0"
+	const referencedImg = "carolynvs/whalesayd@sha256:8b92b7269f59e3ed824e811a1ff1ee64f0d44c0218efefada57a4bebc2d7ef6f"
```
carolynvs, line 80: I was expecting that these extra validations would be added to the Airgapped test above, since this feature supports airgap and testing it requires that we are making it impossible to use the original images by taking the original registry offline.

The test you are editing is for checking that when we archive twice, we get the same digest.

Can you move these checks into the other test?

carolynvs (changes requested):

pkg/porter/publish.go
VinozzZ: That's good point and a great name. I will change it

tests/integration/archive_test.go
VinozzZ: Sounds good

carolynvs (approved):
LGTM

PR #2189, opened 2022-06-22
Set the uninstalled flag on the installation

carolynvs:
I made a mistake and never set the uninstalled flag when I added this field to installation status. It's now being set when a bundle run is applied to the installation.

### DIFF
--- a/pkg/porter/runs_test.go
+++ b/pkg/porter/runs_test.go
@@ -71,32 +71,36 @@ func TestPorter_PrintInstallationRunsOutput(t *testing.T) {
 	}
 
 	for _, tc := range outputTestcases {
-		p := NewTestPorter(t)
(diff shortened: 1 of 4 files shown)


pkg/porter/runs_test.go
```diff
@@ -71,32 +71,36 @@ func TestPorter_PrintInstallationRunsOutput(t *testing.T) {
 	}
 
 	for _, tc := range outputTestcases {
-		p := NewTestPorter(t)
-		defer p.Close()
-		ctx := context.Background()
+		tc := tc
```
carolynvs, line 74: this is a bit unreleated, but I saw that we were calling defer in a for loop and this just fixes that quick

VinozzZ (approved):

PR #2214, opened 2022-06-29
Only load mixin metadata from used mixins

### LINKED ISSUE #2212: Porter calls all the mixins during build
## Describe the bug
When I run porter build, if I have a broken mixin installed (such as the old helm mixin), that can mess things up because every single mixin in my PORTER_HOME is called during build.

## To Reproduce
1. Run porter create
2. Run porter build --debug

Note that it outputs that it is calling every installed mixin and not just exec.

## Expected behavior
Only exec mixin should be called.

## Version
v1.0.0-alpha.22

carolynvs:
# What does this change
When we build a bundle, we record metadata about each mixin used in the bundle. Previously we were first retrieving metadata from every single installed mixin, and then filtering that by the mixins used.

I've changed the build command so that instead we first determine which mixins to query, and then we only ask those mixins for their metadata. This avoids problems and unhelpful error messages when installed, but unused, mixins are queried and return an error. It also just speeds up build which is nice too.

# What issue does it fix
Closes #2212 

# Notes for the reviewer
N/A

# Checklist
- [x] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/porter/build.go
+++ b/pkg/porter/build.go
@@ -18,6 +18,7 @@ import (
 	"get.porter.sh/porter/pkg/tracing"
 	"github.com/Masterminds/semver/v3"
 	"github.com/opencontainers/go-digest"
+	"golang.org/x/sync/errgroup"
 )
 
 type BuildOptions struct {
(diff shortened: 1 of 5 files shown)


pkg/porter/build.go
```diff
@@ -188,19 +189,33 @@ func (p *Porter) preLint(ctx context.Context, file string) error {
 }
 
 func (p *Porter) getUsedMixins(ctx context.Context, m *manifest.Manifest) ([]mixin.Metadata, error) {
-	installedMixins, err := p.ListMixins(ctx)
+	ctx, span := tracing.StartSpan(ctx)
+	defer span.EndSpan()
+
+	usedMixins := make([]mixin.Metadata, len(m.Mixins))
```
VinozzZ, line 195: nit - can We move this to be closer to where it's being used before line 216?

pkg/porter/build.go
```diff
@@ -188,19 +189,33 @@ func (p *Porter) preLint(ctx context.Context, file string) error {
...
+
+	g := new(errgroup.Group)
+	results := make(chan *mixin.Metadata, len(m.Mixins))
```
VinozzZ, line 198: Is there a particular reason why want to have a pointer here and then de-reference it later on line 207

carolynvs: It was just a matter of where we wanted to dereference the pointer that was stored in that PackageMetadata interface. I've updated it to do less conversions.

VinozzZ (approved):

PR #2216, opened 2022-06-30
Playground: Plugin Library

carolynvs:
I'm playing around with making a library for making plugins so that people don't have to implement the boilerplate stuff.

### DIFF
--- a/cmd/porter/aliases.go
+++ b/cmd/porter/aliases.go
@@ -3,6 +3,7 @@ package main
 import (
 	"strings"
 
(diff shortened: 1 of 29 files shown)


carolynvs:
This is the new tiny main function for porter, and it would look the same for a plugin
```go
func main() {
	ctx := context.Background()
	app := porter.New()
	rootCmd := buildRootCommandFrom(app)

	cli.Main(ctx, rootCmd, app)
}
```

It shoves in all the logic we have for telemetry, context cancellation, panic handling, etc into this new package pkg/cli

PR #2117, opened 2022-05-31
change dependency schema in porter

VinozzZ:
# What does this change
This PR updates the dependency schema to prepare for the advanced dependency feature
https://github.com/getporter/proposals/blob/main/pep/003-dependency-namespaces-and-labels.md#porteryaml

# What issue does it fix

[1]: https://porter.sh/src/CONTRIBUTING.md#when-to-open-a-pull-request

# Notes for the reviewer
We want to make this change before v1 release so that users don't have to do another schema update immediately after v1

# Checklist
- [ ] Did you write tests?
- [ ] Did you write documentation?
- [ ] Did you change porter.yaml or a storage document record? Update the corresponding schema file.
- [ ] If this is your first pull request, please add your name to the bottom of our [Contributors][contributors] list. Thank you for making Porter better! 🙇‍♀️

# Reviewer Checklist
* Comment with /azp run test-porter-release if a magefile or build script was modified
* Comment with /azp run porter-integration if it's a non-trivial PR

[contributors]: https://porter.sh/src/CONTRIBUTORS.md

PR #2232, opened 2022-07-06
Fix dependencies v1 definition and conversion

carolynvs:
# What does this change
When I was implementing the dependencies v2 extension (advanced dependencies proposal), I realized that we had a regression in the v1 extension recently when we updated the porter.yaml schema to move reference and version underneath the bundle field, e.g. 

```yaml
dependencies:
  - name: mydep
     bundle:
        reference: getporter/mydep
        version: 1.0.0
```

I had to make two fixes to get the original test passing properly, plus update the test since there was a problem with closures that was making it pass when it shouldn't have.

[Fix yaml tag for dependency version](https://github.com/getporter/porter/commit/5baac5168a216f998e255b30c3711e0b0e737d46)

There was a typo in the yaml tag for a dependency's version in porter.yaml. It has always been version, and we intended to just move reference and version underneath a new bundle field. It was accidentally set to versions, but none of the other code, tests or examples were
updated to that value (yay!)

This reverts the yaml tag back to version (singular). I'm not bumping the schema version since this actually matches what the schema has always been, it's just fixing a bug where version completely didn't work for a bit.

[Set AllowPrerelease on deps v1 extension](https://github.com/getporter/porter/commit/d78859a564e08f84150f1db2bc1bc1b224556d1f)

When we are converting a porter.yaml to a bundle.json and populating the v1 dependencies extension metadata, we accidentally stopped populating AllowPrerelease (which was in use previously). Before we updated the dependency schema in preparation for a v2 of the dependency extension this value was set explicitly in the porter.yaml. When we removed it, we
decided to use the mastermind/semver convention of "if a version constraint includes a prerelease, then we will match against releases", i.e. 1.0.0-0 would allow prereleases, but we missed updating the conversion logic to explicitly set AllowPrerelease in that situation.

I've updated our testdata to work with the new logic and verified that we are now properly setting AllowPrerelease again.

# What issue does it fix
This is a fix/follow-up to #2117 

# Notes for the reviewer

# Checklist
- [x] Did you write tests?
- [ ] Did you write documentation? This fixed the code to match the schema/docs

# Reviewer Checklist

### DIFF
--- a/pkg/cnab/config-adapter/adapter.go
+++ b/pkg/cnab/config-adapter/adapter.go
@@ -11,6 +11,7 @@ import (
 	"get.porter.sh/porter/pkg/manifest"
 	"get.porter.sh/porter/pkg/mixin"
 	"get.porter.sh/porter/pkg/tracing"
+	"github.com/Masterminds/semver/v3"
 	"github.com/cnabio/cnab-go/bundle"
 	"github.com/cnabio/cnab-go/bundle/definition"
 )
@@ -71,7 +72,11 @@ func (c *ManifestConverter) ToBundle(ctx context.Context) (cnab.ExtendedBundle, 
 	b.Outputs = c.generateBundleOutputs(ctx, &b.Definitions)
 	b.Credentials = c.generateBundleCredentials()
 	b.Images = c.generateBundleImages()
-	b.Custom = c.generateCustomExtensions(&b)
+	custom, err := c.generateCustomExtensions(&b)
+	if err != nil {
+		return cnab.ExtendedBundle{}, err
+	}
+	b.Custom = custom
 	b.RequiredExtensions = c.generateRequiredExtensions(b)
 
 	b.Custom[config.CustomPorterKey] = stamp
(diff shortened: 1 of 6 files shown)


pkg/cnab/config-adapter/adapter.go
```diff
@@ -401,10 +406,9 @@ func (c *ManifestConverter) generateBundleImages() map[string]bundle.Image {
 	return images
 }
 
-func (c *ManifestConverter) generateDependencies() *cnab.Dependencies {
-
+func (c *ManifestConverter) generateDependencies() (*cnab.Dependencies, error) {
```
VinozzZ, line 409: Maybe I'm missing something, I can't find where this function is returning an error?

pkg/cnab/config-adapter/adapter.go
```diff
@@ -421,12 +425,18 @@ func (c *ManifestConverter) generateDependencies() *cnab.Dependencies {
 			dependencyRef.Version = &cnab.DependencyVersion{
 				Ranges: []string{dep.Bundle.Version},
 			}
+
+			// If we can detect that prereleases are used in the version, then set AllowPrereleases to true
+			v, err := semver.NewVersion(dep.Bundle.Version)
```
VinozzZ, line 430: Should we return the error here if it's not nil?

pkg/cnab/config-adapter/adapter.go
carolynvs: Sorry the next PR after this needed to be able to return an error. Are you okay with me keeping that in place here so that I don't need to redo both PRs to shuffle when we introduce that change?

VinozzZ (approved):

Issue #1429, opened 2021-01-28
Add verbosity flag


carolynvs:
**What design is being proposed?**
Replace `--debug` and `--debug-plugins` and `porter build --verbose` with a single `--verbosity|v` flag.

https://github.com/getporter/porter/discussions/1353

🚨 This needs a PEP.

carolynvs:
We still need this in v1 for two reasons: backwards incompatible changes to output format and flags, and if anyone is using porter in production we really want the code fully instrumented so we can troubleshoot.

carolynvs:
I'd like to have two configurable settings: a flag to control how verbose the console output is, and another to control how much is sent to the tracer/logger. That way I can run at a level that isn't overwhelming and if a command fails go look up the extra info from the log file or jaeger.

PR #2242, opened 2022-07-12
Context plumbing

carolynvs:
# What does this change
Refactoring that adds context to a bunch of functions and uses it to do logging in preparation for #1429 

# What issue does it fix
Part of #1429

# Notes for the reviewer
I will squash this when it is merged.

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist


pkg/porter/build.go
```diff
@@ -94,24 +94,20 @@ func (o *BuildOptions) parseCustomInputs() error {
 }
 
 func (p *Porter) Build(ctx context.Context, opts BuildOptions) error {
-	ctx, log := tracing.StartSpan(ctx)
-	defer log.EndSpan()
-
-	opts.Apply(p.Context)
```
VinozzZ, line 100: Do you mean to remove this line of code?

VinozzZ, line 100: Do you mean to remove this line of code?

pkg/porter/porter.go
```diff
@@ -100,10 +99,6 @@ func (p *Porter) Connect(ctx context.Context) error {
 
 // Close releases resources used by Porter before terminating the application.
 func (p *Porter) Close() error {
-	if p.Debug {
```
VinozzZ, line 103: Do we not want to print this to debug level log anymore?

pkg/exec/builder/action.go
```diff
@@ -73,20 +75,22 @@ func unmarshalActionMap(actionMap map[string][]interface{}, builder BuildableAct
 //		 err := yaml.Unmarshal(contents, &action)
 //		 return &action, err
 //	 })
-func LoadAction(cxt *portercontext.Context, commandFile string, unmarshal func([]byte) (interface{}, error)) error {
-	contents, err := readInputFromStdinOrFile(cxt, commandFile)
+func LoadAction(ctx context.Context, porterCtx *portercontext.Context, commandFile string, unmarshal func([]byte) (interface{}, error)) error {
+	//lint:ignore SA4006 ignore unused ctx for now
```
VinozzZ, line 79: What's the reason why we aren't just ignoring the return value using `_, span`

pkg/exec/builder/action.go
```diff
@@ -73,20 +75,22 @@ func unmarshalActionMap(actionMap map[string][]interface{}, builder BuildableAct
...
 	result, err := unmarshal(contents)
-	if cxt.Debug {
-		fmt.Fprintf(cxt.Err, "DEBUG Parsed Input:\n%#v\n", result)
-	}
 	if err != nil {
-		return fmt.Errorf("could not unmarshal input:\n %s: %w", string(contents), err)
+		return span.Error(fmt.Errorf("could not unmarshal input:\n %s: %w", string(contents), err))
 	}
 
+	span.Debugf("Parsed Mixin Input:\n%#v", result)
```
VinozzZ, line 93: Will this have potential to print out sensitive data?

pkg/porter/credentials.go
```diff
@@ -293,15 +312,19 @@ type CredentialDeleteOptions struct {
 // DeleteCredential deletes the credential set corresponding to the provided
 // names.
 func (p *Porter) DeleteCredential(ctx context.Context, opts CredentialDeleteOptions) error {
+	ctx, span := tracing.StartSpan(ctx,
+		attribute.String("namespace", opts.Namespace),
+		attribute.String("name", opts.Name),
+	)
+	span.EndSpan()
```
VinozzZ, line 319: missing defer here

pkg/porter/cnab.go
```diff
@@ -331,10 +329,9 @@ func (o *sharedOptions) combineParameters(ctx context.Context) map[string]string
 	}
 
 	//
-	// Default the porter-debug param based on the verbosity level
+	// Default the porter-debug param to --debug
```
VinozzZ, line 332: Did we change the code back to the original code here?

pkg/porter/cnab_test.go
```diff
@@ -190,12 +190,13 @@ func TestSharedOptions_LoadParameters(t *testing.T) {
 }
 
 func TestSharedOptions_CombineParameters(t *testing.T) {
-	ctx := context.Background()
+	c := portercontext.NewTestContext(t)
```
VinozzZ, line 193: same here

pkg/exec/builder/action.go
carolynvs: Let's say that we did that to make lint happy. Then later someone needs to use the context variable. If they just use ctx, they will be using the context for the wrong scope, since we are throwing away the properly scoped context value on this line. They would also need to realize that they need to change that line. I don't want to make it easy to introduce a bug like that just to make lint happy. I'd rather set developers up to be successful and use a comment to make lint be quiet about it. 😁

pkg/exec/builder/action.go
carolynvs: I had removed some other prints because of the same problem and must have missed this. If you see more elsewhere, let's make sure to remove debug statements that could be printing out injected creds/params!

pkg/porter/cnab.go
carolynvs: Yes, per our discussion last week we are going to keep a --debug flag on the commands that run bundles. The PORTER_DEBUG parameter should be set off of that and not the verbosity flag. I have a follow-up PR after this to change how we check debug based on the new flag but it required more refactoring.

pkg/porter/cnab_test.go
carolynvs: I realized that the change I had made to this function (combineParameters) was because I _thought_ I needed access to the verbosity level. Since I really just needed the debug flag, I reverted the change.

I'll back this out and do a separate PR with the changes needed once we have a dedicated --debug flag on the porter install/upgrade/uninstall commands, separate from the debug flag that is being deprecated on the porter context.

pkg/porter/build.go
carolynvs: Yes but it should have been in my follow-up PR. This snuck in too early! I'll revert this line and remove it in the other PR where I clean up where the --debug flag is defined.

pkg/porter/porter.go
carolynvs: Yeah it's not a helpful debug statement and would require that I start passing context.Context to Close which you had previously encouraged me to not do. 😀

pkg/porter/cnab.go
carolynvs: OK, if you look at an individual commit you may see changes (because I'm avoiding a force push while you are reviewing) but if you diff the PR as a whole, this file is no longer modified. That was really confusing! 😅

VinozzZ: Ohh, yes, I was looking through each individual commits. Thanks for the clarification!

pkg/porter/credentials.go
```diff
@@ -328,45 +351,45 @@ func validateCredentialName(args []string) error {
...
-		return err
+		return span.Error(err)
 	}
 
-	if p.Debug {
+	if span.ShouldLog(zapcore.DebugLevel) {
 		// ignoring any error here, printing debug info isn't critical
 		contents, _ := p.FileSystem.ReadFile(o.File)
-		fmt.Fprintf(p.Err, "Input file contents:\n%s\n", contents)
+		span.Debugf("Input file contents:\n%s", contents)
```
VinozzZ, line 366: Will this print out the sensitive hard coded value in the file?

pkg/porter/parameters.go
```diff
@@ -497,45 +500,45 @@ func (p *Porter) printDisplayValuesTable(values []DisplayValue) error {
...
-		return err
+		return span.Error(err)
 	}
 
-	if p.Debug {
+	if span.ShouldLog(zapcore.DebugLevel) {
 		// ignoring any error here, printing debug info isn't critical
 		contents, _ := p.FileSystem.ReadFile(o.File)
-		fmt.Fprintf(p.Err, "Input file contents:\n%s\n", contents)
+		span.Debugf("Input file contents:\n%s", contents)
```
VinozzZ, line 515: same here

carolynvs:
@VinozzZ Ok I've removed those print statements.

VinozzZ (approved):

Issue #1430, opened 2021-01-28
Sweep CLI API for inconsistencies before v1.0


carolynvs:
**What design is being proposed?**
Some of our flags evolved over time and are inconsistent. Our last chance to fix that is before the v1.0 release. Let's look at our entire flag and command set, and create a PEP for any changes.

For example, here are some confusing flags

* Specify a credential set, `--cred|c`
* Specify a parameter set, `--parameter-set|p`
* Specify a parameter key=value pair, `--param KEY=VALUE`

We should change it to the following to be more consistent:

`--credential-set|c`, `--parameter-set|p`, `--param`

I think we can use a trick in cobra to have this not be a breaking change and support --cred and --credential set. https://stackoverflow.com/a/67929738

# Task Roundup
- [x] Rename --cred to --credential-set
- [x] Allow reference as a positional argument to explain|inspect
- [x] Move porter bundle install|upgrade|invoke|uninstall to porter installation VERB, keeping the shortened aliases.

carolynvs:
Another inconsistency which was introduced after we settled on adding an installation command. Some commands related to installations are under the bundle resources in the CLI

```
$ porter bundle --help
Commands for working with bundles. These all have shortcuts so that you can call these commands without the bundle resource prefix. For example, porter bundle install is available as porter install as well.

Usage:
  porter bundles [command]

Aliases:
  bundles, bundle

Commands:
  archive     Archive a bundle from a reference
  build       Build a bundle
  copy        Copy a bundle
  create      Create a bundle
  explain     Explain a bundle
  inspect     Inspect a bundle
  install     Create a new installation of a bundle
  invoke      Invoke a custom action on an installation
  lint        Lint a bundle
  uninstall   Uninstall an installation
  upgrade     Upgrade an installation
```

The following commands should be moved under installation:
* porter installation create  (aliased to porter install)
* porter installation upgrade (aliased to porter upgrade)
* porter installation uninstall (aliased to porter uninstall)
* porter installation invoke (aliased to porter invoke)

porter bundle install should stay the same since it's operating on a bundle, not an existing installation.

Thinking ahead to other bundle commands that we want ([like bundle list](https://github.com/carolynvs/porter-proposals/blob/dependencies-labels-namespaces/pep/003-dependency-namespaces-and-labels.md#porter-bundle-index)), I was looking at explain. I think it should stay the same allowing room for a `porter bundle show REPOSITORY` and you can see information about the bundle and how its being used in your environment.

[...]

PR #2245, opened 2022-07-15
Refactor shared bundle action options

carolynvs:
# What does this change
Previously we had a shared option struct, sharedOptions and BundleAction, that were used for a large grouping of commands that work with bundles, from install to explain.

I've refactored the shared options into more granular groupings so that later I can define debug mode for _just_ commands that run bundles. Now there are shared options for commands that execute bundles and commands that merely reference a bundle (such as explain).

# What issue does it fix
This is refactoring that will be used in upcoming changes such as #1430 and #1429 .

# Notes for the reviewer
N/A

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/porter/lifecycle.go
+++ b/pkg/porter/lifecycle.go
@@ -8,9 +8,13 @@ import (
 
 	"get.porter.sh/porter/pkg/cache"
 	"get.porter.sh/porter/pkg/cnab"
+	"get.porter.sh/porter/pkg/cnab/drivers"
 	cnabprovider "get.porter.sh/porter/pkg/cnab/provider"
 	"get.porter.sh/porter/pkg/encoding"
+	"get.porter.sh/porter/pkg/portercontext"
+	"get.porter.sh/porter/pkg/secrets"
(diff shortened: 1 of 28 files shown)


pkg/porter/lifecycle.go
```diff
@@ -26,22 +30,199 @@ type BundleAction interface {
...
+	if c.Debug {
+		final["porter-debug"] = "true"
+	}
+
+	return final
+}
+
+// defaultDriver supplies the default driver if none is specified
+func (o *BundleExecutionOptions) defaultDriver(p *Porter) {
+	if o.Driver == "" {
```
VinozzZ, line 197: The comments are great! However, I'm still not sure why we need to check `o.Driver == ""` twice here?

carolynvs: oops! 🤦‍♀️ That was a copy/paste error when I moved the code from cnab.go to lifecycle.go. I'll get that fixed!

VinozzZ (approved):

PR #2247, opened 2022-07-18
Allow reference as positional arg to explain/inspect 

carolynvs:
# What does this change
Since the inspect and explain commands operate directly on a bundle, it makes sense to let the user specify it as a positional argument. I have updated both commands to accept either the existing --reference flag (for consistency with other commands and backwards compatibility) and alow support passing the reference as the first positional argument.

So either of the following now work:

porter explain REFERENCE
porter explain --reference

I've updated the docs to prefer the form without the flag, since it's shorter.

I noticed that our ExplainOptions didn't default the output format, so I updated our ParseFormat function to default an empty format to plaintext. This is mainly useful for the unit tests so that we don't need to repeat logic that normally cobra would handle for us.

# What issue does it fix
Part of [#1430](https://github.com/getporter/porter/issues/1430#issuecomment-860863202)

# Notes for the reviewer
N/A

# Checklist
- [x] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/printer/printer_test.go
+++ b/pkg/printer/printer_test.go
@@ -3,6 +3,7 @@ package printer
 import (
 	"testing"
 
+	"get.porter.sh/porter/tests"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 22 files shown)


pkg/printer/printer_test.go
```diff
@@ -8,25 +8,31 @@ import (
...
 			}
 
 			err := opts.ParseFormat()
-			if valid {
+			if tc.wantErr == "" {
 				require.Nil(t, err)
-				require.Equal(t, name, string(opts.Format))
+				require.Equal(t, tc.wantFormat, opts.Format)
 			} else {
 				require.NotNil(t, err)
```
VinozzZ, line 34: nit - I know this isn't part of this change in the PR, but we probably could change this to require.Error instead?

VinozzZ (approved):

pkg/printer/printer_test.go
carolynvs: Fixed!

PR #2248, opened 2022-07-19
Render step template from original yaml

### LINKED ISSUE #650: Switch to ghodss/yaml?
I was looking more closely at https://github.com/ghodss/yaml as a replacement for the standard yaml encoder and one nice perk is that you don't need to do double tag duty. So if you need to support yaml and json, you don't need to add tags for both json and yaml, just json tags are sufficient The encoder drives itself off of the json tags and marshalers which is pretty neat. Maybe something we could think about doing later as we are already using this library for our schema validation.

carolynvs:
# What does this change
This switches from marshaling the current step to yaml to using the original yaml from the porter.yaml file when rendering a step template.

This avoids problems where the round trip through Unmarshal/Marshal/Unmarshal causes the quotes to not be preserved exactly as was typed into the porter.yaml. Switching to using yaml3.Node preserves the original whitespace and quotes since we are always working directly from the original yaml and not a data structure.

This is useful pre-work for #316, which will allow Porter to finally inject values as data types _other_ than strings, which makes the preservation of exactly what the user typed very important.

# What issue does it fix
Closes #650 

# Notes for the reviewer
N/A

# Checklist
- [x] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/runtime/runtime-manifest.go
+++ b/pkg/runtime/runtime-manifest.go
@@ -2,6 +2,7 @@ package runtime
 
 import (
 	"archive/tar"
+	"bytes"
 	"compress/gzip"
 	"encoding/base64"
 	"fmt"
@@ -21,6 +22,7 @@ import (
 	"github.com/cnabio/cnab-go/bundle"
 	"github.com/cnabio/cnab-to-oci/relocation"
 	"github.com/hashicorp/go-multierror"
+	yaml3 "gopkg.in/yaml.v3"
 )
 
 const (
@@ -36,6 +38,10 @@ type RuntimeManifest struct {
 
 	// bundle is the executing bundle definition
 	bundle cnab.ExtendedBundle
+
+	// manifestYAML is the porter.yaml loaded into YQ so that we can
+	// do advanced stuff with the manifest, like just read out the yaml for a particular step.
+	manifestYAML *yaml.Editor
 
 	// bundles is map of the dependencies bundle definitions, keyed by the alias used in the root manifest
 	bundles map[string]cnab.ExtendedBundle
@@ -59,6 +65,11 @@ func (m *RuntimeManifest) Validate() error {
 		return err
 	}
 
+	err = m.loadManifest()
+	if err != nil {
+		return err
+	}
+
 	err = m.loadDependencyDefinitions()
 	if err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


pkg/runtime/runtime-manifest.go
```diff
@@ -403,35 +430,47 @@ func (m *RuntimeManifest) buildSourceData() (map[string]interface{}, error) {
...
+	var stepTemplate bytes.Buffer
+	enc := yaml3.NewEncoder(&stepTemplate)
+	defer enc.Close()
+	if err := enc.Encode(stepNode); err != nil {
+		return fmt.Errorf("error re-encoding porter.yaml for templating: %w", err)
 	}
 
 	if m.Debug {
-		fmt.Fprintf(m.Err, "=== Step Template ===\n%v\n", string(payload))
+		fmt.Fprintf(m.Err, "=== Step Data ===\n%v\n", sourceData)
```
VinozzZ, line 458: why do we not using the logger here?

pkg/runtime/runtime-manifest_test.go
```diff
@@ -6,104 +6,92 @@ import (
...
-	assert.NotNil(t, s.Data)
-	t.Logf("Length of data:%d", len(s.Data))
-	assert.NotEmpty(t, s.Data["Parameters"])
-	for k, v := range s.Data {
-		t.Logf("Key %s, value: %s, type: %T", k, v, v)
-	}
-	pms, ok := s.Data["Parameters"].(map[string]interface{})
-	assert.True(t, ok)
+
+	require.IsType(t, s.Data["mymixin"], map[string]interface{}{})
```
VinozzZ, line 86: Why do we need to check for a type and then do type assertion again?

pkg/runtime/runtime-manifest.go
carolynvs: I decided against redoing the context plumbing here as it would have exploded out this refactoring PR even further. I can address that in a separate PR.

pkg/runtime/runtime-manifest_test.go
carolynvs: I have pushed a new commit that finishes cleaning up how we check for types in the unit test.

Using a cast and checking OK like this isn't great because when the test fails, it just says "false is not true". Using IsType with a error message makes it more clear in the test output what failed and why.

VinozzZ (approved):

PR #2251, opened 2022-07-20
Directory Param feature

richbai90:
# Notes for the reviewer
This is a follow up to #2167 it didn't make sense to continue on the same branch as the implementation is completely different.

Still not working: Ability to modify param values via command line

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist


richbai90:
I did my best to follow the same coding practices I noticed in the existing source, but I can't guarantee I did everything the "right" way. If there is a better way to implement these features please provide advice.

pkg/cnab/provider/driver.go
```diff
@@ -44,6 +48,40 @@ func (r *Runtime) newDriver(driverName string, args ActionArguments) (driver.Dri
 		return nil, err
 	}
 
+	// Handle Directory support for the docker driver
+	// TODO: Handle directory support for other runtimes -- how?
```
carolynvs, line 52: Here's what I suggest:

1. Do not encode mount information into the directory parameter extension other than the information from `DirectoryParameterDefinition`. The rest is docker specific and should be specified by the user running the bundle, not the bundle author.
2. Check if docker is the driver, then handle the directory parameter just like you are now but... change where you get the information about the mount. Iterate over the bundle's parameters and set up mounts for each parameter of type directory.
3. Use the parameter's value as a docker --mount flag formatted string to get the mount details. The target will come from the parameter definition itself.

The extension data will need to provide details about each directory parameter. (will comment more about that on the data structures in question)

pkg/cnab/directory_parameter.go
```diff
@@ -0,0 +1,104 @@
...
+	mount.Mount `yaml:",inline"`
+	Name        string `json:"name,omitempty" yaml:"name,omitempty"`
+}
+
+// DirectorySources represents the sources available to the directory parameter type
+// Currently only mount has been specified, but this could change in the future
+type DirectorySources struct {
+	Mount MountParameterSourceDefn `yaml:"mount,omitempty" json:"mount,omitempty"`
+}
+type DirectoryDetails struct {
```
carolynvs, line 39: I haven't tried this but here's what I think will fit better with how the CNAB spec and Porter works:

* DirectoryDetails should have Name defined as well, or use a map[string]DirectoryDetails when storing the extension data and then key off of parameter name.
* DirectoryDefaults should not have DirectorySources or Kind.
* Do not create a new parameter source of type mount, I think we can get it to work without it and sources should be used for things that the bundle author knows about up-front (like using the output of a bundle as the parameter value), and bundle authors shouldn't be involved with whether or not a directory was injected using a docker mount, or through some other trick (like unpacking a gzip file into the specified directory).

carolynvs (commented):
I didn't do a full review and am just evaluating the best way to represent directory parameters in a bundle, and where we can persist the data we need.

So let's figure those parts out first, and then we can move on to the part where you are stuck. I'm kind of hoping that this will get you unstuck. 🤞

pkg/cnab/directory_parameter.go
```diff
@@ -0,0 +1,104 @@
...
+	// But is also a realistic value for UID/GID thus we need to make the type interface
+	// To detect the case that the values weren't set
+	GID interface{} `yaml:"gid,omitempty" json:"gid,omitempty"`
+	UID interface{} `yaml:"uid,omitempty" json:"uid,omitempty"`
+}
+
+// MountParameterSource represents a parameter using a docker mount
+// As a its source with the provided options
+type MountParameterSourceDefn struct {
+	mount.Mount `yaml:",inline"`
```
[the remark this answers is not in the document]
richbai90: Okay I can use a string. My idea with using a larger object was so that we didn't have to parse the string, but it should otherwise simplify things a bit so I'll make the change.

richbai90:
One follow up, for the user and group feature, what should the effect of this setting be? Should the container run with that user id instead of the randomly generated one, or should it be for informational purposes to be handled by the mixin developer?

carolynvs:
It won't affect the user that the bundle is run as. What I was thinking is that it should affect the owner and group for any files written by the mixin, because otherwise they will be owned by a wonky userid like 1001, when it probably needs to be owned by nginx or whatever is going to use the files written to the host.

richbai90:
> What I was thinking is that it should affect the owner and group for any files written by the mixin

It seems to me that there are two ways to accomplish this.

1. Make this a documented contract that the mixin author needs to follow, but don't enforce it in code.
2. Create a preinstall step to scan the files in a directory, and a post install step to do the same. Compare the outputs and update the permissions of the files that have changed.
Is there another way I'm not considering? Of these options, which do you prefer?

carolynvs:
@richbai90 Now that v1 is out, I was hoping we could collaborate on your PR/branch here and get this into an upcoming patch release. I see a few merge conflicts, let me know if you'd like help with them, I can resolve and push to your branch (if you gave maintainers push access in this PR).

Once that's fixed and we've had a chance to see how the tests are looking, we can figure out what the remaining work is. 👍

richbai90:
Hi Carolyn, I'm sorry for falling off the face of the earth. I'll give whomever access to the PR. Unfortunately, I'm no longer on the team where we were extensively working with this product and I would need time I don't currently have to get up to speed with V1 changes.

carolynvs:
@richbai90 Not a problem and thank you for your time and effort on the feature design and implementation. I'll pick this up and make sure that the feature you worked on gets completed and shipped.

PR #2268, opened 2022-07-29
Fix allow docker host access

carolynvs:
# What does this change
[Fix regression in allow-docker-host-access](https://github.com/getporter/porter/commit/92ae5f3c05a29ddd7b95001527ae2fe8f7807ff3)

I accidentally removed setting allow-docker-host-access from the global
config when the flag was not set in a previous refactoring PR. This
didn't cause the build to fail because there were no regression tests
around that flag.

I've added a regression test for it (hello-test) and at the same time
noticed that the mybuns test bundle had its dependency defined wrong, so
the dependency wasn't running. I fixed that too and updated the mydb
bundle to work when called as a dependency of mybuns.

[Remove check for docker socket with --allow-docker-host-access](https://github.com/getporter/porter/commit/13a7cc547f2ae89c17ececf0060dabaa973a5d19)

We were checking for the existence of the docker socket before allowing
the --allow-docker-host-access to be applied. But Docker treats mounting
the docker socket specially and in some cases, such as on windows or
when connected to a remote docker daemon, it won't be present locally.

I'm removing the check and if there is a problem with granting access to
the docker socket, we'll find out when Docker returns an error when we
start the container.

[Fix injecting bundle dependnecy on windows](https://github.com/getporter/porter/pull/2268/commits/a2ccb2f87713f899b5684b3f8dfcd701bb28351a)

When we inject a bundle dependency's bundle.json into the invocation image,
we must always use unix file paths. This was causing bundles with
dependencies to fail when run on windows.

# What issue does it fix
Follow-up from #2258 

# Notes for the reviewer
N/A

# Checklist
- [x] Did you write tests?

# Reviewer Checklist


pkg/config/datastore.go
```diff
@@ -28,8 +28,19 @@ type Data struct {
...
+	// to ensure that the global config value works even for those commands.
 	RuntimeDriver string `mapstructure:"runtime-driver"`
 
+	// AllowDockerHostAccess grants bundles access to the underlying docker host
+	// upon which it is running so that it can do things like build and run containers.
+	// It's a security risk.
+	// It is both a global variable and a command flag because some of our commands, like porter installation apply,
+	// do not expose all the bundle execution flags. This allows us to later manually use the global config value
+	// to ensure that the global config value works even for those commands.
+	AllowDockerHostAccess bool `mapstructure:"allow-docker-host-access"`
```
carolynvs, line 42: We should consider if we want to duplicate all of the runtime flags on porter installation apply. For now I'm just fixing the bug.

VinozzZ (approved):

PR #2270, opened 2022-07-30
update/validate cnab bundle schema version to 1.2.0

### LINKED ISSUE #1996: Set bundle.json schemaVersion to 1.2.0

VinozzZ:
# What does this change

Update cnab bundle schema version to latest cnab core version and validate the schema version before processing a bundle

# What issue does it fix
Closes #1996

# Notes for the reviewer

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/cnab/config-adapter/adapter.go
+++ b/pkg/cnab/config-adapter/adapter.go
@@ -18,7 +18,7 @@ import (
 	"github.com/cnabio/cnab-go/bundle/definition"
 )
 
-const SchemaVersion = "v1.0.0"
+const SchemaVersion = "v1.2.0"
 
 // ManifestConverter converts from a porter manifest to a CNAB bundle definition.
 type ManifestConverter struct {
--- a/pkg/cnab/config-adapter/adapter.go
+++ b/pkg/cnab/config-adapter/adapter.go
@@ -18,8 +18,6 @@ import (
 	"github.com/cnabio/cnab-go/bundle/definition"
 )
 
-const SchemaVersion = "v1.2.0"
-
 // ManifestConverter converts from a porter manifest to a CNAB bundle definition.
 type ManifestConverter struct {
 	config          *config.Config
@@ -52,7 +50,7 @@ func (c *ManifestConverter) ToBundle(ctx context.Context) (cnab.ExtendedBundle, 
 	}
 
 	b := cnab.NewBundle(bundle.Bundle{
-		SchemaVersion: SchemaVersion,
+		SchemaVersion: cnab.BundleSchemaVersion(),
 		Name:          c.Manifest.Name,
 		Description:   c.Manifest.Description,
 		Version:       c.Manifest.Version,
--- a/pkg/cnab/extended_bundle.go
+++ b/pkg/cnab/extended_bundle.go
@@ -14,7 +14,7 @@ import (
 
 const SupportedVersion = "1.0.0 || 1.1.0 || 1.2.0"
 
-var DefaultBundleSchemaVersion = semver.MustParse(string(bundle.GetDefaultSchemaVersion()))
+var DefaultSchemaVersion = semver.MustParse(string(BundleSchemaVersion()))
 
 // ExtendedBundle is a bundle that has typed access to extensions declared in the bundle,
 // allowing quick type-safe access to custom extensions from the CNAB spec.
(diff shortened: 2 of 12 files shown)


pkg/cnab/extended_bundle.go
```diff
@@ -36,6 +40,24 @@ func LoadBundle(c *portercontext.Context, bundleFile string) (ExtendedBundle, er
...
+	}
+
+	supported, err := semver.NewConstraint(SupportedVersion)
+	if err != nil {
+		return fmt.Errorf("invalid supported version %s: %w", SupportedVersion, err)
+	}
+	defaultVersion, err := semver.NewVersion(string(bundle.GetDefaultSchemaVersion()))
+	if err != nil {
+		return err
+	}
```
carolynvs, line 56: I like to move conversions like this into a build time / compiler error so that I don't have to wonder if it's going to work at runtime.

You can define a package variable like so and use a function that panics to ensure that we are always building with a valid value, and then we don't have to check it each time we validate.

```go
var DefaultBundleSchemaVersion = semver.MustParse(string(bundle.GetDefaultSchemaVersion())
```

pkg/cnab/extended_bundle.go
```diff
@@ -36,6 +40,24 @@ func LoadBundle(c *portercontext.Context, bundleFile string) (ExtendedBundle, er
...
+	_, err = schema.ValidateSchemaVersion(schema.CheckStrategyExact, supported, string(b.SchemaVersion), defaultVersion)
```
carolynvs, line 57: I thought we were going to use the configuration setting so that someone could use a looser check, or skip it entirely. Like we do when validating the manifest schema version?

https://github.com/getporter/porter/blob/07947b99479e50149e9ae898d4d2ca7cc119fb45/pkg/manifest/manifest.go#L1126-L1129

If we do make it configurable, you will want to handle any possible warnings returned (it's being ignored right now):

https://github.com/getporter/porter/blob/07947b99479e50149e9ae898d4d2ca7cc119fb45/pkg/manifest/manifest.go#L184-L190

pkg/porter/examples/install_example_test.go
```diff
@@ -14,6 +14,7 @@ func ExamplePorter_install() {
 
 	// Specify any of the command-line arguments to pass to the install command
 	installOpts := porter.NewInstallOptions()
+	// install a bundle with older cnab bundle schema version. It should succeed
```
carolynvs, line 17: Thanks for the comment here! It helps me understand that this test is also indirectly checking this test case too. 👍

pkg/cnab/config-adapter/adapter.go
```diff
@@ -18,7 +18,7 @@ import (
 	"github.com/cnabio/cnab-go/bundle/definition"
 )
 
-const SchemaVersion = "v1.0.0"
+const SchemaVersion = "v1.2.0"
```
carolynvs, line 21: I just realized that we have two constants for the current schema version, here and in manifest.go. Since that will be very easy to get out of sync, can you consolidate that into one variable, or define them in the same file and then use a string constant to parse and create the semver?

carolynvs (changes requested):
Everything looks great, just one cleanup item and we are all set! 👍

pkg/cnab/config-adapter/adapter.go
VinozzZ: Maybe I'm missing something, I thought the schema version in the manifest.go file is the schema version for porter.yaml. And this schema version is for cnab?

carolynvs: Sorry I mispoke. I am referring to this in the extended bundle file. We are using both a hard coded schema version in adapter, and also using the constant from cnab-go, so these will easily get out of sync. 

```go
var DefaultBundleSchemaVersion = semver.MustParse(string(bundle.GetDefaultSchemaVersion()))
```

VinozzZ: I see. That makes sense. I found where we store most of the cnab aliases and stored it there. Let me know what do you think?

pkg/porter/build_integration_test.go
```diff
@@ -77,10 +77,10 @@ func TestPorter_Build(t *testing.T) {
 	bun, err := p.CNAB.LoadBundle(build.LOCAL_BUNDLE)
 	require.NoError(t, err)
 
-	assert.Equal(t, bun.Name, "porter-hello")
-	assert.Equal(t, string(bun.SchemaVersion), "v1.2.0")
-	assert.Equal(t, bun.Version, "0.1.0")
-	assert.Equal(t, bun.Description, "An example Porter configuration")
+	assert.Equal(t, "porter-hello", bun.Name)
```
carolynvs, line 80: Thanks for switching these around to match the expected, got pattern! 👍

carolynvs (approved):
Looks great, thanks!

PR #2274, opened 2022-08-02
Add linter success message

### LINKED ISSUE #2237: Show success message when porter lint cannot find an error
## Description

Currently, there is no output for the linter when there are no errors on run. We found this confusing as it just outputs a blank line.

Typically when the javascript linter, [eslint](https://www.npmjs.com/package/eslint) is run it usually ends with

```
✨  Done in 20.95s
``` 
even when there is no error or warning to report.

For the python linter [black](https://pypi.org/project/black/) it would typically end with something like this

```
All done! ✨ 🍰 ✨
24 files left unchanged.
```

## Proposed Solution

We think we can add a message that reads something like, "Everything looks good" after a successful run and no errors have been detected.

## Alternative Solution

If it is not desired to return a specific message, we could remove the blank line (ie return nothing) which would match some other popular linters

## Additional Context

We thought about this after looking at #1226

ThorstenHans:
# What does this change

With this PR, `porter bundle lint` will print a success message upon successful linting

# What issue does it fix
Closes #2237

### DIFF
--- a/pkg/porter/lint_test.go
+++ b/pkg/porter/lint_test.go
@@ -93,11 +93,11 @@ exec:
 	testcases := []struct {
 		format         string
 		wantOutputFile string
-		linterResults  *linter.Results
+		linterResults  linter.Results
 	}{
-		{"plaintext", "testdata/lint/results.txt", &lintResults},
-		{"json", "testdata/lint/results.json", &lintResults},
-		{"plaintext", "testdata/lint/success.txt", &linter.Results{}},
+		{"plaintext", "testdata/lint/results.txt", lintResults},
+		{"json", "testdata/lint/results.json", lintResults},
+		{"plaintext", "testdata/lint/success.txt", linter.Results{}},
 	}
 	for _, tc := range testcases {
 		t.Run(tc.format, func(t *testing.T) {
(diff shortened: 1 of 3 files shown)


VinozzZ (changes requested):
Thank you for working on this issue!
The change looks good other than missing tests. Do you think you can add a test case in this test to verify the output is expected? https://github.com/getporter/porter/blob/454b6896d97a77021f53c9f09ccf016cc7e7c6ad/pkg/porter/lint_test.go#L70

ThorstenHans:
@VinozzZ I've just pushed a new commit. I've slightly modified the existing test to support also this scenario

pkg/porter/lint_test.go
```diff
@@ -93,9 +93,11 @@ exec:
 	testcases := []struct {
 		format         string
 		wantOutputFile string
+		linterResults  *linter.Results
```
VinozzZ, line 96: What do you think if we keep this as a variable instead of a pointer? The reason why we want to keep this because if a test case modifies this variable value, it may cause subsequent tests be impacted with the modified value.

ThorstenHans: Didn't thought about that side effect. Makes sense, I'll change it in a minute

VinozzZ (approved):
Thank you again for working on this issue. It looks great to me

PR #2277, opened 2022-08-02
Update .dockerignore template to use proper Docker template name

### LINKED ISSUE #2275: porter create template .dockerignore contains old name
` porter create` generates `template.Dockerfile` and `.dockerignore` but the contents of the latter references the old template name `Dockerfile.tmpl`. Fixing this will reduce confusion and even the size of the invocation image by ~1000 bytes!

This is seen in version `porter v1.0.0-beta.2-5-g07947b99 (07947b99)`.

Although it was an omission in commit `7b98f5062e1c1756ecd9698051b52785af2cf568` and easily fixed, a more thorough global find and replace on `Dockerfile.tmpl` => `template.Dockerfile` touches *many* files in doc.

I'll create a PR with at least the single fix and we can determine whether to proceed with the global replace.

ThorstenHans:
# What does this change

With this PR, `porter create` does create a new `.dockerignore` file, with correct  Dockerfile template name (`template.Dockerfile`)

# What issue does it fix
Closes #2275

### DIFF
--- a/pkg/templates/templates/create/.dockerignore
+++ b/pkg/templates/templates/create/.dockerignore
@@ -1,4 +1,4 @@
 # See https://docs.docker.com/engine/reference/builder/#dockerignore-file
 # Put files here that you don't want copied into your bundle's invocation image
 .gitignore
-Dockerfile.tmpl
+template.Dockerfile


carolynvs (approved):
LGTM

I just realized that there are a few more places in the docs (the [FAQ](https://getporter.org/faq/), [manifest file format document](https://getporter.org/bundle/manifest/)) that still reference the old filename if you'd like to fix that too?

carolynvs:
@ThorstenHans I just realized that the same PR was merged this morning #2276. So your PR can't be merged anymore... Sorry!

PR #2282, opened 2022-08-03
read image pull response to save image to local cache

### LINKED ISSUE #2280: Airgapped images are broken in v1.0.0-beta.2
## Describe the bug
Referencing images to be bundled for airgapped scenario only works if the image is available in the local docker cache.

## To Reproduce
1. Add an image to "images" property in porter.yaml and make sure to reference an image that has not already been pulled locally
2. Run `porter build`
3. See error

## Expected behavior
The image must be pulled from the listed registry.

## Porter Command and Output
```
$ porter build
failed to find image istio/pilot:1.13.4 in docker cache: Error: No such image: istio/pilot:1.13.4
failed to find image istio/pilot:1.13.4 in docker cache: Error: No such image: istio/pilot:1.13.4
unable to generate manifest: failed to find image istio/pilot:1.13.4 in docker cache: Error: No such image: istio/pilot:1.13.4
```

## Version
porter v1.0.0-beta.2 (b2f23785)

VinozzZ:
# What does this change

Previously, we assumed that by pulling an image, the image is automatically cached by docker. However, it seems that the caching only happens when the response from pull is read. 

This PR fixes above issue by reading the full response. I also changed the hello smoke test to always run in a fresh local docker environment to make sure this scenario is tested.

# What issue does it fix
Closes #2280 

# Notes for the reviewer

# Checklist
- [x] Did you write tests?

# Reviewer Checklist


tests/smoke/hello_test.go
```diff
@@ -20,6 +21,8 @@ func TestHelloBundle(t *testing.T) {
 	defer test.Close()
 	require.NoError(t, err, "test setup failed")
 
+	// Set up a fresh docker environment
+	mgx.Must(shx.RunV("docker", "system", "prune", "--all", "-f"))
```
carolynvs, line 25: This is a very expensive call to make on a dev machine, since for example I have 30-60GB of cached images that I would have to repull after every time I run this test. 😱  Plus this is also removing containers and volumes too which is not okay to just delete on people's machines.

A less invasive way to make sure that we don't have that image in our cache is to use the `docker image rm IMAGE` command, without any flags specified so that if it's in use, the dev gets an error and can figure out what is a safe way to cleanup their environment.

carolynvs (changes requested):

pkg/cnab/config-adapter/testdata/mybuns.bundle.json
```diff
@@ -9,6 +9,13 @@
       "image": "localhost:5000/mybuns:332dd75c541511a27fc332bdcd049d5b"
     }
   ],
+  "images": {
+    "whalesayd": {
+      "imageType": "docker",
+      "image": "carolynvs/whalesayd:latest",
+      "description": "Whalesay as a service"
```
carolynvs, line 16: Shouldn't this have persisted the digest of the image now that your PR implementing tag resolution of referenced images during porter build was merged?

tests/smoke/hello_test.go
```diff
@@ -21,8 +21,8 @@ func TestHelloBundle(t *testing.T) {
 	defer test.Close()
 	require.NoError(t, err, "test setup failed")
 
-	// Set up a fresh docker environment
-	mgx.Must(shx.RunV("docker", "system", "prune", "--all", "-f"))
+	// make sure the referenced image is not in local image cache
+	mgx.Must(shx.RunV("docker", "rmi", "carolynvs/whalesayd"))
```
carolynvs, line 25: If the image is not in the cache, this line will fail. In cases where you just need to do cleanup before a test, I suggest not checking the error message (kind of like calling rmdir and then ignoring the error if the directory doesn't exist).

pkg/porter/build_integration_test.go
```diff
@@ -110,7 +110,7 @@ func TestPorter_Build_ChecksManifestSchemaVersion(t *testing.T) {
 			defer p.Close()
 
 			// Make a bundle with the specified schemaVersion
-			p.TestConfig.TestContext.AddTestDirectoryFromRoot("tests/testdata/mybuns", "/")
+			p.TestConfig.TestContext.AddTestDirectoryFromRoot("tests/testdata/mydb", "/")
```
carolynvs, line 113: Can you help me understand why this was changed to user a different bundle? If the test worked before with mybuns, I would expect it to still work with mybuns with a referenced image. I am concerned that changing the test bundle may be hiding a regression or problem in Porter.

carolynvs (changes requested):

pkg/cnab/config-adapter/testdata/mybuns.bundle.json
VinozzZ: I believe this testdata is used for testing the manifest to cnab bundle conversion which does not include the logic to generate the build.LOCAL_MANIFEST which is when the tag is resolved

carolynvs: Gotcha, thanks! 👍

pkg/porter/build_integration_test.go
VinozzZ: I changed it back

carolynvs (approved):
Thanks for adding the caching to our TestRegistry. I realize it may not be what we want for the tests long term, but I appreciate that it preserves our existing regression tests.

PR #2283, opened 2022-08-03
replace bundle.invocationImage with bundle.installerImage

### LINKED ISSUE #1824: Template variable bundle.invocationImage is incorrect
## Describe the bug
The bundle.invocationImage template value that is injected into the bundle at runtime is incorrect. It should reflect the final resolved invocation image, but it is the value built into the bundle at build time.

So if you publish it to a different location, it isn't the right value. Other problems:

* it doesn't take into account relocation mapping files. For example, really for any published bundle the invocation image is an artifact _inside_ the bundle repository. It's always getporter/mybundle:sha256:abc123, and never `getporter/mybundle-installer:v0.1.1`.
* It should use a digest instead of a tag.

## To Reproduce
1. Create a bundle
2. Echo `{{ bundle.invocationImage}}` from the porter.yaml during install
3. Build the bundle and publish it to a local registry.
4. Run the bundle from the published location.

## Expected behavior
The same image that is being executed by porter should be printed. Instead it prints the value built into the bundle at buildtime.

## Version
v0.38.7 and v1.0.0-alpha.5

VinozzZ:
# What does this change

Template variable bundle.invocationImage has been deprecated and replaced with bundle.installerImage

This PR makes this update and make sure that installer image is always referencing to the current running image instead of the original image reference created during build time
# What issue does it fix
Closes #1824 

# Notes for the reviewer

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/docs/content/authors/templates.md
+++ b/docs/content/authors/templates.md
@@ -63,7 +63,7 @@ The bundle variable contains data that was declared in the bundle definition (po
 | bundle.name | The bundle name |
 | bundle.description | The bundle description |
 | bundle.version | The bundle version defined in porter.yaml |
-| bundle.invocationImage | (DEPRECATED) The name of the invocation image |
+| bundle.installerImage | The name of the bundle installer image |
 
 #### custom metadata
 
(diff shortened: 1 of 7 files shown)


carolynvs (changes requested):
Everything is great. You just need to update the documentation page to use the new name of the template variable https://getporter.org/authors/templates/#bundle

carolynvs (approved):
Thank you for the fix! 💖

PR #2286, opened 2022-08-05
pin invocation image to linux/amd64

### LINKED ISSUE #2021: Use --platform when we build invocation images and document what we support
In a world before M1 and IOT, we decided that we would just have bundles create a single platform/arch invocation image (linux/amd64) and it's been working well. Now that a lot of the Porter devs are building on M1, we need to figure out how Porter should build the bundle's image, and if it should be:

1. Fixed as amd64
2. Multi-arch amd64/arm64 (not sure how well we've really planned that out in the CNAB spec but I think it should work since we use buildx under the covers)
3. YOLO (this is what we do today), whatever platform you are on, that's what we build.

The YOLO option isn't acceptable because if I publish a bundle from my M1, someone on linux arm64 can't run it. Docker Desktop kind of hides that this is a problem and does unknown magic to make running the other arch work but we need something that once distributed works for most people.

Updated: For 1.0 if we can do multi-arch without a TON of work, let's do that. Otherwise let's fix the platform for the invocation image to linux/amd64 and then document which architecture works with bundles. Then in a point release after v1.0 we can add multi-arch support. The big question is: what is the default, and how does someone opt-in/out of multi-arch images.

VinozzZ:
# What does this change

Currently, if a bundle is build on a linux/amd64 machine, it won't be able to run on a M1 machine. This PR makes it explicit in terms of what porter supports by default. 

# What issue does it fix
Closes #2021 

# Notes for the reviewer

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/docs/content/build-image.md
+++ b/docs/content/build-image.md
@@ -225,7 +225,7 @@ WORKDIR ${BUNDLE_DIR}
 CMD ["/cnab/app/run"]
 ```
 
-Porter starts the Dockerfile by using a base image. You can customize the base image by specifying a Dockerfile template in the porter.yaml. By default, Porter only supports single platform/architecture(linux/amd64) for invocation image. If you want to use other platform, feel free to change the platform flag in the generated Dockerfile template. Next, a set of CA certificates is added.  Next, contents of the current directory are copied into the bundle directory (/cnab/app) in the invocation image. This will include any contributions from the mixin executables. Finally, an entry point that conforms to the CNAB specification is added to the image.
+Porter starts the [Dockerfile](/bundle/custom-dockerfile) by using a base image. You can customize the base image by specifying a Dockerfile template in the porter.yaml. By default, Porter only targets a single os/architecture(linux/amd64) for invocation image. If you want to use other platform, feel free to change the platform flag in the generated Dockerfile template. Next, a set of CA certificates is added.  Next, contents of the current directory are copied into the bundle directory (/cnab/app) in the invocation image. This will include any contributions from the mixin executables. Finally, an entry point that conforms to the CNAB specification is added to the image.
 
 Once this is completed, the image is built:
 
--- a/docs/content/build-image.md
+++ b/docs/content/build-image.md
@@ -80,7 +80,7 @@ actions:
       description: World 2.0
 
 /Users/sigje/.porter/mixins/exec/exec build --debug
-FROM debian:stretch
+FROM --platform=linux/amd64 debian:stretch-slim
 
 ARG BUNDLE_DIR
 
@@ -97,7 +97,7 @@ WORKDIR ${BUNDLE_DIR}
 CMD ["/cnab/app/run"]
 
 Writing Dockerfile =======>
-FROM debian:stretch
+FROM --platform=linux/amd64 debian:stretch-slim
 
 ARG BUNDLE_DIR
 
@@ -114,7 +114,7 @@ WORKDIR ${BUNDLE_DIR}
 CMD ["/cnab/app/run"]
 
 Starting Invocation Image Build =======>
-Step 1/9 : FROM debian:stretch
+Step 1/9 : FROM --platform=linux/amd64 debian:stretch-slim
  ---> 5738956efb6b
 Step 2/9 : ARG BUNDLE_DIR
  ---> Using cache
@@ -208,7 +208,7 @@ After copying any mixins to the .cnab directory, a Dockerfile is generated:
 
 ```console
 Generating Dockerfile =======>
-FROM debian:stretch
+FROM --platform=linux/amd64 debian:stretch
 
 ARG BUNDLE_DIR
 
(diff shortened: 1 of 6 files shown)


docs/content/build-image.md
```diff
@@ -114,7 +114,7 @@ WORKDIR ${BUNDLE_DIR}
 CMD ["/cnab/app/run"]
 
 Starting Invocation Image Build =======>
-Step 1/9 : FROM debian:stretch
+Step 1/9 : FROM --platform=linux/amd64 debian:stretch-slim
```
carolynvs, line 117: Stuff like this is why we've since moved away from including the output of some commands since they get stale very easily 😀

pkg/templates/templates/create/template.buildkit.Dockerfile
```diff
@@ -10,7 +10,7 @@
 
 # You can control where the mixin's Dockerfile lines are inserted into this file by moving the "# PORTER_*" tokens
 # another location in this file. If you remove a token, its content is appended to the end of the Dockerfile.
-FROM debian:stretch-slim
+FROM --platform=linux/amd64 debian:stretch-slim
```
carolynvs, line 13: This is the absolute best place to provide context and instructions since everyone will have the file even if they never find their way to the right page on the website.

Let's add a comment that explains that we are fixed by linux/amd64 _by default_ and this can be changed by editing the --platform flag.

```
# You can control where the mixin's Dockerfile lines are inserted into this file by moving the "# PORTER_*" tokens
# another location in this file. If you remove a token, its content is appended to the end of the Dockerfile.

# Porter targets linux/amd64 by default. Change the --platform flag to target a different platform.
FROM --platform=linux/amd64 debian:stretch-slim
```

docs/content/bundle/custom-dockerfile.md
```diff
@@ -47,6 +47,8 @@ dockerfile: template.Dockerfile
 It is your responsibility to provide a suitable base image, for example one that has root ssl certificates installed. 
 *You must use a base image that is debian-based, such as debian or ubuntu with apt installed.*
 Mixins assume that apt is available to install packages.
+Porter currently only supports single platform/architecture invocation images, by default it's linux/amd64.
```
carolynvs, line 50:
```suggestion
Porter only supports targeting a single os/architecture when the bundle is built. By default, Porter targets linux/amd64.
```

docs/content/build-image.md
```diff
@@ -225,13 +225,13 @@ WORKDIR ${BUNDLE_DIR}
 CMD ["/cnab/app/run"]
 ```
 
-Porter starts the Dockerfile by using a base image. You can customize the base image by specifying a Dockerfile template in the porter.yaml. Next, a set of CA certificates is added.  Next, contents of the current directory are copied into the bundle directory (/cnab/app) in the invocation image. This will include any contributions from the mixin executables. Finally, an entry point that conforms to the CNAB specification is added to the image.
+Porter starts the Dockerfile by using a base image. You can customize the base image by specifying a Dockerfile template in the porter.yaml. By default, Porter only supports single platform/architecture(linux/amd64) for invocation image. If you want to use other platform, feel free to change the platform flag in the generated Dockerfile template. Next, a set of CA certificates is added.  Next, contents of the current directory are copied into the bundle directory (/cnab/app) in the invocation image. This will include any contributions from the mixin executables. Finally, an entry point that conforms to the CNAB specification is added to the image.
```
carolynvs, line 228: Let's link to the custom dockerfile doc page so that people know where to find what you are referencing.

carolynvs (changes requested):

carolynvs (changes requested):
I just ran this locally:

1. porter create
2. porter build

Then opened .cnab/Dockerfile and saw that it is missing --platform.

There is one more file that needs to be updated, pkg/templates/build/buildkit.Dockerfile. You updated the create template, which is used when you run porter create. We have one more template, build/buildkit.Dockerfile, which is used when the user did not specify template in their porter.yaml.

Can you update that file and then check that we have a test that will validate the platform is set both for people using a custom template and those who aren't? 

I think if you update pkg/build/testdata/buildkit.Dockerfile and custom-dockerfile-expected-output.Dockerfile (the test for that one probably needs updating), then we will have 2 checks that the platform is being set when it should.

pkg/build/testdata/buildkit.Dockerfile
```diff
@@ -1,5 +1,5 @@
 # syntax=docker/dockerfile-upstream:1.4.0
-FROM debian:stretch-slim
+FROM --platform=linux/amd64 debian:stretch-slim
```
carolynvs, line 2: I think that there was a bit of a misunderstanding from my previous feedback. I was only asking that you update this file so that when someone does not have a custom dockerfile specified, the template used during porter build has a platform set.

You should not modify the contents of the custom dockerfile when set (which is what I think you were doing with buildFROMSection?)

carolynvs: > Can you update that file and then check that we have a test that will validate the platform is set both for people using a custom template and those who aren't?

Sorry this was unclear! What I meant was that:

* When the custom dockerfile is generated by porter during `porter create`, the template.Dockerfile should have a platform set.
* When porter build is run and a custom dockerfile is not set in porter.yaml (by the template field), the template used during porter build should include the platform
* When the user has a custom dockerfile specified in their porter.yaml, we should not attempt to insert a platform. This allows them to customize the platform themselves, or do whatever special stuff they require without Porter's interference.

VinozzZ: Thank you for the clarification! I updated the code to reflect it

carolynvs (approved):
Looks great, thank you!

PR #2291, opened 2022-08-09
persist run record for custom actions that produce outputs

### LINKED ISSUE #1980: Record run when a bundle action generates outputs
Right now the logic for when porter should record a bundle run is: 

* the bundle action is not stateless
* the bundle action modifies bundle resources

It is possible for a stateless action to create outputs though, and unless modifies is artificially set to true, the outputs would not be captured by Porter.

A couple things come to mind here:

1. modifies and stateless are CNAB concepts that Porter has exposed without any user experience improvements. I feel like these are concepts that maybe the runtime needs to know but perhaps Porter can help map that better to how a bundle author thinks about what the action does.
2. If we don't tweak how custom actions are defined, we should at least change the runtime to record when a bundle output is created.

Is there a better way to represent the concepts of stateless and modifies in a way that works for us? 

* stateless - does this require the bundle to already be installed? Could be renamed to "allowBeforeInstall".
* modifies - does this change any resources managed by the bundle? We can use that eventually to prevent concurrent runs that modify resources. Users can use this to figure out if it's safe to run an action. Could be renamed to "modifiesBundleResources"

A couple examples of how these two can combine to make real-world actions:

* preflight (stateless: true, modifies: true) - Run a series of checks against an environment before installing. If there are configuration problems identified, correct them.
* dry-run (stateless: true, modifies: false) - Print the actions the bundle _would_ take but don't change anything. Could be run either before or after the bundle is installed.
* status (stateless: false, modifies: false) - Print the status of the bundle resources.

VinozzZ:
# What does this change

When a custom action that does not modify bundle resources and is stateless, its run should be recorded in the database if the action produce any outputs.
This PR implements this behavior.

# What issue does it fix
Closes #1980 

# Notes for the reviewer

# Checklist
- [x] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/docs/content/bundle/manifest/_index.md
+++ b/docs/content/bundle/manifest/_index.md
@@ -400,30 +400,72 @@ are automatically defaulted, such as `dry-run`, `help`, `log`, and `status`. You
 actions unless you want to change the defaults.
 
 You may want to declare your custom action when the action does not make any changes, and its execution should not 
-be recorded (`stateless: true` and `modifies: false`). An example of usecases is:
+be recorded (`stateless: true` and `modifies: false`). An example of the usecase is:
+```yaml
+customActions:
+  myhelp:
+    description: "Print a special help message"
+    stateless: true
+    modifies: false
+
+myhelp:
+  exec:
+    command: ./help.sh
+    arguments:
+      - special-help
+    outputs:
+      - name: help-message
+        regex: "(.*)"
+
+outputs:
+  - name: connStr
+    description: "db connection string"
+    type: string
+    applyTo:
+      - install
+```
+
+* `description`: Description of the action.
+* `stateless`: Indicates that the action is purely informational and can be executed before the install action runs.
+* `modifies`: Indicates whether this action modifies resources managed by the bundle.
+
+In this example, the output `help-message` is not going to be recorded during bundle execution. This is determined by two criterias:
+  - `myhelp` is configured to be stateless and do not modify any bundle resources.
+  - there's no bundle level output applied to `myhelp` action
+
+If a custom action is stateful or modifies bundle resources, its output will be captured by default.
+
+Here's another example to demonstrate how to cature a custom action's output when it's stateless and makes no change to a bundle:
 ```
 customActions:
-  getLicense:
+  myhelp:
     description: "Return the product license information"
     stateless: true
     modifies: false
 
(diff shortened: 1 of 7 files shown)


docs/content/bundle/manifest/_index.md
```diff
@@ -400,21 +400,29 @@ are automatically defaulted, such as `dry-run`, `help`, `log`, and `status`. You
...
 ```
 customActions:
-  myhelp:
-    description: "Print a special help message"
+  getLicense:
+    description: "Return the product license information"
     stateless: true
     modifies: false
+
+getLicense:
```
carolynvs, line 411: Are you trying to demonstrate that this action will be recorded? It's a bit unclear for this example why you are including an output but then not discussing how actions that generate outputs are recorded.

Maybe split this into two custom actions? Keep myhelp as it was, and then add the getLicense one too, along with a bundle output defined so that you can explain why one is recorded and the other one isn't?

docs/content/bundle/manifest/_index.md
```diff
@@ -400,21 +400,29 @@ are automatically defaulted, such as `dry-run`, `help`, `log`, and `status`. You
...
+          regex: 'LICENSE: (.*) \(.*\)'
 ```
+The `help` action is supported out-of-the-box by Porter
+and is automatically defaulted to this definition so you do not need to declare it. If you have an action that is 
+similar to `help`, but has a different name, you should declare it in the `customActions` section.
 
 * `description`: Description of the action.
-* `stateless`: Indicates that the action is purely informational, that credentials are not required, 
-   and that Porter should not keep track of when this action was last executed.
+* `stateless`: Indicates that the action is purely informational and can be executed before the install action runs. It also,means that credentials are not required, and that Porter should not keep track of when this action was last executed.
```
carolynvs, line 425:
```suggestion
* `stateless`: Indicates that the action is purely informational and can be executed before the install action runs.
```

Now that the logic for when a run should be persisted is more complicated, this sentence is confusing since it doesn't address it fully. How about explaining it only once next to the above example? "Porter will record a run when the action modifies resources, is not stateless, or generates an output."

carolynvs (changes requested):
Looks good! The docs just need a few edits and it's ready to go 👍

carolynvs (approved):

PR #2306, opened 2022-08-17
Added unexpected EOF error check for when state file exists but is empty

### LINKED ISSUE #2307: State file unpacking fails if an empty state file exists
## Describe the bug
I'm getting the following error when trying to use a porter bundle that has "state" provided in the manifest when running in the operator:
could not create a new gzip reader for the statefile: unexpected EOF
This is when using "state" to store off terraform tfstate files ex:

state:
  - name: tfstate
    path: terraform/terraform.tfstate
  - name: tfvars
    path: terraform/terraform.tfvars.json
This works when setting the tfstate file as a bundle output and using it that way.
Running this on beta.2 in the operator

## To Reproduce
1. Install a porter bundle in the operator that has "state" defined in the porter.yaml

## Expected behavior
Porter should install successfully for bundles with "state" in the porter.yaml that run in the operator

## Version
porter v1.0.0-beta.2 (b2f23785)

sgettys:
# What does this change
If the state.tgz file exists but is empty then the unpack will fail. Suggested behavior would be that in this instance the unpack just continues without trying to read in the file once it knows its empty. 

# What issue does it fix
Closes #2307 

# Notes for the reviewer

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/runtime/runtime_manifest.go
+++ b/pkg/runtime/runtime_manifest.go
@@ -530,7 +530,21 @@ func (m *RuntimeManifest) unpackStateBag(ctx context.Context) error {
 		m.debugf(log, "No existing bundle state to unpack")
 		return nil
 	}
-
+	bytes, err := m.config.FileSystem.ReadFile(statePath)
+	if err != nil {
+		m.debugf(log, "Unable to read bundle state file")
+		return err
+	}
+	// TODO(sgettys): hack around state.tgz ALWAYS being injected even when empty files mess things up
+	// I'm not sure yet why it's injecting as null instead of "" (as required by the spec)
+	// We want to get the null -> "" fixed, and also not write files into the bundle when unset.
+	// that's a cnab change somewhere probably
+	// the problem is in injectParameters in cnab-go
+	if string(bytes) == "null" {
+		m.debugf(log, "Bundle state file has null content")
+		m.config.FileSystem.Remove(statePath)
+		return nil
+	}
 	// Unpack the state file and copy its contents to where the bundle expects them
 	// state var name -> path in bundle
 	log.Debug("Unpacking bundle state...")
@@ -565,7 +579,7 @@ func (m *RuntimeManifest) unpackStateBag(ctx context.Context) error {
 
 	gzr, err := gzip.NewReader(stateArchive)
 	if err != nil {
-		if err == io.ErrUnexpectedEOF || err == io.EOF {
+		if err == io.EOF {
 			log.Debug("statefile exists but is empty")
 			return nil
 		}
--- a/pkg/runtime/runtime_manifest.go
+++ b/pkg/runtime/runtime_manifest.go
@@ -565,7 +565,7 @@ func (m *RuntimeManifest) unpackStateBag(ctx context.Context) error {
 
 	gzr, err := gzip.NewReader(stateArchive)
 	if err != nil {
-		if err == io.ErrUnexpectedEOF {
+		if err == io.ErrUnexpectedEOF || err == io.EOF {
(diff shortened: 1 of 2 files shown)


VinozzZ:
Hey @sgettys , thanks for working out the fix for this!
Feel free to remove the WIP in the PR title whenever this PR is ready for review

VinozzZ (changes requested):
The code looks good to me. Do you mind to add some tests for this scenario for future proof?

sgettys:
> The code looks good to me. Do you mind to add some tests for this scenario for future proof?

I can look into adding some. There weren't any existing tests around the state bag so I'm not sure if that's due to testing complexity or if it was just not added in initially. Let me dig into the code some more to see how to add some tests around this

VinozzZ:
> > The code looks good to me. Do you mind to add some tests for this scenario for future proof?
> 
> I can look into adding some. There weren't any existing tests around the state bag so I'm not sure if that's due to testing complexity or if it was just not added in initially. Let me dig into the code some more to see how to add some tests around this

I see. We do have a test bundle in `tests/smoke/testdata/mybuns` that we use to make sure porter is behaving correctly. Maybe we can edit that test bundle to include situation when a state file exist but is empty?

sgettys:
> > > The code looks good to me. Do you mind to add some tests for this scenario for future proof?
> > 
> > 
> > I can look into adding some. There weren't any existing tests around the state bag so I'm not sure if that's due to testing complexity or if it was just not added in initially. Let me dig into the code some more to see how to add some tests around this
> 
> I see. We do have a test bundle in `tests/smoke/testdata/mybuns` that we use to make sure porter is behaving correctly. Maybe we can edit that test bundle to include situation when a state file exist but is empty?

I think I have a way to easily inject some test case scenarios around this. Testing out now

sgettys:
> > > The code looks good to me. Do you mind to add some tests for this scenario for future proof?
> > 
> > 
> > I can look into adding some. There weren't any existing tests around the state bag so I'm not sure if that's due to testing complexity or if it was just not added in initially. Let me dig into the code some more to see how to add some tests around this
> 
> I see. We do have a test bundle in `tests/smoke/testdata/mybuns` that we use to make sure porter is behaving correctly. Maybe we can edit that test bundle to include this situation when a state file exist but is empty?

Ok added a couple of tests. Looks like there's actually 2 different code paths for this. 1 if /porter/state.tgz exists but is empty then the NewReader will return "EOF" error. If it has content but not a gzip header then it will return "UnexpectedEOF". This would make it so that an invalid state tarball, whether empty file or no header set, would result in the unpackStateBag returning nil and the initialization continuing. When finalize is ran then that state will be overwritten using the packStateBag method which will properly format the state file. thoughts on this behavior?

sgettys:
> Thank you for adding the tests and clarify the behavioral changes that this change indicates. From my perspective, if we ignore the `io.UnexpectedEOF` error, the state file maybe invalid/corrupted, the execution of the bundle now becomes nondeterministic and user has no way to know that until after the bundle execution. What do you think if we only handle the `io.EOF` error since an empty file is still an valid file but return the `io.UnexpectedEOF`?
> We can then look at how the operator is mounting the state file to figure out why it's mounting an empty file or an file that only contains `\n`

That's a good point, although is there a way to recover from a corrupted or invalid state file? I'm also working on getting this into the operator integration tests so we can have a bundle that defines the "state" keyword as part of the integration tests.

Looks like the volume mounting occurs here:
https://github.com/cnabio/cnab-go/blob/main/driver/kubernetes/kubernetes.go#L431

VinozzZ (approved):
lgtm

Issue #2319, opened 2022-08-31
Tagging invocation image with random tag is not needed and prevents proper garbage collection


jasmdk:
**Is your feature request related to a problem? Please describe.**
Porter v1 no longer pushes the invocation image as a separate repo, but instead it adds a tag with a random value. The tag is obsolete if porter knows the digest of the image already and the tag prevents standard garbage collection routines (like in Harbor) to implement a strategy for removing intermediate bundles using a tag pattern.

**Describe the solution you'd like**
I expect that porter uses the published manifest to determine the image that it must use as invocation image, however the `invocationImages[].image` points to the image using both a tag and a digest which means that the digest takes precedence and the tag is actually ignored.
By not tagging the invocation image, then garbage collection routines that uses tags to determine which artifacts to retain or delete can consistently remove all images in a bundle including the invocation image - otherwise removed bundles will produce dangling invocation images no longer referenced from a bundle, but unable to be garbage collected as they are tagged.
If a tag is needed, could it then be the same tag as the bundle, which again would allow garbage collection routines to identify bundles to delete based on tags.

**Describe alternatives you've considered**

1. Create our own cron-job to identify all CNAB bundles and their invocation images and compare this with all docker images in the same repo, which is tagged. Remove the tag from the images that is not currently referenced by a CNAB manifest and then let Harbor do the actual garbage collection.
2. Accept that storage grows at an even faster rate with unused dangling images

**Additional context**
A docker file on the form `<registry>/<repo>:<tag>@<digest>` which is what is stored in the manifest can also be pulled using `<registry>/<repo>@<digest>` because the digest takes precence over a tag, so the tag is effectively ignored.

carolynvs:
The short answer is that it is impossible to calculate the repository digest of the invocation image without pushing it to the repository first. So we push it to the bundle repository, then retrieve the repository digest and rewrite the bundle to reference it by digest and not the tag.

@VinozzZ I thought that you looked into removing the tag afterwards but I don't remember the details on why we don't do that. I think it was because removing a tag isn't universally supported or something like that?

jasmdk:
If it can not be removed, is there a reason why the tag seems to be a random value and not the bundle tag?

carolynvs:
The tag is a hash of the bundle reference, so not really random. We do not push the image to a location that could be used by other versions of the bundle to avoid race conditions that arise when publishing multiple versions of the bundle to the same registry. Reusing a constant tag value for the image is not an option, sorry.

[...]

PR #2328, opened 2022-09-02
Discard the temporary invocation image tag during publish

carolynvs:
# What does this change
When we publish a bundle, Porter creates a temporary tag for the invocation image, pushes it to the tag, and then updates the bundle to reference the invocation image by its digest.

Right now the reference to the image includes both the temporary tag and the digest, e.g.

example.com/mybundle:temporary-tag@sha256:abc123

This creates a long term reference to the temporary tag, which prevents users from pruning it, and the registry's own internal garbage collector from ever allowing it to be removed since the bundle manifest will always reference it.

The tag was intended to be temporary so I have updated publish so that when we determine the repository digest for the invocation image, we rewrite the invocation image reference to use ONLY the digest, and to discard the temporary tag.

# What issue does it fix
Part of #2319

# Notes for the reviewer

# Checklist
- [x] Did you write tests?

# Reviewer Checklist


VinozzZ (approved):
LGTM!

PR #2333, opened 2022-09-05
Refactor bundle runtime data structures

carolynvs:
This is refactoring work in preparation for PEP003 - advanced dependencies. I want to add more implementations to the cnab package, such as for workflows. So I have renamed the provider package to be more specific about what it is capable of (bundleruntime). It really is the implementation for https://github.com/cnabio/cnab-spec/blob/cnab-core-1.2.0-ga/103-bundle-runtime.md.


carolynvs:
This is on hold until after the 1.0 release

PR #2336, opened 2022-09-06
Make identifiable temporary bundle image tags

### LINKED ISSUE #2319: Tagging invocation image with random tag is not needed and prevents proper garbage collection

carolynvs:
# What does this change
When we publish a bundle, we have to push the invocation image temporarily to a tag, so that we can then retrieve its repository digest. Right now the tag is the hash of the bundle reference, which is impossible for someone maintaining a registry to identify and clean up.

This changes the format of the temporary invocation image tag to:

porter-HASH

For example:

The input text for the invocation image hash is the location of the bundle:

ghcr.io/getporter/examples/porter-hello:v0.2.0

which results in a md5 hash of 3cb284ae76addb8d56b52bb7d6838351, so the resulting temporary location where we push the invocation image is:

ghcr.io/getporter/examples/porter-hello:porter-3cb284ae76addb8d56b52bb7d6838351

I have updated all our code to use the same function to generate the temporary tag, even for referenced images which previously used a different method of computing the hash when publishing from an archive. Referenced images now compute the hash based on the original location of the image.

So if a bundle referenced nginx:v1.0.0 and is used in a bundle located at example.com/mybuns, the temporary location for nginx would be:

example.com/mybuns:porter-55135915b5d3b5cf91b7305ec6a76d6e

where the hash uses the location of the original image (nginx:v1.0.0)

# What issue does it fix
Closes #2319

# Notes for the reviewer
N/A

# Checklist
- [x] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/porter/publish.go
+++ b/pkg/porter/publish.go
@@ -334,6 +334,13 @@ func getNewImageNameFromBundleReference(origImg, bundleTag string) (image.Name, 
 		return image.EmptyName, err
 	}
 
+	// Calculate a unique tag based on the original referenced image. It is safe to
+	// use only the original image, and not a combination of both the destination and
+	// the source to create a unique value, because we rewrite the referenced image
+	// to always use a repository digest. The only time two images will have the same
+	// source value is when they are the same image and have the same content. In
+	// which case it is okay if two bundles both reference the same image and reuse
+	// the same temporary tag because the content is the same.
 	tmpImage, err := cnab.CalculateTemporaryImageTag(origImgRef)
 	if err != nil {
 		return image.EmptyName, err
--- a/pkg/porter/publish.go
+++ b/pkg/porter/publish.go
@@ -2,12 +2,9 @@ package porter
 
 import (
 	"context"
-	"crypto/md5"
-	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
-	"path"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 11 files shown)


pkg/porter/publish.go
```diff
@@ -327,29 +324,29 @@ func pushUpdatedImage(layout registry.Layout, origImg string, newImgName image.N
...
 	if err != nil {
-		return image.EmptyName, fmt.Errorf("unable to parse bundle tag %q into domain/path components: %w", bundleTag, err)
+		return image.EmptyName, err
 	}
 
-	// Use the original image name with the bundle location to generate a randomized tag
-	source := path.Join(path.Dir(bundleName.Name()), origName.Name()) + ":" + bundleName.Tag()
-	nameHash := md5.Sum([]byte(source))
-	imgTag := hex.EncodeToString(nameHash[:])
+	tmpImage, err := cnab.CalculateTemporaryImageTag(origImgRef)
```
VinozzZ, line 337: Would this cause uniqueness issue when a different bundle tag is provided for the same original image? I remember I had to include the bundle tag information for the hash due to some uniqueness issue

VinozzZ: Right! I forgot that we changed to use digest for the referenced image. Then it shouldn't cause an issue anymore. Thank you for explaining it!

VinozzZ (approved):

pkg/porter/publish.go
carolynvs: I'm glad that you asked! I was a bit confused about why we were previously building up a composite string value based on both the source and the destination. 👍 Now we've got it written down somewhere. Let me add a quick comment to capture what we talked about here so we don't forget later.

PR #2360, opened 2022-09-14
Set runtime debug mode

### LINKED ISSUE #2358: Debug mode is not being picked up by the bundle runtime
When I run a bundle and specify --debug, that should cause the bundle to print out additional information, such as the commands being run.

```console
mkdir mybuns && cd mybuns
porter create
porter install test --debug
# bundle output
```

The bundle output should include a line like `/cnab/app ./helpers.sh install` but it doesn't.

There seem to be a couple interrelated problems:

- [ ] The RuntimeConfig.DebugMode value should be initialized to PORTER_DEBUG which is set in the bundle when run with --debug.
- [ ] The verbosity configuration value isn't being overridden when set via --verbosity. Porter is only respecting the config file value and environment variable. This affects what porter prints out when running a bundle.
- [ ] Some runtime structs and their functions have been updated to use telemetry, e.g. span.Info, but I suspect that when the TraceLogger is initialized inside a bundle, it isn't picking up the environment variables we are sending to set verbosity, otel, etc.

carolynvs:
# What does this change
This fixes the bundle output when a bundle is run in debug mode with --debug. It initializes RuntimeConfig.DebugMode (used by the mixins) with the value of the PORTER_DEBUG environment variable. It also configures a logger and puts it on the go context, so that when mixins use tracing.StartSpan or tracing.LoggerFromContext it gets a configured logger that understands the verbosity that was set on the Porter client side.

We are now setting PORTER_VERBOSITY on the bundle's invocation image when run so that --verbosity set on the client is also used in the bundle. When the bundle is run in debug mode (i.e. porter install --debug), we also pass PORTER_VERBOSITY=debug so that a dev can see everything that is happening.

# What issue does it fix
Closes #2358 

# Notes for the reviewer
I am copy/pasting a lot of boilerplate code from porter's main into the exec mixin. In order for the other mixins to get debug mode and verbosity working, they too will need to do this. Long term I would like to use the pkg/cli package to share a common main implementation but @VinozzZ and I agreed this was the more conservative fix since we are in a release candidate.

# Checklist
- [x] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/tests/smoke/hello_test.go
+++ b/tests/smoke/hello_test.go
@@ -31,7 +31,9 @@ func TestHelloBundle(t *testing.T) {
 
 	// Install the bundle and verify the correct output is printed
 	_, output = test.RequirePorter("install", testdata.MyBuns, "--reference", testdata.MyBunsRef, "--label", "test=true", "-p=mybuns", "-c=mybuns", "--param", "password=supersecret")
-	require.Contains(t, output, "Hello, *******")
+	require.Contains(t, output, "Hello, *******", "expected to see output printed from the bundle itself")
+	// Make sure that when the mixin uses span.Debug to print the command it is running, that it's being printed
+	require.Contains(t, output, "/cnab/app ./helpers.sh install", "expected to see output printed from the porter runtime libraries")
 
 	// Should not see the mybuns installation in the global namespace
 	test.RequireInstallationNotFound("", testdata.MyBuns)
--- a/cmd/exec/main.go
+++ b/cmd/exec/main.go
@@ -1,23 +1,56 @@
 package main
 
 import (
+	"context"
+	"fmt"
 	"io"
 	"os"
+	"runtime/debug"
 
+	"get.porter.sh/porter/pkg/cli"
 	"get.porter.sh/porter/pkg/exec"
 	"github.com/spf13/cobra"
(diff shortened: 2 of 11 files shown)


tests/smoke/hello_test.go
```diff
@@ -31,7 +31,9 @@ func TestHelloBundle(t *testing.T) {
 
 	// Install the bundle and verify the correct output is printed
 	_, output = test.RequirePorter("install", testdata.MyBuns, "--reference", testdata.MyBunsRef, "--label", "test=true", "-p=mybuns", "-c=mybuns", "--param", "password=supersecret")
-	require.Contains(t, output, "Hello, *******")
+	require.Contains(t, output, "Hello, *******", "expected to see output printed from the bundle itself")
+	// Make sure that when the mixin uses span.Debug to print the command it is running, that it's being printed
+	require.Contains(t, output, "/cnab/app ./helpers.sh install", "expected to see output printed from the porter runtime libraries")
```
VinozzZ, line 36: Should we also check that `DebugMode` is also working by checking that the `Step Template` is printed out as well?

VinozzZ: Actually, I found another integration test that's testing that behavior. All good

VinozzZ (approved):
lgtm

PR #2366, opened 2022-09-20
Pass docker auth when retrieving bundle metadata

### LINKED ISSUE #2365: Existing bundle check during publish isn't using docker credentials
## Describe the bug
When I publish a new bundle to a registry that uses authentication, it fails when checking if the bundle already exists because the registry returns an auth error (401) instead of 404. We are authenticating with anonymous credentials and not the credentials in the docker login cache.

## To Reproduce
1. Create a bundle
2. Publish it to docker hub, at a location that doesn't already exist
3. Note that publish fails with an auth error

## Expected behavior
My local docker credentials should be used when checking if the bundle exists already at the published location.

## Porter Command and Output
```
$ porter publish --reference docker.io/carolynvs/gcloud-test:bug
error retrieving bundle metadata for carolynvs/gcloud-test:bug: GET https://index.docker.io/v2/carolynvs/gcloud-test/manifests/bug: UNAUTHORIZED: authentication required; [map[Action:pull Class: Name:carolynvs/gcloud-test Type:repository]]
Publish stopped because detection of carolynvs/gcloud-test:bug in the destination registry failed. To overwrite it, repeat the command with --force specified: error retrieving bundle metadata for carolynvs/gcloud-test:bug: GET https://index.docker.io/v2/carolynvs/gcloud-test/manifests/bug: UNAUTHORIZED: authentication required; [map[Action:pull Class: Name:carolynvs/gcloud-test Type:repository]]
```

## Version
1.0.0-rc.1

carolynvs:
# What does this change
Before we publish or copy a bundle, we retrieve the destination bundle's metadata as a way to check if it already exists, and requires --force to overwrite. By default crane passes the current keychain when interacting with registries. Crane uses go-containerregistry under the hood and at the time I didn't realize that it was crane, and not go-containerregistry specifying the authentication. The go-containerregistry library authenticates as anonymous by default.

So when a bundle was published to a new location, we would get a 401 auth error because authentication was not supplied, when we expected a 404 to tell that the bundle didn't exist at the destination location. I have fixed our call to get the bundle metadata so that we pass in the keychain.

# What issue does it fix
Closes #2365 

# Notes for the reviewer
N/A

# Checklist
- [x] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/build/azure-pipelines.integration.yml
+++ b/build/azure-pipelines.integration.yml
@@ -31,6 +31,13 @@ stages:
   - job: integration_test
     displayName: "Integration Test"
     steps:
+      # We log in here because TestRegistry integration test needs a valid docker session
+      # This unfortunately means that this pipeline must be manually triggered for non-maintainers
+      - task: Docker@2
+        displayName: Docker Login
+        inputs:
+          command: login
+          containerRegistry: ghcr.io/getporter
       - task: GoTool@0
         displayName: "Set Go Version"
         inputs:
--- a/build/azure-pipelines.integration.yml
+++ b/build/azure-pipelines.integration.yml
@@ -11,8 +11,20 @@ pr:
 pool:
   vmImage: "ubuntu-latest"
 
-variables:
+variables: # these are really constants
   GOVERSION: "1.18"
+  # Cache go modules and the results of go build/test
+  # Increment the version number prefix of the key and restoreKey to clear the cache
+  GOCACHE: $(Pipeline.Workspace)/.cache/go-build/
+  GOCACHE_KEY: 'v3 | go-build | "$(Agent.OS)" | go.sum'
+  GOCACHE_RESTOREKEYS: |
+    v3 | go-build | "$(Agent.OS)"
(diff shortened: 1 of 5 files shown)


VinozzZ (approved):
wow, that's such a subtle thing. Do you think we could add this discovery as a
comment above the change for the auth keychain in the code? I hope that will
help to explain why that code is needed in the future

carolynvs:
Do you mean the part about what crane does by default vs go-container registry?

carolynvs:
When I first implemented this function, I used go-containerregistry instead of crane because I was returning more than just the digest. Now that the function only returns digest, I can just use crane. I've added another commit that re-implements the function using just crane. That way we don't need to explain the differences between how the two libraries authenticate.

What do you think?

carolynvs:
@VinozzZ I have updated the solution to use crane only. Can you take another look?

VinozzZ:
> @VinozzZ I have updated the solution to use crane only. Can you take another look?

Looks good to me

PR #2371, opened 2022-09-21
Update installation instructions for mixins

carolynvs:
The porter mixins install command by default only installs stable (non-prerelease) versions. Update our mixin instructions to explicitly have the user set a version so that they get the most recent release.


carolynvs:
@VinozzZ  Can you look this over and help check that the versions I listed match the ones we are releasing today?

VinozzZ (approved):
I have double checked the versions and they match with what we are releasing today

PR #2428, opened 2022-10-20
Consistently set and use SchemaType when exporting/importing

### LINKED ISSUE #2269: porter parameters/credentials apply ignores the schemaType
## Describe the bug
It is possible to create a ParameterSet as a YAML file and apply it as a CredentialSet and the other way around. I would expect porter to complain as the schemaType does not match. 

## To Reproduce
1. Run `porter parameters create myparams.yaml`
2. See `creating porter parameter set in the current directory`
3. Run `porter credentials apply myparams.yaml`
4. See error: `Applied NAMESPACE/NAME credential set`
5. Or run `porter credentials create mycreds.yaml`
6. See `creating porter credential set in the current directory`
7. Run `porter parameters apply mycreds.yaml`
8. See error: `Applied NAMESPACE/NAME parameter set`

## Expected behavior
Porter should abort, stating that a CredentialSet can not be applied as a ParameterSet and the other way around.

## Porter Command and Output
See above

## Version
porter v1.0.0-beta.1 (eaf8d0d4)

### LINKED ISSUE #2348: build fails when schemaType is set in porter.yaml
We documented that schemaType was supported for the porter manifest, see https://getporter.org/bundle/manifest/file-format/#example, but when you actually set it, porter fails to validate the manifest because it's an unexpected field.

```
$ porter build
Using buildkit build driver
unsupported property set or a custom action is defined incorrectly: error unmarshaling custom action schemaType: yaml: unmarshal errors:
  line 1: cannot unmarshal !!str `Bundle` into manifest.Steps
```

This is on porter v1.0.0-rc.1

Reproduction Steps:
1. Run porter create
2. Edit the bundle and include `schemaType: Bundle` at the top
3. Run porter build

carolynvs:
# What does this change
All resource files should have both a schemaVersion and a schemaType. The schemaType field serves multiple purposes:
* Porter can validate it when present during the `porter * apply` commands to catch if the wrong file type was passed
* In the future VS Code can provide autocomplete based on the schemaType and schemaVersion

In some cases resources already supported schemaType but didn't validate it or store the value in the database. It is now an official field on CredentialSetSpec, InstallationSpec, ParameterSetSpec, and the bundle Manifest (which isn't persisted to the db).

The schemaVersion for resources has been incremented and porter supports multiple versions of these resources at the same time. We will migrate to the most recent schema version when validating the resource before it's saved to the database.

I've updated our file format docs so that people can see details about a particular version of a schema, and track what changes were made between versions.

# What issue does it fix
Closes #2348 
Closes #2269 

# Notes for the reviewer
N/A

# Checklist
- [x] Did you write tests?

# Reviewer Checklist


pkg/porter/credentials.go
```diff
@@ -233,9 +245,9 @@ func (p *Porter) ShowCredential(ctx context.Context, opts CredentialShowOptions)
 	}
 
 	credSet := DisplayCredentialSet{
-		SchemaType:    "CredentialSet",
 		CredentialSet: cs,
 	}
+	credSet.SchemaType = "CredentialSet"
```
VinozzZ, line 250: I wonder if we can declaring a constant for each `SchemaType` s will help to consolidate the hard coded string into one place

pkg/storage/installation.go
```diff
@@ -30,6 +31,9 @@ type Installation struct {
 
 // InstallationSpec contains installation fields that represent the desired state of the installation.
 type InstallationSpec struct {
+	// SchemaType indicates the type of resource imported from a file.
```
VinozzZ, line 34: Do we need to have a migration step to populate values for this field for an existing database?

carolynvs: SchemaType is not required and supports an empty string for a value, so adding a new optional field doesn't impact existing records.

VinozzZ: I think it's a bit confusing since the value is manually set in the code but not required in the database. If in the future some other area of the code path references the SchemaType, then it's hard to trace at what point during the code execution the value is populated. That's my main concern here

carolynvs: How about I populate the schemaType during apply if it's unset? Or do you only see a full db migration as the acceptable solution?

VinozzZ: populating it during apply for installation record sounds good to me

carolynvs: Thanks! I really appreciate you thinking of this before I messed it up. 😊

carolynvs:
Moving back to draft while I work on bumping the schemaVersion for all resources that had schemaType added and getting a JIT migration built into Validate()

pkg/storage/parameterset.go
```diff
@@ -84,21 +85,63 @@ func (s ParameterSet) DefaultDocumentFilter() map[string]interface{} {
...
+	//lint:ignore SA4006 ignore unused ctx for now
+	ctx, span := tracing.StartSpan(ctx)
+	defer span.EndSpan()
+
+	for s.SchemaVersion != DefaultParameterSetSchemaVersion {
+		switch s.SchemaVersion {
+		// In 1.0.2 we added SchemaType and any resources with it unset should have the default value applied
+		case "1.0.1":
+			span.Info("Migrating parameter set from 1.0.1 to 1.0.2")
+			// There isn't an explicit migration for this, because schemaType is funny
```
VinozzZ, line 136: I'm curious why you moved the code for setting schemaType if it's empty out of this `Migrate` function

VinozzZ (approved):

pkg/storage/credentialset.go
```diff
@@ -77,13 +86,63 @@ func (s CredentialSet) DefaultDocumentFilter() map[string]interface{} {
...
+}
+
+// Migrate checks if the resources is in an older schema version, and if so applies successive migrations until it is up-to-date.
+func (s *CredentialSet) Migrate(ctx context.Context) error {
+	//lint:ignore SA4006 ignore unused ctx for now
+	ctx, span := tracing.StartSpan(ctx)
+	defer span.EndSpan()
+
+	for s.SchemaVersion != DefaultCredentialSetSchemaVersion {
+		switch s.SchemaVersion {
```
sgettys, line 133: nit: This migration (and the other resource migration code) seems like it could get unwieldy quickly and become a bit of a maintenance headache having all of this switch logic here. Is there a way to simplify schema migration for resources using like the strategy pattern or something along those lines?

sgettys (approved):

pkg/storage/parameterset.go
carolynvs: Originally you were concerned about inconsistently setting the schemaType, and having a hard time understanding when it's getting set. When I moved setting the schemaType only when migrating, I realized that only creates more opportunity for inconsistency.

The migration is only triggered if their schemaVersion is on an old version, which leaves open paths for people to still end up with the schemaType unset.

For example:

1. I have a resource with "schemaVersion: 1.0.1" and before applying it, see that I'm on an old schemaVersion, so I update the schemaVersion to 1.0.2.
2. I apply the resource, and since the version is "up-to-date", the schemaType is left empty.

or

1. I have a resource on 1.0.1 without the schemaType set.
2. I apply the resource and it's auto migrated to 1.0.2 and the type is set.
3. I keep using the original resource file, but perhaps realize that the version is out of date, and update it to 1.0.2 but otherwise keep the same file.
4. I apply it again, and now the schemaType in the database is unset.

or

1. I have a new resource that I made by hand. I set the schemaVersion to 1.0.2 but forget to set the schemaType
2. I apply it and the schemaType is empty in the database (because no migration was applied).

If the concern that drove us to set the value in Validate() was consistency, only defaulting in the migration doesn't solve the consistency problem. I originally wasn't concerned about if schemaType is set in the db, which is why I left it off but was willing to add the defaulting to address the consistency concern. So we should either ensure porter sets it consistently or just not bother and go back to the original because this was a lot of changes to remain inconsistent. 😁

carolynvs:
I'm starting to think that I'm going down the wrong path here. The change that made us think we should bump schemaVersion and do a migration is that schemaType changed from being a display only field to something persisted in the db.

I'm mostly concerned with allowing people using the operator to copy/paste the output of a porter show command into the spec of a k8s manifest. I thought having schemaType on the Spec would be helpful (I keep hoping we can eventually embed a Spec in the Operator instead of copying the fields), but I feel like it's causing a lot of trouble for not a lot of good.

Maybe I should go back to having SchemaType just on DisplayInstallation/DisplayCredentialSet/etc and validate on import but schemaType is never persisted to the db? 🤔 

This PR feels like a lot considering the original request for change "throw an error when I import the wrong schemaType" was pretty small...

pkg/storage/parameterset.go
VinozzZ: Thank you for the clarification!

carolynvs:
I spoke with Yingrong and we agreed that schemaVersion is a user facing compatibility guarantee. What we put in the database is an implementation detail. We should make sure that porter doesn't do anything with it's database to not be compatible with the schema but they don't have to match 1:1 at all times. In this case defaulting the schemaVersion in Validate is an implementation detail but doesn't require a schema version bump.

I will do the following:

1. Remove the migration part, since there's nothing to really migrate.
2. Rollback the schemaVersion bump but keep the nice doc changes.
3. Document how we decide when to bump SchemaVersion, that it's a user-facing compatibility guarantee, etc in the contributing guide.

carolynvs:
@VinozzZ @sgettys Sorry for the rebase but I needed to cleanly back out the recent changes around migration and this was easier. I've made the changes requested, can you take another look?

VinozzZ (approved):

PR #2435, opened 2022-10-24
Split Installation into a spec and status

carolynvs:
I have updated Installation to be composed of
* Instance specific fields that make testing difficult, such as the ID since tests have to make those values consistent to compare data structures easily
* Spec - all fields that a user would set on an installation
* Status - our own internal tracking of the installation status.

This makes it easier for us to test or work with just the values the user specified by having a spec that won't have auto-generated values like IDs, timestamps, etc. It also enables us to embed an installation spec in another data structure, like a workflow that runs multiple installations, without having unnecessary fields embedded like the status and id. It lets us think about just the desired state of the installation.

This is refactoring work in support of PEP003 - Advanced Dependencies.

Questions for reviewers:
- Are there any concerns about the impact of this on the operator data structures? I would love to be able to embed specs for any types used in the operator so that they aren't copy/pasted. If we used a different yaml library to marshal them, we don't need to add yaml tags to the source structs even (though I'd rather add yaml tags to the porter structs vs require copy/paste)
- Do we need a general purpose "Meta" structure, or is defining ID/SchemaVersion directly on Installation is okay long term?

### DIFF
--- a/pkg/cnab/provider/credentials_test.go
+++ b/pkg/cnab/provider/credentials_test.go
@@ -49,7 +49,12 @@ func TestRuntime_loadCredentials(t *testing.T) {
 		},
 	})
 
-	args := ActionArguments{Installation: storage.Installation{CredentialSets: []string{"mycreds"}}, Action: "install"}
+	args := ActionArguments{
+		Installation: storage.Installation{
+			InstallationSpec: storage.InstallationSpec{
+				CredentialSets: []string{"mycreds"}},
+		},
+		Action: "install"}
 	gotValues, err := r.loadCredentials(context.Background(), b, args)
 	require.NoError(t, err, "loadCredentials failed")
 
@@ -58,7 +63,12 @@ func TestRuntime_loadCredentials(t *testing.T) {
 	}
 	assert.Equal(t, wantValues, gotValues, "resolved unexpected credential values")
 
-	args = ActionArguments{Installation: storage.Installation{CredentialSets: []string{"/db-creds.json"}}, Action: "install"}
+	args = ActionArguments{
+		Installation: storage.Installation{
+			InstallationSpec: storage.InstallationSpec{
+				CredentialSets: []string{"/db-creds.json"}},
+		},
+		Action: "install"}
 	_, err = r.loadCredentials(context.Background(), b, args)
 	require.Error(t, err, "loadCredentials should not load from a file")
 }
@@ -137,7 +147,12 @@ func TestRuntime_loadCredentials_WithApplyTo(t *testing.T) {
 		require.NoError(t, err, "Save credential set failed")
 
 		b := getBundle(true)
-		args := ActionArguments{Installation: storage.Installation{CredentialSets: []string{"mycreds"}}, Action: "install"}
+		args := ActionArguments{
+			Installation: storage.Installation{
+				InstallationSpec: storage.InstallationSpec{
+					CredentialSets: []string{"mycreds"}},
+			},
+			Action: "install"}
 		gotValues, err := r.loadCredentials(context.Background(), b, args)
 		require.NoError(t, err, "loadCredentials failed")
 		assert.Equal(t, secrets.Set{"password": "mypassword"}, gotValues)
--- a/pkg/porter/apply.go
+++ b/pkg/porter/apply.go
@@ -85,12 +85,12 @@ func (p *Porter) InstallationApply(ctx context.Context, opts ApplyOptions) error
 
 		// Create a new installation
 		installation = storage.NewInstallation(input.Namespace, input.Name)
-		installation.Apply(inputInstallation)
+		installation.Apply(inputInstallation.InstallationSpec)
 
 		log.Info("Creating a new installation", attribute.String("installation", installation.String()))
 	} else {
 		// Apply the specified changes to the installation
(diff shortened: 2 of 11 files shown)


sgettys:
So from an operator Installation CRD perspective it would look like this?
```
apiVersion: porter.sh/v1
kind: Installation
metadata:
  name: hello-llama
  labels:
    porter.sh/testdata: "true"
spec:
  schemaVersion: 1.0.2
  installationSpec:
    namespace: operator
    name: mellama
    bundle:
      repository: getporter/hello-llama
      version: 0.1.1
    parameters:
      name: "Porter Operator"
```

And the InstallationSpec in the operator API installation type becomes:
```
type InstallationSpec struct {
	// AgentConfig is the name of an AgentConfig to use instead of the AgentConfig defined at the namespace or system level.
	// +optional
	AgentConfig *corev1.LocalObjectReference `json:"agentConfig,omitempty" yaml:"-"`

	// PorterConfig is the name of a PorterConfig to use instead of the PorterConfig defined at the namespace or system level.
	PorterConfig *corev1.LocalObjectReference `json:"porterConfig,omitempty" yaml:"-"`

	//
	// These are fields from the Porter installation resource.
	// Your goal is that someone can copy/paste a resource from Porter into the
	// spec and have it work. So be consistent!
	//

	// SchemaVersion is the version of the installation state schema.
	SchemaVersion string `json:"schemaVersion" yaml:"schemaVersion"`

        // InstallationSpec is the full InstallationSpec for the Installation resource
        InstallationSpec porter.InstallationSpec `<whatever these tags are>`

	// Labels applied to the installation.
	Labels map[string]string `json:"labels,omitempty" yaml:"labels,omitempty"`

}
```

I guess the only issue I might have with that is the overloading of spec and how it collides with "spec" in k8s. So essentially it's installation.spec.installationSpec which definitely doesn't read well.

In terms of the Meta field I'm assuming this would be unstructured data and Porter internals would only use the "ID" field out of it? Or would there be potential for other fields to be added later on and meta would just become the dumping ground for unstructured data? 
Would putting a new expected field from porter into meta tick over the schema version? My concern might be that unstructured meta could make schema version less useful

carolynvs:
I would like to embed the InstallationSpec struct in place of the copy/paste fields, but not inside a nested spec. This would keep the current file representation of an Installation in the operator looking exactly the same as it does today.

```go
// the operator's installationspec
type InstallationSpec struct {
	// AgentConfig is the name of an AgentConfig to use instead of the AgentConfig defined at the namespace or system level.
	// +optional
	AgentConfig *corev1.LocalObjectReference `json:"agentConfig,omitempty" yaml:"-"`

	// PorterConfig is the name of a PorterConfig to use instead of the PorterConfig defined at the namespace or system level.
	PorterConfig *corev1.LocalObjectReference `json:"porterConfig,omitempty" yaml:"-"`

        // embed porter's installation spec, do not add another level of "spec:"
	storage.InstallationSpec `yaml:",inline"`
}
```

carolynvs:
> My concern might be that unstructured meta could make schema version less useful

Agreed, I would like to just embed the ID and perhaps the SchemaVersion directly on storage.Installation and not deal with a generic/untyped Meta. I still waffling about if we can move SchemaVersion under spec or not. I'll try updating this patch to have schemaVersion under spec and see if there's any unwanted fallout.

I just raised the question since this PR is basically copying patterns from K8s and they use a Meta. I wasn't sure if there were reasons that we may want to as well.

sgettys:
> > My concern might be that unstructured meta could make schema version less useful
> 
> Agreed, I would like to just embed the ID and perhaps the SchemaVersion directly on storage.Installation and not deal with a generic/untyped Meta. I still waffling about if we can move SchemaVersion under spec or not. I'll try updating this patch to have schemaVersion under spec and see if there's any unwanted fallout.
> 
> I just raised the question since this PR is basically copying patterns from K8s and they use a Meta. I wasn't sure if there were reasons that we may want to as well.

Yeah it can be useful but also can be confusing. We were working with backstage yesterday and their resource definition directly mirrors k8s so it was really confusing thinking are we dealing with a k8s resource or a backstage resource. Although for something like "meta" it's not quite the same since that's much more generic and the porter installation definition is clearly a porter object that can't be confused immediately with a k8s resource definition

carolynvs:
Let me know if you see ways that we can improve usability for concerns like that! We aren't locked into a format yet for the operator so now's the time if we see ways to avoid mistakes from other projects who already integrated with k8s and take advantage of their lessons learned.

I'm hoping the spec change will reduce the effort on us to maintain these structs but really it's not so bad so if this seems like a step backward usability-wise, let me know! I can just limit the change to only porter and not also apply it to the operator. I can open a PR later to see how it looks.

sgettys (approved):

carolynvs:
I went ahead and moved SchemaVersion into Spec as well, since that's a user-specified field.

When I tried to use storage.InstallationSpec instead of the copy/paste fields I immediately remembered why we don't do that. The operator-sdk/kubernetes requires that everything on the CRD implement DeepCopyInto which it can't generate for an external type. If we want to embed porter types in the operator, we will first need to use the generator in the porter repo to implement those copy methods. I'll make a separate issue to track that since I think reuse would be nice.

sgettys (approved):

PR #2449, opened 2022-11-09
Split codeql from trivy

carolynvs:
After we upgraded from CodeQL v1 to v2, the performance of our "security scan" workflow went from 15m to over an hour, usually timing out. When CodeQL is initialized it replaces the go binary with their own custom one, and any builds done with it are included in the analysis.

I have moved everything that doesn't need to be analyzed to before initializing CodeQL, and also split out the cross-compilation required by Trivy (which scans our porter agent image) into a separate workflow. This way only the native builds for the porter client, runtime, exec mixin and agent are scanned by CodeQL.

This has brought performance back to a tolerable level, about 12m again.

### DIFF
--- a/.github/workflows/codeql.yaml
+++ b/.github/workflows/codeql.yaml
@@ -1,0 +1,40 @@
+# Scan our binaries for vulnerabilities
+name: CodeQL
+on:
+  # always do a security scan when a change is merged
+  push:
+    branches-ignore:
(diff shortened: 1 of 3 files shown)


carolynvs:
Trivy found existing alerts about the version of the docker cli we are using. We can't upgrade until Buildx works with recent versions of Docker, so we are pinned to a compatible version.

PR #2454, opened 2022-11-13
Improve applying installation parameters without reparsing as flags

carolynvs:
# What does this change
Do not re-parse parameters set on an installation resource as --param flags and instead apply any --param flag overrides on top of parameters defined on the installation.

# What issue does it fix
Refactoring work for PEP003-advanced-dependencies. This is groundwork for having a bundle workflow use advanced wiring when passing parameters to a dependency (i.e. directing the output of a dependency to a parameter of another).

# Notes for the reviewer
This is a change in behavior that I think we want. Currently, when any --param flags are specified, any parameters previously specified with --param are cleared out. Here's an example:

1. porter install --param env=dev
2. porter upgrade --param logLevel=11

The upgrade will undo the env=dev param initially set in install. This forces the user to always specify every single parameter every time they run porter, or move them all into parameter sets.

I don't think it's clear to users that setting a different parameter would change an unrelated parameter on the installation.

What this PR does instead is applies explicit CLI overrides on top of parameters specified on the installation (i.e. previously explicit CLI overrides, or params set by the new depsv2 feature). So when the upgrade is run and logLevel is changed, the env parameter would not be cleared out.

# Checklist
- [ ] Did you write tests? TODO
- [ ] Did you write documentation? TODO, explain the override, i.e. an installation with lots of params set, and you use --param to change just 1 using porter upgrade.

# Reviewer Checklist


carolynvs:
This was such an awkward change to make, and the code for resolving parameters is spread out throughout the codebase to the point that I really think resolution needs a refactor so that we can just call it once and have it take care of all resolution concerns: defaults, overrides, parameter sets, previous values set on an installation, etc.

Like the change works but ... it's so hard to maintain that I don't have a lot of confidence in it for someone to maintain going forward. I'll timebox a refactor and see if it's a rabbit hole or a great idea.

carolynvs:
Closing in favor of #2476

PR #2476, opened 2022-12-01
Refactor param resolution

carolynvs:
# What does this change
Refactors how we resolve parameters so that it's easier to think about and maintain. Previously the logic for parsing and resolving parameters was spread out all over the place and was composed of many (unfortunately stateful) functions.

I've combined all that logic into a single function that resolves all user-defined parameters (defaults and parameter sources are still done separately since we need to be able to distinguish between the two, and the params are represented in each with different data types).

I had to cheat and use a bit of state to identify parameters that were set on the CLI or through a parameter set that really were parameters for a dependency of the bundle being executed. Dependencies v1 (the current implementation) allows the user to directly set a parameter for a dependency. The new implementation doesn't work that way (the parent bundle always manages passing data to the dependencies instead). Since eventually that code will go away and I really don't want to touch it, I went with the smallest possible fix in that area of the code.

This is refactoring for a) my happiness and that of future maintainers and b) to enable setting parameters differently for dependencies v2.

# What issue does it fix
N/A

# Notes for the reviewer
N/A

# Checklist
- [x] Did you write tests?
- [ ] Did you write documentation? TODO: document more clearly the hierarchy of how we set parameters from the default, parameter source, parameters set from a parameter set, explicit overrides used in previous runs of that installation and the current parameter sets and explicit overrides on the current run.

# Reviewer Checklist


sgettys (approved):

pkg/porter/upgrade.go
```diff
@@ -75,15 +75,22 @@ func (p *Porter) UpgradeBundle(ctx context.Context, opts *UpgradeOptions) error
...
-	if err != nil {
-		return fmt.Errorf("could not apply options to installation: %w", err)
-	}
 	i.Status.Modified = time.Now()
 	err = i.Validate()
 	if err != nil {
 		return err
 	}
+
+	_, err = p.applyActionOptionsToInstallation(ctx, opts, &i)
```
VinozzZ: Should we validate the installation record after applying options?

PR #2478, opened 2022-12-02
Trace sensitive attributes in custom dev builds

carolynvs:
# What does this change
Instead of adding and removing println statements that write out sensitive data to assist with troubleshooting while writing code, only to have to remember to remove them before committing, I've added tracing.SetSensitiveAttributes.

In all normal builds created by go build, or our magefile targets, that function is a noop and does nothing. However if you build with go build -tags traceSensitiveAttributes, then that function will write out the passed attributes (which contain sensitive data) to the listening open telemetry collector (if any).

When porter runs with traceSensitiveAttributes set, a warning is printed to stderr to indicate the different behavior.

I hope this will give us a way to make it easy to debug in development while ensuring that we don't ever actually trace sensitive attributes in real builds.

# What issue does it fix
I really needed this when I was refactoring parameter resolution, and in the past I know we have accidentally committed println statements that we shouldn't have. So this puts in a safer way to help us develop this area of the code.

# Notes for the reviewer
I considered making this a configuration setting, but I thought this did a better job of making it real hard to someone to accidentally leak sensitive data. Most people don't do their own builds, let alone with artisanal build flags so if you end up with a build of Porter that traces sensitive data, it's clear they opted into it.

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist


VinozzZ (approved):
lgtm!

PR #2504, opened 2022-12-28
Install multiple plugins by passing in a file argument to porter plugins install

### LINKED ISSUE #2491: Add ability to install multiple plugins
### Background
Currently porter allows users to install plugin one at a time through command `porter plugins install <plugin-name>`. This works if users are running porter in a bash script or manually. 
However, in porter-operator, we recently implemented the functionality to install plugins through a custom resource. This requires porter to be able to install multiple plugins with one command.

### Implementation
I think we can add a new option `-f config.yaml` into the [`porter plugins install` command](https://github.com/getporter/porter/blob/1da56edd297e83ff6a8c9f0c4f3d1b3003624636/pkg/porter/plugins.go#L153). the file should contain definitions of plugins. 

example plugins config.yaml file
 ```yaml
kubernetes:
  version: v1.0.0
  URL: https://cnd.porter.sh/plugins/atom.xml
azure:
   version: v1.0.0
  URL: https://test/plugins/atom.xml
```

### Requirements
- The input file should use map data structure for plugins configuration. This is to signal users that the order defined in the config file does not control the order in which plugins are installed.
- The order of plugins installation should be sorted using alphabetical order. We need to make sure that all the plugins are always installed in the same order.
- Versions and feedURL should be optional. The default value for version should be "latest". The default value for feedURL should be the official porter plugins feed URL.
- The value set in `PORTER_MIRROR` should be honored when no feed url is specified in the config file.

### PR DESCRIPTION
**Pull request by VinozzZ**
# What does this change

This PR adds a new `--file` flag to the `porter plugins install` command.
This allows users to define multiple plugins in a file and install them with a single porter command.

# What issue does it fix
Closes #2491

# Notes for the reviewer

Maybe we need to define a schema for this plugins config file?

# Checklist
- [x] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/plugins/install.go
+++ b/pkg/plugins/install.go
@@ -1,14 +1,97 @@
 package plugins
 
 import (
+	"encoding/json"
+	"fmt"
+	"sort"
+
 	"get.porter.sh/porter/pkg/pkgmgmt"
+	"get.porter.sh/porter/pkg/portercontext"
 )
 
 type InstallOptions struct {
 	pkgmgmt.InstallOptions
+
+	File string
 }
 
-func (o *InstallOptions) Validate(args []string) error {
+func (o *InstallOptions) Validate(args []string, cxt *portercontext.Context) error {
 	o.PackageType = "plugin"
+	if o.File != "" {
+		if len(args) > 0 {
+			return fmt.Errorf("plugin name should not be specified when --file is provided")
+		}
+
+		if o.URL != "" {
+			return fmt.Errorf("plugin URL should not be specified when --file is provided")
+		}
+
+		if o.Version != "" {
+			return fmt.Errorf("plugin version should not be specified when --file is provided")
+		}
+
+		if _, err := cxt.FileSystem.Stat(o.File); err != nil {
+			return fmt.Errorf("unable to access --file %s: %w", o.File, err)
+		}
+
+		return nil
+	}
+
 	return o.InstallOptions.Validate(args)
+}
+
+// InstallFileOption is the go representation of plugin installation file format.
+type InstallFileOption struct {
+	data map[string]pkgmgmt.InstallOptions
+	keys []string
+}
+
+func (io *InstallFileOption) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	data := map[string]pkgmgmt.InstallOptions{}
+	err := unmarshal(&data)
+	if err != nil {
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: pkg/plugins/install.go [hunk 1]
```diff
@@ -1,14 +1,97 @@
...
+	configs := map[string]pkgmgmt.InstallOptions{}
+	err := json.Unmarshal(data, &configs)
+	if err != nil {
+		return fmt.Errorf("could not unmarshal into plugins.InstallFileOption: %w", err)
+	}
+
+	return io.sort(configs)
+}
+
+// sort returns InstallOptions in alphabetical order.
```
sgettys, line 68: Nit: This sorts the plugins and stores that in the InstallOptions, doesn't return them.

pkg/plugins/install.go
```diff
@@ -1,14 +1,97 @@
...
+	})
+
+	io.data = data
+	io.keys = keys
+
+	return nil
+
+}
+
+// Configs returns InstallOptions list in alphabetical order.
```
sgettys, line 90: This does assume that "sort" was called by either of the unmarshals. Should sort be called here to ensure alphabetical or do you think we're fine assuming that the proper unmarshaler was called?

pkg/plugins/install.go
```diff
@@ -1,14 +1,97 @@
...
+	data := map[string]pkgmgmt.InstallOptions{}
+	err := unmarshal(&data)
+	if err != nil {
+		return fmt.Errorf("could not unmarshal into plugins.InstallFileOption: %w", err)
+	}
+	return io.sort(data)
+}
+
+func (io *InstallFileOption) UnmarshalJSON(data []byte) error {
+	configs := map[string]pkgmgmt.InstallOptions{}
```
sgettys, line 59: nit: call both "data" or "configs" between unmarshalers for consistency

pkg/porter/plugins.go
```diff
@@ -177,3 +190,44 @@ func (p *Porter) UninstallPlugin(ctx context.Context, opts pkgmgmt.UninstallOpti
 
 	return nil
 }
+
+func (p *Porter) getInstallConfigs(ctx context.Context, opts plugins.InstallOptions) ([]pkgmgmt.InstallOptions, error) {
```
sgettys, line 194: Nit: Call this maybe getPluginsInstallConfigs so we don't confuse this with "Installations" down the road

sgettys:
Need docs/schema around new plugins config file format

sgettys (approved):

PR #2505, opened 2022-12-29
Fix logger level setup

VinozzZ:
# What does this change

Currently porter initialize the logger before command-line flags are being applied.This creates an issue with the `--verbosity` flag not being applied to the log level.

This PR fixes the issue by utilize `zap.AtomicLevel` to change log level during runtime.

# What issue does it fix
Closes https://github.com/getporter/hashicorp-plugins/issues/4

# Notes for the reviewer

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/portercontext/context.go
+++ b/pkg/portercontext/context.go
@@ -71,6 +71,7 @@ type Context struct {
 	tracerInitalized bool
 
 	// handles send log data to the console/logfile
+	atom   *zap.AtomicLevel
 	logger *zap.Logger
 
 	// IsInternalPlugin indicates that Porter is running as an internal plugin
@@ -167,9 +168,16 @@ func (c *Context) ConfigureLogging(ctx context.Context, cfg LogConfiguration) {
 		c.logCfg.TelemetryServiceName = c.InternalPluginKey
 		baseLogger = c.makePluginLogger(c.InternalPluginKey, cfg)
 	} else {
-		baseLogger = c.makeConsoleLogger()
+		// currently the root logger is created before flags are being applied
+		// only change the log level when the root logger already initialized
+		if c.atom != nil {
+			c.atom.SetLevel(c.logCfg.Verbosity)
+		} else {
+			atom := zap.NewAtomicLevelAt(c.logCfg.Verbosity)
+			c.atom = &atom
+		}
+		baseLogger = c.makeConsoleLogger(c.atom)
 	}
-
 	c.configureLoggingWith(ctx, baseLogger)
 }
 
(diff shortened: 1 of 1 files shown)


pkg/portercontext/context.go
```diff
@@ -167,9 +168,16 @@ func (c *Context) ConfigureLogging(ctx context.Context, cfg LogConfiguration) {
...
-		baseLogger = c.makeConsoleLogger()
+		// currently the root logger is created before flags are being applied
+		// only change the log level when the root logger already initialized
+		if c.atom != nil {
+			c.atom.SetLevel(c.logCfg.Verbosity)
+		} else {
+			atom := zap.NewAtomicLevelAt(c.logCfg.Verbosity)
+			c.atom = &atom
+		}
+		baseLogger = c.makeConsoleLogger(c.atom)
```
carolynvs, line 179: I had to try this out locally to verify that it does fix the --verbosity flag, but I don't understand why. 😂

When we configure the logger after applying the flags, we don't change the existing logger but make a completely new one. So while this totally fixes it, I am not doubting that, I'm wondering if there's more going on. Like I only _think_ that I'm making a new logger but really the code is using the original one still... 🤔 

I'm going to dig a little bit more to test out that idea.

VinozzZ: It did take a me a while to understand what's going on. The reason why this change is able to fix the verbosity flag is because we are applying the same AtomLevel object when creating a new logger. That's why the `makeConsoleLogger()` takes in a `*zap.AtomicLevel` and apply it to the new logger

carolynvs: Ok yeah what is happening is that Porter stores the RootTraceLogger on the Context after we first create it, and then when we update the logging configuration after the flags are applied, Porter doesn't update the logger on the Context.

That's why your fix of using atomic works, it allows the logger stored in the context to have it's values changed later (since we are still using that logger and not the updated one when we call tracing.LoggerFromContext).

I have a couple different ways to fix this that I've tried out locally and I think it would be good to consider taking some/all of them:

* Move initialization of the logger into Porter.Connect and only call it once. Porter.Close becomes responsible for ending the span and closing the logger instead of main.
* Be okay with not having as much info logged until we finish loading the full Porter config
* "Update" the logger stored on the Context when the root logger is changed, and ensure that the updated context is passed to subsequent calls.

I can push these up to a branch on this repo for you to look at or maybe we can do a zoom and go over which changes seem best for long term maintenance?

VinozzZ: I'm happy to do a zoom call to find a best long term solution

PR #2532, opened 2023-01-24
check for plugins file schema version and type

VinozzZ:
# What does this change

Previously, the `SchemaType` and `SchemaVersion` are not parsed or validated. This PR adds the corresponding logic to properly process these fields

# What issue does it fix

# Notes for the reviewer

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/docs/content/reference/file-formats.md
+++ b/docs/content/reference/file-formats.md
@@ -162,21 +162,22 @@ You can use this [json schema][plugins-schema] to validate a plugins config file
 ```yaml
 schemaType: Plugins
 schemaVersion: 1.0.0
-azure:
-  version: v1.0.0
-  feedURL: https://cdn.porter.sh/plugins/atom.xml
-  url: https://example.com
-  mirror: https://example.com
+plugins:
+  azure:
+    version: v1.0.0
+    feedURL: https://cdn.porter.sh/plugins/atom.xml
+    url: https://example.com
+    mirror: https://example.com
 ```
 
 | Field                | Required | Description                                                                                                                                    |
 |----------------------|----------|------------------------------------------------------------------------------------------------------------------------------------------------|
 | schemaType           | false    | The type of document. This isn't used by Porter but is included when Porter outputs the file, so that editors can determine the resource type. |
 | schemaVersion        | true     | The version of the Plugins schema used in this file.                                                                                     |
-| <pluginName>.version | false    | The version of the plugin.                                                                                                                 |
-| <pluginName>.feedURL | false    | The url of an atom feed where the plugin can be downloaded.
-| <pluginName>.url     | false    | The url from where the plugin can be downloaded.                                                                                                                 |
-| <pluginName>.mirror  | false    | The mirror of official Porter assets.                                                                                                                 |
+| plugins.<pluginName>.version | false    | The version of the plugin.                                                                                                                 |
+| plugins.<pluginName>.feedURL | false    | The url of an atom feed where the plugin can be downloaded.
+| plugins.<pluginName>.url     | false    | The url from where the plugin can be downloaded.                                                                                                                 |
+| plugins.<pluginName>.mirror  | false    | The mirror of official Porter assets.                                                                                                                 |
 
 [cs-schema]: /schema/v1/credential-set.schema.json
 [ps-schema]: /schema/v1/parameter-set.schema.json
--- a/pkg/plugins/install.go
+++ b/pkg/plugins/install.go
@@ -6,7 +6,12 @@ import (
 
 	"get.porter.sh/porter/pkg/pkgmgmt"
 	"get.porter.sh/porter/pkg/portercontext"
+	"github.com/cnabio/cnab-go/schema"
 )
+
+// InstallPluginsSchemaVersion represents the version associated with the schema
+// plugins configuration documents.
+var InstallPluginsSchemaVersion = schema.Version("1.0.0")
 
 type InstallOptions struct {
 	pkgmgmt.InstallOptions
@@ -25,8 +30,9 @@ func (o *InstallOptions) Validate(args []string, cxt *portercontext.Context) err
 			return fmt.Errorf("plugin URL should not be specified when --file is provided")
 		}
 
-		if o.Version != "" {
-			return fmt.Errorf("plugin version should not be specified when --file is provided")
+		// version should not be set to anything other than the default value
+		if o.Version != "" && o.Version != "latest" {
+			return fmt.Errorf("plugin version %s should not be specified when --file is provided", o.Version)
 		}
 
 		if _, err := cxt.FileSystem.Stat(o.File); err != nil {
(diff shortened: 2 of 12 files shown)


docs/content/reference/file-formats.md
```diff
@@ -162,21 +162,22 @@ You can use this [json schema][plugins-schema] to validate a plugins config file
 ```yaml
 schemaType: Plugins
 schemaVersion: 1.0.0
-azure:
-  version: v1.0.0
-  feedURL: https://cdn.porter.sh/plugins/atom.xml
-  url: https://example.com
-  mirror: https://example.com
+plugins:
```
carolynvs, line 165: oops! I'm sorry that I didn't catch that during review...

pkg/plugins/install.go
```diff
@@ -25,8 +30,9 @@ func (o *InstallOptions) Validate(args []string, cxt *portercontext.Context) err
 			return fmt.Errorf("plugin URL should not be specified when --file is provided")
 		}
 
-		if o.Version != "" {
-			return fmt.Errorf("plugin version should not be specified when --file is provided")
+		// version should not be set to anything other than the default value
+		if o.Version != "" && o.Version != "latest" {
```
carolynvs, line 34: When --version is empty, it defaults to latest so I'm not sure I get what this is checking for or guarding against? The error message just says that they can't specify --version at all so can't we just stick with the original check that version was specified?

pkg/plugins/install.go
```diff
@@ -39,20 +45,37 @@ func (o *InstallOptions) Validate(args []string, cxt *portercontext.Context) err
...
-type InstallFileOption map[string]pkgmgmt.InstallOptions
+// InstallPluginsSpec represents the user-defined configuration for plugins installation.
+type InstallPluginsSpec struct {
+	SchemaType    string               `yaml:"schemaType"`
+	SchemaVersion string               `yaml:"schemaVersion"`
+	Plugins       InstallPluginsConfig `yaml:"plugins"`
+}
+
+func (spec InstallPluginsSpec) Validate() error {
+	if InstallPluginsSchemaVersion != schema.Version(spec.SchemaVersion) {
```
carolynvs, line 56: I just remembered that we are missing schemaType validation on our other resources. I had created a [branch to fix](https://github.com/carolynvs/porter/tree/schema-type-validation) that and then got pulled off onto other things and forgot to submit a PR.

Let's do a check here for the schemaType as well, so that if they enter in a value other than what's expected, we don't try to process the file further. For example, they accidentally pass a bundle or installation file to the -f parameter.

https://github.com/carolynvs/porter/blob/ea694aa668a448eb2293775da5f5cfda3567214f/pkg/manifest/manifest.go#L185-L187

I'll submit my PR afterwards that adds the check to the other resources.

carolynvs (changes requested):
Since this is changing the file format used from what was just published in v1.0.5, we'll need to communicate that this is a breaking change and update the release notes on 1.0.5 to call out that it's a bad release effectively.

In addition to the usual announcements, since this is a breaking change (and a bad release), you should draft a blog post to go out at the same time as the new release that explains what went wrong, which version to use, and provides an explanation of the install multiple plugins feature. I recommend saying that's its a great way to quickly set up your Porter environment with your required plugins so that people get an idea of when you would use this feature. Probably don't mention the Porter Operator change that caused us to want this feature since it's not shipped yet, so let's focus on how people can use it right now.

pkg/plugins/install.go
VinozzZ: When no version is specified by users, the `o.Version` is default to `latest`. The original check would fail bc the `o.Version` is not empty.
The check here is to guard that users didn't try to set a version value through the flag. `o.Version` should not have value other than "latest" or empty

carolynvs: Thanks! I didn't realize that we were defaulting to "latest" so early, I'm assuming it's the flag default. Makes sense! 👍

pkg/schema/plugins.schema.json
```diff
@@ -11,41 +11,46 @@
...
-    "properties": {
-      "key": {
-        "type": "string"
-      },
-      "value": {
+    },
+    "plugins": {
+      "description": "The definition of plugins to install",
+      "type": "object",
+      "default": "1.0.0",
```
carolynvs, line 18: This default shouldn't be defined here. I assume you meant to define it directly on schemaVersion?

pkg/schema/plugins.schema.json
```diff
@@ -11,41 +11,46 @@
...
+      "additionalProperties": {
         "type": "object",
         "properties": {
-          "version": {
-            "description": "The version for the plugins.",
-            "type": "string"
-          },
-          "feedURL": {
-            "description": "The URL of an atom feed where the plugin can be downloaded.",
+          "key": {
```
carolynvs, line 22: I don't think this is how the properties json schema field works for objects. Have you tested this against the json schema validator (https://www.jsonschemavalidator.net/) with both matching and incorrect schema to make sure it's correct?

Here is how I have defined an object with unknown keys in porter elsewhere (for custom action metadata):

https://github.com/getporter/porter/blob/2aa80a6670bd0aac7284cb8f5ad375b6b44b3365/pkg/schema/manifest.schema.json#L417C6-L421

pkg/schema/plugins.schema.json
```diff
@@ -11,41 +11,46 @@
...
-          "mirror": {
-            "description": "Mirror of official Porter assets.",
-            "type": "string"
-          },
-          "additionalProperties": false
+          "value": {
+            "type": "object",
+            "properties": {
+              "version": {
+                "description": "The version for the plugin.",
```
carolynvs, line 29: I like to give the user hints that will show up in the hover text / documentation that will help them understand how to fill out the file better:

```suggestion
                "description": "The version for the plugin. Defaults to latest when unspecified",
```

pkg/schema/plugins.schema.json
```diff
@@ -11,41 +11,46 @@
...
+                "type": "string"
+              },
+              "feedURL": {
+                "description": "The URL of an atom feed where the plugin can be downloaded.",
```
carolynvs, line 33:
```suggestion
                "description": "The URL of an atom feed where the plugin can be downloaded. Defaults to the official Porter plugin feed.",
```

pkg/schema/plugins.schema.json
```diff
@@ -11,41 +11,46 @@
...
+                "type": "string"
+              },
+              "url": {
+                "description": "The URL from where the plugin can be downloaded",
```
carolynvs, line 37: This field is hard to guess, so I always like to include an example of how to download from github releases:

```suggestion
                "description": "The URL from where the plugin can be downloaded. For example, https://github.com/MChorfa/porter-helm3/releases/download",
```

pkg/schema/plugins.schema.json
```diff
@@ -11,41 +11,46 @@
...
-  "additionalProperties": {
-    "type": "object",
-    "properties": {
-      "key": {
-        "type": "string"
-      },
-      "value": {
+    },
+    "plugins": {
+      "description": "The definition of plugins to install",
```
carolynvs, line 16: It would help to give people a hint that the field names under plugins should be the name of the plugin

```suggestion
      "description": "A map of plugins to install, keyed by the plugin name",
```

carolynvs (changes requested):

carolynvs (approved):
Looks great!

This reminds me that we need to update the porter vs code extension to check for schemaType and apply the proper json schema to the current document, e.g. if the open document is for porter and has "schemaType: Plugins", then vs code should apply this plugins json schema so that the user gets autocomplete.

I'll create another issue to work out the details but wanted to let you know how this schema will be used elsewhere too. 👍

carolynvs:
Here's the issue to track showing autocomplete using our other json schema documents: https://github.com/getporter/vscode-extension/issues/47

PR #2533, opened 2023-01-25
Use SetupDCO magefile target in magefiles repository

carolynvs:
I have moved the SetupDCO script into our magefiles repository so that we can reuse it more easily.


VinozzZ (approved):
🎉🎉🎉

PR #2534, opened 2023-01-25
Ensure lint warnings are printed

### LINKED ISSUE #2525: Lint warnings are not printed
When I run `porter lint` and there is a lint warning (not an error), the warning isn't printed to the console. Both errors and warnings should be printed.

This is hard to reproduce because I noticed this while testing a pull request that was generating a warning. Once #2434 is merged, you can reproduce this by editing [pkg/linter/linter.go](https://github.com/getporter/porter/blob/6df27907b5a40482ceb8c036fede2fa56a518c16/pkg/linter/linter.go#L155) and change the the level of porter-100 from an error to a warning. Then rebuild and run porter against a bundle that has a parameter named "porter_test". You should see a warning printed out for porter-100 (but currently it doesn't).

3. porter create
4. Edit porter.yaml and define a parameter named "porter_test"
5. Run porter lint

The problem is in https://github.com/getporter/porter/blob/6df27907b5a40482ceb8c036fede2fa56a518c16/pkg/porter/lint.go#L74-L77

We are checking if there are any errors and then exiting the function without printing any results. We should move that check to _after_ printing out the results.

This is hard to unit test because we don't have anything that prints a warning at this time, so you can skip adding a test. Make sure to test manually.

See our [Contributing Tutorial](https://getporter.org/contribute/tutorial) and [New Contributor Guide](https://github.com/getporter/porter/blob/release/v1/CONTRIBUTING.md) for help getting started contributing to Porter.

jmhbnz:
# What does this change

Historically when Porter would print lint test results output would only display if errors were present, meaning if only warnings are present these were being suppressed.

This pull request changes the order of operations in `PrintLintResults()` to ensure output is printed before checking for errors and printing the bundle validation message.

# What issue does it fix
Closes #2525 

# Notes for the reviewer

No additional tests written as mentioned in original issue as there is nothing that intentionally prints warnings at this time.

Given the above, Is there any guidance available on a sequence of manual actions I should follow for manually testing this?

# Checklist
- [x] Did you write tests?

# Reviewer Checklist


jmhbnz:
Apologies - I missed DCO sign-off initially and while rebasing to fix that accidentally added my sign-off to a couple of other commits.

It's all fixed now however there is a bit of mess in the PR history while I sorted it out, I can recreate the PR if that would be preferred just let me know.

carolynvs:
It's okay, I can squash your commits into a single one when the PR is merged. So as long as https://github.com/getporter/porter/pull/2534/files looks like what you intended you don't need to mess with git any further. 😁

carolynvs:
Here's how you can test your change with a unit test:

1. Copy the test `TestPorter_PrintLintResults` in [pkg/porter/lint_test.go](https://github.com/getporter/porter/blob/8f590476a38d3904f1c0e0b98887f7360f7991f4/pkg/porter/lint_test.go#L70) and make a new test from it like `TestPorter_PrintLintResults_Warning`.
2. Change the setup for the test so that you are returning a **warning** instead of an error when faking out the lint results. https://github.com/getporter/porter/blob/8f590476a38d3904f1c0e0b98887f7360f7991f4/pkg/porter/lint_test.go#L73
3. Make new testdata files in https://github.com/getporter/porter/tree/main/pkg/porter/testdata/lint to represent the expected test output with your change. The test runs through 3 times to test out each representation of the data. Let's make sure that people see the warnings in each of the output files.

pkg/porter/lint.go
```diff
@@ -82,7 +82,7 @@ func (p *Porter) PrintLintResults(ctx context.Context, opts LintOptions) error {
 		}
 	}
 
-	if !results.HasError() {
+	if !results.HasError() && opts.Format == printer.FormatPlaintext {
```
carolynvs, line 85: Nice catch! ✨

CONTRIBUTORS.md
```diff
@@ -77,3 +77,4 @@ and we will add you. **All** contributors belong here. 💯
 * [Madhu M Pandurangi](https://github.com/MadhuMPandurangi)
 * [Karanjot Singh](https://github.com/0xquark)
 * [Omar Kohl](https://github.com/omarkohl)
+* [James Blair](https://github.com/jmhbnz)
```
carolynvs, line 80: 🎉

carolynvs (approved):
This is great, thank you! 🚀

PR #2543, opened 2023-01-31
Workaround apple silicon codesigning cache bug

carolynvs:
This is a hack to work around a codesigning problem on Apple Silicon where overwriting a binary that has already been executed doesn't cause the corresponding codesign entry in the OS cache to update. Mac then prevents the updated binary from running because the signature doesn't match. Removing the file first clears the cache so that we don't run into "zsh: killed ..."

I've updated the magefile so that when we run the install target, the original binaries are removed before we copy. I considered fixing this directly in shx.Copy but lots of people use that and this is a bug in just regular old cp too. So I don't want shx.Copy to do stuff that even cp doesn't.

See:
* https://stackoverflow.com/questions/67378106/mac-m1-cping-binary-over-another-results-in-crash
* https://openradar.appspot.com/FB8914231

### DIFF
--- a/magefile.go
+++ b/magefile.go
@@ -519,6 +519,16 @@ func Install() {
 
 	// Copy porter binaries
 	mgx.Must(os.MkdirAll(porterHome, pkg.FileModeDirectory))
+
+	// HACK: Works around a codesigning problem on Apple Silicon where overwriting a binary that has already been executed doesn't cause the corresponding codesign entry in the OS cache to update
+	// Mac then prevents the updated binary from running because the signature doesn't match
+	// Removing the file first clears the cache so that we don't run into "zsh: killed porter..."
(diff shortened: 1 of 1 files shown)


magefile.go
```diff
@@ -519,6 +519,16 @@ func Install() {
 
 	// Copy porter binaries
 	mgx.Must(os.MkdirAll(porterHome, pkg.FileModeDirectory))
+
```
VinozzZ, line 522: nit: maybe we can create a function called clearCache or something to capture this hack in once place?

carolynvs: I'll open something in magefiles that we could do in a later follow-up if it comes up again.

https://github.com/getporter/magefiles/issues/22

VinozzZ (approved):

PR #2544, opened 2023-02-01
Update codeowners with active reviewers 

carolynvs:
I've updated our codeowneres file (which is responsible for automatically asking people for reviews when a PR is submitted) with our most active reviewers.

* Added Steven and Brian
* Removed Vaughn


vdice (approved):
❤️

sgettys (approved):

Issue #2566, opened 2023-02-11
add favicons for all devices


Arhell:
https://realfavicongenerator.net/favicon_checker?protocol=https&site=porter.sh%2F#.Y-dfyHZBy3A

PR #2567, opened 2023-02-11
add favicons for all devices

Arhell:
Closes https://github.com/getporter/porter/issues/2566


carolynvs:
Is the goal to have every icon "green" on that checker? I just ran the preview site through it and it's still unhappy about a couple.

https://realfavicongenerator.net/favicon_checker?protocol=https&site=deploy-preview-2567--porter.netlify.app/

Arhell:
> 

Yes, that's right, I would add more icons to make the result better, but alas, there are no better quality icons.

Arhell:
I took this icon, if there are the same in svg it would be cool
https://github.com/getporter/porter/blob/main/docs/static/images/favicon.png

carolynvs:
We do have more high resolution images, some of which are checked into that same directory, such as [porter-icon-badge.png](https://github.com/getporter/porter/blob/main/docs/static/images/porter-icon-badge.png). Does that one work?

Arhell:
works, but I wanted to use the icon that was originally intended for this, can I replace it?

carolynvs:
Sorry I'm unclear on which image you want to replace?

Just for context: The image favicon.png is kind of low-res because we were having trouble with our preferred icon (the porter-icon-badge.png) looked bad when scaled down to 16px. So if you need to replace favicon, sure that's not a problem though let's make sure the image we use looks at both scaled up on something like a phone and tiny.

Arhell:
I checked, it seems to be ok

carolynvs (approved):
Looks great, thanks!

Issue #2595, opened 2023-02-21
update current year


Arhell:
![Screenshot_1](https://user-images.githubusercontent.com/26163841/220436673-2e625200-99c9-4405-ab85-02bf448ce928.png)

PR #2596, opened 2023-02-21
update current year

Arhell:
# What does this change
_Give a summary of the change, and how it affects end-users. It's okay to copy/paste your commit messages._

_For example if it introduces a new command or modifies a commands output, give an example of you running the command and showing real output here._

# What issue does it fix
Closes https://github.com/getporter/porter/issues/2595


carolynvs (approved):
👍

PR #2610, opened 2023-03-02
Support downloading any release artifact through Netlify

carolynvs:
Update our release binary matching redirect rule on Netlify from hard-coded binary names to attempting to download whatever file name was specified. This will let people download the arm64 binaries that we are now publishing, and also our checksum files.


carolynvs:
You can test this out by downloading an arm64 binary through the netlify preview URL

https://deploy-preview-2610--porter.netlify.app/canary/porter-linux-arm64

which currently doesn't work yet (until this is merged) on our main URL

https://cdn.porter.sh/canary/porter-linux-arm64

PR #2611, opened 2023-03-02
Include arm64 binaries when we publish a mixin

carolynvs:
Include entries for arm64 binaries in the atom feed generated by porter mixins feed generate

### DIFF
--- a/pkg/pkgmgmt/feed/generate.go
+++ b/pkg/pkgmgmt/feed/generate.go
@@ -65,7 +65,7 @@ func (feed *MixinFeed) Generate(ctx context.Context, opts GenerateOptions) error
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


sgettys:
LGTM, just wondering on the flip side will the mixin install fetch the appropriate arch?

sgettys (approved):

carolynvs:
When you install a mixin or plugin, we try to download based on GOOS and GOARCH. If that combination doesn't exist, for macs we fallback to amd64.

https://github.com/getporter/porter/blob/7da6fa735c388adc94efbecbdec0bee79da66724/pkg/pkgmgmt/client/install.go#L103-L108

So I expect that once the mixins/plugins start publishing arm64 to their atom.xml feeds, that porter (even older clients) will use the arm64 binary when available.

Issue #2607, opened 2023-03-01
Accessibility add alt to home page images


Arhell:
/assign

carolynvs:
I've updated your permissions to "triage" so that you can self-assign issues like this one.

PR #2614, opened 2023-03-04
Accessibility add alt to home page images

Arhell:
# What issue does it fix
Closes https://github.com/getporter/porter/issues/2607

### DIFF
--- a/docs/themes/porter/layouts/index.html
+++ b/docs/themes/porter/layouts/index.html
@@ -198,35 +198,35 @@
 
             <div class="row mx-0" style="padding-bottom: 100px;">
               <div class="small-12 columns">
-                <h3><img src="/images/lock.png"/> Credentials and Secrets </h3>
+                <h3><img src="/images/lock.png" alt=""/> Credentials and Secrets </h3>
                 <figure>
                   <figcaption>
                     <p>
                       Credentials and secrets are retrieved from industry standard secret stores and securely injected.
                     </p>
                   </figcaption>
-                  <img src="/images/inject-credentials.png" width=900px />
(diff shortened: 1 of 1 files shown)


Arhell (commented):
@carolynvs take a look please

carolynvs (changes requested):
The purpose of the alt text is to provide sufficient detail about the image for low-vision and blind users so that they understand whatever relevant information the image was providing to sighted users. The alt text in this PR doesn't do that, and instead is acting more like a header text for the image.

If you would like to try again, here's a page that gives examples of how to to write alt text:

https://accessibility.huit.harvard.edu/describe-content-images

Arhell:
I've updated, if you need to add or remove anything let me know

carolynvs:
@Arhell Here's an example of what good alt text would look like for one of the images on Porter's homepage



```
The porter.yaml file, a parameter named aks-cluster-name is circled and an arrow is pointing to where the parameter is used in a template further below, where in the installation action of the bundle, the parameter is passed to the azure ARM mixin.
```

For the icons, a better alt text for "lock image" would be "padlock" as it's more specific to what is displayed (there are lots of types of locks but padlock helps describe what is shown), and the word "image" is removed since alt text only describes images.

Does that help clear up how to add alt text if you are still interested?

carolynvs:
@Arhell Setting the alt text to empty doesn't help someone using a screen reader, so I'm really confused about what you are hoping to accomplish with this PR. I'm going to close this for now since I don't think this is very productive and would best be worked out in an issue first to get agreement on the motivation for the change and how it should look.

Issue #2621, opened 2023-03-09
Accessibility update home page "Get Started" btn


Arhell:
Interactive controls must not be nested

PR #2622, opened 2023-03-11
Accessibility update home page "Get Started" btn

Arhell:
# What issue does it fix
Closes https://github.com/getporter/porter/issues/2621


carolynvs:
@Arhell It's really helpful to explain the problem that is being fixed in the original issue, and then describe in the PR how it solves it. Otherwise it's very difficult for someone to properly review it, since I'm left to guess at both.

For example, right now I don't visually see any problem with the get started button on the live website so I can't tell why a change was required and what this does to improve the site (since it looks the same before/after). Can you help fill in the missing context for me?

nit: I did notice that on small screens, the padding on the get started button isn't enough. The padding on the current button looks better when making the screen small.

Arhell:
@carolynvs 
Tnx
visually, I made it as similar as possible to the original, here the goal is to get rid of nesting, as indicated here https://github.com/getporter/porter/issues/2621

This is how I fixed the issue

Arhell:
Focusable elements with an interactive control ancestor (any element that accepts user input such as button or anchor elements) are not announced by screen readers and create an empty tab stop. That is, you could tab to the element but the screen reader will not announce its name, role, or state.

carolynvs (approved):
Thanks for the explanation, makes sense. 👍

PR #2631, opened 2023-03-16
Update Microsoft usage in adopters.md

carolynvs:
Include notes about how Microsoft is using bundles to distribute bundles on Azure Marketplace, and create environments for Azure TRE.


squillace:
+1

PR #2641, opened 2023-03-21
Improve passing custom values as build arguments

### LINKED ISSUE #2633: Large custom values in porter.yaml break build
## Describe the bug
Porter build fails when the porter.yaml file has very large values stored in the custom field.

## To Reproduce
1. Create a bundle with porter create
2. Set a custom field in porter.yaml and make it giant. For example the base64 encoding of a binary (which is how we found it), I used one with 168,567 characters long on a single line.
3. Run porter build

## Expected behavior
The build should succeed

## Porter Command and Output
```
$ porter build
Copying porter runtime ===>
Copying mixins ===>
Copying mixin exec ===>
Building invocation image
[+] Building 1.3s (5/5) FINISHED
 => [internal] load build definition from Dockerfile                                                                                                          0.0s
 => => transferring dockerfile: 832B                                                                                                                          0.0s
 => [internal] load .dockerignore                                                                                                                             0.0s
 => => transferring context: 35B                                                                                                                              0.0s
 => resolve image config for docker.io/docker/dockerfile-upstream:1.4.0                                                                                       1.2s
 => [auth] docker/dockerfile-upstream:pull token for registry-1.docker.io                                                                                     0.0s
 => CACHED docker-image://docker.io/docker/dockerfile-upstream:1.4.0@sha256:178c4e4a93795b9365dbf6cf10da8fcf517fcb4a17f1943a775c0f548e9fc2ff                  0.0s
error building docker image: failed to solve: rpc error: code = Unknown desc = failed to solve with frontend dockerfile.v0: failed to solve with frontend gateway.v0: exit code: 1
unable to build CNAB invocation image: error building docker image: failed to solve: rpc error: code = Unknown desc = failed to solve with frontend dockerfile.v0: failed to solve with frontend gateway.v0: exit code: 1
```

## Version
porter v1.0.9

## Solution Notes
This is caused by a feature in porter where we pass all custom values as build arguments to the docker build when building the invocation image. The custom value is too large to pass to the build, but no meaningful error message is printed.

We can do a few things:
* Only pass custom values that are used by the Dockerfile
* Truncate value and warn about it

We can't add a size limit to the custom field (i.e. forcing them to put the data elsewhere) because we have existing production users relying on using custom today. But they are not passing the data to the build, so either solution above would suffice.

carolynvs:
# What does this change
Currently porter passes all custom values defined in porter.yaml as build arguments when using docker to build the invocation image.

```yaml
custom:
  app:
    version: 1.2.3
```

The above porter.yaml results in porter passing --build-arg CUSTOM_APP_VERSION=1.2.3

We have some users who are storing longer values in the custom map and have managed to hit the max argument length on their system. To address this, I am no longer passing all custom values as build arguments, and only pass them when the Dockerfile declares a build argument with the same name. In the example above, we only pass the app version when the Dockerfile declares `ARG CUSTOM_APP_VERSION`. In this users's case that's enough to avoid the problem.

In addition to the change above, I have also added a max length check on build arguments so that we can detect that it may be too long and provide a better error message. When we let the large value get passed to docker the error is simply "exit code 1" and we don't get any context into what failed. This will ensure that long values are caught before calling docker and telling the user to save the value in a file, and then read it back out in the Dockerfile instead.

# What issue does it fix
Closes #2633 

# Notes for the reviewer
N/A

# Checklist
- [x] Did you write tests?

# Reviewer Checklist


pkg/build/buildkit/buildx.go
```diff
@@ -59,89 +65,154 @@ func (l unstructuredLogger) Write(p []byte) (n int, err error) {
...
 	}
 
 	return nil
 }
 
+func (b *Builder) getDockerfilePath() string {
+	return filepath.Join(b.Getwd(), build.DOCKER_FILE)
+}
+
+func (b *Builder) determineBuildArgs(ctx context.Context, manifest *manifest.Manifest, opts build.BuildImageOptions) (map[string]string, error) {
```
arschles, line 148: nit: maybe putting args on newlines would be cleaner?

```suggestion
func (b *Builder) determineBuildArgs(
    ctx context.Context,
    manifest *manifest.Manifest,
    opts build.BuildImageOptions
) (map[string]string, error) {
```

arschles (commented):
Just some nit-pick comments and suggestions. Overall, LGTM!

carolynvs:
Thanks @arschles! I've updated the code to be a bit more readable, especially when viewing the raw markdown. The \--build-arg is a Hugo thing, it let's me use a double hypen without it being turned into a dash. But I can see how it's super confusing in markdown and not all previewers support the escape code.

pkg/build/buildkit/buildx_test.go
```diff
@@ -144,3 +149,71 @@ func Test_flattenMap(t *testing.T) {
...
+	}
+	// First we use the standard template, that does not use any custom build arguments
+	c.TestContext.AddTestFileFromRoot("pkg/templates/templates/create/template.buildkit.Dockerfile", "/.cnab/Dockerfile")
+
+	// Try manually passing too big of a --build-arg, this should fail
+	opts := build.BuildImageOptions{
+		BuildArgs: []string{"BIG_BUILD_ARG=" + oversizedArg},
+	}
+	_, err := b.determineBuildArgs(ctx, m, opts)
+	assert.ErrorContains(t, err, "BIG_BUILD_ARG is longer than the max")
```
joshuabezaleel, line 201: is there a specific need here that we used `assert` when the rest is using `require`?

tests/integration/build_test.go
```diff
@@ -1,19 +1,49 @@
...
+	test, err := tester.NewTest(t)
+	defer test.Close()
+	require.NoError(t, err, "test setup failed")
+
+	bunPath := filepath.Join(test.RepoRoot, "tests/testdata/mybuns/*")
+	require.NoError(t, shx.Copy(bunPath, test.TestDir, shx.CopyRecursive))
+	test.Chdir(test.TestDir)
+
+	// Use a unique version for appversion so that docker doesn't cache the result and not print the value used in the Dockerfile
+	appversion := fmt.Sprintf("app.versio=%d", time.Now().Unix())
```
joshuabezaleel, line 32: small typo: i guess it should be `app.version` instead of `app.versio`

carolynvs: Thanks for catching that, it actually pointed out a problem with how I was doing the test, so I've cleaned it up a bit more.

pkg/build/buildkit/buildx_test.go
carolynvs: https://github.com/getporter/porter/pull/2643

carolynvs:
@joshuabezaleel I've fixed the test, can you take another look?

pkg/build/buildkit/buildx_test.go
joshuabezaleel: Got it, make sense, thank you lots for the explanation, Carolyn!

joshuabezaleel:
@carolynvs Thank you for always taking the extra miles in making Porter an accessible and welcoming project to contributors, Carolyn, in which you didn't just [kindly explain in length](https://github.com/getporter/porter/pull/2641#discussion_r1145003973) about the reasoning but also sent a PR to [update the docs](https://github.com/getporter/porter/pull/2643) for others to read, really appreciate that from the very beginning. 

The changes looks awesome to me, all good 👍

Issue #2640, opened 2023-03-21
Accessibility fix list


Arhell:
In a properly structured list, all content is contained within list items. Content includes text and other HTML elements. Certain non-content elements are also allowed.
![Screenshot_1](https://user-images.githubusercontent.com/26163841/226598975-6ef97382-ef4f-4d1a-a14d-4de11175dce3.png)

PR #2642, opened 2023-03-21
Accessibility fix list

Arhell:
# What issue does it fix
More details in issue
Closes https://github.com/getporter/porter/issues/2640


Arhell:
corrected the list + corrected the styles so that everything was the same as it is now

carolynvs (approved):
LGTM

Issue #2656, opened 2023-03-26
Accessibility add label to link


Arhell:
An accessible name is a word or phrase coded in a way that assistive technologies can associate it with a specific user interface object. Assistive technologies can then refer to the object by name, not just by type. When a link doesn’t have an accessible name, people who use assistive technologies have no way of knowing its purpose.
![Screenshot_1](https://user-images.githubusercontent.com/26163841/227770768-2e32107d-d769-49e4-89a9-c3f7449282b2.png)

PR #2657, opened 2023-03-26
Accessibility add label to link

Arhell:
# What issue does it fix
Closes https://github.com/getporter/porter/issues/2656


Arhell:
> I am not confident that this is the most accessible solution for the heading anchor links. Here is an article I found that compares various ways to make the link accessible, and also has references to other discussions on what works well for people with screen readers, and the disadvantages of each solution.
> 
> Before we merge anything, we should validate that we understand what a screen reader will say by testing it out.

I agree, but alas, I have no way to check, I took an example from here https://getbootstrap.com/docs/5.3/getting-started/introduction/#quick-start they use `aria-label` too

PR #2697, opened 2023-04-02
Inject claim and not run

### LINKED ISSUE #2696: Porter is injecting run into /cnab/claim.json
## Describe the bug
Porter is injecting its Run object instead of the CNAB claim representation of the Run (runs have a special function Run.ToCNAB that we should be using instead). The CNAB Claims spec (not core this is additional), says that we should inject a claim to /cnab/claim.json. Porter doesn't use this file at all, and neither does any other implementation that I know of but a bundle _could_ decide to use that file.

So we should fix this to provide the CNAB representation of a run (claim).

## To Reproduce
1. Create a bundle
2. Edit the bundle to cat out /cnab/claims.json
3. Install the bundle and note that it's outputting schemaVersion 1.0.2 (which is Run's schema version) instead of the CNAB claim schema version. Another telltale sign is that the file contains a namespace field, but CNAB doesn't support namespace, the CNAB installation field is NAMESPACE/NAME but we don't see that.

## Expected behavior
The outputted /cnab/claim.json should be a claim and not a run.

## Version
1.0.11

carolynvs:
# What does this change
We were accidentally injecting Porter's run represetation, which is _very_ similar to the CNAB claim but isn't exactly the same. I've updated the code to add a test to ensure that we don't break injecting the claim.

For context, Porter doesn't use the injected claim but it is an optional part of the CNAB Claims spec, so while unused, if we are going to say we support it we should do it to spec.

# What issue does it fix
Closes #2696 

# Notes for the reviewer
We need to communicate this in the release notes in case any bundle authors were relying on the old bugged behavior.

# Checklist
- [x] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/cnab/provider/action.go
+++ b/pkg/cnab/provider/action.go
@@ -65,16 +65,21 @@ func (r *Runtime) SetOutput() cnabaction.OperationConfigFunc {
 
 func (r *Runtime) AddFiles(ctx context.Context, args ActionArguments) cnabaction.OperationConfigFunc {
 	return func(op *driver.Operation) error {
+		if op.Files == nil {
+			op.Files = make(map[string]string, 1)
+		}
+
 		for k, v := range args.Files {
 			op.Files[k] = v
 		}
 
 		// Add claim.json to file list as well, if exists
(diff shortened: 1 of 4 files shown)


sgettys (approved):
LGTM!

PR #2709, opened 2023-04-06
remove resources 

Arhell:
# What does this change
I want to test deleting resources


carolynvs:
I think as long as we have PostCSS installed correctly on Netlify, we can add the resources directory to .gitignore and remove the resources from the repo.

carolynvs (approved):
LGTM, thanks!

PR #2716, opened 2023-04-10
Consolidate test data

carolynvs:
# What does this change
We have a LOT of test bundles that are all slightly different which makes for a lot of test maintenance. I've updated the test bundles so that they are reused more across our unit tests, and removed unnecessary test bundles. Whenever possible let's use the test bundles in tests/testdata so that we aren't constantly testing against minor variations of the hello world bundle everywhere.

* tests/testdata/mybuns - Has as many test cases as we can reasonably stuff into a single bundle. It doesn't make sense but helps us validate and regression test a bunch of stuff at once. This is the default bundle that we should try to use instead of making variations of hello world.
* tests/testdata/mydb - a dependency of mybuns, regression test for v1 of dependencies but shouldn't be used otherwise.
* tests/testdata/myenv - A meta bundle for testing dependendencies v2. It relies on myapp, myinfra and mysqldb bundles which are all part of essentially the same test case, a complex bundle graph. I'd avoid using this for testing anything but dependencies v2 since running it will involve running up to 4 bundles, which will be slower than mybuns.

# What issue does it fix
N/A

# Notes for the reviewer
None

# Checklist
- [x] Did you write tests?

# Reviewer Checklist


sgettys (approved):
Nice cleanup! This makes it much cleaner

carolynvs:
We definitely have some more candidates to remove, let's keep an eye out when working with tests that hit a variation of the "porter-hello" bundle and update it to use a bundle from tests/testdata. 👍

PR #2719, opened 2023-04-11
Fix inconsistent rebuild

### LINKED ISSUE #500: Porter should print a clear message that the bundle has changed and is being rebuilt first
Hi team,

while testing an image with "Docker in Docker", I faced a very strange bug with the latest release:
when running `porter credentials generate`, Porter actually tries to install the bundle:
image

In the screenshot, you can see both attempts from Windows Powershell and WSL (ubuntu).

My "credentials" section looks like this:

```yaml
credentials:
  - name: k3dip
    env: HOST_IP
```
Thanks a lot for your feedback :)

### LINKED ISSUE #2503: Inconsistent (redundant?) bundle builds
## Describe the bug

It looks like porter decides to rebuild a bundle for no consistent reason.
I see this when calling publish right after a build which is rebuilding everything. I also see it with explain...

## To Reproduce
1. Run the following script: `porter build && for i in {1..9}; do porter explain -oyaml > $i.yaml; done`
2. Review the files created and notice that some include the "build prints" which shouldn't be there if every explain command used an existing built bundle.

## Expected behavior
1. All explain in the script above should use the initial built bundle
2. When calling publish right after a build, the bundle isn't rebuilt

## Version

porter v1.0.4 (e3a2a5ef)

### LINKED ISSUE #2718: Not passing --insecure-registry from publish to build
On publish, we check that the bundle has already been built and is up-to-date. If it is not, then we rebuild before continuing with the publish.

Previously, the build implementation always allowed insecure registries but that changed in #2579 when I updated the implementation to use HEAD instead of pulling the entire image. At that time I added --insecure-registry to the build command, but missed hooking up that flag to the same flag on the publish command. So when a rebuild is triggered on publish, if the registry referenced by any images in the bundle is insecure, the rebuild will fail.

This is currently happening intermittently on the airgap smoke tests. Which brings me to another bug! It was originally reported in #2503, where porter would for some reason decide that a bundle is out-of-date for no reason. When the bug above in rebuild cropped up, it consistently reproduced the other bug. So the airgap test would randomly decide the bundle should be rebuilt, but because it's a test specifically for unsecured registries, it would also trigger the bug above.

So both really need to be fixed at the same time.

The root cause of the out-of-date detection incorrectly flagging a bundle was because we compare the list of installed mixins and their version between the cached bundle and the current bundle, and were using a Go map for that instead of a sorted array.

carolynvs:
# What does this change
This patch corrects two interrelated issues that were causing our bundle out-of-date check to not work well:

1. We weren't passing the --insecure-registry flag from publish to build when a rebuild was triggered.
2. We were incorrectly comparing the mixins from the cached bundle and the current bundle, with Go maps instead of a sorted array, so rebuilds would be triggered when nothing had changed.

I have updated the calls to EnsureBundleIsUpToDate so that it correctly passes through the --insecure-registry flag. I have also fixed the mixin comparison to use a sorted array so that it consistently generates the same manifest digest when detecting changes to the bundle.

I have also added a bunch of debug messages to the comparison of the cached and current bundle so that when it's flagged as out-of-date, it will print out why and include trace data to help figure out what's wrong/different.

# What issue does it fix
Closes #2718
Closes #2503
Closes #500

# Notes for the reviewer
N/A

# Checklist
- [x] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/cnab/config-adapter/stamp.go
+++ b/pkg/cnab/config-adapter/stamp.go
@@ -16,6 +16,7 @@ import (
 	"get.porter.sh/porter/pkg/manifest"
 	"get.porter.sh/porter/pkg/portercontext"
 	"get.porter.sh/porter/pkg/tracing"
+	"github.com/Masterminds/semver/v3"
 )
 
 // Stamp contains Porter specific metadata about a bundle that we can place
@@ -86,7 +87,24 @@ func (m MixinRecords) Len() int {
 }
 
 func (m MixinRecords) Less(i, j int) bool {
-	return m[i].Name < m[j].Name
+	// Currently there can only be a single version of a mixin used in a bundle
+	// I'm considering version as well for sorting in case that changes in the future once mixins are bundles
+	// referenced by a bundle, and not embedded binaries
+	iRecord := m[i]
+	jRecord := m[j]
+	if iRecord.Name == jRecord.Name {
+		// Try to sort by the mixin's semantic version
+		// If it doesn't parse, just fall through and sort as a string instead
+		iVersion, iErr := semver.NewVersion(iRecord.Version)
+		jVersion, jErr := semver.NewVersion(jRecord.Version)
+		if iErr == nil && jErr == nil {
+			return iVersion.LessThan(jVersion)
+		} else {
+			return iRecord.Version < jRecord.Version
+		}
+	}
+
+	return iRecord.Name < jRecord.Name
 }
 
 func (m MixinRecords) Swap(i, j int) {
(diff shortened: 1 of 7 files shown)


pkg/cnab/config-adapter/stamp.go
```diff
@@ -69,10 +70,31 @@ func (s Stamp) WriteManifest(cxt *portercontext.Context, path string) error {
...
 }
 
+type MixinRecords []MixinRecord
+
+func (m MixinRecords) Len() int {
+	return len(m)
+}
+
+func (m MixinRecords) Less(i, j int) bool {
+	return m[i].Name < m[j].Name
```
sgettys, line 89: Should this be a combination of name and version?

pkg/cnab/config-adapter/stamp.go
```diff
@@ -152,17 +173,21 @@ func LoadStamp(bun cnab.ExtendedBundle) (Stamp, error) {
 	return stamp, nil
 }
 
-// getUsedMixinsVersion compare the mixins defined in the manifest and the ones installed and then retrieve the mixin's version info
-func (c *ManifestConverter) getUsedMixinsVersion() map[string]string {
-	usedMixinsVersion := make(map[string]string)
+// getUsedMixinRecords compare the mixins defined in the manifest and the ones installed and then retrieve the mixin's version info
```
sgettys, line 176: nit: Update comment now that we're returning the record instead of just version info

carolynvs:
@sgettys I've updated the comparison to take into account name and version to future proof this for when we have mixins as bundles. Please take another look.

Issue #2698, opened 2023-04-03
Resolve run parameters just-in-time


carolynvs:

carolynvs:
As part of this I am also going to calculate a hash of the resolved parameters and credentials. This will help us quickly determine if they have changed (when reconciling an installation) without requiring us to re-resolve the values from an old run, as long as the digest is set.

PR #2722, opened 2023-04-12
Execute bundle from a pre-computed Run

### LINKED ISSUE #2732: Execute a bundle from a pre-computed Run
Currently we pass a lot of information to the Porter bundle execution code, and then leave it up to our CNAB runtime implementation to piece it together and execute the bundle. It does make a number of decisions about how to execute the bundle at that late stage.

In order to support workflows, which are composed of pre-computed Run resources, we need to update our CNAB runtime implementation to accept a pre-computed Run and have it execute the instructions encoded on the Run as defined.

carolynvs:
# What does this change

Update our CNAB runtime implementation to accept a pre-computed Run, which may have already been persisted in a pending state. We now execute it as defined and have moved more of the logic for how the bundle should be executed out of the runtime. This allows us to generate the runs ahead of time, store them on a workflow, and execute later.

# What issue does it fix
Closes #2732 

# Notes for the reviewer
I have carved out in the runtime where we should do JIT resolution of parameters and credentials. Credentials have always been resolved JIT, but parameters are still being resolved earlier. The rest of the fix will come in a follow-up PR for #2698 

I am moving away from using intermediate representations of resolved credentials and parameters in favor of using the existing field on parameter/credential sets that store the resolved value. In a follow-up PR (this was too big to include), I complete the switch over so that we delay resolution as much as possible and avoid excessive type conversions from string to the typed value (e.g. "3" -> 3).

# Checklist
- [x] Did you write tests?

# Reviewer Checklist


kichristensen:
Implemented in #3051

PR #2734, opened 2023-04-19
Add functionality to SourceMapList

carolynvs:
# What does this change
Add methods to SourceMapList (the lists of parameters or credentials in a set) for working with the collection using item keys and converting to a map of resolved values

# What issue does it fix
Prep work for #2698 

# Notes for the reviewer
N/A

# Checklist
- [x] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/secrets/strategy.go
+++ b/pkg/secrets/strategy.go
@@ -5,27 +5,15 @@ import (
 	"errors"
 	"fmt"
 
+	"get.porter.sh/porter/pkg/cnab"
 	"github.com/cnabio/cnab-go/valuesource"
+	"github.com/hashicorp/go-multierror"
 	"gopkg.in/yaml.v3"
 )
 
 // Set is an actual set of resolved values.
 // This is the output of resolving a parameter or credential set file.
 type Set map[string]string
-
-// Merge merges a second Set into the base.
-//
-// Duplicate names are not allow and will result in an
-// error, this is the case even if the values are identical.
-func (s Set) Merge(s2 Set) error {
-	for k, v := range s2 {
-		if _, ok := s[k]; ok {
-			return fmt.Errorf("ambiguous value resolution: %q is already present in base sets, cannot merge", k)
-		}
-		s[k] = v
-	}
-	return nil
-}
 
 // IsValid determines if the provided key (designating a name of a parameter
 // or credential) is included in the provided set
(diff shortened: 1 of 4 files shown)


pkg/secrets/strategy.go
```diff
@@ -130,18 +118,128 @@ func (s Source) MarshalYAML() (interface{}, error) {
...
 }
+
+// HasName determines if the specified name is defined in the set.
+func (l SourceMapList) HasName(name string) bool {
+	_, ok := l.GetByName(name)
+	return ok
+}
+
+// GetByName returns the resolution strategy for the specified name and a bool indicating if it was found.
+func (l SourceMapList) GetByName(name string) (SourceMap, bool) {
```
arschles, line 143: would it be worth returning an `error` with a message like `fmt.Errorf("SourceMapList has no name %s", name)`?

pkg/secrets/strategy.go
```diff
@@ -130,18 +118,128 @@ func (s Source) MarshalYAML() (interface{}, error) {
...
+			return item, true
+		}
+	}
+
+	return SourceMap{}, false
+}
+
+// GetResolvedValue returns the resolved value of the specified name and a bool indicating if it was found.
+// You must resolve the value before calling, it does to do resolution for you.
+func (l SourceMapList) GetResolvedValue(name string) (interface{}, bool) {
```
arschles, line 155: could you use a type param like `T` here and then return a `T`? I'm not sure how deep you have to go into `SourceMap` to make that work, so maybe it's not worth it at the moment, not sure.

pkg/secrets/strategy.go
```diff
@@ -130,18 +118,128 @@ func (s Source) MarshalYAML() (interface{}, error) {
...
+	for _, item := range l {
+		values[item.Name] = item.ResolvedValue
+	}
+	return values
+}
+
+// GetTypedResolvedValues converts the items to a map of key/value pairs, where the value is the type defined by the bundle.
+// Validation errors are accumulated, and parameters are always passed back along with a validation error.
+// If you care about valid parameters, check the error, but otherwise you can use this for display purposes and not have it blow up with no results.
+func (l SourceMapList) GetTypedResolvedValues(bun cnab.ExtendedBundle) (map[string]interface{}, error) {
```
carolynvs, line 178: I copied this from elsewhere in porter while refactoring and just realized that the functionality was only tested higher up in the code, but never as a unit test for the function itself. Adding one now...

carolynvs: Oops, never mind this needs to be defined directly on ParameterSet, and not on the generic SourceMapList. I'll remove it from this PR.

arschles (commented):
Looks good to me overall. I left 2 non-blocking comments.

pkg/secrets/strategy.go
```diff
@@ -130,18 +116,84 @@ func (s Source) MarshalYAML() (interface{}, error) {
 	return s.MarshalRaw(), nil
 }
 
-type StrategyList []SourceMap
+// SourceMapList is a list of mappings that can be access via index or the item name.
+type SourceMapList []SourceMap
```
sgettys, line 120: Should there be a name safe "Add" method?

sgettys (approved):

pkg/secrets/strategy.go
arschles: yea, I agree it shoudn't be a `(SourceMap, bool, error)` return. I wasn't clear enough before, sorry. I was asking if you think we could do a `(SourceMap, error)` return value, and then if the name wasn't found, you could return a `fmt.Errorf(...)` in that case, rather than `false`?

pkg/secrets/strategy.go
carolynvs: yeah right now there's just one place in the code that uses this function, and casting the interface{} to a concrete type is easier than trying to get generics to work in this case.

But I'll keep it in mind if we start using it more. For the most part in Porter, we ignore the actual values of any parameters and just pass them around as interface{} to the bundle (the exception is the porter specific parameter that's inserted into every bundle built by porter, "porter-debug").

pkg/secrets/strategy.go
carolynvs: Right now it wouldn't be used, I don't have anything that mutates sets like that. Normally these are populated by unmarshaling a parameter/credential set.

I can add one though in case it comes in handy later, maybe for our tests.

carolynvs: Actually I'm not sure what the universal behavior of that should even be? Refusing to append if it's already in the list? Overwrite it? I'd rather leave that up to the caller, especially since nothing in code would use it now. We can add something later if there's a common pattern that we want people to use.

pkg/secrets/strategy.go
```diff
@@ -130,18 +116,84 @@ func (s Source) MarshalYAML() (interface{}, error) {
...
+	item, ok := l.GetByName(name)
+	if ok {
+		return item.ResolvedValue, true
+	}
+
+	return nil, false
+}
+
+// GetResolvedValues converts the items to a map of key/value pairs, with the values represented as CNAB-compatible strings
+func (l SourceMapList) GetResolvedValues() map[string]string {
```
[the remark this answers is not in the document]
carolynvs: I'm going to avoid `GetValue()` because it's important to call out that it's a resolved value, i.e. Porter called out to a secret store or read an environment variable using the strategy defined for that item, and this is the final value.

But I agree it would be easy to mistype. I'll rename it to `ToResolvedValues` so that it emphasizes that it's returning a new struct (not just looking up data like the other Get functions).

pkg/secrets/strategy.go
arschles: This struct is acting like an ordered set, do I have that right? If so, do the GetByName/HasName functions need to be faster than linear time? If so, maybe it'd be worth abstracting the list away and adding a function similar to:

Add(key, value, overwrite) 

... that always writes the key/value pair if the key didn't already exist, and only overwrites a key/value pair if the key already existed and overwrite was true. 

Then if you later want/need to make those Get/Has functions faster, you can, and callers are still free to decide the behavior they want? Not sure if this is too much overkill.

carolynvs: The GetByName functionality is just for the tests and is only used once in production code (to read the porter-debug parameter).

The data structure is a map (unique names) represented as a list because it works better UX wise that way when the user is defining the data in Porter (and we can't change that at this point due to compatibility). VS code doesn't do a great job of autocomplete and suggesting what to type when using maps, vs arrays. With maps, when I try to autocomplete, since the key name could be anything, there isn't any autocomplete help. With arrays, I can ask for autocomplete and it creates an entry with all required fields for me, and helps people make bundles faster with less trips to the docs. Happy to show you sometime, and see if there's ways to improve.

Once the data is unmarshaled from porter.yaml into our data structure, we mostly work with it as a single set of data and don't retrieve items by name. However, when importing, detecting duplicate names would be nice.

I think a custom yaml marshal/unmarshal that lets SourceMapList BE a map in Go would get rid of the awkwardness. Let me play around with that idea a bit and see how disruptive that would be. Long term sticking with well understood data structures for this would make it easier to reason about so thank you for pointing out that I was making an odd one!

carolynvs: This feels either like a rabbit hole or a really great idea that we can make generic and use for other parts of porter.yaml that are represented as a list but in code as a map.'

I'll give it one more day to see if it's worthwhile or not. 😂

carolynvs:
Closing in favor of #2750

PR #2737, opened 2023-04-20
Fix test to work on dev machines

carolynvs:
# What does this change
I have updated this test to only check for the desired string, and not attempt to match the entire debug output generated by porter, which can contain stray debug lines depending on the dev env. For example, on my machine I have telemetry enabled but when the backing otel collector is down, porter logs debug lines that it can't connect. The test is failing locally for me cause the test is requiring that ONLY that one line is printed, instead of just checking if the line was printed at all (more accurate and less fragile).

Since the test is intended to check that we are ****'ing out sensitive words, it doesn't need to be that strict.

# What issue does it fix
Avoids test flakes when running integration tests on a local dev env.

# Notes for the reviewer
N/A

# Checklist
- [x] Did you write tests?

# Reviewer Checklist


arschles (approved):
Good idea. LGTM

PR #2750, opened 2023-04-26
Add ArrayEncodedMap

carolynvs:
# What does this change
It's a common pattern in Porter to represent a map of items, with a unique key, as a list in the porter.yaml or a file representation of the resource because the UX for autocomplete is better. We can't change that now, parameters and credentials for example will always be a list, but we can use a map in Porter's code so that it's easier to work with.

I've added ArrayEncodedMap, a data structure that reads in a list and stores it in-memory as a map:

- name: a
  value: stuff
- name: b
  value: things

We can use this as a base data structure initially for Parameter and Credential Sets, storing the mapping of the parameter/credential name to the secret resolution strategy. Over time, it may make sense to use it for the other data structures in porter.yaml as well.

It has support for iterating as a map, or as a sorted list of values. The sorted list is good for unit tests, and anytime we need the output to be stable. The map isn't exposed directly and I have a funky "best you can do with Go" iterator to help keep our for loops looking pretty normal.

# What issue does it fix
Refactoring work for PEP003 as I found working with parameter sets as maps made parameter resolution much more straightforward.

# Notes for the reviewer
This is a replacement for my original PR #2734, which was adding helper methods to work with the set of values by key, but was still storing the values in a slice.

I will follow-up with another PR after this is merged to use it in ParameterSet/CredentialSet. You can see what that looks like in #2749. I wanted to give us a chance to bikeshed this first so that it's not super hard to rework with suggestions. Once we hook it up to ps/cs, the changeset gets pretty large.

Alternative names to consider... 😁 

* MessedUpMap
* ArrayMap (Aaron said this means something in other languages, so probably best to avoid)
* A nonsense name like Blinko, so people don't waste time worrying about what the name implies about the implementation
* NotAMap

# Checklist
- [x] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/encoding/array_encoded_map.go
+++ b/pkg/encoding/array_encoded_map.go
@@ -1,0 +1,226 @@
+package encoding
+
+import (
+	"encoding/json"
+	"fmt"
+	"sort"
+
+	"gopkg.in/yaml.v3"
+)
+
+// MapElement is the in-memory representation of the item when stored in a map.
+type MapElement interface {
+	// ToArrayEntry converts to the representation of the item when stored in an
+	// array.
+	ToArrayEntry(key string) ArrayElement
+}
+
+// ArrayElement is the representation of the item when stored in an array, and
+// includes the key under which the element was stored in the original map.
+type ArrayElement interface {
+	// GetKey returns the unique item key.
+	GetKey() string
+
+	// ToMapEntry converts to the representation of the item when stored in a map.
+	ToMapEntry() MapElement
+}
+
+// ArrayEncodedMap is a map that is represented as an array when marshaled to json/yaml.
+// MapElement is the type of the elements when stored in a map and ArrayElement is the type of the elements when stored in an array.
+type ArrayEncodedMap[T MapElement, K ArrayElement] struct {
+	items map[string]T
+}
+
+// NewArrayEncodedMap initializes an empty ArrayEncodedMap.
+func NewArrayEncodedMap[T MapElement, K ArrayElement]() ArrayEncodedMap[T, K] {
+	return MakeArrayEncodedMap[T, K](0)
+}
+
+// MakeArrayEncodedMap allocates memory for the specified number of elements.
+func MakeArrayEncodedMap[T MapElement, K ArrayElement](len int) ArrayEncodedMap[T, K] {
+	return ArrayEncodedMap[T, K]{
+		items: make(map[string]T, len),
+	}
+}
+
+// Len returns the number of items.
+func (m *ArrayEncodedMap[T, K]) Len() int {
+	if m == nil {
+		return 0
+	}
+	return len(m.items)
+}
+
+// Items returns a copy of the items, and is intended for use with the range
+// operator.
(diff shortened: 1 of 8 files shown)


pkg/encoding/array_encoded_map.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+	return len(m.items)
+}
+
+// Items returns a copy of the items, and is intended for use with the range
+// operator.
+// Use ItemsUnsafe() to directly manipulate the backing items map.
+func (m *ArrayEncodedMap[T, K]) Items() map[string]T {
+	if m == nil {
+		return nil
```
arschles, line 59: return an empty `map` here, so callers don't blow up?

pkg/encoding/array_encoded_map.go
```diff
@@ -0,0 +1,226 @@
...
+
+// ItemsUnsafe returns the backing items map. It is intended for optimizing
+// memory usage when iterating over the items to convert them into alternative
+// representations.
+func (m *ArrayEncodedMap[T, K]) ItemsUnsafe() *map[string]T {
+	if m == nil {
+		return nil
+	}
+
+	if m.items == nil {
```
arschles, line 99: is it ever possible to have `items` be `nil`?

pkg/encoding/array_encoded_map.go
```diff
@@ -0,0 +1,226 @@
...
+	sort.SliceStable(raw, func(i, j int) bool {
+		return raw[i].GetKey() < raw[j].GetKey()
+	})
+	return raw
+}
+
+// UnmarshalRaw is the common Marshal implementation between YAML and JSON.
+func (m *ArrayEncodedMap[T, K]) UnmarshalRaw(raw []K) error {
+	if m == nil {
+		*m = ArrayEncodedMap[T, K]{}
```
arschles, line 157: won't this panic? seems like you'd need to set the value of the pointer to valid memory

```suggestion
		m = &ArrayEncodedMap[T, K]{}
```

pkg/encoding/array_encoded_map.go
```diff
@@ -0,0 +1,226 @@
...
+func (m *ArrayEncodedMap[T, K]) MarshalYAML() (interface{}, error) {
+	if m == nil {
+		return nil, nil
+	}
+	return m.MarshalRaw(), nil
+}
+
+// UnmarshalYAML unmarshals the items in the specified YAML.
+func (m *ArrayEncodedMap[T, K]) UnmarshalYAML(value *yaml.Node) error {
+	var raw []K
```
arschles, line 201: same here RE checking `m == nil`

arschles (changes requested):
Looks good overall. I left a few questions

pkg/encoding/array_encoded_map.go
sgettys: Is this even reachable? If the object is nil then referencing any method on the object should result in a panic before it hits the implementation correct?

sgettys: Never mind just clarified my own question. Go can call methods on nil structs :)

PR #2767, opened 2023-05-23
Porter explain command displays custom section data when output is JSON or YAML

### LINKED ISSUE #2764: Porter explain to display custom section when output is JSON
Automating porter installations would greatly benefit of having this feature. An automation system will have additional requirements that the custom section data in the Porter manifest can satisfy. However, at the moment it is very difficult to extract the information.

It is undestandable that the format `porter explain` command produces by default is not suited to manage arbitrary custom section data. However, when using `--output json` flag this is certainly possible since the data ends up in JSON format in the generated `bundle.json` file anyways.

So, given the following custom section in `porter.yaml`:

```yaml
custom:
  dependencies:
    bundles:
      - name: "bundle-one"
        version: 0.1.0
      - name: "bundle-two"
        version: 0.2.0
        outputs:
          - "output1"
          - "output2"
```

> Note: I am aware of the dependency management supported by Porter, but we may not always want the install action to invoke the dependency installation. Just consider any custom data - don't get hung up on the sample.

What I would hope to find in the output of `porter explain --output json` is:

```json
{
  "dependencies": {
    "bundles": [
      {
        "name": "bundle-one",
        "version": "0.1.0"
      },
      {
        "name": "bundle-two",
        "outputs": [
          "output1",
          "output2"
        ],
        "version": "0.2.0"
      }
    ]
  }
}
```

I have managed to create a suboptimal workaround by creating a custom action that will format the custom section into JSON, but the action invocation requires an existing installation of the bundle, which is not acceptable in my case. Extracting the `bundle.json` file from image is another option, but again, having `porter explain` do the work would be best.

tompaana:
# What does this change

Adds the user-defined [custom](https://getporter.org/bundle/manifest/#custom) section data to the output of `porter explain` command when the output format is JSON or YAML:
- Implements the feature
- Updates unit tests
- Updates documentation

# What issue does it fix

Closes #2764 

# Checklist

- [x] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/porter/explain.go
+++ b/pkg/porter/explain.go
@@ -210,7 +210,8 @@ func generatePrintable(bun cnab.ExtendedBundle, action string) (*PrintableBundle
 		Outputs:       make([]PrintableOutput, 0, len(bun.Outputs)),
 		Dependencies:  make([]PrintableDependency, 0, len(deps)),
 		Mixins:        make([]string, 0, len(stamp.Mixins)),
-		Custom:        bun.Custom,
+		Custom:        make(map[string]interface{}),
+		//Custom:        bun.Custom,
 	}
 
 	for a, v := range bun.Actions {
@@ -301,6 +302,21 @@ func generatePrintable(bun cnab.ExtendedBundle, action string) (*PrintableBundle
 		pb.Mixins = append(pb.Mixins, mixin)
 	}
 	sort.Strings(pb.Mixins)
+
+	// copy custom data to output, but ignore the Porter built-in data
+	for key, value := range bun.Custom {
+		pb.Custom[key] = value
+	}
+	keysToRemove := []string{
+		"io.cnab.parameter-sources",
+		"sh.porter",
+		"sh.porter.file-parameters",
+	}
+	for _, key := range keysToRemove {
+		if _, found := bun.Custom[key]; found {
+			delete(pb.Custom, key)
+		}
+	}
 
 	return &pb, nil
 }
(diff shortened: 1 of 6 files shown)


pkg/porter/explain.go
```diff
@@ -300,6 +302,26 @@ func generatePrintable(bun cnab.ExtendedBundle, action string) (*PrintableBundle
...
+		"io.cnab",
+		"sh.porter",
+	}
+
+	for key, value := range bun.Custom {
+		var ignore bool = false
+
+		for _, ignoreKeyPrefix := range ignoreKeyPrefixes {
+			if strings.HasPrefix(key, ignoreKeyPrefix) {
+				ignore = true
```
sgettys, line 316: nit: Add a break here

pkg/porter/explain.go
```diff
@@ -300,6 +302,26 @@ func generatePrintable(bun cnab.ExtendedBundle, action string) (*PrintableBundle
...
+	// copy custom data to output, but ignore the Porter built-in data
+	ignoreKeyPrefixes := []string{
+		"io.cnab",
+		"sh.porter",
+	}
+
+	for key, value := range bun.Custom {
+		var ignore bool = false
+
+		for _, ignoreKeyPrefix := range ignoreKeyPrefixes {
```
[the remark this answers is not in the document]
tompaana: Will do, but can you elaborate a bit; do you mean something like `func shouldAddCustomKeyToOutput(string key) bool`?

pkg/porter/explain.go
tompaana: The newly created function returns immediately if the condition is true so no break required anymore.

sgettys (approved):

Issue #2773, opened 2023-06-07
cli: "porter list: list INSTALLED bundles"


schristoff:
## Describe the bug
porter list shows installed and uninstalled bundles.

## To Reproduce
the 'helper' for the cli command could be updated, or we can add a flag to porter list

## Porter Command and Output
```
Aliased Commands:
  archive       Archive a bundle from a reference
  build         Build a bundle
  copy          Copy a bundle
  create        Create a bundle
  explain       Explain a bundle
  inspect       Inspect a bundle
  install       Create a new installation of a bundle
  invoke        Invoke a custom action on an installation
  lint          Lint a bundle
 * list          List installed bundles*
```

```
porter list
----------------------------------------------------------------------------
  NAMESPACE  NAME          VERSION  STATE        STATUS     MODIFIED
----------------------------------------------------------------------------
             porter-hello  0.2.0    installed    succeeded  14 minutes ago
             hello         0.2.0    uninstalled  succeeded  2023-05-17
                           0.2.0    installed    succeeded  2023-05-17
```

## Potential Fixes
- just change the tag to show "lists installed and uninstalled bundles"
- remove uninstalled bundles *unless* we pass an --all flag, but this could break any code that relies on it showing all uninstalled bundles

harjyotbagga:
I would like to pick this issue up.

ludfjig:
Closing this since #2888 fixes this

PR #2775, opened 2023-06-08
Listing only INSTALLED packages in `porter list`

harjyotbagga:
Listing only INSTALLED packages in `porter list`

# What does this change
_This change adds a new subcommand `--all-states` to the list subcommand in porter. As per this change, porter would only show installed packages by default. Adding this new subcommand would include all states._

image

# What issue does it fix
Closes [#2773 ](https://github.com/getporter/porter/issues/2773)

# Notes for the reviewer
_I have added the documentation for the sub-command. I have also modified some test cases (adding `all-states` parameter to display all installations) to print and pass all Unit Test cases. If we feel that we should add/modify the test cases, kindly do let me know._

# Checklist
- [x] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -82,3 +82,4 @@ and we will add you. **All** contributors belong here. 💯
 * [Sarah Christoff](https://github.com/schristoff)
 * [Aleksey Barabanov](https://github.com/alekseybb197)
 * [Tomi Paananen](https://github.com/tompaana)
+* [Harjyot Singh Bagga](https://github.com/harjyotbagga)
--- a/cmd/porter/installations.go
+++ b/cmd/porter/installations.go
@@ -47,6 +47,7 @@ The results may also be filtered by associated labels and the namespace in which
 Optional output formats include json and yaml.`,
 		Example: `  porter installations list
   porter installations list -o json
+  porter installations list --all-states
   porter installations list --all-namespaces,
   porter installations list --label owner=myname --namespace dev
   porter installations list --name myapp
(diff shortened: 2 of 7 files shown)


bdegeeter:
Hello @harjyotbagga! First I just wanted to say thanks for the contribution.  It's great to see someone new engage in the project.

Here's a couple of thoughts I have about the change.   The `--all-namespaces` flag is taken from `kubectl`.  I wonder if we want to adopt a pattern similar to kubectl for additional output filtering.  For example, `kubectl get pods --all-namespaces --field-selector=status.phase!=Running` the `--field-selector` while more work to implement offers more flexibility.  (maybe we just capture this as a future enhancement?)

A page could also be taken from `docker` and just support `-a` and `--all` to show any uninstalled or future states that are related to an installation that no longer exists.  What else should we consider for filtering while at the same time avoiding a new flag for each case?

schristoff:
> For example, `kubectl get pods --all-namespaces --field-selector=status.phase!=Running` the `--field-selector` while more work to implement offers more flexibility.

Ah, dang, I didn't even think of doing something like this. I think this would be ideal. Currently we have separate flags for label, name, namespace. If we keep going down this way we will have a ton of flags - but if we implement `--field-selector` then we can delete those other flags. It looks like `-label` starts to implement this a bit.

@harjyotbagga  - if you're up for changing implementation, then perhaps taking a look at [kubectl's implementation of fieldselector](https://github.com/kubernetes/kubectl/blob/a5b0052b8993ed982e8df8c97de7abc4263b3967/pkg/cmd/top/top_pod.go#L180) and using it "as inspiration" would be a good start. Let me know what you think.

harjyotbagga:
Sure @schristoff I would be up for the challenge. It would take me some time to understand how kubectl implemented the field selector (and some advisory help too). But would definitely like to give it a try.

schristoff:
Hey @harjyotbagga  - just wanted to checkin if there's any assistance I can provide to get this PR merged - or if it is okay if someone else does that?

Let us know! We're happy to help!

harjyotbagga:
Hey @schristoff I tried working on this feature. But faced some issues. Someone else can work on this feature. 
I can bring in the changes I was thinking of in a new PR.

PR #2778, opened 2023-06-09
adding functionality to porter create in a specified directory

### LINKED ISSUE #2772: porter create command should create directory with bundle files included
When using porter create <mybundle>, the command will create all the files associated with the bundle in the current directory where ran. For a cleaner file system organization, a directory of the bundle name should be made with the files for the bundle created in that directory. 

**Additional Context**
This functionality saves a step for the user. An example of this is how Rust uses the cargo new <program> command. 

**Risks/Concerns**
None that I'm aware of. 

**What other approaches were considered?**
None

**Implementation details**
portal create mybundle
cd mybundle
ls - porter.yaml, README.md template.Dockerfile, etc.

# Checklist
- [ ] An announcement of this proposal has been sent to the Porter mailing list: https://getporter.org/mailing-list

sbdtu5498:
# What does this change
Appropriate feature has been added and locally tested as well.

# What issue does it fix
Closes #2772 

# Checklist
- [x] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/cmd/porter/bundle.go
+++ b/cmd/porter/bundle.go
@@ -31,14 +31,24 @@ func buildBundleCommands(p *porter.Porter) *cobra.Command {
 }
 
 func buildBundleCreateCommand(p *porter.Porter) *cobra.Command {
-	return &cobra.Command{
-		Use:   "create",
+	cmd := &cobra.Command{
+		Use:   "create [bundle-name]",
 		Short: "Create a bundle",
-		Long:  "Create a bundle. This generates a porter bundle in the current directory.",
+		Long:  "Create a bundle. This generates a porter bundle in the directory with the specified name or in the current directory if no name is provided.",
+		Args:  cobra.MaximumNArgs(1), // Expect at most one argument for the bundle name
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return p.Create()
+			// By default we create the bundle in the current directory
+			bundleName := ""
+
+			if len(args) > 0 {
+				bundleName = args[0]
+			}
+
+			return p.Create(bundleName)
 		},
 	}
+
+	return cmd
 }
 
 func buildBundleBuildCommand(p *porter.Porter) *cobra.Command {
--- a/pkg/porter/create.go
+++ b/pkg/porter/create.go
@@ -4,40 +4,80 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"get.porter.sh/porter/pkg"
 	"get.porter.sh/porter/pkg/config"
 )
 
-func (p *Porter) Create() error {
-	fmt.Fprintln(p.Out, "creating porter configuration in the current directory")
+// Create function creates a porter configuration in the specified directory or in the current directory if no directory is specified.
+func (p *Porter) Create(bundleName string) error {
+	// Normalize the bundleName by removing trailing slashes
+	bundleName = strings.TrimSuffix(bundleName, "/")
 
-	err := p.CopyTemplate(p.Templates.GetManifest, config.Name)
+	// Use the current directory if no directory is passed
+	if bundleName == "" {
+		bundleName = p.FileSystem.Getwd()
+	}
+
+	// Check if the directory in which bundle needs to be created already exists.
+	// If not, create the directory.
+	_, err := os.Stat(bundleName)
+	if err != nil {
+		if os.IsNotExist(err) {
+			// This code here attempts to create the directory in which bundle needs to be created,
+			// if the directory does not exist.
+			// bundleName can handle both the relative path and absolute path into consideration,
+			// For example if we want to create a bundle named mybundle in an existing directory /home/user we can call porter create /home/user/mybundle or porter create mybundle in the /home/user directory.
+			// If we are in a directory /home/user and we want to create mybundle in the directory /home/user/directory given the directory exists,
+			// we can call porter create directory/mybundle from the /home/user directory or with any relative paths' combinations that one can come up with.
+			// Only condition to use porter create with absolute and relative paths is that all the directories in the path except the last one should strictly exist.
+			err = os.Mkdir(bundleName, os.ModePerm)
+			// This error message is returned when the os.Mkdir call encounters an error
+			// during the directory creation process. It specifically indicates that the attempt
+			// to create the bundle directory failed. This could occur due to reasons such as
+			// lack of permissions, a file system error, or if the parent directory doesn't exist.
+			if err != nil {
+				return fmt.Errorf("failed to create directory for bundle: %w", err)
+			}
+			// This error message is returned when the os.Stat call encounters an error other than
+			// the directory not existing. It implies that there was an issue with checking the bundle directory,
+			// but it doesn't mean that the directory creation itself failed.
+			// It could be due to various reasons, such as insufficient permissions, an invalid directory path,
+			// or other file system-related errors.
+		} else {
+			return fmt.Errorf("failed to check bundle directory: %w", err)
+		}
+	}
+
+	err = p.CopyTemplate(p.Templates.GetManifest, filepath.Join(bundleName, config.Name))
 	if err != nil {
 		return err
 	}
 
-	err = p.CopyTemplate(p.Templates.GetManifestHelpers, "helpers.sh")
+	err = p.CopyTemplate(p.Templates.GetManifestHelpers, filepath.Join(bundleName, "helpers.sh"))
 	if err != nil {
 		return err
 	}
 
-	err = p.CopyTemplate(p.Templates.GetReadme, "README.md")
+	err = p.CopyTemplate(p.Templates.GetReadme, filepath.Join(bundleName, "README.md"))
 	if err != nil {
 		return err
 	}
 
-	err = p.CopyTemplate(p.Templates.GetDockerfileTemplate, "template.Dockerfile")
+	err = p.CopyTemplate(p.Templates.GetDockerfileTemplate, filepath.Join(bundleName, "template.Dockerfile"))
 	if err != nil {
 		return err
 	}
 
-	err = p.CopyTemplate(p.Templates.GetDockerignore, ".dockerignore")
+	err = p.CopyTemplate(p.Templates.GetDockerignore, filepath.Join(bundleName, ".dockerignore"))
 	if err != nil {
 		return err
 	}
 
-	return p.CopyTemplate(p.Templates.GetGitignore, ".gitignore")
+	fmt.Fprintf(p.Out, "creating porter configuration in %s\n", bundleName)
+
+	return p.CopyTemplate(p.Templates.GetGitignore, filepath.Join(bundleName, ".gitignore"))
 }
 
 func (p *Porter) CopyTemplate(getTemplate func() ([]byte, error), dest string) error {
--- a/pkg/porter/create_test.go
+++ b/pkg/porter/create_test.go
@@ -1,6 +1,8 @@
 package porter
 
 import (
+	"os"
+	"path/filepath"
 	"testing"
 
 	"get.porter.sh/porter/pkg"
@@ -8,18 +10,18 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestCreate(t *testing.T) {
+func TestCreateInRootDirectory(t *testing.T) {
 	p := NewTestPorter(t)
 	defer p.Close()
 
-	err := p.Create()
+	err := p.Create("")
 	require.NoError(t, err)
 
+	// Verify that files are present in the root directory
 	configFileStats, err := p.FileSystem.Stat("porter.yaml")
 	require.NoError(t, err)
 	tests.AssertFilePermissionsEqual(t, "porter.yaml", pkg.FileModeWritable, configFileStats.Mode())
 
-	// Verify that helpers is present and executable
 	helperFileStats, err := p.FileSystem.Stat("helpers.sh")
 	require.NoError(t, err)
 	tests.AssertFilePermissionsEqual(t, "helpers.sh", pkg.FileModeExecutable, helperFileStats.Mode())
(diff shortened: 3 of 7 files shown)


cmd/porter/bundle.go
```diff
@@ -31,14 +31,23 @@ func buildBundleCommands(p *porter.Porter) *cobra.Command {
...
-		Use:   "create",
+	cmd := &cobra.Command{
+		Use:   "create [bundle-name]",
 		Short: "Create a bundle",
-		Long:  "Create a bundle. This generates a porter bundle in the current directory.",
+		Long:  "Create a bundle. This generates a porter bundle with the directory with the specified name.",
+		Args:  cobra.MaximumNArgs(1), // Expect at most one argument for the bundle name
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return p.Create()
+			bundleName := "mybundle" // Default value
```
troy0820, line 40: Is this necessary? This command will error if bundle name is not passed which I believe is what we want. 

```bash
porter create // will error out
porter create mybundle //will not fail 
```

troy0820:
```suggestion
			var bundleName string 
```

pkg/porter/create.go
```diff
@@ -9,35 +9,41 @@ import (
...
-func (p *Porter) Create() error {
-	fmt.Fprintln(p.Out, "creating porter configuration in the current directory")
+func (p *Porter) Create(bundleName string) error {
+	fmt.Fprintf(p.Out, "creating porter configuration for %s\n", bundleName)
 
-	err := p.CopyTemplate(p.Templates.GetManifest, config.Name)
+	// Create a directory with the given bundle name
+	err := os.Mkdir(bundleName, os.ModePerm)
+	if err != nil {
+		return fmt.Errorf("failed to create directory for bundle: %w", err)
```
troy0820, line 18:
```suggestion
		return fmt.Errorf("failed to create directory for bundle %s: %w", bundleName, err)
```

troy0820 (commented):
Some nits and a question/suggestion about the create command.

cmd/porter/bundle.go
sbdtu5498: @troy0820 currently it will not throw any error. It instead will create a bundle with the default name "mybundle". But if we want to throw an error if the argument isn't passed, then sure, I will fix it.

troy0820: @schristoff @phillipgibson Do we want the behavior to error _(error being a word used to describe not executing the command but show the usage)_ when the argument for the create bundle isn't given to make the user do this? 

```cli
porter create [bundle]
```

Or do we want to create the bundle with the default `mybundle` in the directory if the argument isn't given?

troy0820: >  But if we want to throw an error if the argument isn't passed, then sure, I will fix it.

@sbdtu5498 
I'm not quite sure but I want to see what is the desired behavior that we want and asking the community.  From the issue I believe we want to emulate the same behavior as the below clis. 


`helm create` requires 1 argument 
`cargo new ` requires 1 argument

sbdtu5498: @troy0820 sure thing! Once we get more reviews on how it should be handled, I will add the required changes.

schristoff: My bias is towards helm, and feeling that most users of our product are familiar with helm. 
In that case, let's force the user to do the hard thing, and pick a name. Error if they don't pass in an argument, because making assumptions can cause issues down the line. :)

pkg/porter/create.go
```diff
@@ -9,35 +9,41 @@ import (
 	"get.porter.sh/porter/pkg/config"
 )
 
-func (p *Porter) Create() error {
-	fmt.Fprintln(p.Out, "creating porter configuration in the current directory")
+func (p *Porter) Create(bundleName string) error {
```
schristoff, line 12: Based off the decision above to keep current functionality, but add the ability to pass an argument and make a directory, we should separate out `Create()` `CreateinNewDir()` but try to copy as little code between the two as possible. As always, my names for functions aren't _great_ so feel free to use your own.

sbdtu5498: Sure so basically how about adding a flag in create if it feels good say "--in -I"

sbdtu5498: Or creating a separate function for the purpose

schristoff: I would think of it like helper function. I wouldn't add a flag, maybe a bool - so if they do `porter create` with an argument it sets the directory bool to true. If you want, we could discuss more in the community meeting too :)

troy0820: If you check for the existence of the argument when `porter create` is called, you can call `p.Create()` (which is cwd) if there is no argument. If it does exist then call `p.CreateInDir(dir string)`. 

You can reuse the logic in `p.Create()` only changing  where it’s written to so you don’t have a lot of duplicate logic.

sbdtu5498: That sounds good, so let's go with the "porter create directory", does it sounds good.

cmd/porter/bundle.go
sbdtu5498: Sure thing, this seems feasible.

pkg/porter/create.go
schristoff: To clarify, I think what @troy0820  and I are talking about is just expanding `porter create` to accept `porter create [argument]` that argument being the directory name. Then if that argument is passed in you call `p.CreateInDir(dir string)`. Just re-clarifying because the way your comment there is worded makes it sound like you're making a command called `porter create directory` which is not exactly what I believe we're trying to head towards! :)

troy0820: Exactly, you are not creating another command, you are adding to the command that exists already (porter create).  

You will check the positional args (probably should use `func MaximumNArgs` as 1) and if there is one, you can create ~~change~~ the directory where it is going to be written with the arg passed in.  

That will create the directory and then you can reuse `p.Create()`.

sbdtu5498: cc: @troy0820 @schristoff the required changes have been added. Kindly review.

pkg/porter/create_test.go
```diff
@@ -39,5 +39,37 @@ func TestCreate(t *testing.T) {
 	dockerignoreStats, err := p.FileSystem.Stat(".dockerignore")
 	require.NoError(t, err)
 	tests.AssertFilePermissionsEqual(t, ".dockerignore", pkg.FileModeWritable, dockerignoreStats.Mode())
+}
 
+func TestCreateInDirectory(t *testing.T) {
+	p := NewTestPorter(t)
+	defer p.Close()
+
+	err := p.Create("mybundle")
```
troy0820, line 48: If we pass in `mybundle/yourbundle/superbundle` will this work? 

Helm has a thing where it will create the directory (base directory) if the others exists.  So it `os.Stat`'s the previous directories before making a `helm create` in the last _(superbundle)_ "directory" assuming the last isn't a file.

We are creating the whole path regardless if it's a file at that path to include a directory of that and the bundle name will be `mybundle/yourbundle/superbundle`.  

We may need to Stat the filepath of the directories if they are passed like this to ensure we do not have a bundle name with slashes, etc.  

Helm has an example of this in their create flow.

troy0820 (changes requested):
Just a couple of comments and changes requested.

pkg/porter/create.go
```diff
@@ -9,37 +9,46 @@ import (
...
+	if bundleName == "" {
+		bundleName = "."
+	}
+
+	fmt.Fprintf(p.Out, "creating porter configuration for %s\n", bundleName)
```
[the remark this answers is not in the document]
sbdtu5498: sure this one is done

pkg/porter/create_test.go
sbdtu5498: Oh I am sorry I was a bit busy at the moment to answer. Let me get back on to this 
>If we pass in mybundle/yourbundle/superbundle will this work?
Yes this will work. But there's a catch. It would update the directory if it exists. Now I want to ask should we block the user from running porter create or porter create mybundle if any of the related files exist or should we stick with the behavior that it is in.

troy0820: > Yes this will work. 

I'm curious as what the name will be. Will it be `mybundle/yourbundle/superbundle`

> But there's a catch. It would update the directory if it exists. 

I don't think that's a problem.  Helm does the same thing.  If there exists a directory, it will overwrite what's in there if the target exists. 

> Now I want to ask should we block the user from running porter create or porter create mybundle if any of the related files exist or should we stick with the behavior that it is in.

The existing functionality will allow you to create a bundle in the current directory if anything exits  there and overwrite what is laid down on disks as an overwritten file.  This behavior should stay and expect that to be the case with the target directory _(if new created, if old overwrite)_

sbdtu5498: > I'm curious as what the name will be. Will it be mybundle/yourbundle/superbundle
it will be named superbundle. And will be inside youbundle folder which will be inside, superbundle. It would work exactly like how helm create works
>The existing functionality will allow you to create a bundle in the current directory if anything exits there and overwrite what is laid down on disks as an overwritten file. This behavior should stay and expect that to be the case with the target directory (if new created, if old overwrite)
Sure I have made those small changes will push it in few moments. Let me know if anything else is required.

sbdtu5498: @troy0820 now it is 100% similar to helm create. I have also written the test cases to prove the point and looked at them locally as well. Kindly take a look.

cmd/porter/bundle.go
```diff
@@ -31,14 +31,23 @@ func buildBundleCommands(p *porter.Porter) *cobra.Command {
 }
 
 func buildBundleCreateCommand(p *porter.Porter) *cobra.Command {
-	return &cobra.Command{
-		Use:   "create",
+	cmd := &cobra.Command{
+		Use:   "create [bundle-name]",
 		Short: "Create a bundle",
-		Long:  "Create a bundle. This generates a porter bundle in the current directory.",
+		Long:  "Create a bundle. This generates a porter bundle with the directory with the specified name or in the root directory if no name is provided.",
```
schristoff, line 37: I would leave this as "current directory" or use "working directory" instead of "root directory"
Root directory can have a lot of different meanings depending on the users background, and in my interpretation that is not where I would want my bundle. :)

pkg/porter/create.go
schristoff: One minor nit on this, since this is an exported function can we add a comment describing the function at the top?
i.e 
```
//Create does things! (actually put what the function does here)
func (p *Porter) Create (bundleName string) error {
```

pkg/porter/create.go
```diff
@@ -4,40 +4,59 @@ import (
...
 
-func (p *Porter) Create() error {
-	fmt.Fprintln(p.Out, "creating porter configuration in the current directory")
+func (p *Porter) Create(bundleName string) error {
+	// Normalize the bundleName by removing trailing slashes
+	bundleName = strings.TrimSuffix(bundleName, "/")
 
-	err := p.CopyTemplate(p.Templates.GetManifest, config.Name)
+	if bundleName == "" {
+		bundleName = p.FileSystem.Getwd() // Use the current directory if no directory is passed
```
schristoff, line 18:
```suggestion
        // Use the current directory if no directory is passed
	if bundleName == "" {
		bundleName = p.FileSystem.Getwd()
```

pkg/porter/create.go
```diff
@@ -4,40 +4,59 @@ import (
...
+	}
+
+	fmt.Fprintf(p.Out, "creating porter configuration in %s\n", bundleName)
```
schristoff, line 21: I feel like this should be moved to the end of the function, so if we succeed, this outputs

cmd/porter/bundle.go
```diff
@@ -31,14 +31,23 @@ func buildBundleCommands(p *porter.Porter) *cobra.Command {
...
+		Args:  cobra.MaximumNArgs(1), // Expect at most one argument for the bundle name
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return p.Create()
+			bundleName := "" // Default value
```
schristoff, line 40:
```suggestion
                        // By default we create the bundle in the current directory
			bundleName := ""
```

Minor comment nit

schristoff: By the way it is hard to do suggestions for comments in github because it won't let me use tabs :( so it may look a little weird

pkg/porter/create.go
sbdtu5498: Okay, will move it.

cmd/porter/bundle.go
sbdtu5498: Yeah, but I got it anyways! would fix them up.

schristoff:
Hey @sbdtu5498  - just following up on this to see if you need any help! :)

sbdtu5498:
Extremely sorry for the delays @schristoff. This thing really got out of my mind. Thanks a lot for reminding me as well.

pkg/porter/create.go
```diff
@@ -4,40 +4,66 @@ import (
...
+	if bundleName == "" {
+		bundleName = p.FileSystem.Getwd()
+	}
+
+	// Check if the directory in which bundle needs to be created already exists.
+	// If not return error.
+	// It takes both the relative path and absolute path into consideration.
+	// For example if we want to create a bundle named mybundle in an existing directory /home/user we can call porter create /home/user/mybundle or porter create mybundle in the /home/user directory.
+	// If we are in a directory /home/user and we want to create mybundle in the directory /home/user/directory given the directory exists, 
+	// we can call porter create directory/mybundle from the /home/user directory or with any relative paths' combinations that one can come up with.
```
troy0820, line 28: To clarify, when we do this `porter create scripts/bundles/porter` is the name going to be this as well?  I feel like we may have to split on the `/` to `os.Stat` the directory path if it doesn't/does exist but only use the last name from the `/` when creating the bundle. This seems like it may be doing the former.

pkg/porter/create_test.go
```diff
@@ -39,5 +42,125 @@ func TestCreate(t *testing.T) {
 	dockerignoreStats, err := p.FileSystem.Stat(".dockerignore")
 	require.NoError(t, err)
 	tests.AssertFilePermissionsEqual(t, ".dockerignore", pkg.FileModeWritable, dockerignoreStats.Mode())
+}
+
+// tests to ensure behavior similiarity with helm create
+
+func TestCreateInDirectory(t *testing.T) {
+	// Create a temporary directory for testing
+	tempDir, err := ioutil.TempDir("", "porter-test")
```
troy0820, line 51:
```suggestion
	tempDir, err := os.TempDir("", "porter-test")
```
ioutil.TempDir is deprecated and will not pass the gate for staticcheck/go vet

pkg/porter/create_test.go
```diff
@@ -39,5 +42,125 @@ func TestCreate(t *testing.T) {
...
+	_, err = p.FileSystem.Stat(filepath.Join(bundleDir, ".gitignore"))
+	require.True(t, os.IsNotExist(err))
+
+	_, err = p.FileSystem.Stat(filepath.Join(bundleDir, ".dockerignore"))
+	require.True(t, os.IsNotExist(err))
+}
+
+func TestCreateInChildDirectoryWithExistingParentDirectory(t *testing.T) {
+	// Create a temporary directory for testing
+	tempDir, err := ioutil.TempDir("", "porter-test")
```
troy0820, line 126:
```suggestion
	tempDir, err := os.TempDir("", "porter-test")
```

pkg/porter/create_test.go
```diff
@@ -39,5 +42,125 @@ func TestCreate(t *testing.T) {
...
+	tests.AssertFilePermissionsEqual(t, filepath.Join("mybundle", ".gitignore"), pkg.FileModeWritable, gitignoreStats.Mode())
+
+	dockerignoreStats, err := p.FileSystem.Stat(filepath.Join(bundleDir, ".dockerignore"))
+	require.NoError(t, err)
+	tests.AssertFilePermissionsEqual(t, filepath.Join("mybundle", ".dockerignore"), pkg.FileModeWritable, dockerignoreStats.Mode())
+}
+
+func TestCreateInChildDirectoryWithoutExistingParentDirectory(t *testing.T) {
+	// Create a temporary directory for testing
+	tempDir, err := ioutil.TempDir("", "porter-test")
```
troy0820, line 90:
```suggestion
	tempDir, err := os.TempDir("", "porter-test")
```

pkg/porter/create_test.go
```diff
@@ -1,25 +1,28 @@
 package porter
 
 import (
+	"io/ioutil"
```
troy0820, line 4:
```suggestion
```

troy0820 (changes requested):
Just some nits but passing through the gates for static check we need to change the `ioutil` import and use `os.TempDir`

This is looking good.  Definitely appreciate your patience.  

Also, a question about the bundlename to make sure we don't fail prematurely if we need to walk the directories before creating/not creating them.

pkg/porter/create.go
sbdtu5498: No the name will be let say if you create /home/sbdtu5498/gentoo/nix/os. let say both gentoo and nix exist then it will work and will create bundle in the directory named os. If os doesn't exist, it will create a directory called os. But if nix doesn't exist, then it will throw error no such directory exist. This is the exact behavior that helm commands show. I tried to closely replicate it. Hope it makes sense. I will add few more descriptor comments.

sbdtu5498:
> This is looking good. Definitely appreciate your patience.

No problem at all and sorry for some delays on my part as well.

> Also, a question about the bundle name to make sure we don't fail prematurely if we need to walk the directories before creating/not creating them.

Have answered it above.

> Just some nits but passing through the gates for static check we need to change the ioutil import and use os.TempDir

Fixing it ASAP.

pkg/porter/create.go
sbdtu5498: // Only condition to use porter create with absolute and relative paths is that all the directories in the path except the last one should strictly exist.

sbdtu5498: I will add this line here. Now it will make sense.

sbdtu5498:
@troy0820 please take a look. I think there would be some merge conflicts. I will rebase the branch later if that's fine.

pkg/porter/create.go
```diff
@@ -4,40 +4,80 @@ import (
...
+	if err != nil {
+		if os.IsNotExist(err) {
+			// This code here attempts to create the directory in which bundle needs to be created,
+			// if the directory does not exist.
+			// bundleName can handle both the relative path and absolute path into consideration,
+			// For example if we want to create a bundle named mybundle in an existing directory /home/user we can call porter create /home/user/mybundle or porter create mybundle in the /home/user directory.
+			// If we are in a directory /home/user and we want to create mybundle in the directory /home/user/directory given the directory exists,
+			// we can call porter create directory/mybundle from the /home/user directory or with any relative paths' combinations that one can come up with.
+			// Only condition to use porter create with absolute and relative paths is that all the directories in the path except the last one should strictly exist.
+			err = os.Mkdir(bundleName, os.ModePerm)
```
troy0820, line 35:
```suggestion
			err = os.MkdirAll(bundleName, os.ModePerm)
```
This will make sure that this creates bundles such as `porter create mybundle/newestbundle`

pkg/porter/create.go
```diff
@@ -4,40 +4,80 @@ import (
...
+	if bundleName == "" {
+		bundleName = p.FileSystem.Getwd()
+	}
+
+	// Check if the directory in which bundle needs to be created already exists.
+	// If not, create the directory.
+	_, err := os.Stat(bundleName)
+	if err != nil {
+		if os.IsNotExist(err) {
+			// This code here attempts to create the directory in which bundle needs to be created,
```
troy0820, line 28: This comment is HUGE.  Can we shorten this some?

troy0820 (changes requested):
Last change needed but this is so close @sbdtu5498.  Thank you for your contribution and appreciate your patience with us.

schristoff:
Oh @sbdtu5498  I am so sorry, I never saw that you updated this awhile back. Once those last comments of Troy's are merged I will look over and get this merged ASAP.

ludfjig:
Hi @sbdtu5498 , are you still working on this?

Issue #2777, opened 2023-06-08
update hugo to v0.113.0


Arhell:
https://github.com/gohugoio/hugo/releases/tag/v0.113.0

PR #2780, opened 2023-06-09
update hugo to v0.113.0

Arhell:
# What issue does it fix
Closes https://github.com/getporter/porter/issues/2777


schristoff (approved):

schristoff:
I must have reverted this change somewhere without noticing, that is my bad :)

PR #2824, opened 2023-07-13
Change printer for displayParam to display with list showing values like `porter param show`

### LINKED ISSUE #2785: porter param list doesn't display params
## Describe the bug
Porter param list doesn't display params that porter param show <name> does display

## To Reproduce
1. Run `porter create` to create hello-porter bundle
2. Edit the porter.yaml to uncomment the mysql parameters and build bundle
3. Run this porter command `porter param generate` to create porter parameter set from the bundle
4. Run this porter command `porter param list` to show created params......

## Expected behavior
A list of one that includes the param set you just generated.

## Porter Command and Output
```
squillace@idiopath:~/work/squillace/howdy$ porter param show mysql
Name: mysql
Created: 1 hour ago
Modified: 1 minute ago

-----------------------------------------
  Name        Local Source  Source Type
-----------------------------------------
  mysql_user  RALPH         env
squillace@idiopath:~/work/squillace/howdy$ porter param list
-----------------------------
  NAMESPACE  NAME  MODIFIED
-----------------------------
squillace@idiopath:~/work/squillace/howdy$
```

## Version

porter v1.0.14 (0e739d88)

troy0820:
# What does this change
Fixes the `param list` to show the output of the params and not just the name.  The outputs for `yaml` and `json` show the params when queried and providing that functionality to fix the issue is why this fix is necessary.

# What issue does it fix
Closes #2785

# Notes for the reviewer
_This can be up for discussion if we do not intend to fix this, this way_ 

# Checklist
- [x] Did you write tests?

# Reviewer Checklist


pkg/porter/parameters.go
```diff
@@ -81,17 +81,14 @@ func (p *Porter) PrintParameters(ctx context.Context, opts ListOptions) error {
...
-				cr, ok := v.(DisplayParameterSet)
-				if !ok {
-					return nil
-				}
-				return []string{cr.Namespace, cr.Name, tp.Format(cr.Status.Modified)}
-			}
-		return printer.PrintTable(p.Out, params, printParamRow,
-			"NAMESPACE", "NAME", "MODIFIED")
+		paramsSets := [][]string{}
+		for _, ps := range params[0].Parameters {
```
troy0820, line 85: This is done because each install should have 1 parameter set but the function above returns a []DisplayParameterSet but it's length is one.  Tested by adding more params and it still held true to be 1 per install.

pkg/printer/table.go
```diff
@@ -21,6 +21,18 @@ func NewTableSection(out io.Writer) *tablewriter.Table {
 	return table
 }
 
+func PrintTableParameterSet(out io.Writer, params [][]string, headers ...string) error {
```
troy0820, line 24: Reusing the same method that was used to print before would require a lot of refactoring that would make the function unreadable. Also being that the other method uses it's slice value for the length of the table values (rows) it would always be 1 and append all values on the same row which didn't prove to be what we wanted.  
_(1 ParameterSet can have multiple values, this would use the 1 parameter set as the length and changing that would mess up other methods that call this.)_

Also, using this method doesn't break existing functionality with other parts of the code that uses it.

troy0820:
```python
⇒ porter parameters show porter-hello
Name: porter-hello
Created: 1 hour ago
Modified: 1 hour ago

---------------------------------------------
  Name            Local Source  Source Type
---------------------------------------------
  mysql_user      fake-user     env
  wordpress-user  faker-user    env

⇒ porter parameters list porter-hello -o plaintext
-------------------------------------------------
  NAME            VALUE  STRATEGY    MODIFIED
-------------------------------------------------
  mysql_user      env    fake-user   1 hour ago
  wordpress-user  env    faker-user  1 hour ago

⇒ porter param list porter-hello -o plaintext
-------------------------------------------------
  NAME            VALUE  STRATEGY    MODIFIED
-------------------------------------------------
  mysql_user      env    fake-user   1 hour ago
  wordpress-user  env    faker-user  1 hour ago
```

pkg/porter/parameters.go
troy0820: If this is supposed to handle more than 1 install, I can change this to deal with that behavior.

AGMETEOR:
Hey @troy0820 I hope I am allowed to add my review to this. This looks good and fixes the bug. I just noticed we could maybe also populate the `Type` column when we run `porter parameters show`

pkg/porter/parameters.go
```diff
@@ -81,17 +81,14 @@ func (p *Porter) PrintParameters(ctx context.Context, opts ListOptions) error {
...
+			list := []string{}
+			list = append(list, ps.Name, ps.Source.Strategy, ps.Source.Hint, tp.Format(params[0].Status.Modified))
```
AGMETEOR, line 87: Just a question about order of `ps.Source.Strategy` here and how it aligns columns `NAME", "VALUE", "STRATEGY", "MODIFIED`. It seems in the table columns it should come third after value but here it comes second after the name?

troy0820:
> Hey @troy0820 I hope I am allowed to add my review to this. This looks good and fixes the bug. I just noticed we could maybe also populate the Type column when we run porter parameters show

Yes, please any feedback is valuable. I noticed from the comment that the “strategy” could be `TYPE` which goes with what the other command shows. 

I can change this to that

pkg/porter/parameters.go
troy0820: Good catch, I can arrange the columns to match the requisite “value”. I can do this while changing strategy to TYPE as well as suggested in the other comment. 
Thanks @AGMETEOR

AGMETEOR (approved):

troy0820:
_Updated PR with tests to show `p.PrintTableParameterSet` in `p.PrintParameters` would satisfy the PR requirements_

pkg/porter/parameters.go
```diff
@@ -81,17 +81,14 @@ func (p *Porter) PrintParameters(ctx context.Context, opts ListOptions) error {
...
+			paramsSets = append(paramsSets, list)
+		}
+		return printer.PrintTableParameterSet(p.Out, paramsSets,
+			"NAME", "TYPE", "VALUE", "MODIFIED")
```
schristoff, line 91: I think we should still leave in `namespace` here

troy0820:
@schristoff Here is the issue https://github.com/getporter/porter/issues/2836

I can work on that as well.

AGMETEOR:
> #2836

@troy0820 do you mind if I take that off your hands?

troy0820:
@AGMETEOR I think we want to talk about this in the community office hours.  It's this week but I don't mind if you start with the initial implementation.

AGMETEOR:
> @AGMETEOR I think we want to talk about this in the community office hours. It's this week but I don't mind if you start with the initial implementation.

Could you kindly send me an invitation to `#porter` slack channel?

troy0820:
https://getporter.org/community/

https://communityinviter.com/apps/cloud-native/cncf

troy0820:
```python
-----------------------------------------------
  NAMESPACE  NAME  TYPE    VALUE   MODIFIED
-----------------------------------------------
             foo   secret  ******  1983-04-18
```

This is fixed to show the output being obfuscated if it is an "env" or a "secret".  @AGMETEOR is handling the fix to be more global around params.

pkg/porter/parameters.go
```diff
@@ -81,17 +82,20 @@ func (p *Porter) PrintParameters(ctx context.Context, opts ListOptions) error {
...
-			func(v interface{}) []string {
-				cr, ok := v.(DisplayParameterSet)
-				if !ok {
-					return nil
+		paramsSets := [][]string{}
+		for _, ps := range params {
+			for _, param := range ps.Parameters {
+				list := []string{}
+				hint := param.Source.Hint
+				if param.Source.Strategy == host.SourceEnv || param.Source.Strategy == secrets.SourceSecret {
```
troy0820, line 90: This is not the solution we want but using this for display purposes.  Solution is currently being worked in from issue #2836 as a global option rather than just the plaintext.

troy0820: Test file will match output of command.

pkg/porter/parameters.go
troy0820: Changed to handle more than one install.

pkg/porter/parameters.go
troy0820: Resolving comment.

troy0820:
Because issue #2836 is closed as the value of the secret is never shown but the name of the secret, there is no need to obfuscate the secret(s) when displaying the params list in json, yaml, or plaintext.

schristoff (approved):

schristoff:
ADO pipeline is still broken, this is an Azure issue, not a me issue. Just going to bypass and merge. :/

PR #2855, opened 2023-08-01
UnitTests fail on windows due to path formatting problems 

### LINKED ISSUE #2854: bug: test fails when running on windows due to path formatting
## Describe the bug
Path formatting on windows (backslash vs forwardslash) does not match expected linux format so many test fail

```
--- FAIL: TestStoreRelocationMapping (0.00s)
    --- FAIL: TestStoreRelocationMapping/relocation_file_gets_a_path (0.00s)
        cache_test.go:182:
                Error Trace:    C:/Users/lliljenberg/dev/porter/pkg/cache/cache_test.go:182
                Error:          Not equal:
                                expected: "C:\\Users\\lliljenberg/.porter/cache/887e7e65e39277f8744bd00278760b06/cnab/relocation-mapping.json"
                                actual  : "C:\\Users\\lliljenberg\\.porter\\cache\\887e7e65e39277f8744bd00278760b06\\cnab\\relocation-mapping.json"
```

## To Reproduce
Run `mage testunit`

## Expected behavior
No errors

## Version

porter v1.0.15 (be5c66a9)

ludfjig:
# What does this change
Lots of tests fail on windows because paths are hardcoded using forward slash, and never expects a volume such as "C:". 

# What issue does it fix
Closes #2854 but leaves #2858 for later

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/exec/builder/execute_test.go
+++ b/pkg/exec/builder/execute_test.go
@@ -144,10 +144,6 @@ func TestExecuteStep_HasOrderedArguments(t *testing.T) {
 }
 
 func TestExecuteStep_SpecifiesCustomWorkingDirectory(t *testing.T) {
-	//if runtime.GOOS == "windows" {
-	//	t.Skip() // pwd not available on windows (for some reason the only command that works is "help" ?!?)
-	//}
-
 	ctx := context.Background()
 	c := porterruntime.NewTestRuntimeConfig(t)
 	c.TestContext.UseFilesystem()
--- a/pkg/exec/builder/execute_test.go
+++ b/pkg/exec/builder/execute_test.go
@@ -145,6 +145,10 @@ func TestExecuteStep_HasOrderedArguments(t *testing.T) {
 }
 
 func TestExecuteStep_SpecifiesCustomWorkingDirectory(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip() // pwd not available on windows (for some reason the only command that works is "help" ?!?)
+	}
+
 	ctx := context.Background()
 	c := porterruntime.NewTestRuntimeConfig(t)
 	c.TestContext.UseFilesystem()
--- a/pkg/exec/builder/execute_test.go
+++ b/pkg/exec/builder/execute_test.go
@@ -2,7 +2,6 @@ package builder
 
 import (
 	"context"
-	"fmt"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -145,9 +144,9 @@ func TestExecuteStep_HasOrderedArguments(t *testing.T) {
 }
 
 func TestExecuteStep_SpecifiesCustomWorkingDirectory(t *testing.T) {
-	if runtime.GOOS == "windows" {
-		t.Skip() // pwd not available on windows (for some reason the only command that works is "help" ?!?)
-	}
+	//if runtime.GOOS == "windows" {
+	//	t.Skip() // pwd not available on windows (for some reason the only command that works is "help" ?!?)
+	//}
 
 	ctx := context.Background()
 	c := porterruntime.NewTestRuntimeConfig(t)
@@ -163,8 +162,13 @@ func TestExecuteStep_SpecifiesCustomWorkingDirectory(t *testing.T) {
 		SuffixArguments: []string{},
 	}
 
+	if runtime.GOOS == "windows" {
+		step.TestStep.Command = "cmd.exe"
+		step.Arguments = []string{"/c", "cd"}
+	}
+
 	_, err := ExecuteStep(ctx, c.RuntimeConfig, step)
-	assert.Equal(t, fmt.Sprintln(wd), c.TestContext.GetOutput())
+	assert.Equal(t, wd, strings.TrimRight(c.TestContext.GetOutput(), "\r\n"))
 	require.NoError(t, err, "Execute Step failed")
 }
 
--- a/pkg/mixin/pkgmgmt_test.go
+++ b/pkg/mixin/pkgmgmt_test.go
@@ -16,12 +16,12 @@ func TestRunner_BuildCommand(t *testing.T) {
 		runnerCommand string
 		wantCommand   string
 	}{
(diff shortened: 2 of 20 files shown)


pkg/exec/builder/execute_test.go
```diff
@@ -145,6 +145,10 @@ func TestExecuteStep_HasOrderedArguments(t *testing.T) {
 }
 
 func TestExecuteStep_SpecifiesCustomWorkingDirectory(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip() // pwd not available on windows (for some reason the only command that works is "help" ?!?)
```
ludfjig, line 149: would love any help on this, all commands I tried except `help` resulted in error stating "no such command on %path"

jsturtevant: Is this running in context of powershell or cmd?  `cd` or `chdir` would work on cmd and I think `pwd` is an alias in powershell that will list the current directory

pkg/mixin/pkgmgmt_test.go
```diff
@@ -16,12 +16,12 @@ func TestRunner_BuildCommand(t *testing.T) {
 		runnerCommand string
 		wantCommand   string
 	}{
-		{"build", "build", "/home/myuser/.porter/mixins/exec/exec build"},
-		{"install", "install", "/home/myuser/.porter/mixins/exec/exec install"},
-		{"upgrade", "upgrade", "/home/myuser/.porter/mixins/exec/exec upgrade"},
-		{"uninstall", "uninstall", "/home/myuser/.porter/mixins/exec/exec uninstall"},
-		{"invoke", "status", "/home/myuser/.porter/mixins/exec/exec invoke --action status"},
-		{"version", "version --output json", "/home/myuser/.porter/mixins/exec/exec version --output json"},
+		{"build", "build", "/home/myuser/.porter/mixins/exec/exec build\n\\home\\myuser\\.porter\\mixins\\exec\\exec.exe build"},
```
jsturtevant, line 19: what is happening here?

pkg/porter/cnab_test.go
```diff
@@ -98,6 +101,13 @@ func TestSharedOptions_validateBundleJson(t *testing.T) {
 }
 
 func Test_bundleFileOptions(t *testing.T) {
+	absWantFile, err := filepath.Abs(filepath.FromSlash("/" + config.Name))
```
jsturtevant, line 104: there is some funny handling here I believe, depending on how the it is seen as relative. I don't see anything wrong but double check you are getting the expected values in the test.

pkg/porter/credentials_test.go
```diff
@@ -295,6 +296,10 @@ func TestShowCredential_PreserveCase(t *testing.T) {
 }
 
 func TestCredentialsEdit(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip() // bash and vi not available on windows
```
jsturtevant, line 300: could this be achieved with powershell commands?

pkg/porter/credentials_test.go
```diff
@@ -315,7 +320,8 @@ func TestCredentialsEditEditorPathWithArgument(t *testing.T) {
 
 	p.Setenv("SHELL", "something")
 	p.Setenv("EDITOR", "C:\\Program Files\\Visual Studio Code\\code.exe --wait")
-	p.Setenv(test.ExpectedCommandEnv, "something -c C:\\Program Files\\Visual Studio Code\\code.exe --wait "+filepath.Join(os.TempDir(), "porter-kool-kreds.yaml"))
+	p.Setenv(test.ExpectedCommandEnv, "something -c C:\\Program Files\\Visual Studio Code\\code.exe --wait "+filepath.Join(os.TempDir(), "porter-kool-kreds.yaml")+
+		"\nsomething /C C:\\Program Files\\Visual Studio Code\\code.exe --wait "+filepath.Join(os.TempDir(), "porter-kool-kreds.yaml"))
```
jsturtevant, line 324: the formatting on these lines is hard to read consider using a string literal https://go.dev/ref/spec#String_literals and using `sprintf`

pkg/porter/publish_test.go
```diff
@@ -225,6 +224,11 @@ func TestPublish_RefreshCachedBundle(t *testing.T) {
 	require.NoError(t, err)
 	origBunPathTime := file.ModTime()
 
+	if runtime.GOOS == "windows" {
+		// see https://github.com/getporter/porter/issues/2858
+		time.Sleep(5 * time.Millisecond)
```
jsturtevant, line 229: :-(  we've seen this in k8s too

pkg/test/helper.go
```diff
@@ -31,7 +31,7 @@ func TestMainWithMockedCommandHandlers(m *testing.M) {
 
 			// There may be multiple expected commands, separated by a newline character
 			wantCmds := strings.Split(expectedCmdEnv, "\n")
-
+
```
jsturtevant, line 34: nit: extra white space

pkg/mixin/pkgmgmt_test.go
ludfjig: You can separate expected output by a newline character, so for this test there are two possible expected outputs which both pass

pkg/exec/builder/execute_test.go
ludfjig: Ah I found the problem, they are not executables, they are internal commands. Need to execute them using `cmd.exe /c dir` for example. https://stackoverflow.com/a/71639965

pkg/porter/cnab_test.go
ludfjig: thanks, I believe it's correct

pkg/porter/credentials_test.go
ludfjig: yep you are right, thanks

pkg/mixin/pkgmgmt_test.go
jsturtevant: oh interesting, might be easier to read if it is separated into array for the want command? maybe it isn't possible, not sure how the testing framework works.

PR #2879, opened 2023-08-19
Convert porter/porter to GitHub Actions

schristoff:
Pipeline migrated from [Azure DevOps](https://dev.azure.com/getporter/porter/_build?definitionId=18) :tada:

## Manual steps

Perform the following steps to complete the migration:

### porter/porter
- [ ] Ensure a runner with the following labels exists: $(vmImage)


.github/workflows/porter.yml
```diff
@@ -0,0 +1,172 @@
...
+      run: "./build/doc-only-build.sh"
+      shell: bash
+  Validate-build:
+    name: Native Compile
+    needs:
+    - Setup-pr_type
+    runs-on: ubuntu-latest
+    if: success() && needs.Setup-pr_type.outputs['pr_type.BUILD.DOCS_ONLY'] == 'false'
+    steps:
+    - name: checkout
```
troy0820, line 40: Do we need to checkout twice?  Can we checkout once and see if it is a doc build, etc?

pkg/porter/lifecycle.go
```diff
@@ -302,7 +302,7 @@ func (p *Porter) BuildActionArgs(ctx context.Context, installation storage.Insta
 		AllowDockerHostAccess: opts.AllowDockerHostAccess,
 		PersistLogs:           !opts.NoLogs,
 	}
-
+	// test cache
```
troy0820, line 305:
```suggestion
```

Not sure if this needs to be there.

troy0820 (approved):
One nit but non-blocking.  All the pipelines seem to be working.

PR #2892, opened 2023-09-01
add optional [bundle-name] for `porter create [bundle-name]` 

### LINKED ISSUE #2772: porter create command should create directory with bundle files included

ludfjig:
# What does this change
Takes over inactive PR #2778.

`porter create` now accepts an optional parameter `[bundle-name]`. A directory `bundle-name` will be created if it doesn't already exist, and the bundle will be created in there. The bundle name in porter.yaml will also be set to `bundle-name`. 

**I think we should consider making this argument required instead of optional even though it's a user facing change, see https://github.com/getporter/porter/pull/2778#discussion_r1225689310. It's also what helm does.**

Would appreciate feedback on how to word the command help text, I wasn't able to put it as succinctly as I wanted, and also want to mention that a directory will be created.

# What issue does it fix
Closes #2772 

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/cmd/porter/bundle.go
+++ b/cmd/porter/bundle.go
@@ -34,16 +34,15 @@ func buildBundleCreateCommand(p *porter.Porter) *cobra.Command {
 	return &cobra.Command{
 		Use:   "create [bundle-name]",
 		Short: "Create a bundle",
-		Long:  "Create a bundle. This command creates a new porter bundle with the specified bundle-name, in the directory with the specified bundle-name. If no bundle-name is provided, the bundle will be created in current directory and the bundle name will be 'porter-hello'.",
-		Args:  cobra.MaximumNArgs(1), // Expect at most one argument for the bundle name
+		Long: "Create a bundle. This command creates a new porter bundle with the specified bundle-name, in the directory with the specified bundle-name." +
+			" The directory will be created if it doesn't already exist. If no bundle-name is provided, the bundle will be created in current directory and the bundle name will be 'porter-hello'.",
+		Args: cobra.MaximumNArgs(1), // Expect at most one argument for the bundle name
 		RunE: func(cmd *cobra.Command, args []string) error {
-			// By default we create the bundle in the current directory
-			bundleName := ""
 			if len(args) > 0 {
-				bundleName = args[0]
+				bundleName := args[0]
+				return p.CreateInDir(bundleName)
 			}
-
-			return p.Create(bundleName)
+			return p.Create()
 		},
 	}
 }
--- a/cmd/porter/bundle.go
+++ b/cmd/porter/bundle.go
@@ -31,14 +31,24 @@ func buildBundleCommands(p *porter.Porter) *cobra.Command {
 }
 
 func buildBundleCreateCommand(p *porter.Porter) *cobra.Command {
-	return &cobra.Command{
-		Use:   "create",
+	cmd := &cobra.Command{
+		Use:   "create [bundle-name]",
 		Short: "Create a bundle",
-		Long:  "Create a bundle. This generates a porter bundle in the current directory.",
+		Long:  "Create a bundle. This generates a porter bundle in the directory with the specified name or in the current directory if no name is provided.",
+		Args:  cobra.MaximumNArgs(1), // Expect at most one argument for the bundle name
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return p.Create()
+			// By default we create the bundle in the current directory
+			bundleName := ""
+
+			if len(args) > 0 {
+				bundleName = args[0]
+			}
+
+			return p.Create(bundleName)
 		},
 	}
+
+	return cmd
 }
 
 func buildBundleBuildCommand(p *porter.Porter) *cobra.Command {
--- a/cmd/porter/bundle.go
+++ b/cmd/porter/bundle.go
@@ -31,15 +31,14 @@ func buildBundleCommands(p *porter.Porter) *cobra.Command {
 }
 
 func buildBundleCreateCommand(p *porter.Porter) *cobra.Command {
-	cmd := &cobra.Command{
+	return &cobra.Command{
 		Use:   "create [bundle-name]",
 		Short: "Create a bundle",
-		Long:  "Create a bundle. This generates a porter bundle in the directory with the specified name or in the current directory if no name is provided.",
+		Long:  "Create a bundle. This command creates a new porter bundle with the specified bundle-name, in the directory with the specified bundle-name. If no bundle-name is provided, the bundle will be created in current directory and the bundle name will be 'porter-hello'.",
 		Args:  cobra.MaximumNArgs(1), // Expect at most one argument for the bundle name
 		RunE: func(cmd *cobra.Command, args []string) error {
 			// By default we create the bundle in the current directory
 			bundleName := ""
-
 			if len(args) > 0 {
 				bundleName = args[0]
 			}
@@ -47,8 +46,6 @@ func buildBundleCreateCommand(p *porter.Porter) *cobra.Command {
 			return p.Create(bundleName)
 		},
 	}
-
-	return cmd
 }
 
 func buildBundleBuildCommand(p *porter.Porter) *cobra.Command {
--- a/pkg/porter/create.go
+++ b/pkg/porter/create.go
@@ -1,6 +1,7 @@
 package porter
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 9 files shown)


pkg/porter/create.go
```diff
@@ -1,43 +1,73 @@
...
+func (p *Porter) Create(bundleName string) error {
+	// Normalize the bundleName by removing trailing slashes
+	bundleName = strings.TrimSuffix(bundleName, "/")
 
-	err := p.CopyTemplate(p.Templates.GetManifest, config.Name)
+	// If given a bundleName, create directory if it doesn't exist
+	if bundleName != "" {
+		_, err := os.Stat(bundleName)
+		if err != nil && errors.Is(err, os.ErrNotExist) {
+			err = os.Mkdir(bundleName, os.ModePerm)
```
ludfjig, line 24: Troy wanted a MkdirAll here, but I think it makes more sense with a Mkdir because otherwise the bundlename could contain slashes (/) which I'm not sure is what we want.

cmd/porter/bundle.go
```diff
@@ -32,11 +32,18 @@ func buildBundleCommands(p *porter.Porter) *cobra.Command {
...
-		Use:   "create",
+		Use:   "create [bundle-name]",
 		Short: "Create a bundle",
-		Long:  "Create a bundle. This generates a porter bundle in the current directory.",
+		Long:  "Create a bundle. This command creates a new porter bundle with the specified bundle-name, in the directory with the specified bundle-name. If no bundle-name is provided, the bundle will be created in current directory and the bundle name will be 'porter-hello'.",
+		Args:  cobra.MaximumNArgs(1), // Expect at most one argument for the bundle name
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return p.Create()
+			// By default we create the bundle in the current directory
+			bundleName := ""
```
schristoff, line 41:
```suggestion
			var bundleName string 
```

cmd/porter/bundle.go
```diff
@@ -32,11 +32,18 @@ func buildBundleCommands(p *porter.Porter) *cobra.Command {
...
+			if len(args) > 0 {
```
schristoff, line 42: If create the bundleName string then we can skip the `if len(args) > 0` 
and just do the 
```
bundleName = args[0]
```

pkg/porter/create.go
```diff
@@ -1,43 +1,73 @@
...
-func (p *Porter) Create() error {
-	fmt.Fprintln(p.Out, "creating porter configuration in the current directory")
+// Create creates a new bundle configuration with the specified bundleName. A directory with the given bundleName will be created if it does not already exist.
+// If bundleName is the empty string, the configuration will be created in the current directory, and the name will be "porter-hello".
+func (p *Porter) Create(bundleName string) error {
+	// Normalize the bundleName by removing trailing slashes
+	bundleName = strings.TrimSuffix(bundleName, "/")
 
-	err := p.CopyTemplate(p.Templates.GetManifest, config.Name)
+	// If given a bundleName, create directory if it doesn't exist
```
schristoff, line 20: A couple of comments:
Instead of `if bundleName is not nil` and `ifBundle is nil` it's good to build the function around the concept of 
"If this thing, do this thing. Else, do the other thing"
You can use else, but the logic of the function should implicitly do the else.
Also, we should break this out into another function entirely. 

It would be preferred imo to not change the function signature of `Create()` to `Create(string)` but I have no strong preference.

schristoff (commented):
Looking good so far! 👍🏻

pkg/porter/create.go
troy0820: The MkDirAll would make the bundle at the path defined but not name the bundle the whole path. 
I should be able to make a bundle at `really/cool/path/awesome` and the bundle should be awesome but the path and scaffold should be at the path.

pkg/porter/create.go
```diff
@@ -1,43 +1,73 @@
...
+		if err != nil && errors.Is(err, os.ErrNotExist) {
+			err = os.Mkdir(bundleName, os.ModePerm)
+			if err != nil {
+				return fmt.Errorf("failed to create directory for bundle: %w", err)
+			}
+		}
+	}
+
+	var err error
+	if bundleName == "" {
```
troy0820, line 32: Your logic is repeating itself if it falls through the bundleName case being empty. Returning early with nil would make this check implicit if it isn’t filled out here. 

The logic is the same in both cases it’s just where it gets executed. This can make it easier to understand and less things to write.

pkg/porter/create.go
troy0820: You could make this optional and have the function signature not change by using `…` for the param.

pkg/porter/create.go
```diff
@@ -1,43 +1,73 @@
...
+	// Normalize the bundleName by removing trailing slashes
+	bundleName = strings.TrimSuffix(bundleName, "/")
 
-	err := p.CopyTemplate(p.Templates.GetManifest, config.Name)
+	// If given a bundleName, create directory if it doesn't exist
+	if bundleName != "" {
+		_, err := os.Stat(bundleName)
+		if err != nil && errors.Is(err, os.ErrNotExist) {
+			err = os.Mkdir(bundleName, os.ModePerm)
+			if err != nil {
```
troy0820, line 25: Your error is not nil so you don’t have to check it again. Your logic above can be changed to check if it is nil and the. Check if it is the error you are looking for. If it is still nil, you can fall through to the not nil case and not re check the error.

troy0820: This applies to line 23

pkg/porter/create.go
```diff
@@ -1,43 +1,73 @@
...
+		// create bundle with default name "porter_hello"
+		err = p.CopyTemplate(p.Templates.GetManifest, filepath.Join(bundleName, config.Name))
+	} else {
+		// create bundle with given name
+		err = p.CopyTemplate(func() ([]byte, error) {
+			content, err := p.Templates.GetManifest()
+			if err != nil {
+				return nil, err
+			}
+			content = []byte(strings.ReplaceAll(string(content), "porter-hello", bundleName))
```
troy0820, line 42: Are we removing the porter-hello hardcoded?

troy0820 (changes requested):
Couple of nits and some changes needed but this looking good.

pkg/porter/create.go
ludfjig: something like this? 	
```_, err := os.Stat(dir)
	if err != nil {
		if errors.Is(err, os.ErrNotExist) {
			err = os.MkdirAll(dir, os.ModePerm)
		}
		if err != nil {
			return fmt.Errorf("failed to create directory for bundle: %w", err)
		}
	}
```

cmd/porter/bundle.go
ludfjig: fixed

pkg/porter/create.go
ludfjig: hopefully this is addressed now that it's 2 different functions

pkg/porter/create.go
ludfjig: Yea, the porter-hello will be replaced

pkg/porter/create.go
```diff
@@ -1,43 +1,81 @@
...
+	"strings"
 
 	"get.porter.sh/porter/pkg"
 	"get.porter.sh/porter/pkg/config"
 )
 
+// Create creates a new bundle configuration in the current directory
 func (p *Porter) Create() error {
-	fmt.Fprintln(p.Out, "creating porter configuration in the current directory")
+	destinationDir := "."
```
schristoff, line 16: Little comment for this would be great

pkg/porter/create.go
```diff
@@ -1,43 +1,80 @@
...
+// Create creates a new bundle configuration in the current directory
 func (p *Porter) Create() error {
-	fmt.Fprintln(p.Out, "creating porter configuration in the current directory")
+	destinationDir := "." // current directory
 
-	err := p.CopyTemplate(p.Templates.GetManifest, config.Name)
+	if err := p.CopyTemplate(p.Templates.GetManifest, filepath.Join(destinationDir, config.Name)); err != nil {
+		return err
+	}
+	return p.copyAllTemplatesExceptPorterYaml(destinationDir)
```
schristoff, line 21:
```suggestion
       // pass in current directory
	return p.copyAllTemplatesExceptPorterYaml(".")
```

schristoff (approved):
One last nit and we are good to go

troy0820 (approved):

PR #2915, opened 2023-09-19
Fix bug where unwanted v is prepended to tag name

### LINKED ISSUE #2914: "v" prefix is prepended to tag when it shouldn't be
## Describe the bug
#2890 introduced a bug which where it appends a "v" prefix where it shouldn't. In the case that the tag is "latest", or when a tag is missing.

## To Reproduce
`porter explain -r ghcr.io/getporter/examples/porter-hello` and 
`bin/porter explain -r ghcr.io/getporter/examples/porter-hello:latest` will try fetch tag :v and :vlatest, respectively

## Expected behavior
No v should be prepended in these cases

## Porter Command and Output
```
$ porter [e.g. install --param foo=bar --credential-set azure]
lots of output
```

## Version

[e.g. porter v0.14.1-beta.2 (72dd5df)]

ludfjig:
# What does this change
Fixes bug

# What issue does it fix
Closes #2914 

# Notes for the reviewer

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/porter/lifecycle_test.go
+++ b/pkg/porter/lifecycle_test.go
@@ -3,6 +3,7 @@ package porter
 import (
 	"context"
 	"fmt"
+	"os"
 	"strings"
 	"testing"
 
@@ -585,7 +586,7 @@ func Test_ensureVPrefix(t *testing.T) {
 				assert.False(t, tt.args.opts._ref.Tag()[0] == 'v')
 			}
 
-			err := ensureVPrefix(tt.args.opts)
+			err := ensureVPrefix(tt.args.opts, os.Stdout)
 
 			// after
 			tt.wantErr(t, err, fmt.Sprintf("ensureVPrefix(%v)", tt.args.opts))
@@ -653,7 +654,7 @@ func Test_ensureVPrefix_idempotent(t *testing.T) {
 				assert.True(t, tt.args.opts._ref.Tag()[1] != 'v')
 			}
 
-			err := ensureVPrefix(tt.args.opts)
+			err := ensureVPrefix(tt.args.opts, os.Stdout)
 
 			// after
 			tt.wantErr(t, err, fmt.Sprintf("ensureVPrefix(%v)", tt.args.opts))
@@ -686,7 +687,7 @@ func Test_ensureVPrefix_latest(t *testing.T) {
 			bundleRef: nil,
 		}
 
-		err := ensureVPrefix(&opts)
+		err := ensureVPrefix(&opts, os.Stdout)
 		assert.NoError(t, err)
 
 		// should be unchanged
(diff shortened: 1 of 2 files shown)


schristoff (approved):
gj adding test

schristoff:
if it is telemetry test just merge anyways

pkg/porter/lifecycle_test.go
```diff
@@ -585,7 +586,7 @@ func Test_ensureVPrefix(t *testing.T) {
 				assert.False(t, tt.args.opts._ref.Tag()[0] == 'v')
 			}
 
-			err := ensureVPrefix(tt.args.opts)
+			err := ensureVPrefix(tt.args.opts, os.Stdout)
```
troy0820, line 589: Do we need to make all this noise in the test?  Can we use `io.Discard`?

pkg/porter/lifecycle_test.go
```diff
@@ -668,3 +669,53 @@ func Test_ensureVPrefix_idempotent(t *testing.T) {
...
+			BundlePullOptions: BundlePullOptions{
+				Reference:        latestRef,
+				_ref:             nil,
+				InsecureRegistry: false,
+				Force:            false,
+			},
+			bundleRef: nil,
+		}
+
+		err := ensureVPrefix(&opts, os.Stdout)
```
troy0820, line 690:
```suggestion
		err := ensureVPrefix(&opts, io.Discard)
```
So we don't have all the noise from the test because we are not asserting against the output.

pkg/porter/lifecycle_test.go
```diff
@@ -668,3 +669,53 @@ func Test_ensureVPrefix_idempotent(t *testing.T) {
...
```
troy0820, line 715: Same as above.

pkg/porter/lifecycle_test.go
```diff
@@ -668,3 +669,53 @@ func Test_ensureVPrefix_idempotent(t *testing.T) {
 		})
 	}
 }
+
+// ensure no v is added if specifying :latest tag
+func Test_ensureVPrefix_latest(t *testing.T) {
```
troy0820, line 674: Could these be one tests and make the cases a table test? 

```go
tests := map[string]struct{
latestRef string
}{
"latest":{latestRef:"example/porter-hello:latest"},
"not-latest":{latestRef:"example/porter-hello"},
}
for name, test := range tests {
 t.Run(name, func(t *testing.T) {
  opts := BundleReferenceOptions{
			installationOptions: installationOptions{},
			BundlePullOptions: BundlePullOptions{
				Reference:        test.latestRef,
				_ref:             nil,
				InsecureRegistry: false,
				Force:            false,
			},
			bundleRef: nil,
		}
		err := ensureVPrefix(&opts, os.Stdout)
})
}
```

troy0820 (commented):
You are facing test failures, outside of this changeset.  These are non-blocking but it would be nice if you could get them in.

pkg/porter/lifecycle_test.go
ludfjig: good idea!

PR #2781, opened 2023-06-10
fixing the nesting of tracing

nitishchauhan0022:
# What does this change
Fixes the improper nesting of traces which looks like caused by passing of the same context which is returned after logging the config traces, making the following traces as child traces of config trace.

# What issue does it fix
Closes #2563 

# Notes for the reviewer
I have tested the tracing which is mentioned in the issue is properly nested now.

PR #2916, opened 2023-09-19
Revert "fixing the nesting of tracing (#2781)"

schristoff:
Integ Test `TestTelemetrySetup` keeps failing due to our changes here. I'm reverting this for now with the hopes in the **very near** future we can fix this in a way that doesn't cause this flaky-ness.


troy0820 (approved):
LGTM just append your signing

Issue #2548, opened 2023-02-02
Convert a dependency's bundle interface document into a bundle.json


carolynvs:
This relies on #2323 being merged before this can be started on.

https://github.com/getporter/proposals/blob/main/pep/003-dependency-namespaces-and-labels.md#bundle-interfaces allows for either referencing an existing published bundle to define a bundle interface, or to use an embedded bundle.json. We don't want to allow any arbitrary json to be embeded in the porter manifest for this and will instead let the user define using the same porter syntax, the parameters, credentials and outputs that should be defined on the bundle in order to use it as a dependency.

When we convert a porter.yaml to a bundle.json, we need to convert the bundle interface document from it's porter representation into an (incomplete) CNAB bundle document and embed it in the final bundle.json for the bundle being converted.

## Dependencies
* #2323

ℹ️ Read [PEP003 - Advanced Dependencies](https://github.com/getporter/proposals/blob/main/pep/003-advanced-dependencies.md) for context about how dependencies should work, design details, and notes about desired behavior.

PR #2922, opened 2023-09-22
(feat) Bundles v2: Implement Bundle Interface Document

schristoff:
# What does this change
When setting a sharing mode is turned on for installations we need a way to share param/outputs/creds between them so we can wire them together. 

The way we are doing this without updating the cnab spec is by converting everything into a document (type interface). This document type is something CNAB already implements - so once we get to translating porter yaml -> cnab json we get a smiley face.

Problem: 
I remember talking to Carolyn about issues with using the existing `bundle.Parameters` and `bundle.Credentials` and I recall it being about the required types needed passed into them (for example, path being required for parameters) but I don't remember where we left off with that.
I guess we'll see what happens when we get there.

#2548 

# Checklist
- [x ] Did you write tests?

# Reviewer Checklist


schristoff:
Keeping Carolyn as a co-author on this for obvious reasons. Even months later, you're still the driving force behind this project @carolynvs  ❤️

arschles (approved):

schristoff:
oh i already did that

pkg/cnab/config-adapter/adapter.go
```diff
@@ -166,134 +166,151 @@ func (c *ManifestConverter) generateDefaultAction(action string) bundle.Action {
...
+		kind := "parameter"
+		if sourceParam.IsState {
+			kind = "state"
 		}
-		outputs[output.Name] = o
-	}
 
-	for _, o := range c.Manifest.Outputs {
-		addOutput(o)
+		defName := c.addDefinition(sourceParam.Name, kind, sourceParam.Schema, defs)
```
devigned, line 292:
```suggestion
		p.Definition = c.addDefinition(sourceParam.Name, kind, sourceParam.Schema, defs)
```

pkg/cnab/config-adapter/adapter_test.go
```diff
@@ -636,18 +637,43 @@ func TestManifestConverter_generateDependenciesv2(t *testing.T) {
...
 
 	testcases := []struct {
-		name    string
-		wantDep depsv2ext.Dependency
+		name     string
+		wantDep  depsv2ext.Dependency
+		wantDefs definition.Definitions
 	}{
-		{"all fields", depsv2ext.Dependency{
+		{name: "all fields", wantDep: depsv2ext.Dependency{
```
devigned, line 644: Doesn't look like this has gone through go fmt.

devigned (commented):
This isn't necessarily about this PR, but you may want to consider starting spans for more functions. For example, creating a logger is nice, but the span is still the same span as from the parent (I believe from looking at the implementation of LoggerFromContext), so the attributed log entries on the span will not have the localization to the function as one might expect.

I don't have enough depth to give a functional review. However, there are a couple superficial things that might be good to update, and one more widespread thing to consider, granularity of spans.

pkg/cnab/config-adapter/adapter.go
schristoff: Right now I believe most of our stuff goes to stdOut, I would love to implement a structured logger so we could follow the span throughout multiple functions, appending more data, and being able to see exactly what func errors. 
I believe right now because we abstract away our logging that implementing a structured logger wouldn't be a huge undertaking.

schristoff: Oh hey, this is new https://go.dev/blog/slog

devigned: Structured log is good, but tracing is better. Tracing is structured logs + causality through call hierarchy. This is probably going to be more apparent in the operator as stuff will be less going to stdout and more to some telemetry ingestion pipeline.

pkg/cnab/config-adapter/adapter_test.go
schristoff: I'll fix this next PR

PR #2927, opened 2023-09-25
(docs) expand more on the "default" field for parameters

schristoff:


docs/content/wiring.md
```diff
@@ -51,6 +51,15 @@ You can also provide any other attributes, as specified by the CNAB [parameters]
   default: "wordpress"
 ```
 
+If using the default parameter field it must be set. However, it must be passed in as an empty
```
ludfjig, line 54: it's not clear to me what these two sentences mean, maybe could be clearer?

schristoff: Sure! Is there an exact part that is confusing to you?

ludfjig: What does "it must me set", and why you gotta set them to some empty value

arschles: how about the following:

>If you decide to use the default parameter field, you must set it as the empty value

does this capture everything you want to say?

arschles (changes requested):

docs/content/wiring.md
ludfjig: I still dont understand why it has to be empty. Isn't this in contradiction with the above example? 
- name: database_name
  type: string
  default: "wordpress"

PR #2934, opened 2023-09-29
(feat) [shareddependencies]: move solver logic out of v1 pkg so it can be shared in extendedbundle

schristoff:
We're just moving this logic somewhere that will make it easier to access this functions for v1 + v2 deps packages. :)


schristoff:
Ran the failing integ test locally a couple times and it keeps passing 🙃  - I think ADO is not feeling working on a Friday, fair
``` 
--- PASS: TestDependenciesLifecycle (25.87s)
PASS
```

PR #2940, opened 2023-10-09
Shared Dependencies: the LOGIC 

schristoff:
Goal: If I have a parent bundle and a dependency, and that dependency already exists, don't install the dependency again, just wire them up together. 

Things to note:
- If child dependency exists, and sharing + group match, then **on install** we will get the arguments from the child dependency and give it to the parent when CNAB.execute is called. Child dependency won't be changed otherwise.
- For pre-existing child dependencies, the following occurs:
    -  Update: Updates to the parent are applied, the pre-existing child is not passed any update information
    - Invoke: This is for users to implement they want, but I'm leaving it as a foot gun. If you do an invoke and change the pre-existing child, then all other non-direct parents will need to be invoked as well. As in, we have no reconcile or logic to share these changes, do at your own risk.
    - Delete: For a pre-existing child, the parent is delete, the child is left. Users must delete the child manually.


pkg/porter/parameters.go
```diff
@@ -665,7 +663,7 @@ func (p *Porter) resolveParameterSources(ctx context.Context, bun cnab.ExtendedB
 				outputName = source.OutputName
 			case cnab.DependencyOutputParameterSource:
 				// TODO(carolynvs): does this need to take namespace into account
```
schristoff, line 665: Yes, but not namespace prob group name?

build/testdata/bundles/wordpress/porter.yaml
```diff
@@ -64,13 +64,13 @@ uninstall:
...
-  - name: wordpress-password
-    description: "The Wordpress installation password"
-    type: string
-    default: "default-password"
-    applyTo:
-      - "install"
-      - "upgrade"
-    sensitive: true
-    path: /cnab/app/outputs/wordpress-password
+# outputs:
```
schristoff, line 67: put this back

pkg/porter/parameters.go
schristoff: Yes, but I don't know what to do with it yet. If we name every related bundle with a prefix of the group name, this makes it easier to filter on. But we also have to deal with on deletes how do we handle child/parent relationship, esp. if the child has multiple parents?

docs/content/docs/development/authoring-a-bundle/working-with-dependencies.md
```diff
@@ -29,6 +29,55 @@ dependencies:
         reference: getporter/mysql:v0.1.3
 ```
 
+## Define dependencies v2 (Shared)
+
+DependenciesV2 is under the [**experimental** flag](https://porter.sh/docs/configuration/configuration/#experimental-feature-flags), and therefore shared dependencies
+is an experimental feature. Proceed with caution.
```
arschles, line 35:
```suggestion
The second version of dependencies -- also called "shared dependencies" or DependenciesV2 -- is available under the [**experimental** flag](https://porter.sh/docs/configuration/configuration/#experimental-feature-flags), and therefore an experimental feature. Please proceed with caution.
```

docs/content/docs/development/authoring-a-bundle/working-with-dependencies.md
```diff
@@ -29,6 +29,55 @@ dependencies:
...
+
+To enable DependenciesV2 you must set the experimental flag. This can be done
+by setting an environment variable:
```
arschles, line 38:
```suggestion
You can enable the experimental flag, thus enabling DependenciesV2, by setting an environment variable as follows:
```

docs/content/docs/development/authoring-a-bundle/working-with-dependencies.md
```diff
@@ -29,6 +29,55 @@ dependencies:
...
+```
+PORTER_EXPERIMENTAL=dependencies-v2
+```
+
+The configuration for dependenciesv2 is similar to v1, except now there is the "sharing" section which has the required fields of "mode", and "group" "name".
```
arschles, line 43:
```suggestion
The configuration for DependenciesV2 is similar to that of the first version, except there is now a "sharing" section with the following required fields: `mode`, `group.name`.
```

docs/content/docs/development/authoring-a-bundle/working-with-dependencies.md
```diff
@@ -29,6 +29,55 @@ dependencies:
...
+is an experimental feature. Proceed with caution.
+
+To enable DependenciesV2 you must set the experimental flag. This can be done
+by setting an environment variable:
+```
+PORTER_EXPERIMENTAL=dependencies-v2
+```
+
+The configuration for dependenciesv2 is similar to v1, except now there is the "sharing" section which has the required fields of "mode", and "group" "name". 
+Mode is a boolean, and group name is the identifier which will allow for certain
```
arschles, line 44:
```suggestion
`mode` is a boolean, and `group.name` is the identifier which will allow for certain
```

docs/content/docs/development/authoring-a-bundle/working-with-dependencies.md
```diff
@@ -29,6 +29,55 @@ dependencies:
...
+        group:
+          name: myapp
+      parameters:
+        database-name: wordpress
+        mysql-user: wordpress
+        namespace: wordpress
+```
+
+If there is an existing dependency installed that the parent bundle should connect to, you must create a label for the existing dependency with the key 
+of "sh.porter.SharingGroup" and the value of the group name specified in the parent bundle. The existing dependency **must** be successfully installed. If it is uninstalled this must be deleted by the users before this can proceed.
```
arschles, line 64:
```suggestion
If there is an existing dependency installed that the parent bundle should connect to, you must create a label for the existing dependency with the `sh.porter.SharingGroup` key, and the value of the group name specified in the parent bundle. The existing dependency **must** be successfully installed. If it is uninstalled this key must be deleted by the users before the operation can proceed. 
```

pkg/cnab/dependencies/v1/doc.go
arschles, line 1: should this file just be deleted? I can't tell from this diff it is is or not

pkg/cnab/extended_bundle.go
```diff
@@ -215,3 +225,201 @@ func (b ExtendedBundle) GetReferencedRegistries() ([]string, error) {
...
+	v2, err := bun.ReadDependenciesV2()
+	if err != nil {
+		return nil, fmt.Errorf("error reading dependencies v2 for %s", bun.Name)
+	}
+
+	q := make([]DependencyLock, 0, len(v2.Requires))
+	for name, d := range v2.Requires {
+		d.Name = name
+
+		//todo(schristoff): We should 100% move this logic into
```
arschles, line 274:
```suggestion
		// todo(schristoff): We should 100% move this logic into
```

arschles (approved):
This LGTM. I only have documentation nits.

pkg/cnab/dependencies/v1/doc.go
schristoff: I don't... know. But I think not?

devigned (changes requested):
Mostly documentation feedback. I caught a couple spelling issues. I could have missed others, so might want to run through spellchecker.

Nice work!!

pkg/cnab/extended_bundle.go
```diff
@@ -215,3 +225,201 @@ func (b ExtendedBundle) GetReferencedRegistries() ([]string, error) {
...
+	}
+
+	q := make([]DependencyLock, 0, len(v2.Requires))
+	for name, d := range v2.Requires {
+		d.Name = name
+
+		// todo(schristoff): We should 100% move this logic into
+		// the bundle validator area
+		if d.Sharing.Mode && d.Sharing.Group.Name == "" {
+			return nil, fmt.Errorf("dont do this")
```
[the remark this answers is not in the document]
schristoff: lol whoops thank you

PR #2943, opened 2023-10-14
Check the licenses of dependencies

### LINKED ISSUE #2942: (githubactions/pipelines) Dependency License Checker
We should be checking the licenses of dependencies we import to make sure it stays happy with **our** license.

This could be [useful](https://github.com/google/go-licenses) to achieve this, but to be honest I only read the first two sentences. I'm sure there's pipelines we could copy for this though.

kallydev:
# What does this change

Use github.com/google/go-licenses in GitHub Actions to check the licenses of dependencies.

# What issue does it fix

Closes #2942.

# Notes for the reviewer

https://github.com/kallydev/porter/actions/runs/6518629371

If we need to ignore more packages, we can use the `--ignore` multiple times.

```bash
go-licenses check ./... \
    --ignore github.com/xi2/xz,golang.org/x/sys/unix # https://github.com/xi2/xz/blob/master/LICENSE \
    --ignore github.com/xi2/xz,golang.org/x/sys/unix # https://github.com/xi2/xz/blob/master/LICENSE
```

# Checklist
- [ ] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/.github/workflows/check-licenses.yaml
+++ b/.github/workflows/check-licenses.yaml
@@ -1,0 +1,24 @@
+name: Check licenses
+on:
+  push:
+  pull_request:
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version-file: go.mod
+          cache: true
+      - name: Install go-licenses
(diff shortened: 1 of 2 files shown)


.github/workflows/check-licenses.yaml
```diff
@@ -0,0 +1,24 @@
+name: Check licenses
+on:
+  push:
+  pull_request:
```
schristoff, line 4: Can we set this up so it only runs on pull requests against main?

PR #2949, opened 2023-10-23
Add feature to mount host volumes

### LINKED ISSUE #475: Support mounting volumes for install/uninstall commands
I have a use case in which I would like to bind-mount a file in the install process.
After exploring the codebase, it is not clear whether it is possible or not to define volumes/mount points in the porter.yml or in any other place.

Is this something left to mixins specification?

maniSbindra:
# What does this change
This change adds flag **--mount-host-volume** for commands like **porter install** and **porter uninstall**, which enables making a host file system folder available inside the running container, which executes the action.

For example the following commands makes the host's **$HOME/.hostfolder** folder available inside the bundle (as **read/write** mount) at **/home/nonroot/.hostfolder** .

```
porter install test-install --mount-host-volume "$HOME/.hostfolder:/home/nonroot/.hostfolder:rw"
```

Multiple folders may be mounted. Format for adding a mount is **\<host path\>:\<container path\>\[:\<option\>\]**. If option is **ro** or not provided or invalid, then file system is mounted as **read only**. If option is **rw** then file system is mounted as **read/write** inside the container

# What issue does it fix
Closes #475 

# Notes for the reviewer
This works with the docker driver. By default folders are mounted as read only.

# Checklist
- [X] Did you write tests?

# Reviewer Checklist


lbergnehr:
This is great! I would suggest, if it's not already there, to also enable mounting through the porter config. That way you vould set up mounts you always want, for example `~/.azure` when doing Azure CLI authentication. Would be even more seamless.

pkg/cnab/provider/docker_linux.go
```diff
@@ -47,3 +47,14 @@ func (r *Runtime) mountDockerSocket(cfg *container.Config, hostCfg *container.Ho
 	hostCfg.Mounts = append(hostCfg.Mounts, dockerSockMount)
 	return nil
 }
+
+func (r *Runtime) addVolumeMountToHostConfig(hostConfig *container.HostConfig, source string, target string, readOnly bool) error {
+	mount := mount.Mount{
+		Source:   source,
+		Target:   target,
+		Type:     "bind",
```
sgettys, line 55: Should this be volume instead of bind?

pkg/porter/lifecycle.go
```diff
@@ -92,6 +99,47 @@ func (o *BundleExecutionOptions) GetParameters() map[string]interface{} {
 	return o.finalParams
 }
 
+// Sets the final resolved set of host volumes to be made availabe to the bundle
+func (o *BundleExecutionOptions) GetHostVolumeMounts() []cnabprovider.HostVolumeMountSpec {
```
sgettys, line 103: Will this handle spaces in the windows directory structure?

pkg/cnab/provider/docker_linux.go
maniSbindra: I did try with volume, however then you cannot mount a target folder where the name of the folder starts with a dot, the error received is something like `cannot create container: Error response from daemon: create /abc/.xyz: "/Users/testuser/.xyx" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed`

pkg/porter/lifecycle.go
maniSbindra: Yes it does work. I have now added a test for the same. Thanks

sgettys (approved):

PR #2953, opened 2023-11-09
Prevent panic on invalid configuration structure in exec mixin

### LINKED ISSUE #2946: (panic): exec mixin being passed string instead of a map causes panic
When using the exec mixin on a bundle the following: 
```
exec: echo hello world
```

Will cause a panic 
however

```
exec:
   command: '"echo hello world"'
```

Will not

We should not get to the point where the former will cause a panic and we should just error

HeavyBR:
# What does this change

Modified the `GetDescription()` method to check if the mixin has the correct structure before try to access their fields, with that we prevent panics on invalid mixins and we return a proper error.

# What issue does it fix

Closes #2946 

# Notes for the reviewer

I've started a discussion about this in the issue, but I don´t know if is correct to duplicate the error log as it is doing right now, we're logging in the error span and in the end of the cobra command.

# Checklist

- [X] Did you write tests?
- [X] If this is your first pull request, please add your name to the bottom of our [Contributors][contributors] list. Thank you for making Porter better! 🙇‍♀️

# Reviewer Checklist


schristoff (approved):

HeavyBR:
Hey @schristoff, seems that the integration tests are not passing, I've tried to found what is happening in the Azure Pipelines logs but I couldn't found, also tried to reproduce locally but nothing. Can you help me with that?

schristoff:
Looks like `TestUninstall_DeleteInstallation` failed but it looks like it did so in the Setup stage. Unsure if it is the changes here or in CI. 
If you'd like, I'd encourage you to quickly run the test and see if it fails locally, but I can also poke at it later today too ✨

HeavyBR:
I would't like to bother you with this but I've tried this test on my machine and I have to say "It works on my machine" 🤣 

I'll have to ask for your help again, maybe I'm not seeing something 😢

schristoff:
@HeavyBR  - very late, but you are right, it was CI and it did work on your machine. I'll get this merged soon.

PR #2966, opened 2024-01-24
(chore): cleanup ado, add badge

schristoff:


.github/workflows/porter-install-check.yml
```diff
@@ -25,8 +25,7 @@ jobs:
       shell: powershell
   macos:
     runs-on:
-      - self-hosted
-      - macOS-latest
+      - macos-latest
```
troy0820, line 28: Did we want to do this on Windows? I know that @ludfjig  was having issues with this at times. If it's easier not to, we don't have to.

build/testdata/bundles/mysql/.cnab/bundle.json
troy0820, line 1: This is empty?  Can we delete this?

.github/workflows/build_azure_pipelinesrelease_template.yml
```diff
@@ -78,7 +78,7 @@ jobs:
   Validate-unit_test:
     name: Unit Test
     runs-on: ubuntu-latest
-    if: !(inputs.skipTests)
+    if: ${{inputs.skipTests == false}}
```
troy0820, line 81: Only thing I would say is if we are moving from Azure pipelines, maybe let's rename the  file?

.github/workflows/porter-install-check.yml
troy0820: Nevermind, I see we already do this.

troy0820 (approved):
This LGTM.  The rename of the file is a nit and it's non-blocking.

build/testdata/bundles/mysql/.cnab/bundle.json
schristoff: It gets made when I delete files it expects :/

PR #2530, opened 2023-01-23
WIP: Lint messages reported from porter should not mention mixins

0xquark:
# What does this change
When porter generates a lint message directly, instead of the message coming from a mixin, we should change the message so that it doesn't print out extra "empty" text about mixins that isn't relevant.

# What issue does it fix
Closes #2526 

_If there is not an existing issue, please make sure we have context on why this change is needed. See our Contributing Guide for [examples of when an existing issue isn't necessary][1]._

[1]: https://getporter.org/src/CONTRIBUTING.md#when-to-open-a-pull-request

# Notes for the reviewer
I am not sure why my previous commits are also included. This issue isn't tested as it requires #2434 to be merged.

# Checklist
- [ ] Did you write tests?
- [ ] Did you write documentation?
- [ ] Did you change porter.yaml or a storage document record? Update the corresponding schema file.
- [ ] If this is your first pull request, please add your name to the bottom of our [Contributors][contributors] list. Thank you for making Porter better! 🙇‍♀️

# Reviewer Checklist
* Comment with /azp run test-porter-release if a magefile or build script was modified
* Comment with /azp run porter-integration if it's a non-trivial PR

[contributors]: https://getporter.org/src/CONTRIBUTORS.md

PR #2992, opened 2024-02-09
Lint messages reported from porter should not mention mixins

### LINKED ISSUE #2526: Lint messages reported from porter should not mention mixins
When porter generates a lint message directly, instead of the message coming from a mixin, we should change the message so that it doesn't print out extra "empty" text about mixins that isn't relevant.

Once #2434 is merged, you can do the following to reproduce:

1. porter create
2. edit porter.yaml and define a parameter named "porter_test"
3. porter lint

expected output
```
Running linters for each mixin used in the manifest...
error(porter-100) - Reserved name warning
porter_test has a reserved prefix. Parameters cannot start with porter- or porter_
See https://getporter.org/reference/linter/#porter-100 for more information
```

actual output
```
Running linters for each mixin used in the manifest...
error(porter-100) - Reserved name warning
: 0th step in the  mixin ()
porter_test has a reserved prefix. Parameters cannot start with porter- or porter_
See https://getporter.org/reference/linter/#porter-100 for more information
```

You can tell that a lint result was generated by porter because Location.Mixin is an empty string. The relevant code to change is 

https://github.com/getporter/porter/blob/6df27907b5a40482ceb8c036fede2fa56a518c16/pkg/linter/linter.go#L113-L116

Please update the unit tests in linter_test.go to validate that lint results generated by porter do not print out that extra text ": 0th step in the  mixin ()"

See our [Contributing Tutorial](https://getporter.org/contribute/tutorial) and [New Contributor Guide](https://github.com/getporter/porter/blob/release/v1/CONTRIBUTING.md) for help getting started contributing to Porter.

kichristensen:
# What does this change
When porter generates a lint message directly, instead of the message coming from a mixin, we should change the message so that it doesn't print out extra "empty" text about mixins that isn't relevant.

# What issue does it fix
Closes #2526

# Notes for the reviewer
There is a stale PR, #2530, for the same issue.

# Checklist
- [X] Did you write tests?
- [X] If this is your first pull request, please add your name to the bottom of our [Contributors][contributors] list. Thank you for making Porter better! 🙇‍♀️

# Reviewer Checklist

### DIFF
--- a/pkg/linter/linter.go
+++ b/pkg/linter/linter.go
@@ -111,8 +111,12 @@ type Location struct {
 }
 
 func (l Location) String() string {
-	return fmt.Sprintf("%s: %s step in the %s mixin (%s)",
-		l.Action, humanize.Ordinal(l.StepNumber), l.Mixin, l.StepDescription)
+	mixin := ""
+	if l.Mixin != "" {
+		mixin = fmt.Sprintf("in the %s mixin ", l.Mixin)
+	}
+	return fmt.Sprintf("%s: %s step %s(%s)",
+		l.Action, humanize.Ordinal(l.StepNumber), mixin, l.StepDescription)
 }
 
 // Results is a set of items identified by the linter.
--- a/pkg/linter/linter.go
+++ b/pkg/linter/linter.go
@@ -68,7 +68,9 @@ type Result struct {
 func (r Result) String() string {
 	var buffer strings.Builder
 	buffer.WriteString(fmt.Sprintf("%s(%s) - %s\n", r.Level, r.Code, r.Title))
-	buffer.WriteString(r.Location.String() + "\n")
+	if r.Location.Mixin != "" {
+		buffer.WriteString(r.Location.String() + "\n")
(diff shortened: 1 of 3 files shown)


pkg/linter/linter.go
```diff
@@ -111,8 +111,12 @@ type Location struct {
 }
 
 func (l Location) String() string {
-	return fmt.Sprintf("%s: %s step in the %s mixin (%s)",
-		l.Action, humanize.Ordinal(l.StepNumber), l.Mixin, l.StepDescription)
+	mixin := ""
```
schristoff, line 114: Looking at the original issue, I think the logic should be something along the lines of if the mixin name is "", we should not be returning a string. Let me know if you are reading that from #2526

kichristensen: You are complety correct, I will work on the change

kichristensen: @schristoff It has been changed now to align with your comment

schristoff (approved):

PR #2999, opened 2024-02-14
Update dependcies and update go to 1.21

troy0820:
# What does this change
Update dependencies to a newer version of kubernetes.  This change brought in the need to update go to 1.21 which should be okay being that go 1.22 came out.  

- [x] Update k8s dependencies 
- [x] Update go to 1.21 

# What issue does it fix
N/A

[1]: https://getporter.org/src/CONTRIBUTING.md#when-to-open-a-pull-request

# Notes for the reviewer
_Put any questions or notes for the reviewer here._

# Checklist
- [ ] Did you write tests?
- [ ] Did you write documentation?
- [ ] Did you change porter.yaml or a storage document record? Update the corresponding schema file.
- [ ] If this is your first pull request, please add your name to the bottom of our [Contributors][contributors] list. Thank you for making Porter better! 🙇‍♀️

# Reviewer Checklist
* Comment with /azp run test-porter-release if a magefile or build script was modified
* Comment with /azp run porter-integration if it's a non-trivial PR

[contributors]: https://getporter.org/src/CONTRIBUTORS.md

PR #3000, opened 2024-02-14
change to 1.21.3

troy0820:
Because main is on 1.20.7 and the pipeline is using main's yaml template, we have to update this so the rest of the PRs can do the integration tests and not fail.


troy0820:
These pipelines may fail

schristoff:
is this dependent on #2999 
also happy 3000 🎉

troy0820:
#2999 kinda fixes this but wanted to see if this was merged in first would the int tests pass because it needs the yaml@main to have the right go version.

troy0820:
Also may keep this PR ready for review because it’s 3000 lol

schristoff:
:( Goodbye 3k, you were cool

PR #3008, opened 2024-03-05
(bug): Update Porter/Porter GHA to only run on push into main

schristoff:


kurktchiev (approved):
Lgtm probability of breaking something never 0 :)

PR #3016, opened 2024-03-13
(chore): porter/porter should run on push to main and merge into main 

schristoff:
We grab the badge from porte/porter (now just porter) to put on the readme, and if it runs on pull requests only then that badge can be in a failed status because of something not merged. 
I added for it to run on merge into main so we can use that as a badge instead.


troy0820 (approved):
This looks good to me.

PR #3034, opened 2024-03-25
Lint for parameters being used in actions to which they don't apply

### LINKED ISSUE #1018: Lint for parameters being used in actions to which they don't apply
After #1017, it will be a hard runtime failure if your porter.yaml attempts to use `{{bundle.paramters.PARAM}}` when `PARAM` was defined to not apply to the action where that templated value is used.

We should lint for this and hard fail so that we can detect and correct this problem at build time.

Example:

```yaml
parameters:
- name: connstr
   type: string
   applyTo:
    - install

uninstall:
  - exec:
      description: "Use a non-applicable parameter"
      command: ./helpers.sh
      arguments:
      - "{{ bundle.parameters.connstr }}"
```

This should be an porter level linter, not a mixin specific linter, which would be our first.

kichristensen:
# What does this change
Update linter to check if parameters are being used in actions to which they don't apply, as it would cause a runtime failure.

If a parameter is used in an action to which is doesn't apply `porter lint` will produce the output similar to this:
```
error(porter-101) - Parameter does not apply to action
install: 1st step in the exec mixin (Install Hello World)
Parameter onlyAppliesToUninstall does not apply to install action
See https://porter.sh/docs/references/linter/#porter-101 for more information
---
```

# What issue does it fix
Closes #1018

# Notes for the reviewer

# Checklist
- [X] Did you write tests?
- [X] Did you write documentation?

# Reviewer Checklist


schristoff:
My initial thought is: "Cool, but are we breaking backward compat with this?" - which I think is a a big yes, but I think it may be okay to do it in this case.

kichristensen:
@schristoff Completely forgot about backwards compability with this change. That said, in my opinion, it is not fully backwards compatible.
But as is causes a hard runtime error instead for failing on during build, I think it is acceptable, and even better to fail early.
It is a grey zone 🤷 

No matter what happens with this PR, there should probably be some sort of strategy for adding new lints, as all new lints can be considered breaking changes in the sense that they potentially breaks existing builds, at least for errors. Might be worth taking up on a community meeting.

sgettys:
I agree it's better to hit this at build time than runtime. Any breakages because should only be on already broken bundles I would imagine

pkg/linter/linter.go
```diff
@@ -183,7 +189,29 @@ func (l *Linter) Lint(ctx context.Context, m *manifest.Manifest) (Results, error
...
+		actions = append(actions, action{actionName, steps})
+	}
+	for _, action := range actions {
+		res, err := validateParamsAppliesToAction(m, action.steps, tmplParams, action.name)
+		if err != nil {
+			return nil, span.Error(fmt.Errorf("error validating action: %s", action.name))
+		}
+		results = append(results, res...)
+	}
+	span.EndSpan()
```
sgettys, line 212: This should be a defer otherwise this won't get closed if the return on 208 is hit. Any reason to have 2 different spans in here instead of a single one?

sgettys:
Can you add an integration test for this as well?

pkg/linter/linter.go
kichristensen: You are correct, in both cases. Also I cannot find any good reasons for 2 spans here. I will change it

kichristensen:
> Can you add an integration test for this as well?

@sgettys Sure, it has been added

schristoff:
Since our next release is 1.1.0 (I know, breaking compat should be 2.0+ but this is a small compat thing) I'm fine with merging this sooner rather than later.

schristoff (approved):
Looks good once conflicts are resolved

PR #3052, opened 2024-04-01
Improve error message on malformed step

### LINKED ISSUE #370: poor error message: more than one mixin specified
When the yaml for an action step is malformed, and more than one key is at the root of the step, you trigger "more than one mixin specified". But the error message isn't clear on what is wrong and how to fix it.

Let's help people out by:
* Improving the error message
* Adding the error message to the FAQ and explain how to fix

# Reproduction Steps
1. make a new directory and inside it run porter create
2. Edit porter.yaml and mess up the indentation of the install step (note that description and the other fields should be indented underneath exec)
    ```yaml
    install:
        - exec:
          command: ./helpers.sh
          arguments:
            - install
    ```
3. Run porter build

```console
$  porter build
Using buildkit build driver
1 error occurred:
	* validation of action "install" failed: more than one mixin specified
```

Ideally the error message should say that the first step in the install action is malformed and they should check that all of the mixins's child fields are indented underneath the name of the mixin.

kichristensen:
# What does this change
Instead of giving a cryptic error message, "more than one mixin specified", better describe the most likely cause of the problem, making it easier for people to understand what to fix.

# What issue does it fix
Closes #370

# Checklist
- [X] Did you write tests?

# Reviewer Checklist

### DIFF
--- a/pkg/manifest/manifest.go
+++ b/pkg/manifest/manifest.go
@@ -23,6 +23,7 @@ import (
 	"github.com/cbroglie/mustache"
 	"github.com/cnabio/cnab-go/bundle"
 	"github.com/cnabio/cnab-go/bundle/definition"
+	"github.com/dustin/go-humanize"
 	"github.com/hashicorp/go-multierror"
 	"github.com/opencontainers/go-digest"
 	"go.opentelemetry.io/otel/attribute"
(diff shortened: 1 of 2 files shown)


pkg/manifest/manifest.go
```diff
@@ -23,6 +23,7 @@ import (
 	"github.com/cbroglie/mustache"
 	"github.com/cnabio/cnab-go/bundle"
 	"github.com/cnabio/cnab-go/bundle/definition"
+	"github.com/dustin/go-humanize"
```
schristoff, line 26: Nit: This package seems to be not very maintained, but it doesn't seem to really need a lot. We've been using it already, so it's really nbd, but with recent security things going on I think I'm just hyper-aware for a bit :)
Just thinking out loud here.

schristoff (approved):

Issue #2662, opened 2023-03-28
Support outputs template variable in dependency outputs field


carolynvs:
When evaluating a dependency's outputs for template syntax, we should support a shorter way of addressing the current dependency's outputs

Instead of ${bundle.dependencies.CURRENT_DEP.outputs.OUTPUT_NAME} the author should be able to use  ${outputs.OUTPUT_NAME}. These should be equivalent variables and use the same backing map when the template is resolved.

This variable is only applicable to the outputs section of a dependency. Each template must be evaluated separately, if we want to support this new variable since it is specific to the context in which it's running.

## Questions
- [ ] Is this too similar to `${bundle.outputs.OUTPUT_NAME}` which stores different data? like it could be `${dependency.outputs.OUTPUT_NAME}` instead. 🤷‍♀️ 

ℹ️ Read [PEP003 - Advanced Dependencies](https://github.com/getporter/proposals/blob/main/pep/003-advanced-dependencies.md) for context about how dependencies should work, design details, and notes about desired behavior.

PR #3058, opened 2024-04-04
Add support for mapped dependency outputs

kichristensen:
# What does this change
Implements support for dependency output mapping in Dependencies V2.
This is precursor to #2662 as it requires that dependency outputs can be mapped and templated. In this implementation only the "old" format is implemented, `${outputs.OUTPUT_NAME}` is not supported yet, but will be implemented in a follow-up PR, where the documentation also would be updated.

# What issue does it fix
Closes no issues but is a precursor to #2662

# Notes for the reviewer

# Checklist
- [X] Did you write tests?


pkg/runtime/runtime_manifest.go
```diff
@@ -428,6 +452,13 @@ func (m *RuntimeManifest) ResolveStep(ctx context.Context, stepIndex int, step *
 		return log.Error(fmt.Errorf("unable to build step template data: %w", err))
 	}
 
+	mustache.AllowMissingVariables = false
+
+	err = m.buildAndResolveMappedDependencyOutputs(sourceData)
```
schristoff, line 457: This should be behind a DepsV2 flag

pkg/runtime/runtime_manifest_test.go
```diff
@@ -406,6 +406,95 @@ install:
...
+	pCtx.Setenv("PORTER_MYSQL_PASSWORD_DEP_OUTPUT", "password")
+
+	mContent := `schemaVersion: 1.0.0
+dependencies:
+  requires: 
+  - name: mysql
+    bundle:
+      reference: "getporter/porter-mysql"
+    outputs:
+      mappedOutput: Mapped ${ bundle.dependencies.mysql.outputs.password }
```
schristoff, line 456: Is there a way to remove the addtl `Mapped`

pkg/runtime/runtime_manifest_test.go
```diff
@@ -406,6 +406,95 @@ install:
 	assert.Equal(t, []string{"mysql-password", "password"}, gotSensitiveValues, "Incorrect values were marked as sensitive")
 }
 
+func TestResolveStep_DependencyMappedOutput(t *testing.T) {
```
schristoff, line 409: I think having a test leveraging golden files like [we did here](https://github.com/getporter/porter/blob/main/pkg/cnab/testdata/bundle-depsv2.json) would be helpful :)

schristoff (changes requested):
I appreciate you taking action on this :)

pkg/runtime/runtime_manifest_test.go
kichristensen: @schristoff Are you thinking about changing this test, or adding another?
I'm asking because it feels a little wrong to me to change this one, because it is about generating a single part of the runtime manifest, and the other tests in this file follow the same pattern as used here.

schristoff: Adding another is fine (always needs more tests)

schristoff (approved):

PR #3071, opened 2024-04-09
Resolve bundle digests when DepV2 is enabled

### LINKED ISSUE #2676: Resolve dependency digests during build
During `porter build`, we should resolve the repository digest for each dependency that has a default bundle reference specified. The reference in the final porter.yaml stored in ./cnab/app/porter.yaml should use the digest instead of the tag.

# Questions
[ ] It is still TBD how we will preserve the original tag in the bundle to use later for relocation mapping (#2670) 

ℹ️ Read [PEP003 - Advanced Dependencies](https://github.com/getporter/proposals/blob/main/pep/003-advanced-dependencies.md) for context about how dependencies should work, design details, and notes about desired behavior.

kichristensen:
# What does this change
When Dependencies V2 is enable bundle dependencies with a default bundle reference will be resolved to digests.

# What issue does it fix
Closes #2676

# Notes for the reviewer

# Checklist
- [X] Did you write tests?


pkg/porter/generateManifest.go
```diff
@@ -111,15 +113,83 @@ func (p *Porter) generateInternalManifest(ctx context.Context, opts BuildOptions
...
+		}
+	}
+
 	return e.WriteFile(build.LOCAL_MANIFEST)
 }
 
+func (p *Porter) resolveDependencyDigest(ctx context.Context, e *yaml.Editor, opts cnabtooci.RegistryOptions) error {
+	// find all referenced dependencies that does not have digest specified
+	// get the digest for all of them and update the manifest with the digest
+	return e.WalkNodes(ctx, "dependencies.requires.*", func(ctx context.Context, nc *yqlib.NodeContext) error {
```
schristoff, line 133: My only concern is
1) Should we resolve dependencies that aren't required (I feel yes?) - and:
2) Are there any concerns with this and  "shared dependencies", where the dependency is already installed and running (I think this should be okay)

pkg/porter/generateManifest.go
```diff
@@ -111,15 +113,83 @@ func (p *Porter) generateInternalManifest(ctx context.Context, opts BuildOptions
...
+	if p.IsFeatureEnabled(experimental.FlagDependenciesV2) {
+		if err = p.resolveDependencyDigest(ctx, e, regOpts); err != nil {
+			return err
+		}
+	}
+
 	return e.WriteFile(build.LOCAL_MANIFEST)
 }
 
+func (p *Porter) resolveDependencyDigest(ctx context.Context, e *yaml.Editor, opts cnabtooci.RegistryOptions) error {
```
schristoff, line 130: Could this be reworked to leverage existing [dependency resolution](https://github.com/getporter/porter/blob/main/pkg/manifest/manifest.go#L868) logic?  - moving this into "validate" logic may help us reduce LOC and improve efficiency :)

pkg/porter/generateManifest.go
```diff
@@ -111,15 +113,83 @@ func (p *Porter) generateInternalManifest(ctx context.Context, opts BuildOptions
 			}
 		}
 		return e.SetValue(path+"digest", digest.String())
-
 	})
 	if err != nil {
 		return err
 	}
 
+	if p.IsFeatureEnabled(experimental.FlagDependenciesV2) {
```
schristoff, line 121: Same as below (?) - I would love to see this moved into [here](https://github.com/getporter/porter/blob/main/pkg/manifest/manifest.go#L120) - [we can use this](https://github.com/getporter/porter/blob/main/pkg/manifest/manifest.go#L174) too :)

schristoff (changes requested):
Thank you for working on this. I've taken a stab at it before but would get sidetracked. I would love to see it using the validate() logic, which makes this a little straightforward but long term could improve efficiency and be more "predictable" in placement.

pkg/porter/generateManifest.go
kichristensen: 1. I'm not sure I fully understand what you mean with "dependencies not required"
2. I think it should be okay too, but I will create a test to verify it. It makes sense long term anyway

kichristensen:
> Thank you for working on this. I've taken a stab at it before but would get sidetracked. I would love to see it using the validate() logic, which makes this a little straightforward but long term could improve efficiency and be more "predictable" in placement.

@schristoff Are you suggesting to use `validate()` to do the digest replacement or something else?

pkg/porter/generateManifest.go
kichristensen: 2. Should be covered by the existing integration for shared dependencies. Do you agree?

schristoff (approved):
We've have a lot of conversations about this, but ultimately my forgetfulness has been holding this PR up. Let's just merge it and if I ever remember, I can refactor.

PR #3088, opened 2024-04-28
Perform full checkout of repository during release

### LINKED ISSUE #3087: When producing canary release the version is calculated incorrectly
## Describe the bug
When publishing the canary release the magefile tries to determine the version of the commit. That is not possible as Github Actions by default checks out the repository with a depth of 1, making it impossible to determine the correct version using `git describe --tags`.
See https://github.com/getporter/porter/actions/runs/8861567104/job/24333736800 for an example of a failing action

## To Reproduce
1. Merge to `main`
2. See that the [porter-canary](https://github.com/getporter/porter/actions/workflows/porter-canary.yml) action fails

## Expected behavior
The action should succeed

kichristensen:
# What does this change
Perform full checkout of repository during release
Otherwise `git describe --tags` won't work correctly, because of the missing history, causing the canary build to not be published.

# What issue does it fix
Closes #3087

# Checklist
- [ ] Did you write tests?

### DIFF
--- a/.github/workflows/build_pipelinesrelease_template.yml
+++ b/.github/workflows/build_pipelinesrelease_template.yml
@@ -167,6 +167,8 @@ jobs:
     steps:
     - name: checkout
       uses: actions/checkout@v4.1.0
+      with:
+        fetch-depth: 0
     - uses: actions/setup-go@v4
       with:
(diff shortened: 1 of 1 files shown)


schristoff (approved):
Huh, good catch

PR #3089, opened 2024-04-30
(feat): show descriptions for params/creds at gen time

schristoff:


schristoff:
As going through this I found the `Hint:` and `Strategy:` fields in Secret.Source. This would be the optimal way to implement this. I'm going to see if we can figure out how much wiring is needed to get those together.

pkg/storage/parameterset_test.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 func TestNewParameterSet(t *testing.T) {
-	ps := NewParameterSet("dev", "myparams",
+	ps := NewParameterSet("dev", "myparams", "",
```
kichristensen, line 18: Wouldn't it make sense assert that the description is added correctly like this test does with the other inputs?

pkg/storage/credentialset_test.go
```diff
@@ -17,7 +17,7 @@ import (
 )
 
 func TestNewCredentialSet(t *testing.T) {
-	cs := NewCredentialSet("dev", "mycreds", secrets.SourceMap{
+	cs := NewCredentialSet("dev", "mycreds", "", secrets.SourceMap{
```
kichristensen, line 20: Wouldn't it make sense assert that the description is added correctly like this test does with the other inputs?

pkg/storage/parameterset_test.go
schristoff: Correct, I'll update these

schristoff:
closing for #3112

PR #3094, opened 2024-05-04
Remove workflow_dispatch from test-porter-release action

kichristensen:
# What does this change
Parameters cannot be used when triggered by anything other than workflow_dispatch. As this action is also triggered on PRs targeting the main branch, it cannot use parameters. An example of a failing action can be found here https://github.com/getporter/porter/actions/runs/8909019136.
Instead the parameters are no longer dynamic and the default values have been hardcoded.

**UPDATE**:
As the content of this workflow is already covered by the `porter/porter-canary` workflow, `test-porter-release` is deleted instead.

# Checklist
- [ ] Did you write tests?


schristoff:
Now this is showing "test-porter-release" job on PRs, and I think we only want it on nightly or when there is a merge into main :/

kichristensen:
@schristoff Good point, actually we might not want it at all, as the flow is also "tested" when producing the canary build on merge to main. That could be an argument for deleting the workflow all together. What to you think?

schristoff:
I'm cool with deleting it :)

kichristensen:
@schristoff it is deleted instead

schristoff (approved):

PR #3100, opened 2024-05-07
Support skipping parameters and credentials during generate

### LINKED ISSUE #3074: Parameter generation should allow for acceptance of default value
Currently we are allowed to define a `default: some_value` filed for each parameter. However, when generating a parameter-set the process does not allow me to take the defined default as... well a default.

During generation, a `skip` option should be added to choices.

**Additional context**
[Related Slack discussion
](https://cloud-native.slack.com/archives/CN8NA4F8V/p1712596822247469)

### LINKED ISSUE #3075: Credential generation should allow for skipping a non-required field and allow for setting a default
Currently we are allowed to define a `required: bool` filed for each credential. However, when generating a credentials-set the process does not allow me to skip the credential or define a default value it should/can hold.

During generation, a `skip` option should be added to choices and a default should be accepted

**Additional context**
[Related Slack discussion
](https://cloud-native.slack.com/archives/CN8NA4F8V/p1712596822247469)

kichristensen:
# What does this change
Adds a `skip` option to the prompt allow users to skip parameter and credential generation of specific entries.

The prompt will look like this:

```
Generating new credential test2 from bundle porter-hello

==> 2 credentials required for bundle porter-hello

? How would you like to set credential "kubeconfig"
   [Use arrows to move, enter to select, type to filter]
  secret
  specific value
  environment variable
  file path
  shell command
> skip
```

# What issue does it fix
Closes #3074 and closes #3075

# Notes for the reviewer
No tests have been created for this as it requires user interaction

# Checklist
- [ ] Did you write tests?


pkg/generator/generator.go
```diff
@@ -52,7 +53,7 @@ func genSurvey(name string, surveyType SurveyType) (secrets.SourceMap, error) {
 	// extra space-suffix to align question and answer. Unfortunately misaligns help text
 	sourceTypePrompt := &survey.Select{
 		Message: fmt.Sprintf("How would you like to set %s %q\n ", surveyType, name),
-		Options: []string{questionSecret, questionValue, questionEnvVar, questionPath, questionCommand},
+		Options: []string{questionSecret, questionValue, questionEnvVar, questionPath, questionCommand, questionSkip},
```
schristoff, line 56: How are you handling if required is set to true?

kichristensen: Go catch, no currently was focusing on the parameters, noticed that is was used for both credentials and parameters, and forgot about. I will update

kichristensen: @schristoff Updated, to only allow skipping parameters or credentials that aren't required. Discover that the CNAB bundle has a required parameter for both credentials and paramters. All parameters without a default value is required in the accroding the CNAB bundle.

schristoff: How do you feel about changing this to passing in an `opts ...Options` and those options contain description and required?

kichristensen: @schristoff that could make sense, it would allow us to easily extend the generator with new options over time. My only "concern" is if it is a refactor that gives value long term. I mean how many options would there be? But after the talk about more metadata on credentials it might actually be worth it.
I will update try to update the PR to use options, then we can see how it would look and feel.

schristoff: If it is too much work to implement - nbd on choosing to move forward with this way.

kichristensen: @schristoff I have updated this PR to use options, what do you think?

schristoff (approved):

Issue #2902, opened 2023-09-12
Signing of Porter Bundles


phillipgibson:
Providing the ability to sign a Porter bundle as an additional verification on data integrity of the bundle. 

Related issues  #1457 & #2041

PR #3135, opened 2024-05-28
Run signing integration test

kichristensen:
# What does this change
In order parallelize integration tests, the Github workflow doesn't use Mage to run all integration tests directly. Instead each integration test file is run in different jobs, requiring new integration test files to be manually added to the workflow.
The integration test for signing images was missing and is added in this PR.

For some reason running the Notation and Cosign test in parallel fails on in the Github workflow, so parallelization of those two tests are also disabled.

Related to #2902 

# Notes for the reviewer

# Checklist
- [ ] Did you write tests?

### DIFF
--- a/.github/workflows/porter-integration-pr.yml
+++ b/.github/workflows/porter-integration-pr.yml
@@ -124,3 +124,33 @@ jobs:
         PORTER_INTEG_FILE: uninstall_test.go
       run: go run mage.go -v TestIntegration
       shell: bash
+  signing_test_integ:
+    runs-on: ubuntu-latest
+    steps:
+    - name: checkout
+      uses: actions/checkout@v4.1.0
+    - uses: actions/setup-go@v4
+      with:
+        go-version: "${{ env.GOVERSION }}"
+        cache: true
+    - name: Set up Docker Buildx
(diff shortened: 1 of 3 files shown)


.github/workflows/porter-integration-pr.yml
```diff
@@ -124,3 +124,33 @@ jobs:
         PORTER_INTEG_FILE: uninstall_test.go
       run: go run mage.go -v TestIntegration
       shell: bash
+  signing_test_integ:
```
schristoff, line 127: Should we make another reusable workflow file?

schristoff (approved):
Approving but maybe we should make another reusable workflow file

.github/workflows/porter-integration-pr.yml
kichristensen: Didn't even think of that, it would be a great idea. I will update the PR

kichristensen: Unfortunately it is not possible. The documentation states that it is a maximum of 20 reusable workflows from the top-level workflow file 😢
https://docs.github.com/en/actions/using-workflows/reusing-workflows#limitations

Issue #3073, opened 2024-04-10
Canary builds are not published


kichristensen:
## Describe the bug
The website, https://porter.sh/docs/getting-started/install-porter/#canary, describes how to download the canary build, but the canary build has not been uipdate for a very long time, https://github.com/getporter/porter/releases/tag/canary.

## To Reproduce
1. Install the canary release
2. Run this porter command `porter version`
3. See that version v1.0.0-73-g43fa0c5f is installed and that the mixin installation fails

## Expected behavior
The canary release should be install correctly in the version produced by the main branch or the documentation should not mention how to download the canary build

## Porter Command and Output
```
$  curl -L https://cdn.porter.sh/canary/install-linux.sh | bash
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  1110  100  1110    0     0    846      0  0:00:01  0:00:01 --:--:--  3100
Installing porter@v1.0.0-73-g43fa0c5f to /home/kim/.porter from https://cdn.porter.sh
curl: (22) The requested URL returned error: 404
```

schristoff:
🤔  Is the mixin feed not being updated for canary either? I wonder if the github workflow needs the same update as you did for primary releases

kichristensen:
@schristoff I have looked a little at this issue. It seems to be a combination of two things, but I can only verify one of them.
1. The [porter-canary pipeline](https://github.com/getporter/porter/blob/main/.github/workflows/porter-canary.yml) always fails because of missing inputs when triggered by a push to main. This is pretty easy to fix.
2. It looks like that when the pipelines was running in Azure Devops that something updated the Git tag `canary`, see https://dev.azure.com/getporter/porter/_build/results?buildId=8747&view=logs&j=8561663e-1bfb-5abd-6b97-f5ad2d809477&t=2af5fe51-7771-5e5d-91a7-c278b90bd4a2&l=97. I cannot find any trace of the output in any mage files or similar, and it looks like it is important that the `canary` tag is updated in order for the canary build to be published

kichristensen:
Reopened as the canary build is still not published correctly

PR #3146, opened 2024-06-03
Use PUBLISH_TOKEN for publishing mixins

kichristensen:
# What does this change
GITHUB_TOKEN only have access to the current repository, when publishing mixins we need permission to create commits in the packages repo.

# What issue does it fix
Related to #3073 

# Notes for the reviewer
Because the secret is only available to workflows, and requires a canary release or tagged release to be triggered, it has not been tested if the secret is still valid. Worst case it will fail just as much as it did before.

# Checklist
- [ ] Did you write tests?


schristoff (approved):

schristoff:
whats up with DCO

schristoff (approved):

Issue #3136, opened 2024-05-30
Specifying version of mixin in porter.yaml


KurtSchenk:
**Is your feature request related to a problem? Please describe.**
Porter has mixins which are similar to plugins or packages that are listed in the porter.yaml file. However, the specific version of the mixin to install can only be provided in the Porter command. As a developer who has experience working with npm, NuGet modules, I find this limitation to be problematic. 

**Describe the solution you'd like**
It would be more efficient to track mixin versions in the yaml file, as it is done in files like package.json and csproj, especially when it comes to debugging and dealing with broken CI. When it is in the porter.yaml, you can be sure the right version of the mixin matches what you have tested: the package behaves the same all through the development and deployment lifecycle.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

schristoff:
Hey! 
To add clarification for **future me** and the community: Right now Mixins allow for `clientVersion` which is the version of the tool we download in the mixin. Ex: In the Terraform Mixin if we set ClientVersion to 1.1.0, then we are downloading Terraform 1.1.0 - but this is not the version of the mixin *itself*
What is being proposed is to specify the mixin version, so for instance:

```
mixin:
    terraform: 
     clientVersion: 1.1.0
     version: 1.0
```
     
As discussed this would be a schema change. We will have to add a function that checks what mixins are installed and if the version we want is installed. We currently have [this function](https://github.com/getporter/porter/blob/main/pkg/cnab/config-adapter/stamp.go#L196) which we may be able to dive into and extend

dgannon991:
Can we close this issue now? I know there was talk of maybe having it auto download the right version, or maybe having many installed at the same time and selected the right one?

kichristensen:
@KurtSchenk @dgannon991 implemented syntax implemented for specifying that a specific version of a mixin have to be installed when building the bundle. The version of the bundle can be specified using the following syntax:

```yaml
mixins:
  terraform@v1.0.0:
    clientVersion: 1.1.0
```

The correct version of the mixin have to be installed manually, but if you don't have the correct version installed, linting will fail.

Do you believe it solves you request?

kichristensen:
Assuming this was solved in #3215. If this is not the case please create a new issue

PR #3148, opened 2024-06-04
Added validation if someone specifies a version on a bundle

dgannon991:
# What does this change
Added a check on version numbers if specified in your porter YAML

# What issue does it fix
Delivers towards #3136

# Notes for the reviewer
This is just a draft to get some feedback. I intend to refactor the Lint() function and add some unit tests, but I wanted to check if I'd gone down the right route first :)

### DIFF
--- a/pkg/linter/linter.go
+++ b/pkg/linter/linter.go
@@ -12,7 +12,7 @@ import (
 	"get.porter.sh/porter/pkg/portercontext"
 	"get.porter.sh/porter/pkg/tracing"
 	"get.porter.sh/porter/pkg/yaml"
-	"github.com/Masterminds/semver"
+	"github.com/Masterminds/semver/v3"
 	"github.com/dustin/go-humanize"
 )
 
--- a/pkg/linter/linter.go
+++ b/pkg/linter/linter.go
@@ -12,6 +12,7 @@ import (
 	"get.porter.sh/porter/pkg/portercontext"
 	"get.porter.sh/porter/pkg/tracing"
 	"get.porter.sh/porter/pkg/yaml"
+	"github.com/Masterminds/semver"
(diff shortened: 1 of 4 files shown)


dgannon991:
I've added some tests and done a small refactor on the bottom of Lint(). Let me know if we want to break further parts of the function out!

pkg/linter/linter.go
```diff
@@ -263,9 +264,55 @@ func (l *Linter) Lint(ctx context.Context, m *manifest.Manifest) (Results, error
...
+		return nil, span.Error(err)
+	}
+
 	return results, nil
 }
 
+func (l *Linter) validateVersionNumberConstraints(ctx context.Context, m *manifest.Manifest) error {
+	for _, mixin := range m.Mixins {
+		if mapConfig, ok := mixin.Config.(map[string]interface{}); ok {
+			if v, exists := mapConfig["version"]; exists {
```
kichristensen, line 279: Are there currently anything ensuring that mixins cannot use a config field called `version`?
I'm thinking of the situation where mixins already provide a field called `version`, which might result in unintended behaviour

dgannon991: No, it was simply the suggested field name on the issue. We could always use `mixinVersion` or something more verbose to avoid clashing?

kichristensen: I personally like `version` the best, but it could break some existing plugins.
@schristoff do you know of any way we can easily handle a "reservation" of a mixin configuration property?

KurtSchenk: @kichristensen could a similar approach such as porter-state, porter-debug be taken? 
image

dgannon991:
I think we can close this one, it was done in a different way.

kichristensen:
Closes in favor of #3215

PR #3155, opened 2024-06-11
Call host secrets plugin directly when resolving secrets

### LINKED ISSUE #2223: Automatically call the host secrets plugin
Right now when Porter resolves values for a Parameter/Credential set, we call the secrets plugin for everything, even those where the source is not secret, like env, path, etc. Those other sources are all handled by the host secrets plugin and we shouldn't require every custom secrets plugin to handle figuring out when to call back to the host secrets plugin.

When resolving a ps/cs, Porter should check the source, and if it is a source that is associated with the host secrets plugin: command, value, env, path, then Porter should resolve it directly from the host secrets plugin. The secrets plugin should only be called for other sources. Right now the only other source is secret, but if in the future we add more source types, then we should also allow custom plugins to handle that if they choose.

kichristensen:
# What does this change
We should not require all secret plugins to call the host secrets plugin for non secret values. Instead we should call the host secrets plugin directly.

# What issue does it fix
Closes #2223 

# Notes for the reviewer
No integration test was added, as there already is an existing integration test covering the case
https://github.com/getporter/porter/blob/75d49d290e28fac1b54561513fd933ea1b67c190/tests/integration/install_test.go#L44

# Checklist
- [X] Did you write tests?


pkg/cnab/provider/credentials_test.go
```diff
@@ -145,3 +146,42 @@ func TestRuntime_loadCredentials_WithApplyTo(t *testing.T) {
...
+		Action: cnab.ActionInstall,
+		Credentials: storage.NewInternalCredentialSet(secrets.SourceMap{
+			Name: "password",
+			Source: secrets.Source{
+				Strategy: host.SourceValue,
+				Hint:     "mypassword",
+			},
+		}),
+	}
+	err := r.loadCredentials(context.Background(), b, &run)
```
schristoff, line 176: Think we could make a testcontext and pass that in here, but we don't have to (the func is like `NewPorterTestContext` or `NewTestContext` ?)

pkg/storage/helpers.go
```diff
@@ -23,3 +24,7 @@ func NewTestStore(tc *config.TestConfig) TestStore {
 func (s TestStore) Close() error {
 	return s.testPlugin.Close()
 }
+
```
schristoff, line 27: We should comment on what this logic is reasoning about

schristoff (approved):
two minor nits otherwise g2g

PR #3200, opened 2024-08-04
Do not reference CNAB invocation images in documentation

### LINKED ISSUE #175: Limit CNAB references throughout documentation
It's a bit confusing right now how we talk about CNAB, the spec, how porter maps back to it, etc in our docs right now. Because we can't make assumptions about what people know, and if people are coming to a page because they want to know about a porter feature or because they want to know about how porter is conforming to the spec.

I spoke with @jeremyrickard and we agreed on how to organize the docs a bit:

* Add a page explaining CNAB and how porter uses it
* Add a page explaining how the manifest maps to a bundle.json, where we conform to the spec and where we have unique features/data
* On each page, call out interesting CNAB tidbits in a subsection titled `CNAB`.
* Don't start any doc with references to CNAB, or assume that the reader already knows what it is. Assume that each porter doc (and the reader) is there for the porter feature first.

kichristensen:
# What does this change

As CNAB is an implementation detail as this point, we no longer want to use the term 'invocation image'. Instead we should use the term 'bundle image'.
This PR ensures that the current references to 'invocation image' is changed to 'bundle **image'.**

# Notes to reviewer
The Check Links check fails because the renamed paged has not been pushed yet

# What issue does it fix
Closes #175

# Checklist
- [ ] Did you write tests?
- [X] Did you write documentation?


pkg/porter/publish.go
```diff
@@ -153,11 +153,11 @@ func (p *Porter) publishFromFile(ctx context.Context, opts PublishOptions) error
 		m.Reference = ""
 	}
 
-	// Update invocation image and reference with opts.Reference, which may be
+	// Update bundle image and reference with opts.Reference, which may be
 	// empty, which is fine - we still may need to pick up tag and/or registry
 	// overrides
 	if err := m.SetInvocationImageAndReference(opts.Reference); err != nil {
```
schristoff, line 159: should change function sig maybe?

pkg/porter/publish.go
```diff
@@ -203,7 +203,7 @@ func (p *Porter) publishFromFile(ctx context.Context, opts PublishOptions) error
 
 	bundleRef.Digest, err = p.Registry.PushImage(ctx, imgRef, regOpts)
 	if err != nil {
-		return log.Errorf("unable to push CNAB invocation image %q: %w", m.Image, err)
+		return log.Errorf("unable to push CNAB bundle image %q: %w", m.Image, err)
 	}
 
 	bundleRef.Definition, err = p.rewriteBundleWithInvocationImageDigest(ctx, m, bundleRef.Digest)
```
schristoff, line 209: One second thought - it may drastically increase scope of an already large PR. I'll let you decide

schristoff (approved):

pkg/porter/publish.go
kichristensen: Done. Make sense and it wasn't a big change.

pkg/porter/publish.go
kichristensen: Done. I tried it and it wasn't too bad.

PR #3205, opened 2024-08-14
Bundle script for Offline Installation of Porter

### LINKED ISSUE #3198: Offline Environment Error: Unable to Pull Mongo Image in Porter v1.1.0
## Describe the bug
When using Porter in an offline environment, it attempts to pull the `mongo:4.0-xenial` image, resulting in an error due to the lack of internet connectivity. This prevents the successful execution of Porter commands that rely on this image.
## To Reproduce
1. Install the Porter binary in an offline environment.
2. Run the command `porter list`.
3. See the error message regarding the inability to pull the `mongo:4.0-xenial` image.
## Expected behavior
I expected the `porter list` command to execute successfully and list the installations without attempting to pull any external dependencies, as the environment is offline.
## Porter Command and Output
```
$ porter list
rpc error: code = Unknown desc = error pulling mongo:4.0-xenial: 
could not list installations: could not read storage schema document: rpc error: code = Unknown desc = error pulling mongo:4.0-xenial: 
could not list installations: could not read storage schema document: rpc error: code = Unknown desc = error pulling mongo:4.0-xenial:
```
## Version
```
porter v1.1.0
```

jmcudd:
# What does this change
_Give a summary of the change, and how it affects end-users. It's okay to copy/paste your commit messages._

The new `bundle-linux.sh` script automates the creation of a tarball containing all the necessary artifacts for installing the Porter CLI, including the MongoDB Docker image and the Porter installation files. It pulls a specific version of the MongoDB image and packages all the downloaded assets into a compressed tarball, providing clear instructions for extracting the bundle and executing the installation script. This makes it easy for users to set up Porter in air-gapped environments.

The `install-linux.sh` script has been updated to handle offline installations by installing local binaries and the MongoDB image if they are present. It now checks for the existence of the `mongo_image.tar` file, loads the MongoDB image into Docker, and runs a MongoDB container during installation if it is not already running. Additionally, the script allows the use of a local `porter-linux-amd64` binary if it is available, allowing for an fully offline installation.

# What issue does it fix
Closes #3198

# Notes for the reviewer

The `bundle-linux.sh` script pulls the latest version of the `install-linux.sh` from the CDN, so when testing you will need to replace the downloaded `install-linux.sh` file with the one from this branch.

# Checklist
- [ ] Did you write tests?

### DIFF
--- a/docs/content/docs/getting-started/install-porter.md
+++ b/docs/content/docs/getting-started/install-porter.md
@@ -33,7 +33,7 @@ curl -L https://cdn.porter.sh/$VERSION/install-mac.sh | bash
 
 **Linux**
 
-```bash 
+```bash
 export VERSION="v1.1.0"
 curl -L https://cdn.porter.sh/$VERSION/install-linux.sh | bash
 ```
--- a/docs/content/docs/getting-started/install-porter.md
+++ b/docs/content/docs/getting-started/install-porter.md
@@ -32,7 +32,8 @@ curl -L https://cdn.porter.sh/$VERSION/install-mac.sh | bash
 ```
 
 **Linux**
-```bash export VERSION="v1.1.0"
+```bash 
+export VERSION="v1.1.0"
 curl -L https://cdn.porter.sh/$VERSION/install-linux.sh | bash
 ```
 
--- a/docs/content/docs/getting-started/install-porter.md
+++ b/docs/content/docs/getting-started/install-porter.md
@@ -32,9 +32,7 @@ curl -L https://cdn.porter.sh/$VERSION/install-mac.sh | bash
 ```
 
 **Linux**
-
-```bash
-export VERSION="v1.1.0"
+```bash export VERSION="v1.1.0"
 curl -L https://cdn.porter.sh/$VERSION/install-linux.sh | bash
 ```
 
@@ -112,6 +110,21 @@ You will need to create a [PowerShell Profile][ps-link] if you do not have one.
 
 ```
 iwr "https://cdn.porter.sh/latest/install-windows.ps1" -UseBasicParsing | iex
+```
+## Air-gapped Install
+
+To run the installation script in an air-gapped environment, create an air-gapped installation package with the `bundle-linux.sh` script.
+
+```
+curl -L https://raw.githubusercontent.com/getporter/porter/main/scripts/bundle/bundle-linux.sh | bash
+```
+
+Then extract and install Porter.
+
+```
+echo "tar -xzf porter-air-gapped-install-latest.tar.gz -C ."
+echo "cd porter-air-gapped-install-latest"
+echo "bash install-linux.sh"
 ```
 
 ## Canary
--- a/docs/content/docs/getting-started/install-porter.md
+++ b/docs/content/docs/getting-started/install-porter.md
@@ -32,6 +32,7 @@ curl -L https://cdn.porter.sh/$VERSION/install-mac.sh | bash
 ```
 
 **Linux**
+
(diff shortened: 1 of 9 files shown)


scripts/bundle/bundle-linux.sh
```diff
@@ -0,0 +1,37 @@
...
+
+mkdir -p $DOWNLOAD_DIR
+
+echo "Pulling MongoDB image..."
+docker pull "$MONGO_IMAGE_URL"
+echo "Saving MongoDB image to $MONGO_IMAGE_FILE..."
+docker save -o "$DOWNLOAD_DIR/$MONGO_IMAGE_FILE" "$MONGO_IMAGE_URL"
+echo "Downloading Porter from $PORTER_MIRROR/$PORTER_VERSION/install-linux.sh"
+curl -fsSLo "$DOWNLOAD_DIR/install-linux.sh" "$PORTER_MIRROR/$PORTER_VERSION/install-linux.sh"
+echo "Downloading Porter from $PORTER_MIRROR/$PORTER_VERSION/porter-linux-amd64"
```
kichristensen, line 25: Why is Porter being downloaded two times? Once with `install-linux.sh` and once more directly from Github releases? It also affects the logic in the `install-linux.sh` script

kichristensen (commented):
Thank you very much for the PR

scripts/bundle/bundle-linux.sh
jmcudd: Porter is only downloaded once. The install-linux.sh has been modified to only download the `porter-linux-amd64` binary if it doesn't exist locally in the current directory. Here is a break down of how `bundle-linux.sh` would be used in practice.

1. Run `curl -L https://raw.githubusercontent.com/getporter/porter/main/scripts/bundle/bundle-linux.sh | bash` in a connected environment. This will download and bundle the assets required to install porter in an offline environment.
2. Bring the `porter-air-gapped-install-1.1.0.tar.gz` file to offline environment and untar it.
3. Run `bash install-linux.sh` and it will use the locally available files to install porter and mongo. The `install-linux.sh` script will not curl anything during the install.

This process is outlined in [docs/content/docs/getting-started/install-porter.md](https://github.com/getporter/porter/pull/3205/files#diff-3c3b63308e68595f8ea0578a6616c3045495d24e39e31bd776c600461f7ad915R114-R128)

kichristensen: Ahh yes, makes sense

docs/content/docs/getting-started/install-porter.md
```diff
@@ -32,9 +32,7 @@ curl -L https://cdn.porter.sh/$VERSION/install-mac.sh | bash
 ```
 
 **Linux**
-
-```bash
-export VERSION="v1.1.0"
+```bash export VERSION="v1.1.0"
```
kichristensen, line 35: `export VERSION="v1.1.0" should be on newline, otherwise it won't show up

kichristensen (commented):
I do understand the change you are implementing here, but I'm a little concerned about the complexity it adds to the installation script.
Also preferably it should be implemented for the Powershell installation script.
Would it be possible for you to join the [community meeting](https://porter.sh/community/#dev-meeting), so we can discuss the different options here? The next community meeting is on Thursday

docs/content/docs/getting-started/install-porter.md
jmcudd: Good catch! This should be resolved now.

jmcudd:
Closed in favor of #3210

PR #3210, opened 2024-08-20
Offline Porter Installation

### LINKED ISSUE #3198: Offline Environment Error: Unable to Pull Mongo Image in Porter v1.1.0

jmcudd:
# What does this change

The new `bundle-linux.sh` script automates the creation of a tarball containing all the necessary artifacts for installing the Porter CLI, including the MongoDB Docker image and the Porter installation files. It pulls a specific version of the MongoDB image and packages all the downloaded assets into a compressed tarball, providing clear instructions for extracting the bundle and executing the installation script. This makes it easy for users to set up Porter in air-gapped environments.

The `install-linux.sh` script has been updated to local or remote binaries, to facilitate an offline installation. This is accomplished by setting the `PORTER_BINARY_URL` to a `file://<absolute-path-to-linux-binary>` instead of the Porter CDN.

# What issue does it fix
Closes #3198

# Notes for the reviewer

This PR directly addresses feedback about complexity of the installation script in PR #3205

The `bundle-linux.sh` script pulls the latest version of the `install-linux.sh` from the CDN, so when testing you will need to replace the downloaded `install-linux.sh` file with the one from this branch. Use the following command.

```
cd scripts/bundle/
PORTER_SCRIPT_URL="file://$(realpath ../install/install-linux.sh)" bash bundle-linux.sh
```

# Checklist
- [ ] Did you write tests?


kichristensen:
@jmcudd Thanks for the PR, I will do some testing locally tomorrow and get back to you. But overall I like it

scripts/bundle/bundle-linux.sh
```diff
@@ -0,0 +1,52 @@
...
+curl -fsSLo "$DOWNLOAD_DIR/porter-linux-amd64" "$PORTER_BINARY_URL" 
+
+cat << EOF > $DOWNLOAD_DIR/install-bundle.sh 
+export PORTER_BINARY_URL="file://\$(realpath ./porter-linux-amd64)" 
+export PORTER_SCRIPT_URL="file://\$(realpath ./install-linux.sh)" 
+bash install-linux.sh
+docker load -i "$MONGO_IMAGE_FILE"
+if ! docker ps -q -f name=porter-mongodb-docker-plugin; then
+  echo "Running MongoDB container..."
+  docker run --name porter-mongodb-docker-plugin -d -p 27018:27017 -v mongodb_data:/data/db --restart always mongo:4.0-xenial
```
kichristensen, line 34: Any specific reason for starting MongoDB during installation?

kichristensen:
> @kichristensen I took your feedback, in PR #3205, about the added complexity in the install script and came up with an alternative that keeps things fairly simple.
> 
> One thing to note about this is it doesn't solve installation of mixins, I'm not sure how that would be solved.

Could you update the documentation with a note, saying the mixins aren't handled by the script? Just to make it clear for users that mixins aren't currently handled

kichristensen:
I also think it would make a lot if sense to add the script to the release just like we do with `install-linux.sh` (https://github.com/getporter/porter/blob/main/magefile.go#L491), and adding the Netlify redirect (like this one https://github.com/getporter/porter/blob/main/docs/static/_redirects#L64). That would allows to have the script versioned like the rest of the installation scripts.

scripts/bundle/bundle-linux.sh
jmcudd: Yes, this specifically addresses issue #3198.

jmcudd:
Here is what I've addressed in my latest changes:
- Added note in docs about mixins not being included in the bundle
- Updated release to include the bundle-linux.sh script and update the PORTER_VERSION with sed
- Updated netlify to do the redirects

scripts/bundle/bundle-linux.sh
kichristensen: Do you think it would make sense to mention that it starts MongoDB in the documentation? It feels a little unexpected.
Side question: If MongoDB is not running next time porter is run, will it work when the container have been started once, or will it require a manual start of the container?

jmcudd: A manual restart may be required, but only if the docker container is stopped or destroyed. Mongo is deployed with `--restart always` so hopefully it won't be an issue that arises that often.

jmcudd: I'll update the docs to explain mongo, also I'll see what the behavior is currently on `main` branch when you stop mongo and run porter a second time.

kichristensen: @jmcudd I know we all are busy, but was wondering if you had any update on this?

schristoff:
This looks great, aside from Kim's documentation request. Thank you :)

docs/content/docs/getting-started/install-porter.md
```diff
@@ -114,6 +115,29 @@ You will need to create a [PowerShell Profile][ps-link] if you do not have one.
...
+tar -xzf porter-air-gapped-install-latest.tar.gz -C .
+cd porter-air-gapped-install-latest
+bash install-bundle.sh
+```
+
+Note: The `install-bundle.sh` script starts mongo. If mongo ever has to be restarted use the following command.
+
+```
+docker run --name porter-mongodb-docker-plugin -d -p 27018:27017 -v mongodb_data:/data/db --restart always mongo:4.0-xenial
+```
```
jmcudd, line 138: I've added a note here about mongo and how to restart it if it is stopped for any reason.

kichristensen: Thanks looks very good

kichristensen (approved):

PR #3215, opened 2024-09-01
Support for version number validation against mixins

dgannon991:
# What does this change
Added a check on version numbers if specified in your porter YAML

# What issue does it fix
Delivers towards #3136 

# Notes for the reviewer
@kichristensen - this is just a draft to make sure I'd got the right idea, before adding the checks into the linting side. I was fairly sure I'd understood, but thought it was worth double checking as it's been a little while :)

### DIFF
--- a/pkg/manifest/manifest.go
+++ b/pkg/manifest/manifest.go
@@ -640,8 +640,30 @@ func (l Location) IsEmpty() bool {
 }
 
 type MixinDeclaration struct {
-	Name   string
-	Config interface{}
+	Name    string
+	Version *semver.Constraints
+	Config  interface{}
+}
+
+func extractVersionFromName(name string) (string, *semver.Constraints, error) {
+	parts := strings.Split(name, "@")
+
+	// if there isn't a version in the name, just stop!
+	if len(parts) == 1 {
+		return name, nil, nil
+	}
+
+	// if we somehow got more parts than expected!
+	if len(parts) != 2 {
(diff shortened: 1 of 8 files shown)


kichristensen:
@dgannon991 Yes the format looks like what I thought about. Didn't really think about adding the full semver constraint, but I think it makes sense, and it would allow to later implement logic that could automatically download the right version if needed

dgannon991:
Nice. I’ll try and get the other half of it down for Thursday. Cheers!

pkg/manifest/manifest.go
```diff
@@ -676,7 +711,12 @@ func (m *MixinDeclaration) UnmarshalYAML(unmarshal func(interface{}) error) erro
 	}
 
 	for mixinName, config := range mixinWithConfig {
-		m.Name = mixinName
+		name, version, err := extractVersionFromName(mixinName)
```
kichristensen, line 714: Won't the name and version already have been set at line 690? Is it needed to also set it here?

kichristensen (commented):
Looks really good, added a few small comments

pkg/manifest/manifest.go
dgannon991: Ah sadly yes it is, there are two different logic branches (I think!) One for if you just specify a mixin name, like:
```yaml
mixins:
  - exec
  - terraform
```

And another if you provide config for the mixin, like:
```yaml
mixins:
  - terraform:
    clientVersion: 1.3.0
```

kichristensen (approved):

PR #3225, opened 2024-09-24
Treat object parameters as objects in templating

### LINKED ISSUE #1508: Direct way of consuming parameters/credentials as JSON objects?

lbergnehr:
# What does this change
This change makes it possible to access a parameter in templating, not only to the top-level, i.e. the parameter itself, but also -- if the parameter is of type "object" -- to access properties of the object value. For example:

```yaml
parameters:
  - name: foo
    type: object
    default:
      a: 1
      b: 2

install:
  mymixin:
    some_property: ${ bundle.parameters.foo.a }
```

In this case, the `.a` was not possible before as `foo` was merely a JSON string at the point of template resolution.

# What issue does it fix
Closes #1508 (partly?)

# Notes for the reviewer
This does not solve the same issue for credentials. I'm not sure if that scenario is as common as with parameters though.

I wasn't sure how to handle `sensitive` along with a parameter of type object. Right now it's required for the type to be `string` is `sensitive` is to work. What would be a good way to handle it? Serialize the object to a string again and then sending it to the sensitive machinery?

# Checklist
- [x] Did you write tests?

### DIFF
--- a/pkg/runtime/runtime_manifest.go
+++ b/pkg/runtime/runtime_manifest.go
@@ -6,6 +6,7 @@ import (
 	"compress/gzip"
 	"context"
 	"encoding/base64"
+	"encoding/json"
 	"fmt"
 	"io"
 	"os"
@@ -131,15 +132,27 @@ func (m *RuntimeManifest) loadDependencyDefinitions() error {
 	return nil
 }
 
-func (m *RuntimeManifest) resolveParameter(pd manifest.ParameterDefinition) string {
+func (m *RuntimeManifest) resolveParameter(pd manifest.ParameterDefinition) (interface{}, error) {
+	getValue := func(envVar string) (interface{}, error) {
+		value := m.config.Getenv(envVar)
+		if pd.Type == "object" {
+			var obj map[string]interface{}
+			if err := json.Unmarshal([]byte(value), &obj); err != nil {
+				return nil, err
+			}
+			return obj, nil
+		}
+		return value, nil
+	}
+
 	if pd.Destination.EnvironmentVariable != "" {
-		return m.config.Getenv(pd.Destination.EnvironmentVariable)
+		return getValue(pd.Destination.EnvironmentVariable)
 	}
 	if pd.Destination.Path != "" {
-		return pd.Destination.Path
+		return pd.Destination.Path, nil
 	}
 	envVar := manifest.ParamToEnvVar(pd.Name)
-	return m.config.Getenv(envVar)
+	return getValue(envVar)
 }
 
 func (m *RuntimeManifest) resolveCredential(cd manifest.CredentialDefinition) (string, error) {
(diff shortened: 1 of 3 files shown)


kichristensen (changes requested):
@lbergnehr Thank you very much for the PR. Will take a closer look.
Quickly skimming through the related issue, could you add a test that verifies this PR won't change the existing behaviour (unless such a test already exist)? See https://github.com/getporter/porter/issues/1508#issuecomment-1157910618

lbergnehr:
> @lbergnehr Thank you very much for the PR. Will take a closer look. Quickly skimming through the related issue, could you add a test that verifies this PR won't change the existing behaviour (unless such a test already exist)? See [#1508 (comment)](https://github.com/getporter/porter/issues/1508#issuecomment-1157910618)

Thanks @kichristensen! I added the test and also had to change the implementation to make it pass (great test! :)). There's a caveat in that nested properties will not be interpolated as JSON, but the actual parameter object will (see commit message for more info).

(FYI: the force push was only to add my signature.)

pkg/runtime/runtime_manifest.go
```diff
@@ -271,9 +297,12 @@ func (m *RuntimeManifest) buildSourceData() (map[string]interface{}, error) {
...
 
 		pe := param.Name
-		val := m.resolveParameter(param)
+		val, err := m.resolveParameter(param)
+		if err != nil {
+			return nil, err
+		}
 		if param.Sensitive {
-			m.setSensitiveValue(val)
+			m.setSensitiveValue(val.(string))
```
lbergnehr, line 305: One possible solution to only handling `string`s would be to do a `fmt.Sprint(val)` here, since that will invoke the `Format` function of object-type parameters and make them "strings".

kichristensen:
Thanks for the update @lbergnehr. I will double-check that the behaviour is the same as in the current version, mean while could you take a look at the failing ["Vet and Lint"](https://github.com/getporter/porter/actions/runs/11075975800/job/30778059446?pr=3225)?

lbergnehr:
> Thanks for the update @lbergnehr. I will double-check that the behaviour is the same as in the current version, mean while could you take a look at the failing ["Vet and Lint"](https://github.com/getporter/porter/actions/runs/11075975800/job/30778059446?pr=3225)?

Thanks for having a look @kichristensen! I think I fixed the linting error (although I couldn't run `mage lint` locally to verify).

kichristensen (approved):

pkg/runtime/runtime_manifest.go
kichristensen: Will this mean that object properties cannot be marked as sensitive, and still work? Could you add a test that it works for sensitive values?

schristoff (approved):

pkg/runtime/runtime_manifest.go
kichristensen: Ok, that makes sense. Could you create an issue for this so we don't forget?
And thank you very much for the work on this

lbergnehr: Created https://github.com/getporter/porter/issues/3227.

kichristensen: Thank you very much

PR #3340, opened 2025-02-21
Support complex types as parameter set values

### LINKED ISSUE #3331: Parameter sets `value` only supports `string` type
If a bundle takes a parameter that is not a string type, e.g. `object` or `integer`, that is the type that should be possible to enter in the parameter set definition. Currently, the value gets blanked out if the type is anything else than a string.

That being said, if the value is "converted" to a string, e.g. creating a string json representation, then it still works as a parameter since Porter will convert it when setting the parameter on the bundle. This is not very intuitive though, especially when dealing with numbers, which are likely to get entered verbatim (without quotes) and thus will get blanked out.

Support any type supported by Porter parameters in the `value` property of a parameter in parameter sets, for example:

```yaml
...
schemaType: ParameterSet
schemaVersion: 1.0.1
namespace: ""
name: omni-overlord-hive-ohive
parameters:
  - name: some-object
    source:
      value:
          some-prop: 1
status:
  created: "0001-01-01T00:00:00Z"
  modified: "2025-02-05T15:20:52.633252075+01:00"
```

In the example above, the object set to `value` is an object with a property `some-prop`.

Currently, the workaround to this is setting the value like this:

```yaml
...
  - name: some-object
    source:
      value: |
          { "some-prop": 1 }
...
```

I think the implementation could be as simple as converting the `value` objects to strings in the code, which would not require a lot of refactoring since the string type would be type same in the rest of the code.

A better, but more involved way, might be to actually treat the `value` as any type in all related parts of the code.

erikced:
# What does this change
This change updates the parameter sets to support `number`, `object`, `array` and `boolean` values for `value` sources, e.g.
```yaml
schemaType: ParameterSet
schemaVersion: 1.1.0
name:  my-params
parameters:
  - name: some-param
    source:
      value: 
        a: [1, 2, 3, 4]
        b: abcd
  - name: some-other-param
      source:
        value: true
```
so that the parameter set value can be of the same type as the bundle parameter.

Currently this is only supported by manually converting the values to their JSON representation, which is less intuitive
```yaml
  - name: some-param
    source: 
      value: '{"a": [1, 2, 3, 4], "b": "abcd"}'
  - name: some-other-param
      source:
        value: "true"
```

It is implemented by updating the unmarsalling code for `Source` so that when parameter with the `value` strategy is encountered and the hint is not a `string`, it is converted to its corresponding JSON string representation.
# What issue does it fix
Closes #3331

# Notes for the reviewer
Since the unmarshalling - and hence value transformation - is performed before the parameter set is stored in the parameter store the value stored will not necessarily match the input parameter set when running `porter parameters apply`, which might not be desirable(?). Even though porter will not itself write values that are not strings to the parameter store, it can read such values from the store as the BSON is converted to JSON and then unmarshalled in the same way as parameter sets from file.

# Checklist
- [x] Did you write tests?


pkg/secrets/strategy.go
```diff
@@ -69,6 +69,13 @@ func (s *Source) UnmarshalRaw(raw map[string]interface{}) error {
...
 				s.Hint = value
+
+			} else if s.Strategy == "value" {
+				value, err := unmarshalRawValue(v)
+				if err != nil {
+					return err
+				}
+				s.Hint = value
 			} else {
 				s.Hint = fmt.Sprintf("%v", s.Hint)
```
erikced, line 80: I am a bit confused by this, as the hint should be an empty string at this point?

kichristensen:
@erikced Thanks for the PR. This is not the part the I'm most familiar with, maybe @schristoff has some more insights. But I have a few questions.
Will the old complex "stringified" parameters still work, when passing them through the commandline? And what will be the expected way to pass complex types from the command line?
It would be great if you also could update the documentation to describe this functionality?
To make me more certain that this works end-to-end I would really like to see a simple integration test or two for covering this change, would that be possible?

kichristensen:
@erikced Looks good to me. Am I correct when I think the new format will technically also work in version 1.0.1 of the parameter file, but it will fail  schema validation if it is validated?

erikced:
> @erikced Looks good to me. Am I correct when I think the new format will technically also work in version 1.0.1 of the parameter file, but it will fail schema validation if it is validated?

@kichristensen yes, that should be the case.

As a side note, I am not entirely certain whether the version should be 1.1.0 or 1.0.2, from a semver-like perspective it should probably be 1.1.0 since it is an extension of the functionality and not a patch (whatever that may be for a document) but other document versioning standards would probably treat it as 1.0.2 since all valid 1.0.0/1.0.1 documents are still valid.

schristoff:
I like `1.1.0` 👍

PR #2890, opened 2023-08-29
add v prefix

ludfjig:
# What does this change
`porter explain -r getporter/hello-llama:0.1.1` now prepends a `v` to the version (0.1.1), so that it will find the bundle if it is tagged v1.2.3 in the registry.

# What issue does it fix
Closes #2886 

[1]: https://getporter.org/src/CONTRIBUTING.md#when-to-open-a-pull-request

# Notes for the reviewer
_Put any questions or notes for the reviewer here._

# Checklist
- [x] Did you write tests?
- [ ] Did you write documentation?
- [ ] Did you change porter.yaml or a storage document record? Update the corresponding schema file.
- [ ] If this is your first pull request, please add your name to the bottom of our [Contributors][contributors] list. Thank you for making Porter better! 🙇‍♀️

# Reviewer Checklist
* Comment with /azp run test-porter-release if a magefile or build script was modified
* Comment with /azp run porter-integration if it's a non-trivial PR

[contributors]: https://getporter.org/src/CONTRIBUTORS.md

Issue #1431, opened 2021-01-28
porter build --version doesn't accept semver with commit hash properly


squillace:
## Describe the bug
porter is documented as accepting a SemVer --version option. The format to append a commit hash would be from https://github.com/Masterminds/semver might look like: `porter build --version 0.1.7+$(git rev-list --max-count=1 HEAD)`. 

However, the output is:
```
build --version 0.1.7+$(git rev-list --max-count=1 HEAD)
Error: could not set bundle tag to "v0.1.7+58d98af56c3a4c40c69535654216bd4a1fa701e7": invalid tag format
```

## To Reproduce
1. Run `porter build --version 0.1.7+$(git rev-list --max-count=1 HEAD)`.
2. See error

## Expected behavior
Porter should build and tag the bundle and installer with the complete version string.

## Version
Copy the output of `porter version` below

porter v0.32.0 (8e03c7da)

vdice:
Thanks for the issue, @squillace .  My thinking is that Porter should sanitize the provided semver `version` value to a valid [Docker tag](https://docs.docker.com/engine/reference/commandline/tag) when tagging the invocation image and bundle.

For example, if the bundle version is: `0.1.0-beta.1+abcd1234`, the image/bundle tags could be sanitized to comply with Docker tag restrictions.  In this case, `0.1.0-beta.1-abcd1234` would comply.  (`+` is not allowed, but `-` and `_` are).

carolynvs:
I like Vaughn's suggestion to default the tag to a sanitized version string when necessary, and printing a warning when we do that. We should also clearly document the limitations of docker tags on our website and on the existing blog post about using CI/C with these flags.

We should probably research a bit on what replacement character to use, in case there's already an accepted workaround for this semver -> docker tag problem.

squillace:
related to #1432 as well

MikkelHegn:
I'm running into this as well currently. What's the reasoning behind making semver a requirement, as opposed to just a recommendation?

I'm considering only using the commit id to identify my bundle.

vdice:
> What's the reasoning behind making semver a requirement, as opposed to just a recommendation?

Good question.  This originates from the CNAB Spec for the underlying bundle.json that each Porter bundle is represented as.  Currently, the spec states that the version field [must be a SemVer2 string](https://github.com/cnabio/cnab-spec/blob/master/101-bundle-json.md#name-and-version-identifying-metadata).

squillace:
[...]

PR #3348, opened 2025-03-10
Only add `v` prefix to semver tags

erikced:
# What does this change
#2890 added implicit `v` prefixes to all image tags if they were missing, #2915 adds a few exceptions and this MR changes the behavior to only prefix actual semver tags with `v`, so that custom tags are not altered. Tags `1.2.3` and `1.2.3-pre_build` will be converted to `v1.2.3` and `v1.2.3-pre_build` respectively, whereas `latest`, `v1.2.3` or `my-branch-build` will not be altered.

Interally support for buildmetadata has also been added to `OCIReference` and it has been made responsible for version<->tag mapping where applicable. Tests related to `ensureVPrefix` have also been simplified.

# What issue does it fix
Closely related to #2914, also touches on the implementation of #1431.

# Notes for the reviewer

# Checklist
- [x] Did you write tests?

### DIFF
--- a/pkg/cnab/oci_reference.go
+++ b/pkg/cnab/oci_reference.go
@@ -163,7 +163,7 @@ func (r OCIReference) HasVersion() bool {
 	}
 
 	if tagged, ok := r.Named.(reference.Tagged); ok {
-		_, err := semver.NewVersion(tagged.Tag())
+		_, err := semverFromTag(tagged.Tag())
 		return err == nil
 	}
 	return false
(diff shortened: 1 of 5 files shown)


kichristensen:
Overall it LGTM and it does make sense, but it will be a breaking change, maybe adding a parameter and/or config value would make sense such that it can be turned on or off. What do you think @schristoff?

schristoff (approved):

Issue #2019, opened 2022-04-13
Support --cache-from


carolynvs:
When building in CI, where the cache is starting from scratch, builds are slow. It would be great if we supported the --cache-from flag and be able to point to a registry to grab layers from and speed things up.

I'm not exactly sure what we want to support, or if there are differences between docker build --cache-from and what we can do with buildx+buildkit. So whoever picks this up will need to do research first!

https://docs.docker.com/engine/reference/commandline/build/#specifying-external-cache-sources
https://github.com/docker/build-push-action/blob/master/docs/advanced/cache.md

kichristensen:
Implemented in #3371

PR #3371, opened 2025-05-12
Expose more buildkit build options

erikced:
# What does this change
This change introduces a number of new build flags, `--cache-to`, `--cache-from`, `--builder` and `--output` which map to the corresponding `docker buildx build` / buildkit options. `--output` is restricted compared to its `docker build` counterpart in that it does not allow overriding the `type` and `name` parameters as those must be controlled by porter for builds to be successful.

It also changes default builder selection slightly to use the default builder for the current docker context instead of the builder with the same name as the current context and enables the `docker-container` buildkit driver. By using the `docker-container` build driver instead of the default `docker` driver it is possible to export cache to the repo directly instead of storing it inline in the built image.

With these changes it is possible to use caching and custom builders, e.g.
```bash
porter build --cache-to type=inline --cache-from my-repo/bundle:latest --output compression=estargz,force-compression=true --builder my-container-builder 
```

# What issue does it fix
n/a

# Checklist
- [x] Did you write tests?

### DIFF
--- a/cmd/porter/bundle.go
+++ b/cmd/porter/bundle.go
@@ -93,6 +93,8 @@ The docker driver builds the bundle image using the local Docker host. To use a 
 		"Define additional build context with specified contents (format: NAME=PATH). May be specified multiple times.")
 	f.StringArrayVar(&opts.CacheFrom, "cache-from", nil,
 		"Add cache source images to the build cache. May be specified multiple times.")
+	f.StringArrayVar(&opts.CacheTo, "cache-to", nil,
+		"Add cache target images to the build cache.")
 	f.StringArrayVar(&opts.SSH, "ssh", nil,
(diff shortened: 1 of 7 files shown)


kichristensen:
Thank you for the PR @erikced.
Is it possible to add some integration tests that verifies it works as expected?
On another note, during publish a bundle will automatically be built, unless the `autobuild-disabled` flag is provided. Not a lot of parameters are available to control the auto build behaviour on the publish command, do you think it make sense to also add them to publish in case of the the bundle is rebuilt?

kichristensen:
@erikced Do you have time to fix the conflicts?
It would also solve #2019

erikced:
@kichristensen this should be ready now, please have a look.

PR #3372, opened 2025-05-12
Fix resolution of file parameters

### LINKED ISSUE #3349: Porter always resolves file parameters using the `value` strategy
## Describe the bug
When resolving file parameters, the source strategy (i.e. type) e.g. `env`, `value` etc is ignored and porter always assumes that the provided hint is either a path to a file or a base64-encoded string with the file contents.
## To Reproduce
Use the following bundle
```yaml
schemaType: Bundle
schemaVersion: 1.0.1
name: porter-file-param-test
version: 0.1.0
registry: localhost:5000
mixins: [exec]
parameters:
  - { name: data-file, type: file, path: data-file, applyTo: [install] }
  - { name: data-string, type: string, path: data-string, applyTo: [install] }
install:
  - exec: { description: "Install", command: /usr/bin/bash, arguments: [-c, 'printf "data-file: %s\ndata-string: %s\n" "$(cat data-file)" "$(cat data-string)"'] }
uninstall:
  - exec: { description: "Uninstall", command: /bin/true }
```
and parameter set
```yaml
parameters:
  - { name: data-file, source: { env: DATA } }
  - { name: data-string, source: { env: DATA } }
```
1. Create example data files: `echo this is DATA > DATA && echo this is my-data > my-data`
2. Apply the parameter set `porter apply parameters.yaml`
3. Install the bundle `DATA=my-data porter install --parameter-set test test`
4. Note that the string data has been resolved to the value of `DATA` (`my-data`) whereas the file data has been read from a file with the same name as the environment variable.
```
❯ DATA='my-data' porter install --parameter-set=test test --force
Just-in-time resolving credentials...
Just-in-time resolving parameters...
executing install action from porter-file-param-test (installation: /test)
Install
data-file: this is DATA
data-string: my-data
```

## Expected behavior
Porter resolves the value using the provided strategy, then checks if it is the path to a local file or base64-encoded data.

## Version
porter v1.2.1

erikced:
Remove parameter resolution short-cut in `loadParameterSets` so that the name (or value) of file parameters are resolved in the same way as other parameters, unless the parameter strategy is `path`. Later, in `finalizeParameters`, 
 getUnconvertedValueFromRaw` will perform special handling to either load the file or pass the data along assuming it is base64 encoded. 

# What does this change
This makes porter first resolve the parameter value as usual for all parameter types except `path`, then check whether the resolved value is a file path instead of assuming the value is a path (essentially ignoring the parameter resolution strategy).

# What issue does it fix
Closes #3349 

# Notes for the reviewer

# Checklist
- [x] Did you write tests?

### DIFF
--- a/pkg/porter/parameters.go
+++ b/pkg/porter/parameters.go
@@ -23,6 +23,7 @@ import (
 	dtprinter "github.com/carolynvs/datetime-printer"
 	"github.com/cnabio/cnab-go/bundle"
 	"github.com/cnabio/cnab-go/bundle/definition"
+	"github.com/cnabio/cnab-go/secrets/host"
 	"github.com/olekukonko/tablewriter"
 	"go.mongodb.org/mongo-driver/bson"
 )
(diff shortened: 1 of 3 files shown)


kichristensen:
Thank you for the PR @erikced.
Would it be possible to write a test for this functionality? Could be similar to the description in the issue if possible.

erikced:
> Thank you for the PR @erikced. Would it be possible to write a test for this functionality? Could be similar to the description in the issue if possible.

Sure, I will also check what is going on with the failing integration test.

erikced:
Restored the behaviour for parameters using the `path` strategy so that porter does not attempt to load the file, then use that data as input to the file loading in `getUnconvertedValueFromRaw`. Also updated the integration test to take a file name from an environment variable as well.

kichristensen (approved):

PR #3407, opened 2025-07-08
Initial draft of mongo healthcheck

dgannon991:
@kichristensen - this is a really rough sketch of what we could do. Essentially add healtchecks and wait for status. It might be nicer to use the docker library to do this, but it's getting late here so I wanted to show /something/ :D


kichristensen:
@dgannon991  Looks like a fair approach to me. Looks a but convoluted, but if it works I think it is good for for now, we can improve later. Does it affect the startup performance a lot?

dgannon991:
@kichristensen  - this is the best version I can come up with, it's still a little more complex, but I think a chunk of that is that the method could do with a heavier refactor if I'm being honest!

PR #3431, opened 2025-08-12
feat: create state file parent directory if it doesn't exist

### LINKED ISSUE #3430: Porter does not create state files' parent dir
## Describe the bug
Given state file, e.g. `/path/to/some_state` porter does not create the parent directory prior to attempting to restore the state file resulting in `error creating state file /path/to/some_state: open /path/to/some_state: no such file or directory`.

## To Reproduce
1. Use the porter.yaml from [this gist](https://gist.github.com/erikced/f04953ac9071d836b5cc64fe9947d256)
2. Run `porter install state-path-test`
3. Run `porter upgrade state-path-test` (or `install --force`)

## Expected behavior
Porter creates the necessary parent directories to be able to inject the state.

## Version
porter v1.3.0

kichristensen:
# What does this change
This commit introduces a new test, `TestStateBagUnpackWithParentDirectoryCreation`, which verifies that the parent directory for state files is created correctly during the unpacking process. Additionally, it includes a helper function, `createTestStateArchive`, to facilitate the creation of tar.gz archives for testing.

Changes in `runtime_manifest.go` ensure that the parent directory is created before state files are written, enhancing the robustness of the state management functionality.

# What issue does it fix
Closes #3430

# Notes for the reviewer

# Checklist
- [X] Did you write tests?

### DIFF
--- a/pkg/runtime/runtime_manifest.go
+++ b/pkg/runtime/runtime_manifest.go
@@ -630,7 +630,14 @@ func (m *RuntimeManifest) unpackStateBag(ctx context.Context) error {
 		dest := stateFiles[name]
 		m.debugf(log, "  - %s -> %s", name, dest)
 
-		f, err := os.OpenFile(dest, os.O_CREATE|os.O_RDWR, os.FileMode(header.Mode))
+		// Ensure parent directory exists before creating the file
+		parentDir := filepath.Dir(dest)
+		if err := m.config.FileSystem.MkdirAll(parentDir, pkg.FileModeDirectory); err != nil {
+			return log.Error(fmt.Errorf("error creating parent directory for state file %s: %w", dest, err))
+		}
+
+		// Create the file using the file system abstraction instead of os.OpenFile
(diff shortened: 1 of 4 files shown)


kichristensen:
> You beat me to it, thanks 🙂 A small note is that `afero.MemMapFs` which is used when creating the unit test config/context creates parent dir structures on demand, i.e. it's hard to create a failing test case since using an actual file system is not possible either. The test still serves a purpose though as there is no previous unit test that validates that the state file contents are unpacked to the expected destination.

That is a good point. The integration test is the "real" test here, ensuring that it actually works :)
Now I'm wondering what the unit test is actually testing, as we already have the test `TestStateBagUnpack`. Maybe the test doesn't test anything we didn't already test? Any opinions on that?

erikced:
> Maybe the test doesn't test anything we didn't already test?

It does! `TestStateBagUnpack` only tests three different empty or invalid state files, the new test has an actual archive and tests that a state file is extracted to its desired location, I definitely think it adds value.

PR #3459, opened 2025-09-22
Create an SBOM when publishing

### LINKED ISSUE #2930: (feat): Generate SBOM for invocation image + bundle
Being able to offer generating an SBOM for the invocation image and bundle would be a great step in our security focus.
Talking to Syft on their Slack, they pointed us to [this example](https://github.com/anchore/grype/blob/main/grype/pkg/syft_provider.go#L11) for ingesting Syft as a Go lib.

Ideally, when users run `porter publish` command we would take a `sbom` (bool) flag. If true, then within `pkg/porter/publish` we would call a separate `pkg/porter/publish/sbom.go` that would do the generation required. We could separate out this feature into it's own package, or keep it within publish.  The porter package is already kinda chunky, so maybe best to try and keep all the imports needed for this feature separate. 

Note: The linked code generates a Syft SBOM, but we need an SPDX SBOM, so we would need to incorporate the use of [formats](https://github.com/anchore/syft/blob/main/syft/formats/formats.go) into this code.

dgannon991:
# What does this change
Adds a flag onto the publish command which will generate an SBOM for the published image.

# What issue does it fix
Closes #2930 

# Notes for the reviewer
This is currently an extreme draft version to get feedback, none of the following have been done :D

# Checklist
- [ ] Did you write tests?


kichristensen:
@dgannon991 Does the SBOM also handled other images part of the bundle, in the `images` section? If not I think it is fine, but we should create a follow up issue for this feature.

pkg/porter/publish.go
```diff
@@ -8,6 +8,8 @@ import (
 	"path/filepath"
 	"strings"
 
+	_ "modernc.org/sqlite" // required for rpmdb and other features
```
kichristensen, line 11: Can this be moved into `sbom.go` instead? The module has an `init` function which will be triggered on import

pkg/porter/sbom.go
```diff
@@ -0,0 +1,70 @@
...
+	if err != nil {
+		return log.Errorf("failed to create directory for SBOM %s: %w", sbomPath, err)
+	}
+
+	f, err := s.fileSystem.Create(sbomPath)
+	if err != nil {
+		return log.Errorf("failed to create SBOM file %s: %w", sbomPath, err)
+	}
+	defer func() {
+		err = errors.Join(err, f.Close())
```
kichristensen, line 56: Do you want this error to be returned from the function? If so have you tested that, I don't think this is the error that will be returned. I'm pretty sure the deferred function is ran after the return value have been created, meaning this isn't the error that will be returned. Not a 100% sure

kichristensen (commented):
How big an impact does this feature have on the size the porter binary? A lot of new dependencies have been pulled in.

docs/content/docs/references/cli/publish.md
```diff
@@ -43,6 +43,7 @@ porter publish [flags]
       --preserve-tags        Preserve the original tag name on referenced images
   -r, --reference string     Use a bundle in an OCI registry specified by the given reference.
       --registry string      Override the registry portion of the bundle reference, e.g. docker.io, myregistry.com/myorg
+      --sbom-file string     Generate a Software Bill of Materials (SBOM) for the bundle at this location.
```
kichristensen, line 46: Should we mention the SBOM file format here?

PR #3487, opened 2025-11-18
Object file parameter

### LINKED ISSUE #3257: Object parameters cannot be read from a file
## Describe the bug
According to
https://github.com/getporter/porter/blob/f61648fcb851ab4bf6d062b593b3d5ccd6e91bdb/docs/content/docs/bundle/manifest/_index.md?plain=1#L182,
an object parameter should, much like the `file` type parameter, be able to be
taken from a file directly when specified using the `--param` option in e.g.
`porter install`.

## To Reproduce
1. Use this porter.yaml:

   ```yaml
   schemaType: Bundle
   schemaVersion: 1.0.1
   name: porter-repro
   version: 0.1.0
   registry: "localhost:5000"

   # Declare and optionally configure the mixins used by the bundle
   mixins:
     - exec

   parameters:
     - name: object_parameter
       type: object
       path: /cnab/app/object_parameter.json

   # Define the steps that should execute when the bundle is installed
   install:
     - exec:
         command: echo
         arguments:
           - install ${ bundle.parameters.object_parameter }

   # Define the steps that should execute when the bundle is upgraded
   upgrade:
     - exec:
         description: "World 2.0"
         command: ./helpers.sh
         arguments:
           - upgrade

   # Define the steps that should execute when the bundle is uninstalled
   uninstall:
     - exec:
         description: "Uninstall Hello World"
         command: ./helpers.sh
         arguments:
           - uninstall
   ```

1. Put the following in `object_parameter.json`:

   ```json
   {
     "key": "value"
   }
   ```
1. Run `porter install --param object_parameter=./object_parameter.json`
1. Observe the error:

   ```
   unable to convert parameter's object_parameter value ./object_parameter.json to the destination parameter type object: could not unmarshal value ./object_parameter.json into a json object: invalid character '.' looking for beginning of value
   ```

## Expected behavior
The file should be read and marshalled into the object parameter
representation if the parameter given in the `--param` option is pointing to a
file.

## Version

```
porter v1.2.0 (45c17c15)
```

kichristensen:
# What does this change

This change adds support for loading object-type parameters from JSON files using the `@` prefix syntax. When passing object parameters via CLI or parameter sets, users can now reference an external JSON file instead of providing inline JSON.

**Example usage:**

```bash
# Install with object parameter from file
porter install --param config=@config.json

# Upgrade with object parameter from file
porter upgrade --param config=@config.json

# Use with parameter sets
porter install --parameter-set myparam
```

Where a parameter set might contain:
```yaml
parameters:
  - name: config
    source:
      value: "@/path/to/config.json"
```

**Key features:**
- Only works for parameters defined with `type: object` in the bundle
- Validates that the file contains valid JSON before processing
- Works with both `--param` CLI flags and parameter set `value` sources
- Security: Only applies to user-provided values, not default values in bundle definitions
- Provides clear error messages when files are not found or contain invalid JSON

# What issue does it fix

Closes #3257

# Notes for the reviewer

- The `@` prefix logic is implemented in `pkg/porter/parameters.go:getUnconvertedValueFromRaw()` 
- Security consideration: The `@` prefix only works for user-provided parameter values (CLI/parameter sets), not for default values defined in bundle manifests, preventing potential security issues with malicious bundles
- The implementation validates JSON syntax immediately after reading the file to provide clear error messages
- Non-object parameters ignore the `@` prefix, treating it as a literal string value for backward compatibility

# Checklist
- [x] Did you write tests?

### DIFF
--- a/.github/workflows/porter-integration-pr.yml
+++ b/.github/workflows/porter-integration-pr.yml
@@ -81,6 +81,11 @@ jobs:
     uses: getporter/porter/.github/workflows/integ-reuseable-workflow.yml@main
     with:
       test_name: outputs_test
+  object_param_integration_test:
+    name: Object Param Integration Test
+    uses: getporter/porter/.github/workflows/integ-reuseable-workflow.yml@main
+    with:
+      test_name: object_param_test
   publish_integration_test:
     name: Publish Integration Test
     uses: getporter/porter/.github/workflows/integ-reuseable-workflow.yml@main
--- a/.github/workflows/porter-integration-release.yml
+++ b/.github/workflows/porter-integration-release.yml
@@ -92,6 +92,12 @@ jobs:
     with:
       test_name: outputs_test
       registry: ${{inputs.registry}}
+  object_param_integration_test:
+    name: Object Param Integration Test
+    uses: getporter/porter/.github/workflows/integ-reuseable-workflow.yml@main
+    with:
+      test_name: object_param_test
+      registry: ${{inputs.registry}}
   publish_integration_test:
     name: Publish Integration Test
     uses: getporter/porter/.github/workflows/integ-reuseable-workflow.yml@main


lbergnehr:
In parameter sets, there's `source.path` in the reference documentation. Would it make sense to use that in parameter sets, where we have that ability to make the distinction? I like the `@` notation as it's commonly used elsewhere on the CLI (e.g. curl).

dgannon991 (approved):
LGTM, and the comments for why it's safe were useful, cheers!

PR #3488, opened 2025-11-18
Optimize bundle size

kichristensen:
# What does this change

This PR introduces an **experimental feature flag** `optimized-bundle-build` that reduces bundle image sizes by approximately **54%** (from ~446MB to ~204MB) by restructuring how bundles are built.

## How it works

The optimization changes the build context and eliminates duplicate layers:

**Before (legacy build):**
- Build context: project root directory
- Creates duplicate layers: copies `.cnab` directory twice (once as `.` then again as `.cnab`)
- Uses RUN commands to delete files and set permissions (creates additional metadata layers)
- Result: ~446MB bundle images

**After (optimized build):**
- Build context: `.cnab` directory
- Uses named build context `userfiles` to selectively copy user files
- Uses `COPY --chown` and `COPY --chmod` to set permissions in a single layer
- Eliminates duplicate 121MB layer from copying `.cnab` twice
- Eliminates 121MB metadata layer from `RUN chgrp/chmod` commands
- Result: ~204MB bundle images

## Enabling the feature

Set the experimental flag via environment variable:

```bash
export PORTER_EXPERIMENTAL=optimized-bundle-build
porter build
```

Or in your Porter config (`~/.porter/config.toml`):

```toml
experimental = ["optimized-bundle-build"]
```

## Migration for custom Dockerfiles

If you use a custom Dockerfile, you'll need to update the COPY syntax:

**Legacy syntax:**
```dockerfile
COPY . ${BUNDLE_DIR}
```

**Optimized syntax:**
```dockerfile
COPY --from=userfiles --link . ${BUNDLE_DIR}/
```

See updated documentation in `docs/content/docs/bundle/custom-dockerfile.md` for complete migration guide.

## Example

```bash
# Enable the feature
export PORTER_EXPERIMENTAL=optimized-bundle-build

# Build a bundle
porter build

# Check the resulting image size
docker images | grep mybundle
# Before: mybundle:v0.1.0  446MB
# After:  mybundle:v0.1.0  204MB
```

# What issue does it fix

This is an optimization enhancement that doesn't fix a specific issue. Bundle images can be unnecessarily large due to duplicate layers in the Dockerfile build process. This change provides an opt-in experimental feature to reduce image sizes significantly.

# Notes for the reviewer

- The feature is **opt-in** via experimental flag to maintain backward compatibility
- All existing tests pass in legacy mode (default behavior unchanged)
- Added comprehensive test coverage:
  - Unit tests for both build modes (`pkg/build/dockerfile-generator_test.go`)
  - Smoke tests validate both modes end-to-end (`tests/smoke/*_test.go`)
  - Custom Dockerfile handling tested with both modes
- Documentation updated to explain both build modes and migration path
- The optimization eliminates two major sources of bloat:
  1. Duplicate .cnab directory copy (121MB saved)
  2. Metadata layer from RUN chgrp/chmod (121MB saved)

**Breaking change considerations:**
- Users with custom Dockerfiles will need to update COPY syntax when enabling the flag
- Migration is straightforward and documented
- Legacy mode remains the default until feature is proven stable

# Checklist
- [x] Did you write tests?

### DIFF
--- a/pkg/build/buildkit/buildx.go
+++ b/pkg/build/buildkit/buildx.go
@@ -151,7 +151,7 @@ func (b *Builder) BuildBundleImage(ctx context.Context, manifest *manifest.Manif
 	// Use optimized build context when feature flag is enabled
 	if b.IsFeatureEnabled(experimental.FlagOptimizedBundleBuild) {
 		// Validate that user hasn't defined a "porter-internal-userfiles" named context
-		if _, exists := buildContexts["porter-internal-userfiles"]; exists {
+		if _, exists := namedContexts["porter-internal-userfiles"]; exists {
 			return span.Error(fmt.Errorf("the named context 'porter-internal-userfiles' is reserved by Porter when using the optimized-bundle-build experimental feature; please rename your build context"))
 		}
 


pkg/build/dockerfile-generator.go
```diff
@@ -167,23 +167,58 @@ func (g *DockerfileGenerator) getBaseDockerfile(ctx context.Context) ([]string,
 }
 
 func (g *DockerfileGenerator) buildPorterSection() []string {
+	if g.IsFeatureEnabled(experimental.FlagOptimizedBundleBuild) {
+		// Optimized build: Copy user files from the userfiles named context
+		copyUserFiles := "COPY . ${BUNDLE_DIR}"
+		if g.GetBuildDriver() == config.BuildDriverBuildkit {
```
dgannon991, line 173: Should we add in a validation for someone using the optimised bundle build but /not/ using the BuildKit driver?

pkg/build/buildkit/buildx.go
```diff
@@ -144,14 +145,27 @@ func (b *Builder) BuildBundleImage(ctx context.Context, manifest *manifest.Manif
 		return span.Errorf("error parsing the --output flag: %w", err)
 	}
 
+	namedContexts := toNamedContexts(buildContexts)
+	var contextPath string
+
+	// Use optimized build context when feature flag is enabled
+	if b.IsFeatureEnabled(experimental.FlagOptimizedBundleBuild) {
+		// Build from .cnab directory as context, add parent dir as named context for user files
+		namedContexts["userfiles"] = buildx.NamedContext{Path: b.Getwd()}
```
dgannon991: Does this if a user had a namedContext called "userfiles" it would be overwritten? Do we need validation on this?

pkg/build/dockerfile-generator.go
kichristensen: Good question. Actually the only build driver supported is Docker, https://github.com/getporter/porter/blob/86145eb060600340cdaa4d8efe2a90284d9e81f7/pkg/porter/porter.go#L159-L160
So technically this is not possible

pkg/build/buildkit/buildx.go
kichristensen: Good point, I have added validation and renamed the context, to reduced the chance for collisions

pkg/build/buildkit/buildx.go
```diff
@@ -144,14 +145,32 @@ func (b *Builder) BuildBundleImage(ctx context.Context, manifest *manifest.Manif
 		return span.Errorf("error parsing the --output flag: %w", err)
 	}
 
+	namedContexts := toNamedContexts(buildContexts)
+	var contextPath string
+
+	// Use optimized build context when feature flag is enabled
+	if b.IsFeatureEnabled(experimental.FlagOptimizedBundleBuild) {
+		// Validate that user hasn't defined a "porter-internal-userfiles" named context
+		if _, exists := buildContexts["porter-internal-userfiles"]; exists {
```
dgannon991:
```suggestion
		if _, exists := namedContexts["porter-internal-userfiles"]; exists {
```
Should this be checking namedContexts instead of buildContexts?

PR #3517, opened 2025-12-13
Skip build when bundle is up-to-date

### LINKED ISSUE #690: Skip build when nothing has changed
I am building a GUI experience around `porter install`.  As part of that, I need to parse the parameters out of the bundle to display an editor UI.  Very well, you say, you have the `porter.yaml`, parse away.  BUT WAIT!  Porter also supports installing a bundle from an OCI registry!  And that will not have a `porter.yaml` but only a `bundle.json`.  And, lazy as I am, I don't want to write two sets of parsing code.

Now I could work around this by doing a `porter build` to get a `.cnab/bundle.json` and parse that.  But then I incur an unnecessary and lengthy build step if the output was already up to date.  Porter knows enough to skip unnecessary builds when doing `install`, but it doesn't currently provide a way for _me_ to say 'build if you need to (but not if you don't)'.  It would be a useful enhancement to surface this conditional build logic through the API.  Thanks!

kichristensen:
# What does this change

This PR adds conditional rebuild logic to the `porter build` command. By default, `porter build` now checks if the bundle is up-to-date before building. If nothing has changed, it skips the build and displays "Bundle is up-to-date!". This saves time when repeatedly running build commands during development.

A new `--force` flag was added to force a full rebuild from scratch, ignoring any cached data. When `--force` is specified, it also sets `--no-cache` for the docker build.

**Example usage:**

```bash
# First build - builds normally
$ porter build
Building bundle ===>
...

# Second build with no changes - skips build
$ porter build
Bundle is up-to-date!

# Force rebuild despite no changes
$ porter build --force
Building bundle ===>
...
```

# What issue does it fix

Closes #690

# Notes for the reviewer

- The conditional build logic reuses the existing `IsBundleUpToDate()` function from `pkg/porter/stamp.go`
- When `--force` is specified, it automatically enables `--no-cache` to ensure a full rebuild
- Two integration tests were added:
  - `TestBuild_SkipWhenUpToDate` - verifies build is skipped when bundle unchanged
  - `TestBuild_ForceRebuild` - verifies `--force` triggers rebuild

# Checklist
- [x] Did you write tests?

### DIFF
--- a/tests/integration/build_test.go
+++ b/tests/integration/build_test.go
@@ -352,4 +352,8 @@ func TestBuild_ForceRebuild(t *testing.T) {
 	// Build again with --force, should rebuild
 	_, output := test.RequirePorter("build", "--force", "--name=porter-test-force-rebuild")
 	require.NotContains(t, output, "Bundle is up-to-date!", "build with --force should not skip the build")
+
+	// Verify no cache is used when --force is specified
+	require.NotContains(t, output, "importing cache manifest", "build with --force should not import cache")
+	require.NotContains(t, output, "preparing layers for inline cache", "build with --force should not export cache")
 }


pkg/porter/build.go
```diff
@@ -110,6 +113,23 @@ func (p *Porter) Build(ctx context.Context, opts BuildOptions) error {
 
 	span.Debugf("Using %s build driver", p.GetBuildDriver())
 
+	// If --force is specified, enable --no-cache for docker build
+	if opts.Force {
+		opts.NoCache = true
```
dgannon991, line 118: Really minor one, do we need a test to confirm this behaviour? (I.E. when forcing we also skip the docker cache?)

kichristensen: @dgannon991  Good point I have update the test

dgannon991 (approved):

PR #3532, opened 2025-12-27
Add config management commands (show, edit, set)

### LINKED ISSUE #1077: CLI: Porter config management commands
Currently, [Porter's config file](https://porter.sh/configuration/) can be created/edited manually via its expected location (`~/.porter/config.toml`).

It would be nice if we had a `porter config` suite of CLI commands to create, edit and view this file, without necessarily directly accessing its location mentioned above.

A new suite of commands under `porter config`:
 - `porter config create` - to create the configuration file (not sure if Porter creates it behind-the-scenes already?)  I might be convinced that this subcommand isn't necessary (Perhaps `porter config edit` would just create the file if it didn't already exist.)
 - `porter config show` - to show the current configuration
 - `porter config edit` - to edit the current configuration
 - Other? Different?

As mentioned in the description, the alternative is to create/open/edit the file via its location in the Porter home directory.

**Additional context**
N/A

kichristensen:
# What does this change

Adds three new commands for managing Porter's configuration file:

- `porter config show [--output FORMAT]` - Display current config
- `porter config edit` - Edit config in $EDITOR  
- `porter config set KEY VALUE` - Set individual config values

## Examples

**Show config (displays defaults if no config exists):**
```bash
$ porter config show
allow-docker-host-access = false
build-driver = "buildkit"
verbosity = "info"
...

$ porter config show -o json
{
  "build-driver": "buildkit",
  "runtime-driver": "docker",
  ...
}
```

**Set individual values:**
```bash
$ porter config set verbosity debug
Config saved to /home/user/.porter/config.toml
Set verbosity = debug

$ porter config set logs.level info
Config saved to /home/user/.porter/config.toml
Set logs.level = info
```

**Edit in your preferred editor:**
```bash
$ porter config edit
# Opens config.toml in $EDITOR
```

## Features

- Multi-format support: toml (default), yaml, json
- Format preservation: modifications preserve existing file format
- Format conversion: `show -o` converts between formats
- Type-safe value setting with automatic conversion (string, bool, int, nested fields)
- Domain validation with helpful error messages
- Creates default config if none exists

# What issue does it fix

Closes #1077

# Notes for the reviewer

**Key design decisions:**
- Reflection-based field discovery (no manual field mappings)
- Tag-based validation using `validate:"oneof=..."` struct tags
- Format detection from file extension
- Preserves existing format for set/edit, allows conversion with show

# Checklist
- [x] Did you write tests?
- [x] Did you write documentation? (help text with examples)

### DIFF
--- a/pkg/config/setter.go
+++ b/pkg/config/setter.go
@@ -11,12 +11,14 @@ import (
 // Supports simple fields and nested fields like "logs.level" or "telemetry.enabled".
 // Uses reflection to dynamically set fields based on struct tags.
 func SetConfigValue(data *Data, path string, value string) error {
-	if err := ValidateConfigValue(data, path, value); err != nil {
+	// Validate and normalize the value (validation also normalizes case for oneof fields)
+	normalizedValue, err := ValidateAndNormalizeConfigValue(data, path, value)
+	if err != nil {
 		return err
 	}
 
 	parts := strings.Split(path, ".")
-	return setFieldByPath(reflect.ValueOf(data).Elem(), parts, value)
+	return setFieldByPath(reflect.ValueOf(data).Elem(), parts, normalizedValue)
 }
 
 // setFieldByPath recursively navigates and sets a field using the path segments
@@ -124,12 +126,18 @@ func setFieldValue(field reflect.Value, value string) error {
 
 // ValidateConfigValue validates that a config path exists and the value is valid for its type
 func ValidateConfigValue(data *Data, path string, value string) error {
+	_, err := ValidateAndNormalizeConfigValue(data, path, value)
+	return err
+}
+
+// ValidateAndNormalizeConfigValue validates and normalizes (case) the value for oneof fields
+func ValidateAndNormalizeConfigValue(data *Data, path string, value string) (string, error) {
 	if path == "" {
-		return fmt.Errorf("config path cannot be empty")
+		return "", fmt.Errorf("config path cannot be empty")
 	}
 
 	if value == "" {
-		return fmt.Errorf("config value cannot be empty")
+		return "", fmt.Errorf("config value cannot be empty")
 	}
 
 	parts := strings.Split(path, ".")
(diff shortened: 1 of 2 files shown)


pkg/config/setter.go
```diff
@@ -0,0 +1,221 @@
...
+
+// findFieldByTag finds a struct field by matching against toml/yaml/json tags
+func findFieldByTag(v reflect.Value, name string) (reflect.Value, error) {
+	field, _, err := findFieldByTagWithMeta(v, v.Type(), name)
+	return field, err
+}
+
+// findFieldByTagWithMeta finds a struct field and returns both the value and StructField metadata
+func findFieldByTagWithMeta(v reflect.Value, t reflect.Type, name string) (reflect.Value, reflect.StructField, error) {
+	for i := 0; i < v.NumField(); i++ {
```
dgannon991, line 61: My editor is showing a warning here, saying: "If the passed reflect.Value is not a struct (or is a pointer), v.NumField() will panic"

pkg/config/setter.go
```diff
@@ -0,0 +1,221 @@
...
+		field := t.Field(i)
+
+		// Check toml, yaml, and json tags
+		for _, tagKey := range []string{"toml", "yaml", "json"} {
```
dgannon991, line 65: Does this mean if we miss an annotation in a config file, it won't work here? (E.G. I provide TOML + YAML, but the user requests JSON)

pkg/config/setter.go
```diff
@@ -0,0 +1,221 @@
...
+// validateWithTag validates a value against the validate struct tag
+func validateWithTag(sf reflect.StructField, value string, path string) error {
+	validateTag := sf.Tag.Get("validate")
+	if validateTag == "" {
+		return nil // No validation tag
+	}
+
+	// Parse the validate tag - currently only support "oneof=value1 value2 ..."
+	if strings.HasPrefix(validateTag, "oneof=") {
+		validValues := strings.Split(strings.TrimPrefix(validateTag, "oneof="), " ")
```
dgannon991, line 222: Just a question, would this be better as a CSV (or similar) instead of space separated?

dgannon991 (changes requested):
Additional tests AI recommend adding;

- Attempt to set a non-primitive field (slice/map/struct) and assert a descriptive error.
- Attempt to set an exported field whose underlying type is a named numeric/enum to ensure SetString/SetInt works for named types.
- Tests that simulate FileSystem failures for SaveConfig and CreateDefaultConfig.
- An integration-style test for porter.SetConfig path that exercises loading existing file + saving changes (you already have this partly, but exercise the CLI path end-to-end).
- Test case-insensitive inputs if you decide to normalize values (or add docs that values are exact).

pkg/config/setter.go
kichristensen: If for example we something like this, where the JSON annotation is missing:

```go
type Data struct {
  BuilDriver string `toml:"build-driver" yaml:"build-driver"
}
```

This function will find anyway because it loops through all of them, the requested format is not considered here. It currently works because I assume that the value will be same across all formats.
The annotation is used during marshaling to the requested format, meaning it must there for each supported configuration format.
For this specific function it will only "fail" if the field annotation differs between formats, which is a little confusing.
All input is appreciated, I'm not 100% sure what I think myself.

pkg/config/setter.go
kichristensen: Inspiration was taken from https://github.com/go-playground/validator where they use spaces, don't know why.
I'm happy to change if you want to.
The reason the library isn't use is that it primarily works on full structs, not individual values, which would require a lot of code to handle

pkg/config/setter.go
kichristensen: Good catch. I have added a guard and a unit test

dgannon991: Ace, cheers!

pkg/config/setter.go
dgannon991: Ah OK, that makes a lot more sense. Is there any form of validation we could add for the values being the same across the different format types?

pkg/config/setter.go
dgannon991: That makes sense, I'd say given that there's is an established standard, we stick with it, and if there are any questions as to why, we can just signpost people to them. Cheers! :)

PR #3542, opened 2026-01-22
#2930: Attempt a plugin using the syft CLI

### LINKED ISSUE #2930: (feat): Generate SBOM for invocation image + bundle

dgannon991:
This closes #2930 with a slightly different approach. Copying the signing plugin approach of using a CLI tool in a plugin this time :)


kichristensen:
@dgannon991 Unfortunately this PR have some conflicts, while you fix those I will take a closer look at the changes

dgannon991:
@kichristensen - thank you, but please hold off until I get a chance to run through the copilot comments. Most are about copy/pasted comments, so they should be quick wins :) I'll ping you when it's all tidied up :)

pkg/porter/publish.go
```diff
@@ -243,7 +266,25 @@ func (p *Porter) publishFromFile(ctx context.Context, opts PublishOptions) error
...
+		log.Infof("Bundle details are: %+v", bundleRef)
+		if err := p.SBOMGenerator.Generate(
+			ctx,
+			bundleRef.String(),
+			opts.SBOMPath,
+			opts.InsecureRegistry,
+		); err != nil {
+			return log.Error(err)
+		}
+	}
```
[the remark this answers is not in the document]
dgannon991: It's a good version 1 IMO as the bundle contains the invocation images.

pkg/sbom/pluginstore/grpc.go
```diff
@@ -0,0 +1,78 @@
...
+
+import (
+	"context"
+
+	"get.porter.sh/porter/pkg/portercontext"
+	"get.porter.sh/porter/pkg/sbom/plugins"
+	"get.porter.sh/porter/pkg/sbom/plugins/proto"
+)
+
+var _ plugins.SBOMGeneratorProtocol = &GClient{}
```
schristoff, line 11: Is this necessary? I was under the impression that Syft gave us examples and we could just import the necessary libraries.
But to be fair.. that was like 3 years ago.

schristoff: (I mean this in referencing to the proto being generated and used throughout - I clicked the wrong area to leave this comment)

dgannon991: So when I did it that way, it increased the size of the porter binary by approx 50% (from memory, it was a while ago!) and brought in loads of extra dependencies that we wouldn't have needed otherwise. So we tried it this way instead, so that it was opt-in. Happy to go back to bundling it all in (as this is being a massive PITA!)

schristoff: You are right
I totally forgot that we discussed this before
I will forget again so prepare to remind me -_-

PR #3554, opened 2026-02-27
Only v-prefix tags with at least a major and minor version

erikced:
# What does this change

Update auto v-prefixing on pull: only prefix tags with at least major and minor version, to avoid enforcing v-prefix on all-digit refs like hashes and build numbers.

`semver.NewVersion` is relatively lax in its intepretation of what is a semver tag, causing automatic v-prefixing to trigger on bundles tagged only digits, e.g. CI build numbers or hashes. This MR adds an additional check that there has to be at least _one_ dot in the version number. `semver.StrictNewVersion` was not used as changing that would alter the behavior when converting bundle versions to tags if the version in the manifest is partial semver.

# Checklist
- [x] Did you write tests?

### DIFF
--- a/pkg/porter/lifecycle.go
+++ b/pkg/porter/lifecycle.go
@@ -379,7 +379,7 @@ func ensureVPrefix(opts *BundleReferenceOptions, out io.Writer) error {
 	}
 
 	// Do nothing for empty tags, tags that do not contain at least major and minor versions and non-semver tags
-	if !ociRef.HasTag() || !tagIsSemverLike(ociRef) || !ociRef.HasVersion() {
+	if !tagIsSemverLike(ociRef) || !ociRef.HasVersion() {
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/porter/lifecycle.go
```diff
@@ -397,8 +397,9 @@ func ensureVPrefix(opts *BundleReferenceOptions, out io.Writer) error {
 	return nil
 }
 
-func tagStartsWithNumber(ociRef *cnab.OCIReference) bool {
-	return ociRef.HasTag() && ociRef.Tag()[0] >= '0' && ociRef.Tag()[0] <= '9'
+func tagIsSemverLike(ociRef *cnab.OCIReference) bool {
+	tag := ociRef.Tag()
+	return tag[0] >= '0' && tag[0] <= '9' && strings.Contains(tag, ".")
```
dgannon991: If the image doesn't have a tag, the `.Tag()` method will return `""` which will throw a runtime panic here I think. Would it be possible to:
- Add a test for an image without a tag
- Add a check for tag being an empty string here?

Cheers

erikced: Updated `tagIsSemverLike` to be a bit more resilient and removed the corresponding check from `ensureVPrefix`, I did not add any additional tests as I think it is covered by the [_ignores references with no tag_](https://github.com/getporter/porter/blob/424ae2c502f3b2e9c9aeb8546862f856dddb0962/pkg/porter/lifecycle_test.go#L570-L573)  `ensureVPrefix` test and `tagIsSemverLike` is internal but I can add tests for it too if you wish 🙂

dgannon991: Ah perfect, thank you. Yup you're absolutely right, I'd missed that because it wasn't in the merge preview, sorry about that!

PR #3565, opened 2026-03-08
fix(config): resolve secrets only for selected context

### LINKED ISSUE #3558: Difficult to mix secret plugins/sources in new config contexts
### What is the issue?

Multi-context configs resolve `${secret.*}` references from all
contexts, not just the selected one. This makes multi-context
configs unusable when different contexts use different secrets
plugins or key vaults.

### Expected behaviour

When a context is selected (via `current-context`, `--context`,
or `PORTER_CONTEXT`), only `${secret.*}` references within that
context's `config:` block should be resolved. References in other
contexts should be ignored.

### Actual behaviour

All `${secret.*}` references across the entire config file are
collected and resolved using the selected context's secrets
plugin. If context A references a secret that only exists in
context B's key vault, loading fails even though context A is
the one selected.

### Steps to reproduce

Given this config at `~/.porter/config.yaml`:

```yaml
schemaVersion: "2.0.0"
current-context: local
contexts:
  - name: local
    config:
      default-secrets-plugin: "filesystem"
  - name: prod
    config:
      default-secrets: "prod-vault"
      default-storage: "prod-db"
      storage:
        - name: "prod-db"
          plugin: "mongodb"
          config:
            url: "${secret.prodMongoConnectionString}"
      secrets:
        - name: "prod-vault"
          plugin: "azure.keyvault"
          config:
            vault: "kv-prod"
            subscription-id: "..."
```

Running:

```
porter config show
```

Fails with:

```
could not render config file because ${secret.prodMongoConnectionString}
could not be resolved: rpc error: code = Unknown desc = error reading
secret from filesystem: open ~/.porter/secrets/prodMongoConnectionString:
no such file or directory
```

The `local` context doesn't use this secret at all, but Porter
still tries to resolve it using the `local` context's filesystem
plugin.

Even `--context prod` fails the same way if there are secrets in
other contexts that `prod`'s key vault doesn't contain.

### Root cause

In `pkg/config/loader.go`, the config loading sequence is:

1. **Lines 119-145**: The entire YAML file is parsed as a liquid
   template. `listTemplateVariables(tmpl)` (line 140) captures
   all `${secret.*}` references from the full file, across all
   contexts.

2. **Lines 158-175**: `extractContextConfig()` isolates the
   selected context's config block — but the template variable
   list was already captured from the entire file in step 1.

3. In `pkg/config/config.go`, `loadFinalPass()` (lines 395-420)
   iterates through all captured template variables and resolves
   every `${secret.*}` reference, including those belonging to
   unselected contexts.

The fix would be to either:

- Extract the selected context's config block before capturing
  template variables, so only the selected context's secrets are
  resolved
- Or scope `listTemplateVariables()` to only the selected
  context's YAML subtree

### Environment

- Porter version: v1.4.0-37-g424ae2c5 (canary)
- OS: macOS

SergioChan:
# What does this change
- Scopes template variable discovery to the selected context's `config:` block in multi-context config files.
- Prevents secret resolution from traversing `${secret.*}` references that belong to other contexts.
- Adds a regression test proving loading `current-context: local` does not attempt to resolve secrets that are only referenced under `prod`.

This makes `porter config show` and normal config loading work when contexts use different secret stores.

# What issue does it fix
Closes #3558

# Notes for the reviewer
- The fix is intentionally minimal: legacy flat config behavior is unchanged.
- For multi-context files, `cfg.templateVariables` is now recomputed from the selected context map so second-pass secret resolution only evaluates relevant keys.

# Checklist
- [x] Did you write tests?
- [ ] If this is your first pull request, please add your name to the bottom of our [Contributors][contributors] list. Thank you for making Porter better! 🙇♀️

### DIFF
--- a/pkg/config/loader.go
+++ b/pkg/config/loader.go
@@ -15,6 +15,7 @@ import (
 	"github.com/osteele/liquid/render"
 	"github.com/spf13/viper"
 	"go.opentelemetry.io/otel/attribute"
+	"gopkg.in/yaml.v3"
 )
 
 var _ DataStoreLoaderFunc = NoopDataLoader
@@ -175,6 +176,10 @@ func LoadFromViper(viperCfg func(v *viper.Viper), cobraCfg func(v *viper.Viper))
 				return log.Error(err)
 			}
 
+			// Scope template variables to the selected context only so that
+			// `${secret.*}` from other contexts are not resolved.
+			cfg.templateVariables = listTemplateVariablesForContext(contextConfigMap)
+
 			ctxViper := viper.New()
 			ctxViper.SetFs(cfg.FileSystem)
 			if err := setDefaultsFrom(ctxViper, cfg.Data); err != nil {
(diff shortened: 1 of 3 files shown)


SergioChan:
Thanks for calling that out — I added an integration test in commit `f9e2537` to cover this path (`tests/integration/config_test.go`).

The new case exercises selected-context secret resolution behavior so the regression is validated beyond unit-level coverage.

Issue #2284, opened 2022-08-04
Add ability to start a span without writing to logger


VinozzZ:
Currently, porter has a `TraceLogger` wrapper that contains both tracing and logging ability. When a span is created, any information addes to the span through the `TraceLogger` will also writes to the logger. This creates issue like #2278 which causes the same error being written multiple times in the console.
To avoid big changes throughout the entire codebase from a complete redesign and having too much boilerplate, we would like to solve this problem by introducing additional methods onto `TracerLogger` so that we can create spans without also writing to our logger. 
One way to achieve this is to change logger sink when creating a span so that writing to logger becomes a no-op operation.  We could use some exploration time to try out some options and come up with a long-term solution

PR #3592, opened 2026-05-17
Print errors once, not per span level

### LINKED ISSUE #2278: Only print error messages once to the console
I had to disable TestCLI because it was enforcing that an error message is only printed once. I originally added that test because I only wanted main to write errors to the console, not cobra.

Now that we are using open telemetry with our TraceLogger everywhere, functions are calling span.Error() in the function that originated the error. Now errors are being written out as they occur, and when an error is wrapped and returned, it ends up being printed multiple times, including the printing by main.

We need to decide who is responsible for printing the error, and do so in a way that errors are only printed once. I think this means that span.Error should not write to the log, and wait to write to the log after the error has completely propagated up the call stack.

### LINKED ISSUE #3355: Porter default error messages are too verbose for an installer experience
## Describe the bug
Porter error output too verbose for an installer experience
In the output image below, it can been seen that the error is in the first line alone `error: no matching resources found`. However, further verbose output is provided which can be avoided or removed or provided for debug purposes alone.

## To Reproduce
1. Run any porter install that can result into an error
2. See error as given in the output

## Expected behavior
It would great for the user to provide this error message over what porter provides

## Porter Command and Output
Image

## Version
porter v1.2.1 (aa97c590)

kichristensen:
# What does this change

Errors in Porter were printed multiple times to the console — once per call stack layer — because `TraceLogger.Error()` wrote to both the OTel span and the Zap console logger, and error wrapping caused each layer to re-log the growing chain:

```
error: exit status 1
error: running script: exit status 1
error: running bundle: running script: exit status 1
```

This PR removes the Zap logger call from `TraceLogger.Error()` so it only records to the OTel span. Both `cmd/porter/main.go` and `cmd/exec/main.go` now print the final error to stderr exactly once after the command fails.

The rule is now clear:
- `log.Error(err)` — records to telemetry/tracing only, no console output
- `log.Debug/Info/Warn` — prints to console and records to span
- `main.go` — sole place that prints the final error to the user

The long-disabled `TestCLI` integration test is re-enabled.

The new error output will look like this:

```
$ porter explain -r=[ghcr.io/getporter/missing-bundle

unable to pull bundle: failed to resolve bundle manifest "ghcr.io/getporter/missing-bundle": object required
```

instead of the the current:

```
$ porter explain -r=[ghcr.io/getporter/missing-bundle

unable to pull bundle: failed to resolve bundle manifest "ghcr.io/getporter/missing-bundle": object required
unable to pull bundle: failed to resolve bundle manifest "ghcr.io/getporter/missing-bundle": object required
```

# What issue does it fix

Closes #3355
Closes #2278

# Notes for the reviewer

Issue #2284 proposed adding a new span-without-logger API. This PR takes a simpler approach: remove console output from `Error()` entirely and let `main` own the user-facing error print. No new API surface, no call-site changes needed.

# Checklist
- [x] Did you write tests?


pkg/portercontext/context_test.go
```diff
@@ -71,10 +71,10 @@ func TestContext_PluginVerbosityLevel(t *testing.T) {
...
 		wantNumberOfLogLines int
 	}{
-		{"debug level", zapcore.DebugLevel, 4},
-		{"info level", zapcore.InfoLevel, 3},
-		{"warn level", zapcore.WarnLevel, 2},
-		{"error level", zapcore.ErrorLevel, 1},
+		{"debug level", zapcore.DebugLevel, 3},
+		{"info level", zapcore.InfoLevel, 2},
+		{"warn level", zapcore.WarnLevel, 1},
+		{"error level", zapcore.ErrorLevel, 0},
```
schristoff, line 77: What is the reason for the levels change here?

pkg/portercontext/testdata/expected-output.txt
```diff
@@ -1,4 +1,3 @@
 Writing logs to /.porter/logs/0.json
 a thing happened
 a weird thing happened
-a bad thing happened
```
schristoff, line 4: I still believe these should be capture-able, are they no longer so?

kichristensen: Errors are still captured to the log file, but not by the console per-span. This file only tracks console output, the error is shown once, printed by `main.go`

pkg/portercontext/context_test.go
kichristensen: I think this was a mistake, I will take a closer look

kichristensen: Yes it was, I have reverted that part

schristoff:
When we put this in the release we are gonna have to call that this was a breaking change is all

schristoff (approved):

PR #3627, opened 2026-07-12
migration: error instead of panic when an installation has no claims

### LINKED ISSUE #3629: Migrate panics on installation with no claims
## Summary
`Migration.migrateInstallation` indexes `claimIDs[0]` after `listItems` returns an empty slice for an installation that has no claims (e.g. an `installations/<name>` marker with no matching `claims/<name>/` dir). This panics the migration instead of failing loudly.

## Root cause
`migrateInstallation` builds `claimIDs` then sets `inst.ID = claimIDs[0]` with no length guard, so an installation that legitimately has zero claims triggers an index-out-of-range panic.

## Reproducer
1. Create a Porter home with `installations/noclaims/` present but no `claims/noclaims/` entry.
2. Run `porter storage migrate`.
3. Migration panics with index out of range on `claimIDs[0]`.

## Fix
PR #3627 adds a length check that returns a clear error (`installation <name> has no claims to derive a stable id from`) so the migration fails with a useful message instead of panicking.

AruneshDwivedi:
migrateInstallation indexed claimIDs[0] after listItems returned an empty slice for an installation with no claims, panicking the migration. It now returns a clear error so the migration fails loudly instead of crashing.

The zero-claims case is covered by a test that adds an isolated no-claims installation to the test home and asserts Migrate returns the error. The shared testdata fixture is left untouched so the standard migrations keep succeeding.

Closes #3629

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -107,3 +107,4 @@ and we will add you. **All** contributors belong here. 💯
 - [Geeta Chavan](https://github.com/geetachavan1)
 - [Stephen Augustus](https://github.com/justaugustus)
 - [Erik Cederberg](https://github.com/erikced)
+- [Arunesh Dwivedi](https://github.com/AruneshDwivedi)
--- a/pkg/storage/migrations/migration.go
+++ b/pkg/storage/migrations/migration.go
@@ -224,6 +224,9 @@ func (m *Migration) migrateInstallation(ctx context.Context, installationName st
 
 	// Sort the claims earliest to latest and assign the installation id
 	// to the earliest claim id. This gives us a consistent installation id when the migration is repeated.
+	if len(claimIDs) == 0 {
+		return fmt.Errorf("installation %s has no claims to derive a stable id from", installationName)
+	}
 	sort.Strings(claimIDs)
 	inst.ID = claimIDs[0]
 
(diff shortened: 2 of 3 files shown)


kichristensen:
Thanks for the fix — nice catch, and good root-cause writeup in the PR description.

Could you add a test covering the zero-claims case? `TestMigration_Migrate` currently
only exercises fixture installations that all have claims, so this exact path stays
uncovered. A small addition to `tests/testdata/porter_home/v0/` (an `installations/<name>`
marker with no matching `claims/<name>/` dir) plus an assertion that `Migrate` returns
the new error would do it.

Also, could you fill out the [PR template][2] (checklist + `Closes #`)? We don't have a
tracking issue for this yet, so it'd help to link one or explain the context per the
[Contributing Guide][1] if you open one.

[2]: https://github.com/getporter/porter/blob/main/.github/pull_request_template.md

kichristensen:
@AruneshDwivedi thanks for the changes. The comment from Copilot makes sense, can you please handle that.
Also please add your name to the bottom of our [Contributors](https://porter.sh/src/CONTRIBUTORS.md) list. Thank you for making Porter better! 🙇‍♀️

AruneshDwivedi:
Done. Switched the assertion to require.ErrorContains so a wrong or nil error fails right at that line, and added myself to the contributors list.

kichristensen (approved):

PR #3638, opened 2026-07-21
Skip image push when digest already matches

### LINKED ISSUE #2337: Can we improve existing bundle detection?
In #2331, we added a check that if a bundle already existed at the published location that the user must use the --force flag to overwrite it.

I spent a while trying to figure out how we can compute a local hash or otherwise figure out that the bundle locally is the same as the published bundle and ran into the usual problems with repository digests and tarred/untarred layer digests. 

Let's figure out if there is a reliable way to compare local content with a remote artifact in an OCI repository and if they are the same (without having to pull the remote artifact).

kichristensen:
# What does this change
`porter publish` now skips re-pushing an image when the destination
already has the exact same content, instead of always re-pushing:

- When publishing from source (`porter publish`), before pushing the
  invocation image Porter compares the local Docker cache's repository
  digest for that image against the digest currently published at the
  destination (a single HEAD request, no pull). If they match, the push
  is skipped and the existing digest is reused when rewriting
  `bundle.json`. `--force` always repushes and bypasses this check.
- When publishing from an archive (`porter publish --archive`), the same
  idea applies to `relocateImage`/`pushUpdatedImage`, which pushes both
  the invocation image and any referenced images extracted from the
  archive's OCI layout. Here the digest to push is already known exactly
  from the layout, so it's an even simpler check: one HEAD request against
  the destination tag before pushing.

Example: republishing a bundle whose invocation image content hasn't
changed since the last publish now logs:
```
Bundle image myregistry/mybuns:temp-abc123 already published with matching content, skipping image push
```
instead of re-uploading the image.

# What issue does it fix
Closes #2337

# Notes for the reviewer
This only covers the invocation image (both publish paths) and the
referenced images relocated via the `--archive` path. It does not yet
cover referenced images relocated by `PushBundle`'s underlying
`cnab-to-oci` `FixupBundle` call in the normal (non-archive) publish
path. That's mostly already handled upstream — `cnab-to-oci` itself
skips re-copying a referenced image manifest when it already exists at
the destination digest — except for referenced images that only exist
in the local Docker daemon and have never been pushed anywhere, which
still get unconditionally re-pushed (`pushLocalImage` in cnab-to-oci's
fixup chain). Closing that gap needs either an upstream change to
`cnab-to-oci` or Porter pre-populating the relocation map itself; I
scoped it out of this PR and can follow up separately if useful.

# Checklist
- [x] Did you write tests?

### DIFF
--- a/pkg/cnab/cnab-to-oci/helpers.go
+++ b/pkg/cnab/cnab-to-oci/helpers.go
@@ -13,15 +13,16 @@ import (
 var _ RegistryProvider = &TestRegistry{}
 
 type TestRegistry struct {
-	MockPullBundle        func(ctx context.Context, ref cnab.OCIReference, opts RegistryOptions) (cnab.BundleReference, error)
-	MockPushBundle        func(ctx context.Context, ref cnab.BundleReference, opts RegistryOptions) (bundleReference cnab.BundleReference, err error)
-	MockPushImage         func(ctx context.Context, ref cnab.OCIReference, opts RegistryOptions) (imageDigest digest.Digest, err error)
-	MockGetCachedImage    func(ctx context.Context, ref cnab.OCIReference) (ImageMetadata, error)
-	MockListTags          func(ctx context.Context, ref cnab.OCIReference, opts RegistryOptions) ([]string, error)
-	MockPullImage         func(ctx context.Context, ref cnab.OCIReference, opts RegistryOptions) error
-	MockGetBundleMetadata func(ctx context.Context, ref cnab.OCIReference, opts RegistryOptions) (BundleMetadata, error)
-	MockGetImageMetadata  func(ctx context.Context, ref cnab.OCIReference, opts RegistryOptions) (ImageMetadata, error)
-	cache                 map[string]ImageMetadata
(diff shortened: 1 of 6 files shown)


kichristensen:
Follow-up scope note (not addressed in this PR):

Referenced images (`bundle.json`'s `images`, not the invocation image) in the normal source-build `porter publish` path go through `PushBundle` → `cnab-to-oci`'s `remotes.FixupBundle`. That library already skips re-copying a referenced image when the destination already has a manifest at the resolved digest (`mount.go`'s `createCopyTask` does a `resolve(targetRepo@digest)` before copying), so the common case is already handled upstream, no Porter change needed.

The one remaining gap: referenced images that exist *only* in the local Docker daemon and have never been pushed anywhere fall into `cnab-to-oci`'s `pushLocalImage` fixup, which unconditionally re-pushes with no digest pre-check — the same bug this PR fixes for the invocation image, but living inside the vendored `cnab-to-oci` dependency rather than in Porter. Closing it would need either an upstream `cnab-to-oci` change (preferred, since Porter doesn't own that code path) or Porter pre-populating the relocation map itself by replicating that internal digest logic. Leaving this as follow-up work rather than folding it into this PR.

PR #3658, opened 2026-08-08
Add porter-112 lint: block dep outputs in action steps

### LINKED ISSUE #3650: Restrict bundle.dependencies.* output refs to the dependencies wiring block in lint
Per [proposals#21](https://github.com/getporter/proposals/pull/21), PEP003 now restricts `bundle.dependencies.DEP.outputs.NAME` (and the `outputs.NAME` shorthand) to inside `dependencies.requires[]` — i.e. wiring a dependency's output to another dependency's parameter/credential, or promoting it to a bundle output. These variables are no longer allowed directly in action steps (install, upgrade, uninstall, custom actions); those steps must reference the promoted value via `bundle.outputs.NAME` instead.

porter-106..109 (#2672) already scope their checks to `dependencies.requires[].parameters/credentials/outputs`, so they're unaffected. But nothing currently checks action steps: `validateParamsAppliesToAction` only inspects parameter template vars, so a v2 bundle referencing `${bundle.dependencies.mysql.outputs.dbCon}` directly in an `install` step passes lint silently today.

Add a new lint check (gated behind the `dependencies-v2` experimental flag, same as porter-106..109) that scans action steps for `bundle.dependencies.*` output references and errors, pointing the author at promoting the output via `dependencies.requires[].outputs` and referencing it with `bundle.outputs.NAME` instead.

Note: this only applies to `dependencies-v2` bundles. Dependencies v1 already resolves `bundle.dependencies.DEP.outputs.NAME` directly in action steps via parameter/credential sources at runtime (see `wiring.md`/`using-templates.md`) and that behavior is unaffected.

ℹ️ Read [PEP003 - Advanced Dependencies](https://github.com/getporter/proposals/blob/main/pep/003-advanced-dependencies.md) for context.

kichristensen:
# What does this change
Adds a new lint check, porter-112, gated behind the `dependencies-v2`
experimental flag. It flags `bundle.dependencies.DEP.outputs.NAME`
template variables referenced directly in action steps (install,
upgrade, uninstall, custom actions). Per PEP003, that form is only
valid inside `dependencies.requires[]` wiring (mapping one
dependency's output into another's parameter/credential, or
promoting it to a bundle output); action steps must reference the
promoted value via `bundle.outputs.NAME` instead.

Example that now fails lint under dependencies-v2:

```yaml
install:
  - exec:
      command: ./helpers.sh
      arguments:
        - install
        - "${ bundle.dependencies.mysql.outputs.dbCon }" # porter-112
```

Also fixes a pre-existing bug: every lint result URL except
porter-101 pointed at `porter.sh/reference/linter/...` (missing
`docs/`) instead of `porter.sh/docs/references/linter/...`. Fixed in
`linter.go`, `linter_test.go`, and `pkg/porter/lint.go` (porter-105).

# What issue does it fix
Closes #3650

# Notes for the reviewer
- New check reuses `m.GetTemplateVariables` (not
  `ScanManifestTemplating`) to scan each step's raw data, since
  `ScanManifestTemplating` injects every dependency-output shorthand
  variable found anywhere in `dependencies.requires[].outputs` into
  every scan result when dependencies-v2 is enabled — using it here
  would cause false positives on unrelated steps. There's a
  regression test covering this.
- Dependencies v1 (flag off) is unaffected; it resolves
  `bundle.dependencies.DEP.outputs.NAME` in action steps at runtime.

# Checklist
- [x] Did you write tests?


pkg/linter/linter.go
```diff
@@ -188,7 +188,7 @@ func (l *Linter) Lint(ctx context.Context, m *manifest.Manifest, config *config.
 					Code:    "porter-100",
 					Title:   "Reserved name error",
 					Message: param.Name + " has a reserved prefix. Parameters cannot start with porter- or porter_",
-					URL:     "https://porter.sh/reference/linter/#porter-100",
+					URL:     "https://porter.sh/docs/references/linter/#porter-100",
```
dgannon991: [non-blocking] Should the base of this URL be a constant somewhere? (just thinking about the number of updates you had to do here?)

dgannon991 (approved):
Sorry to drive-by this after being silent for months, but I got a ping in GitHub whilst I happened to be logged in, and thought I could be useful :D 

One non blocking comment on the URL base being pasted everywhere.

And just a note as well if it happens to be in the same area, the text for https://porter.sh/docs/references/linter/#exec-100 is garbled around the URL. It reads `https://porter.sh/(/docs/best-practices/exec-mixin/)#use-scripts` but I think it should be `https://porter.sh/docs/best-practices/exec-mixin/#use-scripts`

pkg/linter/linter.go
kichristensen: Good idea, PR updated.

PR #3661, opened 2026-08-12
Don't hide sensitive output persist failures

### LINKED ISSUE #3657: State-secret store mismatch when installation succeeds but saving secret does not
## Describe the bug
When an installation of a bundle has successfully went through and the metadata of the installation is saved, Porter goes on to save any sensitive output values in the configured secret store. However, if this fails, it appears Porter does not roll back the metadata in the installation and it consequently contains a reference to a sensitive output in the secret store, an output which does not exist.

## To Reproduce
- Run a bundle installation that has a sensitive output.
- During the installation,  make sure to make the configured secret store unavailable so that it will fail when saving the sensitive output.
- Observe the error in the log, but the installation marked as "Successful".
- Try to `output list` the installation.
- An error occurs stating that the output of the last run cannot be found in the secret store.

## Expected behavior
The installation where the output fails to be saved should be set to failed as it's effectively impossible to access it due to the mismatch between the installation state and the state of the secret store.

## Version
Porter Version: v1.5.0
Porter Commit: a81c51a8

kichristensen:
# What does this change
When a sensitive output fails to write to the secret store, Porter used to
still mark the run/installation as succeeded and insert an `Output` record
with a `Key` pointing at a secret that was never created. Later, `output
list` (and anything else resolving that output) would fail with a confusing
"secret not found" error instead of the real cause, and nothing in
`installations list`/`show` hinted that anything was wrong.

This change:
- Stops persisting the dangling secret `Key` reference. `storage.Output` gets
  a `PersistError` field recording why the value is unavailable.
- Reorders `SaveOperationResult` to sanitize outputs *before* persisting
  `Result`/`Installation`, so both get a new `OutputPersistFailed` flag in
  the same write (no extra queries needed to display it later).
- `installations list`/`show` and `runs list` now show a warning, e.g.:
  ```
  $ porter installations list
  NAMESPACE   NAME      VERSION   STATE       STATUS                            MODIFIED
              mybuns    0.1.0     installed   succeeded (output persist failed) 3 minutes ago
  ```
- `output list` no longer fails its entire output on one bad output; the
  broken one shows a `⚠ value unavailable: ...` warning instead, others
  still display normally. `output show <name>` on a broken output now
  returns an actionable error instead of a raw secret-store "not found".
- Adds an opt-in `--fail-on-output-warnings` flag (install/upgrade/invoke/
  uninstall, default off) for CI/automation that wants the command to exit
  non-zero when this happens, instead of just warning.

# What issue does it fix
Closes #3657

# Notes for the reviewer
- Direction follows the discussion on the issue, with one deviation: rather
  than querying the outputs collection per-installation in `installations
  list` (N+1 query), `OutputPersistFailed` is stored directly on
  `Result`/`InstallationStatus`, computed once in `SaveOperationResult`.
- Default for `--fail-on-output-warnings` is off/exit-0: today's accidental
  non-zero exit on a persist failure (while the stored status says
  "succeeded") was itself part of the confusing mismatch, so the default now
  matches the stored status.
- No JSON schema update needed: the internal storage documents (`Output`,
  `Result`, `Installation.Status`) aren't covered by
  `pkg/schema/installation.schema.json`, which only describes the
  user-authored `installation.yaml`.

# Checklist
- [x] Did you write tests?

### DIFF
--- a/cmd/porter/installations.go
+++ b/cmd/porter/installations.go
@@ -436,6 +436,8 @@ func addBundleActionFlags(f *pflag.FlagSet, actionOpts porter.BundleAction) {
 		"Controls if the bundle should have access to the host's Docker daemon with elevated privileges. See https://porter.sh/configuration/#allow-docker-host-access for the full implications of this flag.")
 	f.BoolVar(&opts.ForceRun, "force-run", false,
 		"Run the bundle even if the installation has another incomplete run in progress.")
+	f.BoolVar(&opts.FailOnOutputWarnings, "fail-on-output-warnings", false,
+		"Fail the command if a sensitive output could not be persisted to the secret store, even though the bundle itself installed successfully.")
 	f.StringArrayVar(&opts.HostVolumeMounts, "mount-host-volume", nil, "Mount a host volume into the bundle. Format is <host path>:<container path>[:<option>]. May be specified multiple times. Option can be ro (read-only), rw (read-write), default is ro.")
 	f.BoolVar(&opts.NoLogs, "no-logs", false,
 		"Do not persist the bundle execution logs")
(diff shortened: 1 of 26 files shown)


kichristensen:
@lbergnehr What do you think I about this solution?

lbergnehr:
> @lbergnehr What do you think I about this solution?

I think it matches what I had in mind!

PR #3665, opened 2026-08-18
Do not validate config file schema version for internal plugins

erikced:
Disable config file schema version validation for internal plugins, as they receive a resolved config and do not load a config file, but can also receive `PORTER_CONTEXT` as env is forwarded which causes the validation to fail.

# What does this change
Validation that the config file has schema version 2 is bypassed if `Context.IsInternalPlugin` is set.

# What issue does it fix
```fish
❯ set -x PORTER_CONTEXT my-context
❯ porter installation list
  could not list installations: could not read storage schema document: could not load storage plugin: could not connect to the storage.porter.mongodb plugin: plugin stderr was --context/PORTER_CONTEXT requires a versioned config file; add schemaVersion: "2.0.0" and wrap settings under contexts
  : Unrecognized remote plugin message:

  This usually means that the plugin is either invalid or simply
  needs to be recompiled to support the latest protocol.
```

# Checklist
- [ ] Did you write tests?

### DIFF
--- a/pkg/config/loader.go
+++ b/pkg/config/loader.go
@@ -219,8 +219,9 @@ func LoadFromViper(viperCfg func(v *viper.Viper), cobraCfg func(v *viper.Viper))
 			}
 			cfg.viper = ctxViper
 		} else {
-			// Legacy flat format — existing path unchanged.
-			if cfg.ContextName != "" {
+			// Legacy flat format — existing path unchanged. Bypass schema version validation for internal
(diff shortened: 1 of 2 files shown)


kichristensen:
@erikced Thank you for the PR, looks good to me. Would it be possible to create a test to ensure we don't regress in the future?

kichristensen (approved):

Issue #2645, opened 2023-03-23
Generate an execution plan from a workflow


carolynvs:
- [ ] Resolve dependency graph
- [ ] Determine bundle actions
- [ ] Create installation records
- [ ] Create run records in the `Pending` state
- [ ] Create credential and parameter sets if necessary to share data between runs or just associate passed credential/parameter sets with each installation that needs them

ℹ️ Read [PEP003 - Advanced Dependencies](https://github.com/getporter/proposals/blob/main/pep/003-advanced-dependencies.md) for context about how dependencies should work, design details, and notes about desired behavior.

PR #3677, opened 2026-09-03
Build workflow stages from dep graph

kichristensen:
# What does this change
Adds `buildWorkflowSpec` (`pkg/porter/workflow_builder.go`), which converts
a resolved dependency `*Graph` (`GraphBuilder.BuildDependencyGraph`) into a
`storage.WorkflowSpec`: one `Job` per graph node, grouped into `Stages` by
longest-path layering over `DependsOn` (derived from both structural
requires-edges and data-flow wiring-edges).

Every node gets a `Job`, even one that will later turn out to already be in
sync and need no bundle action - this keeps a stable job ID for
`DependsOn`/wiring references regardless of what action a later step
decides on. `Job.Action`, `Job.Installation`, and `Job.Credentials` are
intentionally left unset here; this change only establishes structure
(job identity, stage/ordering, alias). Populating those fields is later
work also tracked under #2645.

This is purely a new set of internal functions - nothing calls them yet,
so there's no user-visible behavior change.

# What issue does it fix
Part of #2645 (does not close it - this is the first of several planned
chunks; see that issue's checklist).

# Notes for the reviewer
This is the first of ~5 small PRs implementing #2645, split along that
issue's own checklist. Chunks 2-5 (determine job actions, build job
installation records, create pending runs, wire dependency
parameters/credentials) build on this one and will follow as separate PRs
once this lands.

# Checklist
- [x] Did you write tests?

### DIFF
--- a/pkg/porter/workflow_builder.go
+++ b/pkg/porter/workflow_builder.go
@@ -8,9 +8,13 @@ import (
 	"get.porter.sh/porter/pkg/storage"
 )
 
-// buildJobIDs assigns a stable job ID to every node in a topologically
-// ordered node list (see Graph.TopologicalOrder), minted in dependency
-// order so IDs are deterministic for a given graph traversal.
+// buildJobIDs assigns a job ID to every node in a topologically ordered
+// node list (see Graph.TopologicalOrder), minted in dependency order.
+// cnab.NewULID() is random per call, so these IDs are not reproducible
+// across separate builds of the same graph -- they're only stable within
+// the single map this returns, which callers thread through
+// buildDependsOn/buildStages so a given node maps to the same ID
+// throughout one build.
 func buildJobIDs(order []*Node) map[NodeKey]string {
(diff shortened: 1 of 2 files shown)


kichristensen:
Addressed in 6b303946:
- Reworded the buildJobIDs comment to not imply reproducibility across separate builds (cnab.NewULID() is random per call) - it's only stable within one build's returned map.
- Reworded buildDependsOn's comment from "reachable" to "directly targeted", since it's only immediate edges, not the transitive closure.
- buildDependsOn now explicitly filters to EdgeKindRequires/EdgeKindWiring instead of trusting every edge kind returned by EdgesFrom.

kichristensen:
Addressed in 8637f0f2:
- seen now uses map[string]struct{} instead of map[string]bool.
- Added a test for DependsOn deduping a requires-edge + wiring-edge to the same target.
- Added a test for nodeAlias picking the first sorted alias when a shared dependency has two parents under different aliases.

Issue #2647, opened 2023-03-23
Run a workflow


carolynvs:
- [ ] Reconcile the installations in the workflow to understand what actions are necessary (i.e. an installation may already be in the desired state) #determine
- [ ] Generate an execution plan
- [ ] Vet the execution plan
- [ ] Persist the workflow
- [ ] Run the execution plan
- [ ] Support --dry-run to output the proposed (unsaved) execution plan

## Dependencies
* #2645 
* #2628 
* #2646 

ℹ️ Read [PEP003 - Advanced Dependencies](https://github.com/getporter/proposals/blob/main/pep/003-advanced-dependencies.md) for context about how dependencies should work, design details, and notes about desired behavior.

Issue #2608, opened 2023-03-02
Track references to a dependent installation


carolynvs:
When we create an installation from a dependency or reuse an installation to satisfy a dependency, we should keep track of who is referencing it and what for. This will allow us to do a few things:

* Know when it's safe to delete an installation, based on if it is still used
* When upgrading an installation, find the associated installations that represent its dependencies and act upon them as a unit or just display the dependencies.

To do this we need to keep these references up to date, both when the relationship is created, and later when the parent installation is updated (perhaps it stops depending on another bundle) or deleted.

Tentatively I think we can store this information on the dependent installation's status:

```yaml
schemaType: Installation
name: mysqldb
status:
  references:
    - installation: myinfra # the name of the installation that is using this one
      dependency: db # the dependency that it satisfies
```

ℹ️ Read [PEP003 - Advanced Dependencies](https://github.com/getporter/proposals/blob/main/pep/003-advanced-dependencies.md) for context about how dependencies should work, design details, and notes about desired behavior.

PR #3684, opened 2026-09-08
Build job installation records

kichristensen:
# What does this change
Adds `buildJobInstallations` (`pkg/porter/workflow_installations.go`):
builds the `Installation` each workflow `Job` carries.

- root node: caller's installation, used as-is
- existing installation (`Node.ResolvedInstallation`): a copy of it
- new dependency: fresh `storage.NewInstallation`, labeled like
  `runDependencyv2` (parent label, sharing-group label)

Also records `AddReference` (#2608) for every requires edge, so
dependency-reference bookkeeping stays correct even in a diamond
(two parents requiring the same dependency).

In-memory only - no `Insert*`/`Update*` calls. Persisting these
records is a later step (#2647). Nothing calls this yet.

# What issue does it fix
Part of #2645 (does not close it). Third of the planned chunks -
builds on #3677 and #3682 (merged).

# Notes for the reviewer
Chunk 3 of ~5 PRs implementing #2645. Chunks 4-5 (pending runs,
wire dependency params/credentials) build on this and follow next.

# Checklist
- [x] Did you write tests?

### DIFF
--- a/pkg/porter/workflow_installations.go
+++ b/pkg/porter/workflow_installations.go
@@ -1,0 +1,75 @@
+package porter
+
+import (
+	"fmt"
+
+	"get.porter.sh/porter/pkg/storage"
+)
+
+// parentInstallationLabel records, on a freshly created dependency
+// installation, the root installation whose dependency tree it belongs to.
+// Mirrors the "sh.porter.parentInstallation" label runDependencyv2 sets
+// (dependencies.go), but always points at the root rather than the
+// immediate parent: in a transitive graph a node can be several levels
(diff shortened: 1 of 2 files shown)


kichristensen:
Fixed in 26d06364:
- Defensive-copy References before mutating (was sharing backing array with Node.ResolvedInstallation, spare capacity could alias). Added regression test proving it: fails without the fix, passes with it.
- Renamed child to dependency (edge.To is the dependency, per the graph's From-depends-on-To convention).

